PR #51, opened 2015-04-24
Fix Bug where Plant Arbitrator was trying to arbitrate for unsown crops ...

HamishBrownPFR:
...that don't have phenology.

Have altered the test for IsEmerged so a crop has to be sown for it to be true and if that crop has phenology it has to also be emerged


HamishBrownPFR:
This resolves #52 so it can be closed

hol353:

hol353:
test this please

PR #70, opened 2015-05-01
Clear old error messages before start of run. Resolves #69.

### LINKED ISSUE #69: Error messages not being cleared between runs
If a run fails due to an error, the GUI correctly displays an error message. However, if the user corrects the error and re-runs the simulation, the old error message is displayed at the end of the run, even though the simulation was successful.

zur003:
Resolves #69.


zur003:
Also added changes for handling right-click events. Resolves #71. I probably should either have waited for the first changes to be merged in MAIN/master, or started a new branch.

PR #82, opened 2015-05-06
Handle errors or missing values in met files a little more robustly.

zur003:
Addresses part of #6.


zur003:
Ooops.  Should have re-based first.

PR #89, opened 2015-05-11
Fix #19: Potential failure in SimpleLeaf.cs

### LINKED ISSUE #19: Potential failure in SimpleLeaf.cs
From Bangyou (2015-03-12):

Function "CoverGreen" in the SimpleLeaf.cs using CoverFunction or ExtinctionCoefficientFunction to calculate green cover. Both "CoverFunction" and "ExtinctionCoefficientFunction" are optional. "CoverGreen" only checks whether "CoverFunction" exists. 

Program will fail if both "CoverFunction" and "ExtinctionCoefficientFunction" don't exist in the model. 

       public double CoverGreen
        {
            get
            {
                if (CoverFunction == null)
                    return 1.0 - Math.Exp((-1 * ExtinctionCoefficientFunction.Value) * LAI);
                return Math.Min(Math.Max(CoverFunction.Value, 0), 1);
            }
        }
 
 
        [Link(IsOptional = true)]
        IFunction CoverFunction = null;
 
        [Link(IsOptional = true)]
        IFunction ExtinctionCoefficientFunction = null;

byzheng:
Resolve #19.

### DIFF
--- a/Models/Plant/Organs/SimpleLeaf.cs
+++ b/Models/Plant/Organs/SimpleLeaf.cs
@@ -412,6 +412,10 @@ namespace Models.PMF.Organs
             if (Plant.IsEmerged)
             {
                 FRGR = FRGRFunction.Value;
+                if (CoverFunction == null & ExtinctionCoefficientFunction == null)
+                {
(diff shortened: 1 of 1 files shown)


hol353:
Retest this please

PR #107, opened 2015-05-14
Added functionality to show object types on tree for PMF objects

aldoherty:
probably should merge with develop to test functionality rather than master. I will log task as soon as I can.  It is good practice to log a patch against an issue however.


hol353:

hol353:
All, this needs to reference an issue via the words - resolves hash character and a issue number. Look at the other pull requests. Also, I'll only merge with master when the issue has been fully resolved.

hol353:
Test this

PR #124, opened 2015-05-25
Changes to allow GridViews to automatically resize.

### LINKED ISSUE #122: Changes to GridView
GridView does not resize correctly when the APSIM window is small. Implement fixes to allow resizing and add scroll bars where the window is too small to handle the content.

fai04d:
Resolve #122.


hol353:
Justin

Can you make a small change to this please?

Can you have the presenter pass the Image through to GridView LoadImage method as an argument. i.e. GridView method becomes:

public void LoadImage(Bitmap bitmap)

This will allow other presenters to pass whatever image they want and it doesn't have to come from a resource.

Dean

PR #151, opened 2015-06-17
Cleaning up Crop model development

HamishBrownPFR:
There is lots of stuff on the pull request.
- Moved some unfinished crop models from tests to prototype folders and removed .xml from resource folders
- Added some unfinished crop models to protype folders that were on branches
- Some source code changes to fix problems in PMF and IDE icons

Sorry this is to much to put in one pull request but I am trying to tidy up the mess that has occured from developing crop models and changing source code on multiple branches.


hol353:
Hamish

Some of these user interface changes are old I suspect and we don't want them e.g. constantpresenter.cs, phaselookvaluepresenter.cs. Al and co are using PropertyPresenter instead.

Can you just put the fieldpea stuff on one pull request, the plant source changes on another etc? Have separate GitHub issues for each so I know what you're trying to do.

HamishBrownPFR:
I was aiming to do it in smaller chunks but I am not really sure how to separate things out.  The problem is I would make some source changes on one branch (say field pea) and then whan those changes on another branch (say MCSP) so would pull one branch onto another branch to get those changes.  Now they are all intermingled and I can't work out how to separate them, or even if I can.

Having the prototype folder will mean we can put up smaller pull requests and share these changes through the upstream master which will be a lot nicer.  However, this is the mess I got in in its absence.

hol353:
Can't you just manually copy the files you want to keep (eg. just the PMF source changes) stick them in a folder somewhere, get your branch clean (or clone an entirely new ApsimX repo/folder), create a new branch, copy the files in and commit and push.

HamishBrownPFR:
I have set my branch to master, made sure it was set to the tip revision upstream/master, copied the entire folder (ApsimXMaster), changed my branch to back to MCSP-testing and diffed my ApsimX folder with ApsimXMaster.  This quite quickly showed up which files that were different because I pulled them from another branch and they then got behind when further changes were pulled in to master.  So I have overwriten the outof date versions on this branch with the correct versions from the master.  

This was a relatively quick way of getting rid of the stuff I didn't want out of my branch but it seems a bit dirty to have to take a copy of the master and diff it outside of sourcetree.  It also means that the merge request is showing up alot of files that are different which are not actually different.  So it is difficult to see what is actually being changed.

I did have some success trying to do the same thing through source tree.  By clicking on the working most recent commit of this branch in the log window and then holding down ctrl and clicking on the upstream/master commit the file window displays files that differ between the two versions.  By right clicking a file to bring up the menu, selecting 'log selected' brings up a window with all the commits for that file.  Right clicking the commit you want to revert and clicking 'reset to commit' reverts the file to the version you want.  This also removes the file from the list of changed files on the branch and in the pull request.  However, there were a number of files where the log was incomplete and the most up to date commit (that I wanted to rest to was not listed).  I was not able to solve this problem so not able to use this approach to solve the problem, which is annoying because it would have kept the list of changed files in the pull request limited to those that are actually different and it would have removed the need to take a full copy of the master to diff with.

HamishBrownPFR:
This is nearly clean now.  I did a re-base which got all the commits updated and then reverted all the files that shouldn't have had changes in them the to tip commit on the upstream/master.

There are still a few files showing up as changed that are not changed.  They are the files in the documentation folder and They do not show up as changed on my branch but do in the pull request.

Perhaps my branch is doing its compare with the origin master and this is out of date from the upstream master.  I will check this now

HamishBrownPFR:
Not sure what is going on with this pull request.  If I go to my branch on github (https://github.com/PlantandFoodResearch/ApsimX/tree/MCSP-testing) and go to compare I get 38 changed files.
image

However if I go the to pull request on https://github.com/APSIMInitiative/ApsimX/pull/151/files
51 changed files are showing.

image

PR #157, opened 2015-06-19
Hamish tidy up

HamishBrownPFR:
Added Field pea model
Further development of MCSP
Removed PMFSlurp from resources file and moved to prototypes
Fixed some problems with Maize model
Source code changes to PFM


HamishBrownPFR:
Phew!

Issue #166, opened 2015-06-23
Convert Simple Crop Uptake Model (SCUM, the crop model formerly known as MCSP) to APSIMX


HamishBrownPFR:
This model is to be built for the single purpose of estimating nitrogen and water balances on New Zealand cropping farms.  It may be applied for other purposes but the user will need to keep in mind that it is empirical and the parameters might need changing for crops from different areas.  Yield is an input not a prediction.  

See the attached file for full description

PR #180, opened 2015-06-28
Mcsp #166

HamishBrownPFR:
IMPORTANT!!!!

There is a bug fix to the phenology class of PMF in here:
    1. Phenology class was not adding the fraction of progress through a phase to the Phenology.Stage variable on the first day inside each phase
   2.  Phenology class would only allow progress through one stage per day.  Have changed so if short phases take less than the days Tt to complete the model will pass through unlimited phases each day

This change will cause model tests to fail.   Should provide a good test case as to how we deal with model infra structure changes that affect multiple crop models

Also some progress on


HamishBrownPFR:
I have separated the bug fix into another pull request and issue

PR #182, opened 2015-06-28
Phenology bug fix #181

### LINKED ISSUE #181: Fix bugs in PMF phenology
1. Phenology class was not adding the fraction of progress through a phase to the Phenology.Stage variable on the first day inside each phase.  Needs minor change to fix this.  It will affect model results as a number of things are functions of Phenology.Stage however the effect should be minor as this value will only change on days when phases change

2. Phenology class would only allow progress through one stage per day. Have changed so if short phases take less than the days Tt to complete the model will pass through unlimited phases each day.  This will probably not affect any current PMF models but was causing problems in MCSP (prototype)

HamishBrownPFR:
to resolve issue #181 this will cause problems as there will be some small changes to test results.

What is the process for checking off changes when tests fail and allowing bug fixes like this into the master


hol353:
The process is that you go through the diffs and if you're happy with all of them then rename the .db to .db.baseline and commit it.

HamishBrownPFR:
Need to have issue #183 resolved to work out what is failing

PR #201, opened 2015-07-14
Phenology bug fix #181

### LINKED ISSUE #181: Fix bugs in PMF phenology

HamishBrownPFR:
Have fixed the phenology bug and the flow on issues in wheat model.  Expect other test will fail on first try


hol353:
HI Hamish

There is a fodderbeet XML file in the resources directory that needs to be removed. Also the wheat validation .apsimx file has a replacements in it. We don't want that either. The validation must use the wheat XML in the resources.

HamishBrownPFR:
Those problems fixed now, should be good to go?

Issue #43, opened 2015-04-23
Summary file no longer rendering in html in the UI view


HamishBrownPFR:

hol353:
The seem to work ok for me. Which simulation are you looking at?

HamishBrownPFR:
All the simulations I look at on my computer and that I have looked at on Edmars are displaying the summary file in raw text

zur003:
The summary view has quite a few problems. It seems to be very slow to load and draw initially; the current style looks very "raw" - possibly this is a style sheet issue (could that be the source of Hamish's issue?); backslashes in front of text "swallow" the following word (because it's now trying to parse RTF rather than HTML?), so file paths display as just the file extension.

zur003:
Looks like things were changed to use RTF for display rather than HTML (hence avoiding the problem of finding a good HTML display engine). Is this what Hamish is describing? It certainly doesn't look as polished.

hol353:
The HTML component has having all sorts of issues so we decided to ditch it in favour of something (simple RTF control) that would run on MONO as well. I suspect that Hamish is seeing that the tables of numbers are no longer HTML tables. This is because I didn't have the time to support tables.

BTW: I notice the font has changed to a proportional one. This upsets the formatting of tables. Can we put it back to the way it was?

HamishBrownPFR:
This is what it looks like for me.  The summary file view stops scrolling part way down the file so you can't see all the contents

![image](https://cloud.githubusercontent.com/assets/7295858/7465146/920b1bda-f31f-11e4-8468-670aa208e2c0.png)

PR #202, opened 2015-07-15
Use MigraDoc to generate the RTF used to display the SummaryReport.

### LINKED ISSUE #106: Curly braces not appearing in summary file Script listings.
The HTML to RTF conversions now being used for generation and display of the Summary has some flaws. Specifically, "curly braces" are not appearing in the text, which is a problem with manager scripts. Also, it should be possible (I think) to convert the HTML tables to RTF tables. This should enhance readability.

zur003:
Resolves #106. Also resolves the old (already closed) #43.


hol353:
Retest this please Jenkins

PR #203, opened 2015-07-15
Made graph series and regression separate in the user interface tree.

### LINKED ISSUE #192: Rearrange the graphing node and split up the individual series into separate nodes in the tree.
This will simplify the code, be backwards compatible and also allow a regression stats node to be put onto individual series. Should be more flexible. It will also speed up the construction of graphs in the user interface.

hol353:
Resolves #192


hol353:
Retest this please Jenkins

PR #220, opened 2015-07-21
Set all projects to use x86 for debug and AnyCPU for release mode. 

### LINKED ISSUE #218: Set projects to use AnyCPU when in release mode.
I need a GitHub library called 'Octokit' which allows accessing the GitHub system via it's API. This library though doesn't support .NET 4.0. The newer version of 4.5 is compatible with all Windows and MONO systems.

hol353:
Also changed to VStudio 2015.

Resolves #218


zur003:
So we will need to update to VS 2015? Presumably the free "Community" edition will suffice, as APSIM is open source. Will Apsim 7 also be moving the VS 2015?

hol353:
Yes I am suggesting we move to VS 2015.

One problem I've got is if we move to AnyCPU for release builds. When I do that none of the manager scripts compile. This line:
CompilerResults results = Provider.CompileAssemblyFromSource(Params, source);
in reflectionUtilities.cs throws an exception: UnAuthorized. Google doesn't seem to provide an answer.

Perhaps we don't go to AnyCPU.

hol353:
I've solved the manager compiling issues by turning off the creation of debug systems e.g.

Params.IncludeDebugInformation = true;

The next problem though is that sqlite3.dll is a 32 bit DLL and won't load from a 64bit application.

For now I'm closing this issue and sticking with x86.

PR #235, opened 2015-07-31
Fix problem with disappearing labels. Resolves #222.

### LINKED ISSUE #222: Text Box labels disappear for Observed Predicted Object
If I select the predicted-observed object when the GUI window is not maximised I find that the labels for the text boxes on the view for it disappear

image

zur003:
Resolves #222.


hol353:
I'm in the middle of resolving this a different way i.e. use the propertypresenter instead.

PR #258, opened 2015-08-20
Add check for duplicate simulation names when running a sim.

### LINKED ISSUE #4: ApsimX: Report always shows results of last sim run.
From Justin (2014-07-17):
Tested with OilPalmValidation.apsim with no existing .db file. Simulations are all called base; they reside in a factorial node named after the experiment number. Each Experiment node is in a seperate folder. 

Run the Base simulation for Expt 504 which generates a report. Clicking on the MonthlyOutput report shows the results of the run. If you now click on the MonthlyOutput report for Expt324, you will get the results for 504 show (This sim has not yet been run). Run Expt 324 and the MonthlyOutput will show the expected results. However, if you now click the MonthlyOutput node for 504, you will get the results for 324. Running 504 again gives results for 504 regardless of which experiment you click. 

This could be happening because all simulations are named 'Base'. Might need to either expand the uniqueness search to the whole XPath, or disallow identical simulation names even though they're in different places.

hol353:
Resolves #4

### DIFF
--- a/Models/Core/Simulations.cs
+++ b/Models/Core/Simulations.cs
@@ -337,18 +337,6 @@ namespace Models.Core
             NumToRun = simulationsToRun.Length;
             NumCompleted = 0;
 
-            // See if any simulations are named the same as other ones.
-            List<IModel> allSimNames = Apsim.ChildrenRecursively(this);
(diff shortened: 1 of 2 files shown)


hol353:
retest this please Jenkins

PR #267, opened 2015-08-21
Added YieldBuPerAcre output to Maize and fixed bug in report

### LINKED ISSUE #129: New Maize model should have a yield output in Bu/Ac
To be relevant to users and students in USA, yield should also be available for output in Bu/Ac.  This can be provided by a simple function in the grain object that multiplies yield by a conversion factor.

hol353:
Resolves #129


hol353:
retest this please jenkins

Issue #273, opened 2015-08-25
Implement a biomass removal interface for all PMF organs


HamishBrownPFR:
Currently the amount of biomass that is removed by a organ is hard wired in.  Need to be able to send remove fractions with harvest and cut method calls.

hol353:
In Pull request #277 you have manager code like:

//Harvest with manager specified removal fractions
ManagerRemovalData[0].FractionRemoved = 0.3;
ManagerRemovalData[1].FractionRemoved = 0.1;
ManagerRemovalData[0].FractionToResidue = 0.3;
ManagerRemovalData[1].FractionToResidue = 0.7;
MCSP.Harvest(ManagerRemovalData);

Having an array of biomass removals (one for each organ) assumes the order in this manager script exactly matches the order of organs in the plant. Also, you can't tell by looking at the script which fraction relates to which organ. Is there a better way to do this?

HamishBrownPFR:
I am hoping there is a better way as this is not really satisfactory.  Is there a way of finding array positions based on a name rather than number?  The other option is to send text strings from the manager like "Leaf.FractionRemoved = 0.6" and pass them into the appropriate location in the source code.  This would be better for users as you wouldn't have to construct the ManagerRemovalData object in manager script either.

hol353:
We could create a new class (say RemovalFractions) to contain the array of ManagerRemovalData objects. This class could have methods:

void SetFractionRemoved(string organName, double fraction);
void SetFractionToResidue(string organName, double fraction);

called from manager:

RemovalFractions removals = new RemovalFractions();
removals.SetFractionRemoved("leaf", 0.1);
removes.SetFractionToResidue("leaf", 0.6);
...

and then pass this into Harvest. Is this better?

HamishBrownPFR:
Looks better, I will try it and see how it works.

HamishBrownPFR:
Where do I put the RemovalFractions class so it is accessible in the manager.  Do I have to construct it in the manager or can I put it somewhere in the source code

HamishBrownPFR:
Don't worry, I have worked it out

HamishBrownPFR:
Have pushed some changes up that do it the way you suggested.  Is tidier I think so thanks for the suggestion.

[...]

PR #277, opened 2015-08-27
Specify harvest remove proportions

HamishBrownPFR:
Working on #273.  On further inspection the approach for removing biomass from PMF organs was in need of some work.

* Added FractionRemoved and FractionToResidues to the organ interface.
* Created an OragnBiomassRemovalType class which contains the above two properties
* Added a RemoveBiomass property (setter) organ interface.
* When an event is called that removes biomass (currently Harvest and Cut but will add Graze and Prune to this) it calls the RemoveBiomass property sending the FractionRemoved and FractionToResidue values.  
* When these are recieved by the organ classes the appropriate amounts of biomass are sent to surfaceOM and soilOM and removed from the organs biomass.
* Each organ has default FractionRemoved and FractionToResidues values for harvest and cutting events which are used in the absence of others.
* FractionRemoved and FractionToResidues can be sent as arguement in the Harvest() and Cut() methods when called in the manager.
* These will overwrite the defaults and can be used where specific behavior is desired.

I suspect this will cause some tests to fail because biomass returns have not been dealt with properly in PMF till now.

There are still some outstanding issues
1. The leaf organ needs some functionality to respond to cut and harvest using the new infrastructure
2. Grazing and Pruning events need to be plumbed into the new infrastructure
Neither of these are big jobs so it would be good to merge this in when I get it running green.  Also I need these changes for completing the SCUM release procedure.
However, keep the issue open until these final points are resolved


hol353:
retest this please Jenkins

hol353:
Hmm I don't really like the setup method.

The error you were probably getting was because you didn't have a default constructor (i.e. one with no arguments). e.g. this should work:

public RemovalFractions(IOrgan[] Organs)
     {
         OrganList = new OrganBiomassRemovalType[Organs.Length];
         int k = 0;
         foreach (IOrgan o in Organs)
         {
             OrganList[k] = new OrganBiomassRemovalType();
             OrganList[k].NameOfOrgan = o.Name;
             k += 1;
         }
     }

public RemovalFractions() { }

Here there are two constructors, your normal one and the empty bottom one that the serialisation engine will use. Does this work?

HamishBrownPFR:
Yes you are correct, including the default constructor works.  Have made changes and done another push.

PR #287, opened 2015-08-30
Mcsp #166

HamishBrownPFR:
working on #166 

Impovements made to docummentation functions in a number of PMF classes.  Some issues still outstanding.

*  Can't find a way to insert large block of markdown written in another editor into docummentaiton without loosing most of the formatting.
* Another lower level, un-numbered heading (Bold Itallic) needed.
* Root organ doco function still needs work.
* Graphs needs some commentary


HamishBrownPFR:
Here is what I mean by needing another heading, using GenericOrgan (parameterised as product)

This highlighed section of code I want to put a unnumbered bold itallic heading in.  Currently (see doco screen shot below) it doesn't put anything in.
image

image

HamishBrownPFR:
And here is what I mean by inserting large blocks of markdown.  There is a description of the arbitration process in C:\ApsimX\Documentation\Arbitrator.md which I set up in a MD editor.  I want to either stick this into the source code and have the doco pick it up with the same formating or be able to reference the file so the doco inserts it in the correct place.

The screen shot below is of OrganArbitrator.cs.  I have tried using a markdown file reference (I am not sure I have it right and it certianly doesn't work) with no luck.  I have also tried breaking it up in to a text string (see the commented out section in image below) but all the formatting is lost an the autodocummentor doesn't seem to be picking up any of the md formating.

image

HamishBrownPFR:
I will continue to work on the root doco and graph descriptions but will need some help with the other two issues to get this ready for RP.

hol353:
I'll look into these issues ASAP. I notice that you have a potatovalidation.apsimx in your pull request. Is this intentional?

HamishBrownPFR:
Thanks Dean.  I Didn't intend to change potato but did open it so it is probably some ui changes

hol353:
Spacing is important it seems with markdown. There needs to be an empty line before and after a ordered or unordered lists. Image paths need to be relative to the directory where the .apsimx file is (no absolute paths - Jenkins doesn't run in c:\ApsimX). Also remember to double quote the back slash characters in paths. Bold and italics seems to need having two stars together without a space between them.

I don't think we really want to have a separate file called Arbitrator.md. This separates the doc from the source code which we're trying to avoid.

I've committed a push and change that will fix some doc issues. Once you have this the following worked for me in Organ Arbitrator:



    /// <summary>
    /// The Arbitrator class determines the allocation of total dry matter(DM) and Nitrogen components of Biomass between each of the organs in the crop model.Eacn organ potentially has three pools of biomass:
    ///
    /// * Structural biomass which is fixed within an organ once it is partitioned to an organ
    /// * Non - structural biomass which is available for re - translocation to other organs with high demand and is re - allocated to other organs when this organ senesces.
    /// * Metabolic biomass which is generally fixed in an organ but is able to be reallocated and may be re - translocated in some cases.
    /// 
    /// The process followed for biomass arbitration are shown in Figure 1.Arbitration responds to events broadcast daily by the central APSIM infrastructure: 
    /// 
    /// 1. **doPotentialPlantGrowth**. When this event is broadcast the attached method executes code to determine the potential growth of each organ, the extent of moisture stress that a crop encounters and the potential biomass supplies and demands of each organ based on these.In addition to demands for structural, non - structural and metabolic biomass(DM and N) each organ may have the following biomass supplies: 
    ///     * Fixation supply.From photosynthesis(DM) or symbiotic fixation(N)
    ///     * Uptake supply.Typically uptake of N from the soil by the roots but could be uptake by other organs.
    ///     * Retranslocation supply.Non - structural biomass that may be moved from one organ to meet demands of other organs
    ///     * Reallocation supply.Biomass that can be moved from senescing organs to meet the demands of other organs.
    /// 2. * *doPotentialPlantPartitioning.* *On this event the Arbitrator first executes the DoDMSetup() to establish the DM supplies and demands from each organ.Then it executes the DoPotentialDMAllocation() method which works out how much biomass each organ would be allocated assuming N supply is not limiting and sends these allocations to the organs.Each organ then uses their potential DM allocation to determine their N demand(how much N is needed to produce that much DM) and the arbitrator calls DoNSetup() establish N supplies and Demands and begin N arbitration.Firstly DoNReallocation() is called to redistribute N that the plant has available from senescing organs.After this step any unmet N demand is considered the plants demand for N uptake from the soil (N Uptake Demand).
    /// 3. * *doNutrientArbitration * *When this event is broadcast by the model framework the soil arbitrator gets the N uptake demands from each plant (where multiple plants are growing in competition) and their potential uptake from the soil and determines how nuch of their demand that the soil is able to provide.This value is then passed back to each plant instance as their Nuptake and doNUptakeAllocation() is called to distribute this N between organs.
    /// 4. * *doActualPlantPartitioning * *On this event the arbitrator call DoNRetranslocation() and DoNFixation() to satisify any unmet N demands from these sources.Finally, DoActualDMAllocation is called where DM allocations to each organ are reduced if the N allocation is insufficient to achieve the organs minimum N conentration and final allocations are sent to organs.
    /// 
    /// Alt Text
    /// 
    /// * *Figure 1.* *Schematic showing procedure for biomass partitioning arbitration.Orange boxes contain properties that make up the organ / arbitrator interface.  Green boxes are organ specific properties, pink boxes are events that are broadcast each day by the model infrastructure and blue boxes are methods that are triggered by these events.
    ///  For both DM and then N the arbitrator steps through allocation of biomass from ReAllocaiton, Uptake, Retranslocation and then fixation.Biomass is only allocated from subsequent sources if demands have not already been meet from the preceeding source.";
    /// </summary>

and then put this line in the document routine:
       
            AutoDocumentation.GetClassDescription(this, tags, indent);

HamishBrownPFR:
Agree we don't want doco in seperate files.  The location of the margin is really important for markdown so putting the doco in the summary tag as you have done works well.

HamishBrownPFR:
I have just pushed some more changes up.  The doco is looking OK now so If these run clean can they be merged in please and I think this will be adequate to go back to RP.  I still need to document the root and provide some sensibility tests for N and irrigation responses but will work on that tonight.  

Some of the legends on the graphs are getting pushed off the side of the page which will need looking at.

hol353:
Sorry to be picky but can we get rid of arbitrator.md and potatovalidation.apsimx files out of this pull request?

HamishBrownPFR:
With the latest commit I have dealt with all of Neils comments done a bit more tidying up.

I need to sit down now for half a day a work through the document finding oddities and fixing them.  However this isn't going to happen before the RP meeting tomorrow so please meage this in it if is clean and OK.

PR #308, opened 2015-09-02
Expanding doco and adding more tests

HamishBrownPFR:
working on #166 

Adding some extra stuff into the up front introduction better explaining the scope and limitations of the model.

Have added a section about how to use the model in a simulation and tried to put some code boxes from manager to show how it was done but they are not rendering properly.  See highlights below for example.  

Have added a test against the rainshelter maize irrigation x N trial but observations not picking up on graph yet

Not for merging but for help to resolve this doco problem.  I wonder if it would be better to have a doco describing how to use sow, harvest etc in manager and reference that rather that putting stuff in each model which could get out of date

image


HamishBrownPFR:
Added code needed for doing grazing and regrowth.
- PMF in old APSIM we had seperate phases for regrowth which were given stage numbers according to the point where they were to fit into the phenology.  This wouldn't work in APSIMX because stage numbers are done sequentially.  Instead have given regrowth phases a property called parallel phase which is the point they are to be inserted into the sequence.  This functionality is there but I don't think it will be used much because I have also implemented something else....
There is now a SetStage property that is sent with harvest, cut, graze and prune events which tells the model where this event should rewind the phenology to.  Have added a test in SCUM tests.  Seems to work well.

HamishBrownPFR:
Still some work to do to get the graphs working in the maize test.  If you are happy with this can it be merged in as I need to get the defoliation stuff to a colleague.

PR #332, opened 2015-09-09
Water graphs now show series name correctly in legend.

fai04d:
It looks the XML model of OxyPlot has changed slightly. Modified graph XML to match new structure.
Resolves #332


hol353:
Justin

The issue number above is wrong. Also, are you meaning to commit a changed AgroForestry.apsimx?

Issue #243, opened 2015-08-10
Set tests and documentation for SLURP and resolve which slurp to use


HamishBrownPFR:
There are two SLURP models, one the I set up in PMF and another that Neil and Justin have wrote.  Neither have been through reference panel.  We should probably drop one of these and get the docummentation sorted for SLURP.  We wrote in the PMF paper that SLURP is implemented in PMF so my vote of to keep this version.  

Is there any functionality in the other SLURP that the PMFSlurp is not covering?

PR #340, opened 2015-09-09
Slurp doco and tests

HamishBrownPFR:
working on #243 

Finalised structure of PMF based Slurp
Added a test against real data and a sensibility test
Doco still needs some work and test need a polish.

Please merge if all ok


HamishBrownPFR:
Closed and replace with one that will work

PR #351, opened 2015-09-10
Slurp development

HamishBrownPFR:
working on #243 

Completed structural set up of PMF baased SLURP.  added a test with some data, a sensibility test and some doco.

Doco needs more work and test needs a polish.

Please merge if all OK


hol353:
Hamish, in root.cs, why has a bunch of functions been made optional? I don't really like optionals.

HamishBrownPFR:
Because they are parameters that can take a default value so they don't have to be specified in the crop.xml.  Keeps the xml and doco tidier

Why do you not like optionals??

hol353:
Because the code usually has a bunch of if statements testing for nullness and the XML is no longer a complete record of a crop parameterisation. Some values are in the XML and some are specified in the source code. It also means we can't write a tool down the track for going through the XML and flagging missing functions as errors as so many functions are optional. It's not a big issue for me. Maybe something we can talk about when you're over here.

I'll merge this in now.

HamishBrownPFR:
Ok, there must be a better way to achieve this, lets nut it out in October

PR #371, opened 2015-09-17
Added ability for auto-doc to look for type of model rather than just…

### LINKED ISSUE #359: Null reference when trying to document the Agroforestry model.
This model does not have code to enable auto documentation. Needs to be added.

hol353:
… name.

Resolves #359


hol353:
I have also created a auto-documentation page:
https://www.apsim.info/Documentation/ApsimX/Design/Models/Autodocumentation.aspx

PR #437, opened 2015-10-12
Fix ResolveLink function to correctly resolves [Link]s

### LINKED ISSUE #350: PMF organs should not resolve [Link] for functions when there is no exact match
PMF is finding links for functions that don't exist.  This bug has been fixed once before so not sure why it has come back.

The test I used was to rename the NitrogenDemandSwitch on the Root organ.  to NitrogenDemandSwitch123412.  In the code this is not optional so should throw an object reference error.  But it doesn't it just finds another function and uses its value,

image

hol353:
Resolves #350


hol353:
retest this please Jenkins

Issue #453, opened 2015-10-14
New PMF Wheat model


hol353:

HamishBrownPFR:
Inprovements identified at workshop with Chapman etal

- N response in tillering
- externalise leaf size function parameters
- factor to reduce the size of tiller leaves early
- externalise N demand calculations

PR #471, opened 2015-10-16
Changed Wheat Micromet settings.  Added some missing irrigation to Li…

HamishBrownPFR:
working on #453


HamishBrownPFR:
We can have conversations through the pull request window

<<Neil Wrote>>
I'll merge the wheat micromet one - I might suggest that we don't derive wheat from maize - mostly because it will confuse people (using a summer C4 to to a winter C3).  But we can clean this up when I finish getting my bowen ratio data in there - I'll check the daily wheat ET values with measured. 

My implementation is not really satisfactory but Dean is keen to refactor it and put those properties in plant so will get tidied up then.

PR #491, opened 2015-10-20
Reworked the drag/drop mechanism

### LINKED ISSUE #479: Method to add PMF classes when building crop models
In the old PMF we had a tool box with all off the PMF classes so they could be dragged into a model setup to build the model.  Currently we don't have this so they have to be dragged out of another crop model or written into the xml manually.

The problem with the old system was that the tool box objects got out of sink with the source code so often a toolbox object would not work properly. 

Need to add some system that enables adding PMF objects that keeps in sync with source code. Perhaps a right click menu item??

hol353:
Resolves #479


hol353:
retest this please Jenkins

Issue #56, opened 2015-04-24
Addition of Agroforestry Model 


fainges:
An agroforestry model is required. Implementation will be by Neil Huth and Justin Fainges. The implementation will involve the creation of a multi-Zone aware tree model and associated GUI.

fai04d:
Moved model to Prototypes directory.

hol353:
This is the initial reviewers report by Michael Battaglia dated 30th August 2016

sarahcleary:
@hut104 - can you please provide an update on where this is at?

sarahcleary:
@hol353 to discuss progress with @hut104

sarahcleary:
Please note action from RP 8/19 action: @hol353 to discuss with @hut104 and provide update

hol353:
From @hut104: Agroforestry model still to be incorporated into release sometime this financial year.

hut104:
@hol353  and @sarahcleary  please see the response to the reviewer's report.

[Agroforestry_System_Model_APSIM_ReviewersReport_v1 3 Response.docx](https://github.com/APSIMInitiative/ApsimX/files/4089197/Agroforestry_System_Model_APSIM_ReviewersReport_v1.3.Response.docx)

hol353:
Looks good to me. @sarahcleary - can this be signed off by the RP out of session?

sarahcleary:
@APSIMInitiative/reference-panel - please note the response to the reviewer's report above.  Please review and confirm this is ready for release.

peter-devoil:
Fine by me.

sarahcleary:
@APSIMInitiative/reference-panel  - would be good to be able to sign this off out of session - but need a representative from each AI organisation to confirm.  At this stage I only have UQ's approval.

hol353:
I'm happy for Agroforestry to be included in the release.

jbrider:
Approved.

sarahcleary:
@APSIMInitiative/reference-panel - still awaiting approval from AgR and USQ

LouisAK:
Approved

sno036:
I will try to look at this before the meeting but tomorrow is pretty packed - might be later in the week

sarahcleary:
@sno036 - just a reminder we are awaiting AgR's approval.

sarahcleary:
@sno036 - just a reminder we are awaiting AgR's approval.

sno036:
[...]

PR #538, opened 2015-10-27
Agroforestry distance from tree changes

fai04d:
Working on #56 
Distance from tree now calculated at three points; each edge of the Zone and the Zone midpoint.
Root length density is now a weighted interpolation of these three points.


hol353:
retest this please Jenkins

PR #550, opened 2015-11-10
Moved projects to AnyCPU and removed bad references to MathNet.

### LINKED ISSUE #548: Visual Studio configuration problems
In Visual Studio, the projects have a mix of 'platform' configurations; some are Any CPU and some are x86. We want all to be AnyCPU for both debug and release builds.

Also, I am getting warnings (and so is Jenkins) about MathNet.Numerics:

C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.Common.targets(1578,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "MathNet.Numerics, Version=3.7.1.0, Culture=neutral, processorArchitecture=MSIL". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\Jenkins\workspace\1. GitHub pull request\ApsimX\ApServer\ApServer.csproj]

fai04d:
Resolves #548 
Changed all project platforms to AnyCPU.
Removed redundant references to MathNet from projects that don't need them.


fai04d:
Doesn't look like this will work. Some of the 3rd party DLLs we're using are 32-bit. This results in the following error:

An attempt was made to load a program with an incorrect format.

We'll have to run in x86 for now. I don't expect this will be a problem; APSIM doesn't use much memory.

fai04d:
Resolves #548 
Moved to x86 due to incompatibility with binary 32-bit DLLs.

fai04d:
Resolves #548
Moved to x86 due to incompatibility with binary 32-bit DLLs.

fai04d:
Resolves #548 
Removed all configurations except x86 to attempt to force that as the default.

PR #561, opened 2015-11-23
Fix filter expression so it doesn't throw an error when there is no f…

### LINKED ISSUE #560: Graph filter expression throws error when no filter expression is defined

HamishBrownPFR:
…ilter expression

resolves #560


HamishBrownPFR:
Hi Dean.  I hope you don't mind me making UI changes.  I really needed this fixed for today and fortunately it was quite easy. 

The fix is not very robust as it will only work if the filter expression uses a single column.  really needs to be changed to search the full length of the filter expression and add all columns in the expression to the datatable if there is more than one.

PR #601, opened 2016-01-25
Axis for regression stats were being passed in the wrong way.

### LINKED ISSUE #600: Regression line on graphs is backwards.
This is due to the axis being reversed when they're being passed to the calculation function.

fai04d:
Resolves #600 
Reversed order of parameters.


hol353:
retest this please Jenkins

PR #614, opened 2016-02-02
Adding tests for Maize, Potato, Slurp and Scum.  …

### LINKED ISSUE #612: Re-establish model tests
A number of models are not being tested since the implementation of the new testing system.  These need to go back in.

HamishBrownPFR:
Having issues with the Slurp and Scum predicted/Observed tables

resolves #612


HamishBrownPFR:
More detail on the issue with SLURP.  This relates specifically to tests set up on the https://github.com/PlantandFoodResearch/ApsimX/tree/SetTests branch.

In Tests/Slurp.apsimx the predicted/observed object is not generating a table.  I think the problem is due to the format of the date column which is being used for matching with predicted data.  This format does not seem to worry the graph object which is able to graph this data correctly.  I have tried different things with the format in the excel sheet that the observed data is being read from but nothing seems to work.  The wheat test also uses the excel reader to create an observed data table and it does not change the date format but I can't work out why it is working and the Slurp one is not.

image
image
image

PR #616, opened 2016-02-02
Set tests

### LINKED ISSUE #615: Tests fail when Observed data missing
When observed data is missing from a PO table, the test model rightly ignores it, however the null test object is still returned.

fai04d:
Resolves #615 
Tests model now checks for null stats and removes them along with corresponding columns in the table.


fai04d:
Sorry, it's not missing data, that's handled fine. The problem was a non-numeric field was being picked up (date).

Same resolution; null stats are removed.

PR #618, opened 2016-02-09
Fixed resizing of Fpool arrays

### LINKED ISSUE #604: SurfaceOrganicMatter model scrambles its internal records when a new OM type is added.
The attached zip file includes a test simulation that shows the problem.
One simulation adds wheat residue every year
Second adds maize every year
third is same as wheat but add maize in one year.  Adding a second residue type screws up the 
[test.zip](https://github.com/APSIMInitiative/ApsimX/files/106794/test.zip)

The following lines are the cause of the problem
            frPoolC = Resize2DArray<double>(frPoolC, maxFr, numSurfom);
            frPoolN = Resize2DArray<double>(frPoolN, maxFr, numSurfom);
            frPoolP = Resize2DArray<double>(frPoolP, maxFr, numSurfom);

Resizing the array assigns frPoolC to different residue types and leaves lots of zeros and so when you add residue, the 0 fractions results in C being essentially thrown away.

hut104:
resolves #604

### DIFF
--- a/Models/SurfaceOrganicMatter.cs
+++ b/Models/SurfaceOrganicMatter.cs
@@ -2774,9 +2774,9 @@
             Array.Resize(ref no3ppm, numSurfom);
             Array.Resize(ref po4ppm, numSurfom);
             Array.Resize(ref specific_area, numSurfom);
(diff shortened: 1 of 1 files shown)


hol353:
retest this please jenkins

Issue #637, opened 2016-02-23
Reduce the number of units using the System.Windows.Forms namespace


zur003:
Ideally, only Views code in the UserInterface should need to make use of anything in Windows.Forms. This makes porting code to other windowing platforms a lot easier.

PR #641, opened 2016-02-23
Eliminate use of the System.Windows.Forms namespace from a number of files

zur003:
Working on #637. Trying this pull request business afresh.


hol353:

zur003:
Going to try going back to the original pull request.

PR #774, opened 2016-04-18
Reworked the startup screen in the GUI

### LINKED ISSUE #757: Suggestion: Upgrade button should be on "home" window
The Upgrade button appears only after opening a simulation or toolbox - why? I suggest placing it on the "home" window in the standard section next to e.g. "Open an example".

hol353:
Resolves #757


zur003:
Implementing this in GtkSharp will be left as an exercise for the student.

hol353:
Yes, I wondered about GtkSharp when I was doing this. In the end I decided that Eric would sort it because he loves GtkSharp.

hol353:
Actually I need to do some more refactoring of this view to fix a few other problems.

zur003:
I think the time has come for me to share the love.

Issue #626, opened 2016-02-16
What direction should we take with the ApsimX GUI?


zur003:
I've been looking into the cross-platform operation of ApsimX, and thought this might be a good place to jot down some of my thoughts and observations, and also solicit the views of others.

The current GUI, based on Windows Forms, works reasonably well on Windows (as expected, though there are still a few rough spots), but rather poorly on Linux or Macintosh. My (admittedly limited) tests on Linux indicate rather poor responsiveness, the need to use a version of ICSharpCode.TextEditor more recent than that available via NuGet, and poor (that is, slow and with incorrect layout) display of the summary file. The situation is similar on the Mac which, although more forgiving in use of the TextEditor, tends to fail completely on any attempt to view the summary file. For both Linux and Mac, the worst performing aspect of GUI performance currently is the display of the summary file, which is implemented via an RTF representation. I'd like to see us get rid of this in any case: RTF is a difficult format to use. HTML would be much better, if we can find a suitable rendering engine that works across platforms.

If we really want to achieve good cross-platform performance, I think we would need to move away from Windows Forms completely, and move to GtkSharp. With GtkSharp, we should have access to the various controls we need with decent cross-platform support. These include: tabs, trees, grids, charts (Oxyplot has a GtkSharp implementation), context-sensitive code editing (using Mono.TextEditor from MonoDevelop). I'm a little less certain about rendering "rich" text, but I think cross-platform HTML rendering under GtkSharp should be possible via WebKit.

So the positive side of moving to a GtkSharp-based GUI would be a much better cross-platform experience. But there are negative sides as well. Here are a few I can think of:

1. Installation is likely to be much larger and more complex, as it would likely require installing the GtkSharp framework, WebKit, etc.
2. No one currently working on ApsimX has much GtkSharp experience (that I'm aware of), so we would all be facing the cost of learning a different technology. A lot of things work quite differently from their Windows counterparts (how controls are positioned; how the content of a tree control or grid is held and described).
3. There may be versioning issues. The underlying Gtk technology is now at version 3, but most GtkSharp support is at version 2, or currently migrating to version 3. (I suggest we start with version 2, and transition to version 3 only when it is more broadly supported.) There may also be some risk of DLL version conflicts with previously installed versions of Gtk.
[...]

PR #793, opened 2016-04-20
Tell the runner to ignore ApsimNG folder when looking for .apsimx

hol353:
files to run. 
Working on #626.


zur003:
Thanks, Dean.

PR #815, opened 2016-04-29
Use proper syntax for file filter.

### LINKED ISSUE #813: Excelinput throws error
When clicking the "EXCEL file name" - "value" field, Apsim throws an error. I've tried this in my simulations as well as in the standard toolbox, with the same result.

zur003:
Resolves #813 .

This small change fixes the immediate problem, which was a consequence of bad syntax in the file filter being passed to the FIleOpen dialog, but I don't want to mark it as resolved just yet, because what's really needed is a mechanism for passing in the correct file filter. In this specific case, we want to be able to look for *.xlsx files, not all files.

### DIFF
--- a/Models/PostSimulationTools/ExcelInput.cs
+++ b/Models/PostSimulationTools/ExcelInput.cs
@@ -46,6 +46,10 @@ namespace Models.PostSimulationTools
             }
         }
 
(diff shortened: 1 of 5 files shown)


zur003:
OK. The second set of changes should actually resolve the problem a bit better.

PR #844, opened 2016-05-16
Use "Management toolbox" as the label for the management toolbox.

### LINKED ISSUE #821: Wrong tab-name of "management toolbox"
The management toolbox tab is wrongly named "standard toolbox" (instead of "management toolbox"), see picture. This is confusing, because the "real" standard toolbox is also named "standard toolbox" in the tab.
management-toolbox

zur003:
Resolves #821


hol353:
retest this please Jenkins

Issue #65, opened 2015-04-30
Develop Chicory model


ndb01:

rcichota:
Chicory (Cichorium intybus) is a perennial forage herb that has increasingly been used as a forage in New Zealand and elsewhere.  The chicory plant produces leafy tops and has a thick deep taproot, it can be drought and pest tolerant.  Chicory is also characterised by high nutritional quality and potential to produce high dry matter yields.  It can be used for animal grazing or for silage production (in mixed swards), with good regrowth after harvest/graze.
Given the increasing use of chicory as forage, there is a good demand for a model that is able to simulate its growth and regrowth in APSIM. It is one of the forage crops considered in a New Zealand study with the objective of reducing N losses from grazing systems by manipulating the mix of forages grown, thus influencing N and water usage as well as animal diet. Chicory grows in temperate and subtropical climates, the forage and therefore the model will have a wide range of interest worldwide.

zur003:
Also of considerable interest in grazing systems in Australia.

PR #883, opened 2016-06-08
Feature/chicory model#65

rcichota:
Working on #65  - Chicory model, added few outputs to organ arbitrator


rcichota:
retest this please Jenkins

hol353:
Seems to be more than just a few outputs. Perhaps @HamishBrownPFR  can have a look at this given that it looks like changes to the organ arbitrator.

Also, I don't like the removal of the exception in MicroClimate. Need to reinstate that somehow.

HamishBrownPFR:
I will have a close look as soon as I can.

PR #885, opened 2016-06-08
Feature/specific root length#73

rcichota:
Working on #73 - setting specific root length and initial root DM as IFunction


hol353:
Rogerio, can you reinstate the MicroClimate exception.

hol353:
Rogerio, this pull request hasn't been merged yet and can't be because it is not up-to-date. If you want to keep these changes, you'll need to do a pull and then a push on this branch.

rcichota:
Getting this branch up to date, changes in this issue are:
1) Changed InitialDM and SpecificRootLength from variables to IFunction, the first to expose it in the interface and to conform with SimpleLeaf, the second to allow changes as function of plant age. I also changed the units of SpecificRootLength from mm/g to m/g. This gives a more sensible magnitude and it is the units used in most publications
2) Changed the test for GreenCover in MicroClimate (again) to use tolerance of 1^-10 (1^-9 is the maximum send by plants so some values send that should be within (equal) MicroClimate's limits, but aren't because of numerical precision...)

Issue #877, opened 2016-06-06
New Stock Model


her123:
Add the GrazPlan Stock component.

her123:
The Stock component is now functioning. It eats supplement, and transfers urine to SoilNitrogen and faeces to SurfaceOrganicMatter.

hol353:
Neville

This now needs to go to the Reference Panel. They will want to see some documentation and validation that demonstrate that it works. What tests have you got?

sarahcleary:
@APSIMInitiative/reference-panel - this is ready for review.  Call for possible reviewers?  or offers to review?

hol353:
Nearly ready for review - later today it should be. I'll let you know.

hol353:
@sarahcleary This is now ready for review. Possible reviewer: Matthew Harrison.

sarahcleary:
@APSIMInitiative/reference-panel -  any other suggestions or confirmation with @hol353's suggestion?

peter-devoil:
Di Mayberry?

hol353:
Di helped with creating some of the examples and was involved in some of the early workshops so not completely independent.

hol353:
What about @peter-devoil as a reviewer?

her123:
What about Lindsay?

sno036:
@peter-devoil  and Matt Harrison I reckon

sarahcleary:
@APSIMInitiative/reference-panel - any comments on the above suggestions?

jbrider:
Pete and Matt sound like great choices.

Are there more validation datasets available?

hol353:
No unfortunately. At a workshop a couple of years ago Andrew Moore was tasked with trying to find additional validation datasets. Apparently they are not easy to come by.

sarahcleary:
@m8harrison and @peter-devoil - are you both happy (and have time) to review this new model?

peter-devoil:
No. Such a formative component needs an international reviewer with broader scope / experience than I. I'm still asking around for one over email.

m8harrison:
Happy to do this but it will need for be next financial year as I have several commitments before June 30. Thanks for the invitation Sarah.

sno036:
Thanks @m8harrison - next FY is fine!

peter-devoil:
[...]

PR #886, opened 2016-06-10
GrazPlan Stock

her123:
working on  #877

retest this please Jenkins


hol353:
An excellent start. Needs a refactor at some point to make it look more APSIM like. e.g. prefixing type and variable names with a T or b for boolean or f for float etc isn't recommended by Microsoft. Also there seems to be XML parsing happening. Nothing else in ApsimX has a need to parse XML.

her123:
It's mostly complete. No interface from plant or AgPasture yet. The animals can eat supplement, deposit urine to SoilNitrogen and faeces to SurfaceOrganicMatter. The variable naming is because it's a direct port from the original code. This refactoring can be done. Apologies for the xml parsing. Future task. Will shortly work up a test simulation for a feedlot that replicates the original AusFarm outputs.

PR #888, opened 2016-06-13
Fix units for KL and XF

### LINKED ISSUE #853: Units in Soil Water UI are incorrect for KL and XF
see chickpea validation for an example

hol353:
Resolves #853


hol353:
retest this please Jenkins

PR #894, opened 2016-06-16
Allow plant to die as population decrease and retranslocation on SimpleLeaf

rcichota:
Working on #65  - Chicory model
1) I've changed Population from a field to property, this allows to check the value being set, if population is zero then clear the SowingData (effectively setting IsAlive to False).
 2) I've changed the methods DMSupply and NSupply on GenericOrgan (and so in SimpleLeaf) so that retranslocation can be set to leaf too. I did this by removing the calculations from those methods and created functions to compute reallocated and retranslocated (N and DM), which are called. For consistency I did it too for NUptake. I am not sure not whether these function should be set to overrideable. Ideally Root and Leaf would also have the retranslocation enables, but I did not do it now. Should they inherit GenericOrgan and so have that capability?

### DIFF
--- a/Models/Plant/Organs/GenericOrgan.cs
+++ b/Models/Plant/Organs/GenericOrgan.cs
@@ -193,25 +193,36 @@ namespace Models.PMF.Organs
                 PotentialDMAllocation = value.Structural + value.Metabolic;
             }
         }
+
         /// <summary>Gets or sets the dm supply.</summary>
         /// <value>The dm supply.</value>
         public override BiomassSupplyType DMSupply
         {
             get
             {
-                double _DMRetranslocationFactor = 0;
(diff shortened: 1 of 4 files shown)


hol353:
Rogerio

Can you make your new methods start with an uppercase e.g. availableNReallocation?

rcichota:
I have fixed the case of new methods (I had them initially as private, hence the lower case)
I also changed a bit the death of plant due to population decline, now if population is near zero it will call the EndCrop method to actually end the plant (so any procedure needed at this stage is covered without need to reduplicate code. That includes transfer of any plant material to surface OM, zeroing variables etc)

rcichota:
And had to make sure the plant is alive before allowing it to die...

PR #897, opened 2016-06-20
Feature/chicory model#65

rcichota:
Working on #65  - Chicory model
1) Set maximum green cover sent to MicoClimate to 1-10^-9, and set the test in microClimate to this value. This is a bit less stringent than before, which was based of float not double type (Talked to Val about this)
2) Added outputs for DetachedWt and DetachedN (the amounts send by plant to either SurfaceOM or SoilFOM). So one can now track mass balance more easily
3) Change the SetFraction method in plant to allow setting the biomass type (live or dead). I've added the extra argument as optional, with default on 'live' (as it seems most organs were using only live (only leaf seem to have explicitly consider both live and dead). Several simulations had to be updated to account for this change. I expect some diffs in this first merge attempt.
4) Organised the return of dead material from roots to soil. I've removed some redundant code by setting a method to do the return. There were three places where the code was doing just that, now they only call the new method (and sorry, I did shuffle the declaration of few variables and methods to put some order on the file and make easy to follow (for me)).

### DIFF
--- a/Models/MicroClimate.cs
+++ b/Models/MicroClimate.cs
@@ -516,9 +516,9 @@ namespace Models
             {
                 CanopyType componentData = Canopies[j];
 
-                if (MathUtilities.FloatsAreEqual(Canopies[j].Canopy.CoverGreen, 1.0, 1E-05))
+                if (MathUtilities.FloatsAreEqual(Canopies[j].Canopy.CoverGreen, 1.0, 1E-09))
                 {
-                    throw new Exception("Unrealistically high cover value in MicroMet i.e. > -.9999");
+                    throw new Exception("Unrealistically high cover value in MicroMet i.e. > 0.999999999");
                 }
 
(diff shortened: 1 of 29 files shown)


hol353:
Not sure about these new events in root.cs:
    public event NitrogenChangedDelegate NitrogenChanged;
    public event WaterChangedDelegate WaterChanged;

I know that's how we did it in APSIM 7.x but I'm trying to move away from events where possible. Why do you need events? Why not just set NO3, NH4 and SW directly?

rcichota:
I found couple of issues related to splitting the FractionToRemove and FractionToResidue into live and dead material. I fixed the comparison where the default value is overwritten by user values, and I made sure the routines in Leaf (actually one in LeafCohort) and BaseOrgan are the same (Before baseOrgan only considered one fraction). I also added outputs for RemoveWt and RemoveN to ease tracking DM changes.

PR #929, opened 2016-07-17
Refactor leaf appearance code

HamishBrownPFR:
working on #930 

@hol353 @byzheng @hut104  @aldoherty  

The code in structure and leaf that controls leaf cohort initiation and appearance was a mess.  It was trying to enable things to be done multiple ways but was to complex and difficult to maintain.  I have refactored the code to address this problem and enable a bug to be fixed.  

The concept is now that the structure class controls both the initiation of new cohorts and their appearance and subsequent growth.  Leaf looks after the cohorts after they are initiated and deals with them when they appear and start growing.  Doing this meant we could get rid of the reciporical setting of variables between leaf and structure.  Now leaf does not set anything in sturcture.

The parameter central to everything is Structure.LeafTipsAppeared which was called MainStemNodeNo but I have renamed it to a more accurate description.  This drives the initiation of new cohorts and the appearance of these cohorts when their tips become visiable.  

At Germination the number of cohorts that are specified in the LeafClass are initialised.
At Emergence any of these cohorts that have a non-zero area are appeared.
After emergence, every time LeafTipsAppeared increases by 1 an new cohort is initiated and a new leaf appears.  This way the number of cohorts says infront of the number of tips by the number of cohorts that were initiated
Cohorts are initiated until cohort number equals final leaf number.  This is the other major change as previously cohorts were initiat at the rate specified by MainStemPrimordiaInitionRate.  Thus a funal leaf number could be set by terminating primordia initiation.  This is no longer possible and a MainStemFinalLeafNumber must be set.  This is most impotant for Maize and I will describe this a bit more below.  I have gone away from using a primordia initation rate because in most cases we were parameterising it to be committed vegetative organs, rathere that primordia, which was causing confustion.  It also made the code complex and hard to maintain.
Leaves continue to appear until final leaf number is reached.

I have checked all models.  this will only affect models using the cohorting leaf model (and the structure class which is a dependency).
The biggest change is for maize.  Previously final leaf number was predicted by the number of cohorts initiated at emergence (6) and the number that were initiated between emergence and floral initiation.  This was determined by the primordia initiation rate (half of the phyllochorn) and the duration of the phases from emergence to FloralInitiation.  Once the crop reached FI cohort initiation stopped and FLN was therefore set.
Now there are two cohorts at emergence and they are initiated at the same rate as the leaf tips appear until they reach the value of MainStemFinalLeafNumber.  Therefore MainStemFinalLeafNumber is modeled more explicitly.  To do this in maize MainStemFinalLeafNumber is calculated as the sum of 6 and the mean plastochron from emergence to terminal spikelet divided by the tt duration from emergence to terminal spikelet.  This provides a prediction of FLN that gradually changes as the crop develops because phyllochron (what plastochron is derived from) changes and the tt duration can also change. 
FLN is used in other functions in the model and the change in the shape of FLN over time has influences some other parameterisations.  These have been refitted and the performance of the model is as good (even slightly better) than it was prior to the change.

For other models FLN was a simple constant and this change has had minor impact.


HamishBrownPFR:
retest this please jenkins

byzheng:
@HamishBrownPFR Thanks for refactoring leaf cohort model. Current model is much simple to understand. I am going through your changes to understand them.

I noticed you add a new parameter `MinimumLeafNumber` with default value 9 for most of genotypes. 

We have conducted a field experiment with pre-vernalization and extended photoperiod in 2014 for 84 wheat cultivars. The final leaf number can be expected as the minimum leaf number. 

The results indicated the minimum leaf number are 7 or 8 for most of genotypes. Only a few cultivars has leaf number 9 (see attachment).

rplot

It will be worth to add our observations into  APSIMX. Can send you the full table if you need.

HamishBrownPFR:
Thanks Bangyou.

The current parameters were just copied across all cultivars to get it running and yet to be set properly.

I agree that the Min leaf number for Australian varieties will be 6-8

PR #996, opened 2016-07-31
Bugfixes

### LINKED ISSUE #413: Wheat model germination code uses incorrect soil water trigger
The plant model code uses soil water.esw as the trigger for germination.  It should be the esw at the sowing depth rather than for the entire profile.

### LINKED ISSUE #435: Errors in Plant organs
1) The default values for fractions of crop residues removed and returned to the soil surface are not correct and they do not appear to be changing with crop type.
e.g. Leaf defaults to zero for both.

2) The populationbaseddmdemand function did not reset it's accumulated thermal time.  We need to check that this is done correctly in all functions.

3) To test for these, we need a series of multiple year simulations to ensure that things are working correctly.

HamishBrownPFR:
resolves #413 
resolves #435


HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
retest this please jenkins

PR #1014, opened 2016-08-03
Phenologychart

### LINKED ISSUE #811: Graphing of Phenological stages
It is really helpful to see when phenology events occurred in interpreting some graphs, particularly for debugging.  There is a string output called Phenlolgy.CurrentStagename which is given a non-"" value on the day phase changes occur.  Is there some way of plotting these names against the date that they occur so we can quickly see when they occured.

hol353:
Resolves #811


hol353:
retest this please jenkins

Issue #985, opened 2016-07-28
AgPasture working across multiple zones


sno036:
FRNL work on urine patches

Hypotheses/research questions
The overall aims are (a) to evaluate differences in urine patch size (and associated edge-effects) and N concentration, to enable identification of the urine-specific effects of pasture species mixture on nitrate leaching losses and (b) to use experimental data and literature reviews to improve simulation methodology in APSIM for representing the impact of root systems on urine patch N uptake and leaching from diverse pastures.

PR #1028, opened 2016-08-08
Multi zone root

hol353:
Working on #985 

This pull request is actually for PMF not AgPasture but it will provide code that may be used in AgPasture to do the same thing


hol353:
retest this please Jenkins

PR #1029, opened 2016-08-09
Handle (ignore) exceptions raised when trying to clean up old tempora…

### LINKED ISSUE #1020: Cannot open two UIs at the same time
When trying to load a second UI one gets the error message as shown below. It seems that the Temp\APSIMX file is locked by the first UI and this prevents the second to open properly.

An unhandled exception of type 'System.IO.IOException' occurred in mscorlib.dll
Additional information: The process cannot access the file 'C:\Users\....\AppData\Local\Temp\ApsimX' because it is being used by another process.

zur003:
…ry files.
Resolves #1020


hol353:
retest this please jenkins

PR #1030, opened 2016-08-09
Remove blank factor specifications under an Experiment.

### LINKED ISSUE #1008: Factor Node blank Specifications tag causes problems
If you go to, for example, Prototypes\Wheat\WheatValidation\NewZealand\Lincoln1992 and go to the Nit experimental node, type something, change view, go back and delete what you have typed and then change view again.  The Factorial will fail because the UI will have written a blank <String\> tag into the <Specifications> tag on the Factor and the Experiment object will be looking for something called "" but won't be able to find it

hol353:
Resolves #1008


hol353:
retest this please jenkins

Issue #575, opened 2015-12-10
Add new Sugarcane Model


ver078:
This is a port of the old Sugar fortran model into the new Apsim Next Generation.

zur003:
Reopening temporarily to resolve issue with file name problem on case-sensitive systems. (SugarCane.png should be Sugarcane.png.)

PR #1033, opened 2016-08-09
Fix auto-doc issues with formating of headings.

hol353:
Working on #575


hol353:
retest this please jenkins

PR #1034, opened 2016-08-09
Fix splitter on main window

### LINKED ISSUE #798: UI: bottom panel overlaps left and main panel
The bottom panel/window (where e.g. "Simulations complete" appears, see image) overlaps instead of resizing the other panel. When much is expanded in the left panel, the last entry is covered by the bottom panel. Same goes for the main panel (I don't know if there is something below "max. number of records")
apsimx_bottom_panel

zur003:
Resolves #798


zur003:
Oops. Issued this pull request after Dean had fixed the problem, but not yet merged into the main repository.

PR #1059, opened 2016-08-17
Restore parent after serialization, as intended.

### LINKED ISSUE #1058: Simulations in experiments are forgetting their parents.
Bug uncovered by @HamishBrownPFR. When simulations are run within experiments, those simulations lose track of their parent, and so subsequently associated reports and graphs fail because they can't locate the data store.

zur003:
Resolves #1058


zur003:
retest this please, jenkins

PR #1064, opened 2016-08-21
Add condition to deal with array variables for initial summary proper…

### LINKED ISSUE #1063: Summary report can't handle array variables
When summary report gets an array value it prints double[] in the report table instead of the array values

HamishBrownPFR:
resolves #1063.  Not pretty, but works well enough for my purposes


HamishBrownPFR:
@hol353 Can you review this one please Dean.  I have strayed into an area I don't normally go to

Issue #1057, opened 2016-08-16
Multiple Porosity Soil Water Model


HamishBrownPFR:
The exact details of the model are still in flux and will continue to be as the code and written and interactions with soil scientists who need to parameterise it continue.  But in general....

Soil will be split into arbitrary layer thicknesses as for current soils but the pore volume will be split into an (also arbitrary) number of compartments representing different sized pores in the soil.  Each pore will have different properties for how soil water flows into and out of them to try and capture variation that is observed in re-wetting following high intensity irrigation events and for soils exhibiting hydrophobic behavior.  The design philosophy of the model is to provide something that fulfills all the current functions of SoilWat with added functionality for macro pore flow but avoiding the mathematical complexity of a Richards equation model.  

Daily procedure will be something like this:
- Extrapolate rainfall and irrigation data out to hourly values to give granularity needed to capture intensity effects
- Residue to intercept precipitation reaching the surface (ideally Micro climate will do this at some stage in the future)
- Water reaching the surface added to surface pond.
- Water redistribution into the profile will work at an internal hourly time step.
- Potential absorption of water into each layer is calculated from the hydraulic conductivity into each pore compartment and the air filled volume in the pore that may hold water.
- Potential percolation below each layer is calculated from Saturated hydraulic conductivity and used to work out how much water may infiltrate to each depth in the profile for the coming infiltration event
- Infiltration is calculated as the minimum of the potential calculated from the profiles current status and the amount of water on the surface that may infiltrate.  Water that does not infiltrate accumulates in the surface pond and will be able to run off (order of runoff in events still to be determined)
- Infiltration is distributed into the soil from the top down, filling each layer to its maximum absorption before putting water to the next layer.  Note: Maximum absorption differs from saturation as low hydraulic conductivity into certain pore compartments may not allow the layer to reach saturation in the duration of the wetting event. 
- Any infiltration that reaches the bottom of the profile is counted as drainage.
- Drainage occurs from the top layer down, using hydraulic conductivity out of each pore compartment to work out how much water may drain and then discharging the minimum of this and the profiles potential percolation below to the next layer down.  This proceeds until the bottom layer where any discharge is counted as drainage.
- Evaporation, Crop uptake, upward and downward diffusion will be calculated after infiltration and drainage at a daily time step (maybe hourly, if necessary).

PR #1076, opened 2016-08-25
Hydraulic properties

### LINKED ISSUE #1075: Externalise SWIM ksat and theta functions
@fai04d @hut104  @hol353 

For #1057 I need to values of ksat and theta and SWIM provides nice functions to calculate these from the standard information in the soil water node.  

I am thinking to move these functions into a separate class.  The code is currently in APSIMX in Models\Soils\SWIM.cs but I am not sure if this code is doing anything in APSIMX at the moment and I am not sure what your plans @fai04d are with connecting the new SWIM to parameters in APSIMX.  Anyone got any thoughts on how this should be done or if anything else is in track here

HamishBrownPFR:
resolves #1075 

New Class (Hydraulic Properties) contains:
- private parameters for the spline to fit moisture release curve to soil data (SAT, DUL, Ll15) and fit hydrulic conductivity using moisture release and kSat.  This h
- private methods to set these parameters when the soil properties are set 
- public methods to return theta and ks for specified suctions, thetas (respectively) and layers
- Links dirrectly to the Water node under soil to get SAT, DUL, Ll15 and Ksat data
- Has two parameters, suction at DUL and ks at DUL which are entered on the grid view for the class

This class must be added onto the soil node to be available to soil water models and the soil water model must call the SetHydraulicProperties() method to initialise the class.  I have done it like this rather than initialing on a particular clock event so if changes are made to a soils properties mid simulation, this can be called to update hydraulic properties data.

### DIFF
--- a/Models/Core/Model.cs
+++ b/Models/Core/Model.cs
@@ -87,6 +87,7 @@ namespace Models.Core
         [XmlElement(typeof(Memo))]
         [XmlElement(typeof(Folder))]
         [XmlElement(typeof(Replacements))]
+        [XmlElement(typeof(Soils.HydraulicProperties))]
         [XmlElement(typeof(Soils.MultiPoreWater))]
         [XmlElement(typeof(Soils.Water))]
         [XmlElement(typeof(Soils.SoilCrop))]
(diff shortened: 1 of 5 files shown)


HamishBrownPFR:
@hol353 @fai04d @hut104 I am going to merge this in because I need it to progress work on #1057.  If anyone takes exception to anything I have done or can suggest some improvement... Open another issue and assign it to me so I can deal with it.

PR #1111, opened 2016-09-02
Report2d

### LINKED ISSUE #1084: Report can't deal with 2d arrays where the dimension of the 2nd array is different to that of the first
For an example see C:\ApsimX\Prototypes\MultiPoreWater\ReportDetail where the PW output is a 2d array[7][8] with the 8th member of the 2nd dimension not currently reporting

When writing the dataframe, the report.cs calls AddColumnsToTable which works out of the value is an array and sets a variable describing the size of the array.  For 2d arrays this is the size of the first dimension (8 in the screen shot below)
image
Then it goes into FlattenValue and steps through the columnIndex itterator up to the value of maximumNumberArrayElements (the size of the first array) and for each array element calls the AddRange method.
image
By some majic I dont understand (if I did I could fix this) it then runs back through the FlattenValue method and seperates off all the the elements of the 2nd dimension of the array into their own column and gives them an appropriate name.
image
The problem is that on the 2nd time through the variable maximumNumberArrayElements has not been changed to represent the size of the 2nd dimension of the array so the method adds blank rows if it is smaller and leaves of the end values if it is larger.

fai04d:
Resolves #1084 
Added check for 2D jagged arrays and code to handle each dimension correctly.


Models/Report/ReportColumn.cs
```diff
@@ -560,7 +565,7 @@ private List<FlattenedValue> FlattenValue(object value, string name, Type type)
             {
                 // Array
                 Array array = value as Array;
-                for (int columnIndex = 0; columnIndex < this.maximumNumberArrayElements; columnIndex++)
+                for (int columnIndex = 0; columnIndex < (name.Contains("(") ? this.maximumNumberJaggedElements : this.maximumNumberArrayElements); columnIndex++)
```
hol353: Justin, my pet hate is an if statement all on one line. Can you change line 568?

PR #1127, opened 2016-09-05
Incorporate speedup changes made in WindowsForms GUI into GtkSharp.

zur003:
Working on #626

### DIFF
--- a/UserInterface/Menus/ContextMenu.cs
+++ b/UserInterface/Menus/ContextMenu.cs
@@ -131,25 +131,23 @@ namespace UserInterface.Presenters
                      ShortcutKey = "F5")]
         public void RunAPSIM(object sender, EventArgs e)
         {
-            if (this.explorerPresenter.ApsimXFile.FileName == null)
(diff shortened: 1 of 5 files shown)


UserInterface/Menus/ContextMenu.cs
```diff
@@ -131,25 +131,23 @@ public void OnMoveDownClick(object sender, EventArgs e)
                      ShortcutKey = "F5")]
         public void RunAPSIM(object sender, EventArgs e)
         {
-            if (this.explorerPresenter.ApsimXFile.FileName == null)
-                this.explorerPresenter.Save();
-            
-            List<string> duplicates = this.explorerPresenter.ApsimXFile.FindDuplicateSimulationNames();
-            if (duplicates.Count > 0)
+            if (this.explorerPresenter.Save())
```
hol353: I don't want to do a save like this. Doing a save is slow and I want to do it asynchronously elsewhere. Looks like this file is the old one before my changes.

PR #1193, opened 2016-09-20
Get rid of duplicate Converter code, and re-establish Version numbers…

### LINKED ISSUE #1166: ApsimX files always query whether they should be saved, even when no change has been made
I think this is a problem with the "Version" information in ApsimX files. A change made about a month ago to Models\Core\Simulations.cs commented out the "Version" attribute, so when new .apsimx file are generated by serialisation, they no longer hold a version number. However, when they are next opened, Converter.ConvertToLatestVersion gets called, sees no version number, so it default to a version of 0, then tries to update this to version 2. This "updated" version then overwrites the existing one, but has the Version attribute in place. When the file is closed, the serialised version doesn't have the version number, this is seen as a difference, and a "save?" query is issued.

The updating logic also seems a bit odd to me, in that there are two versions: one in Models\Core\Converter.cs and another in Models\Core\APSIMFileConverter.cs. They seem to be trying to do roughly the same thing, but the one in APSIMFileConverter.cs thinks the current version is 1, while the one in Converter.cs says the current version is 2. This ought to be tidied up.

zur003:
… for Simulations.
Resolves #1166


zur003:
@hol353 At some point should get the APSIMFileWriter and APSIMFileReader stuff handling Simulations Version correctly, and reinstate their unit test.

Issue #1194, opened 2016-09-20
Refactoring of PMF Leaf Class


hut104:
The leaf class is in need of refactoring to ensure it remains maintainable.

PR #1195, opened 2016-09-20
Clean up leaf class

hut104:
working on #1194


HamishBrownPFR:
I am guessing you are adding that back in because the wheat tests failed.  There is one simulation that sets that but sets it to one.  Would be better to remove the manager that is setting it from the wheat simulation

hut104:
Do you still want it?

HamishBrownPFR:
No
image

PR #1240, opened 2016-09-29
Fix SQL bug, and ensure similar "start of simulation" errors are repo…

### LINKED ISSUE #1239: Database insert fails causing simulation to fail with no messages.
DataStore.cs line 236 
RunQueryWithNoReturnData("INSERT INTO [Simulations] (Name) VALUES (" + sql + ")");
fails with an error message that is not propagated to the user interface. The simulation fails because the errors cause a transaction to remain open. The begin/end queries for transactions should be in try finally blocks.
These errors appear to occur when a new database is created for a simulation that has multiple simulations.
The errors generated are not being reported to the user interface. This needs attention.

### LINKED ISSUE #1329: GTK UI - Too many graphs on a page
The standard UI shows a maximum of 4 graphs per page when a folder of graphs is selected.  I currently get 5.5 graphs down x 2 across.

zur003:
…rted to the user.
Resolves #1239

### DIFF
--- a/ApServer/Program.cs
+++ b/ApServer/Program.cs
@@ -30,7 +30,7 @@ namespace ApServer
 
         public static void Main(string[] args)
         {
-            Server server = new Server();
+            new Server();
         }
 
         public Server()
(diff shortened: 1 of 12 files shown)


zur003:
Retest this please Jenkins
I had the wrong issue number initially.

zur003:
Withdrawing this because I got two different set of commits combined by accident. I'd rather have them pulled in separately.

PR #1256, opened 2016-10-10
Kill updater exe if it is already running.

### LINKED ISSUE #1226: Problems with upgrading APSIM Next Generation
When installing apsim Next Gen I found that
1) I get an error when it uninstalls, reinstalls, and then tries to run the UserInterface.exe - it always says that the exe is missing
2) If I install version 1166 I get the error above, but then when I go back to the updater I see a version 1187 to install.  If I install that version, and go back to upgrade I then see that the latest version to install is 1166 again.  If I then install 1166 it then says I can install 1187..... and so on.

hol353:
Resolves #1226


hol353:
retest this please jenkins

hol353:
retest this please jenkins

hol353:
retest this please jenkins

Issue #1205, opened 2016-09-21
Documenting units and descriptions in ApsimX


sno036:
In the last Reference Panel we discussed the desirability of having some consistency in the units used across the various models in ApsimX.  The first stage (and if it gets too hard it might also be the last stage!) of this is to properly document what the current units are and get understandable descriptions.  Dean has started this by developing a tool that pulls out the current information and using that tool I have tidied up MicroClimate relatively easily.  

As a next step, Russel and I have gone through SoilWater.  The spreadsheet attached contains the current units/descriptions and a draft of the proposed units/descriptions.  Some of these have queries on them.  Can you look through this list and comment, improve or modify as sensible?  Please alert others that might contribute to this discussion.

[SoilWater.xlsx](https://github.com/APSIMInitiative/ApsimX/files/483872/SoilWater.xlsx)

@hol353 @fai04d @hut104 @rcichota @HamishBrownPFR @josharp22 @zur003 @her123

rcichota:
That is a great start, one concern I have is whether we should be making explicit the 'stuff' in the units. That is, should we use units of 'mm water' or simply 'mm'. It does make clear what the units refer to, but makes some units to be too long, but most importantly it adds the need for 'spaces' in the units and that can create lots of trouble when writing scripts to interpret the outputs or in the way some programs parse the text into tables... I favor keeping the units as simple as possible and using the description to make clear what the 'stuff' is.
On a more pedantic note, it would be good to follow a general style on the description. I'd suggest not start a description with an article (use 'Amount of water' instead of 'The amount of water'). Also avoid acronyms when possible and especially avoid references to variables or parameters (use 'Reduction in curve number' instead of 'Reduction in CN'). Any other??  how about writing 'day' instead of 'd'.
I have made a few comments/suggestions on the table.

rcichota:
File:
[Uploading SoilWater.xlsx…]()

HamishBrownPFR:
Good stuff.

My initial thoughts below.
I don't like using dd-mmm as a unit.  It is really day of year and dd-mmm is a format in which it can be represented.
0-1 and 1-100 are ranges for dimensionless variables rather than units.  Should variables have a specified range also?
[...]

PR #1283, opened 2016-10-26
Updated descriptions, units, bounds for SoilWater.cs

mcauliffer:
Working on #1205 
Updated descriptions, units and bounds for SoilWater.cs

### DIFF
--- a/Models/Soils/SoilWater.cs
+++ b/Models/Soils/SoilWater.cs
@@ -55,7 +55,8 @@ namespace Models.Soils
         /// <value>
         /// The summer date.
         /// </value>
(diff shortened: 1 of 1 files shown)


hol353:

hol353:
add to whitelist jenkins

hol353:
When I pull down this pull request onto my computer and then click on 'SoilWater' under a soil in a simulation I get 'Object reference not set ...' errors. When you put [Description] tags on properties, the user interface will try and show that property to the user. I don't think we want the user to see all these properties. Need to discuss with Val.

mcauliffer:
Thanks Dean. I’ll check it out. Sorry.

PR #1286, opened 2016-10-26
Changed map model to honour scope

### LINKED ISSUE #1163: Maps do not seem to use scope
Maps can be added into apsimx files to show where simulations are based.  However, if I put a map into a folder, it includes all simulations in the map, rather than just the simulations in scope.

I do wonder if we may want to document simulations, or folders of simulations, with maps to show where this particular piece of work is from.  That would require the mapping to be aware of scope.

hol353:
Resolves #1163


hol353:
retest this please jenkins

PR #1315, opened 2016-10-31
Add error check to make sure Microclimate module is present

### LINKED ISSUE #1307: Plant doesn't raise an error if microclimate isn't present

HamishBrownPFR:
resolves #1307 

@hol353 I am not sure I have implemented this in the best possible way.  the problem is that Plant had no awareness of the presence is Microclimate because Microclimate just sets properties inside Plant and if it is not there the model runs happily with zeros all the way.  I have added links to microclimate into the leaf organs and and an error to throw if Microclimate is null.  Couldn't work out how to have the leaf discover the presence/absence of Microcliamte without putting this redundant link in.  Better suggestions welcomed


HamishBrownPFR:
retest this please jenkins

hol353:
Not sure about this one. Putting an unneeded link into all three leaf models seems like overkill. I'll need to think about a better way to check for MicroClimate / Leaf comms.

hol353:
Rather than having a link, perhaps we could set a flag in the setter for PotentialEP and then test that flag in OnDoPotentialPlantGrowth. Gets rid of an unnecessary [Link]

HamishBrownPFR:
@hol353 I have implemented it as you suggested (I think).  Can you please review

hol353:
Still not really clean but for now it is OK. The fact that we have to put it into three organs says we have a problem with code duplication across organs.

Issue #1292, opened 2016-10-27
grapevine model


junqi108:
I will add more details soon.

sarahcleary:
Duplicate of - https://github.com/APSIMInitiative/ApsimX/issues/1822

PR #1349, opened 2016-11-08
added the berry model

junqi108:
working on #1292


HamishBrownPFR:
have fixed problem with biomass removal class and will do another pull request

Issue #1254, opened 2016-10-07
Implement new version of SWIM


fainges:
SWIM is currently in a stand alone state for development. This is the first main repo push to ensure the code is somewhere other than my repo. It is code complete (functionality mirrors original FORTRAN code) but does not give the same outputs due to bugs in the multi-soil flux matrices and is not integrated with APSIM yet.

Some unit tests have been disabled until these bugs have been sorted out.
This push also includes all of the original FORTRAN code from Peter Ross.

JJguri:
I have taken the SWIM3 module from the Hudson simulation in the source code. I am comparing the same APSIMX potato simulation using Soilwat and SWIM3 for one specific experiment. I got LAI under-estimations using the latter. Appear that the sensibility of LL, DUL and KL are really high using SWIM3. Do you know if these parameters should be the same using SWIM3 or Soilwat in the water submodel? I already checked Huth, et al., 2012 and I could not find information about the setting of these parameters in both soil water modules. Find attached both simulations
[SWIM3 vs Soilwat.zip](https://github.com/APSIMInitiative/ApsimX/files/3141170/SWIM3.vs.Soilwat.zip).

hol353:
Link to project plan: https://github.com/orgs/APSIMInitiative/projects/6

sno036:
@APSIMInitiative/reference-panel 

Attached is the presentation for the review meeting (if you want a preview)
[SWIM in NextGen Review.pptx](https://github.com/APSIMInitiative/ApsimX/files/9719586/SWIM.in.NextGen.Review.pptx)

PR #1360, opened 2016-11-11
Swim frame

fai04d:
Working on #1254 
SWIM conversion complete. Still need to integrate with APSIM.


hol353:
Hmmm this looks suspect to me. It says you've modified 159 files. When I look at the diffs, there are a lot of non SWIM changes. Looks like a merge gone wrong. Can you please check.

fai04d:
Blasted git... I thought I fixed all that... I'll have another look.

PR #1374, opened 2016-11-20
Font size

### LINKED ISSUE #1335: GTK UI - Font sizes are too small in general
Fonts as smaller in treeview, property view, and just about everywhere.  Can we get them back to the standard UI size?

hol353:
Resolves #1335


zur003:
Tread carefully. Loading the .gtkrc file from MainView may create some cross-platform problems. We will probably want different .gtkrc settings on each platform.

PR #1392, opened 2016-11-27
Change default GUI from "UserInterface" to "ApsimNG".

### LINKED ISSUE #1387: When using the Updater, the "old" user interface is invoked instead of the new one.

zur003:
Resolves #1387

### DIFF
--- a/Updater/Program.cs
+++ b/Updater/Program.cs
@@ -40,7 +40,7 @@ namespace Updater
         /// <param name="newInstallDirectory"></param>
         private static void DoUpate(string uninstallDirectory, string newInstallDirectory)
         {
-            WaitForProcess("UserInterface");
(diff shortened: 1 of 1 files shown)


hut104:
Are we sure we want to do this?  There seem to be more bugs in this version than the original.  I have users using the old one because the new one doesn't run reliably.
Perhaps the bugs have been fixed.

zur003:
Is it a case of "doesn't run reliably" or "doesn't look the way I expect"? If there are bugs that go beyond the merely cosmetic, they certainly need to be fixed, and putting this version on the pedestal provides a bit of motivation for those fixes to be made.

hut104:
We found that we couldn't create graphs and other things.  It just wasn't usable.  However, Dean may have fixed these things because we put in defect reports.  It's worth checking.

PR #1404, opened 2016-12-08
Trap errors when exporting DataStore to csv

### LINKED ISSUE #1395: APSIM dies if you export to csv files when the csv file is already open in excel
It should throw an exception but not die

hol353:
Resolves #1395


hol353:
retest this please jenkins

PR #1438, opened 2016-12-20
Allow the user to abort all simulation runs

zur003:
Resolves #1437


zur003:
Withdrawn until supporting function is added to APSIM.Shared

PR #1439, opened 2016-12-23
Allow to user to stop any running simulations

### LINKED ISSUE #1436: Need something in the GUI to allow the user to stop simulations
I don't think we currently have any mechanism to allow the user to abort running a set of simulations. This could be useful in situations where the user realises that a large set of factors is going to take a day or two to complete.
We also need better feedback about the progress of execution of multiple simulations. There is a progress bar, but the associated text could indicate the number of completed simulations as well as the total number (e.g., "500 out of 800 completed").

zur003:
Resolves #1436


zur003:
Withheld until APSIM.Shared is updated

Issue #1453, opened 2017-01-22
Handling and reporting Wt and N have few problems


rcichota:
I am adding this issue with a somewhat generic description to enable fixing a few (relatively small) issues on how the variables Wt and N and the Biomass class are handled by the organs. A few points I noticed to date:
1) There is no test for Wt being zero when computing Nconc for organs;
2) There is no test when removing biomass to prevent it to go negative. Wt and N can go below zero on RemoveBiomass if fractions given by user are not balanced (there is some test when gathering default fractions, this could be better organized to avoid duplicate and test the fractions);
3) Message given when removing biomass from GenericOrgan is wrong (fraction are doubled);
4) Reported values for Removed and Detached biomass are wrong, the first is always zero and the second is cumulative over time (seems to be due to passing variables as values instead of as reference, and not zeroing them at the beginning of the day - in GenericOrgan);
5) Missing outputs for biomass allocated (actual growth) and senesced in any given day (these complement Detached and Removed providing information about biomass changes);
6) Test for negative DM or N allocation should be better placed, supply should be tested when calculated while allocations should be in the arbitrator;
7) There is no provision in Biomass.cs for Wt and N value to be very small. This can eventually lead to absurd Nconc values (>100%) as Wt and N decrease. Which is the case when an organ dies, but the plant keeps growing (e.g. flowers after a reproductive phonological phase ends). This seem to cause no errors in the model results (so far), but reported values are wrong/misleading;
8) Initial Wt is set at sowing for root, but at emergence for other organs;
9) There is no test in Biomass.cs for Wt or N to go below zero, also the class is not aware of minimum or maximum Nconc values given for organs, so there is no way to test whether any change has violated those limits (this may not be important if all processes are done carefully, but having a test in Biomass would reduce the chances of errors, and have a test in only one place);

ric394:
This very old. Closing. If this is still an issue, please feel free to comment.

PR #1454, opened 2017-01-22
Feature/chicory model#65

rcichota:
Working on #1453 
- Added tests for Wt = 0 when computing Nconc
- Added test for removal fractions adding up to > 1.0
- Fixed message on RemoveBiomass for GenericOrgan

### DIFF
--- a/Models/Plant/Biomass.cs
+++ b/Models/Plant/Biomass.cs
@@ -229,12 +229,12 @@ namespace Models.PMF
         /// <summary>Clears this instance.</summary>
         virtual public void Clear()
         {
-            _StructuralWt = 0;
-            _NonStructuralWt = 0;
-            _MetabolicWt = 0;
-            _StructuralN = 0;
-            _NonStructuralN = 0;
-            _MetabolicN = 0;
+            _StructuralWt = 0.0;
+            _NonStructuralWt = 0.0;
+            _MetabolicWt = 0.0;
+            _StructuralN = 0.0;
+            _NonStructuralN = 0.0;
+            _MetabolicN = 0.0;
         }
         /// <summary>Adds the specified a.</summary>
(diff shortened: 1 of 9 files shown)


rcichota:
- Fixed reporting of Detached and Removed biomass
- Added outputs for Senesced and Allocated biomass
- Added a BiomassTolleranceValue to check for small values for Wt and N

rcichota:
I have noted few issues with the way the variables 'Wt' and 'N' are handled and made few changes to address some of them. These imply adding checks for negative values or division by zero and tidying up the report of variables for biomass flow (Allocated, Senesced, Detached, and Removed). I guess most is pretty harmless, but there are a couple of things that I want to make you aware and check that I did not misunderstood anything:
- As I add some tests for negative values, I notices that retranslocation supply could potentially be negative (all small it seems), I have set up a max function to prevent this. Please advise if negative values should be allowed.
- I have set up a variable (BiomassTolleranceValue = 10E-10) to test the values of Wt and N against. Of some importance here, this is used when senescence and detachment are calculated, if the remaining value is below that threshold than all biomass senesces or detaches. This was needed because otherwise the biomass of an organ never really becomes zero. Very small values were causing spurious outputs for Nconc (>100%) in Chicory after an organ dies (inflorescence especially). This should not interfere with annual crops, but it is certainly important for perennials.
Finally I have done this changes mostly in GenericOrgan (I added the outputs also to BaseOrgan and Root). Leaf, PerennialLeaf and the organs derived from BaseOrgan have some code I could not follow straight-away or I did not have time time go through (some do not use things like retranslocation, so not relevant). Do we want to leave the issue open till the other organs are updated? Any other issues related to this I did not picked up?
@hol353 @HamishBrownPFR @hut104 @fai04d

PR #1466, opened 2017-01-29
Consolidating Supply and Demand functions

rcichota:
Working on #1465  - Testing, not to be merged yet
Consolidating DM and N supply function for GenericOrgan and Root (SimpleLeaf inherits GenericOrgan)
- Added DMReallocationFactor in GenericOrgan and a function to used it;
- Added reallocation and retranslocation for N and DM in Root;


rcichota:
I guess this is now ready to merge.
I tried to clean up/consolidate the functions for DM and N supply and demand. I have done this mostly on GenericOrgan (and by extention on Simpleleaf), but also had a go in Root.
I made sure they compute (or at least have dummy calculations) for all the fractions in which biomass is divided (Structural, NonStructural and Metabolic). This meant I needed to add some basic parameters (CriticalNConc, StructuralFraction, etc.) to these organs (I update also their descriptions). I have also moved the calculations for Supply and Demand out of properties and into methods. This should make easy to change them (I think demand is not quite right, but need to test more and get feedback from Hamish and others). Parallel to that, in IArbitrator these properties (DMDemand, DMSupply, Ndemand, NSupply) are defined as {get; set;}, but I don't think they are meant to be settable...
I had to change FodderBeet as the developers had set StructuralFration in the UI for Root, although it was never defined in the code (now it is). I will talk to Edmar or Edith about this.
@hol353

hut104:
Dean will send an email as a follow up to this.

Issue #1464, opened 2017-01-29
Need to consolidate Supply and Demand functions in organs


rcichota:
The functions computing supply and demand for both DM and N need consolidating, each organ has different versions (mostly doing exactly the same by coded a bit differently), some use different variable with same meaning, and a few may have errors.
- The main issue with Supply is not making explicit the components that are zero, and lacking uniformity;
- For Demand, non-explicit zeros and non-uniformity are also there, but goes beyond with the use of variables that could lead to errors in the future (this is the case of using PotentialDMAllocation to compute StructuralNDemand for GenericOrgan. PotentialDMAllocation equal PotentialStructuralDMAllocation + PotentialMetabolicDMAllocation, with the second being zero - no this is not an error currently, but might cause issues in the future. we should be using simply PotentialStructuralDMAllocation).
The lack of uniformity is worse for the case of N; the use of local variables in different ways and the placing of brackets seem to have introduced errors comparing the logic between GenericOrgan, SimpleLeaf, and root. The boiled-down versions below (note brackets placements in the first and second and the doubling of StructuralDemand in the third):
NonStructuralDemand = Math.Max(0.0, MaximumNConc.Value * (Live.Wt + PotentialStructuralDMAllocation + PotentialMetabolicDMAllocation) - Live.N - StructuralDemand);
NonStructuralDemand = Math.Max(0.0, MaximumNConc.Value * (Live.Wt + PotentialStructuralDMAllocation + PotentialMetabolicDMAllocation) - Live.N) - StructuralDemand;
NonStructuralDemand = Math.Max(0.0, MaximumNConc.Value * (Live.Wt + PotentialStructuralDMAllocation + PotentialMetabolicDMAllocation) - Live.N - StructuralDemand - StructuralDemand)

ric394:
This very old. Closing. If this is still an issue, please feel free to comment.

PR #1493, opened 2017-02-12
Allow plant to use up any DM supply

rcichota:
Working on #1464 
Small change on the test made when DM is allocated by organ arbitrator. Original test was made against TotalFixationSupply (which is from photosynthesis only), however the test fails (wrongly) if DM reallocation or retranslocation are enabled. The test should be against TotalPlantSupply (from all sources)


rcichota:
I have solved the conflict in organ arbitrator, I like that arbitrator is now separated.
Just reinterating: the change I made was simply to modify the tests made when allocation of N and DM are completed. The test did not included N or DM reallocated or retranslocated, so when I used these in chicory, the simulation crashed. The test is now against TotalPlantSupply.
@hol353 @hut104

PR #1550, opened 2017-03-09
Restore the variable AreaLargestLeaf

### LINKED ISSUE #1549: Restore variable AreaLargestLeaf

byzheng:
resolves  #1549


hol353:
Why is this needed Bangyou?

byzheng:
I need to access the largest leaf area to calculate the max leaf area after terminal spikelet

PR #1552, opened 2017-03-12
Calculate the daily Photothermal Quotient with moving sum

### LINKED ISSUE #1551: Calculate the daily Photothermal Quotient with moving sum
* Export a new variable of daily Photothermal Quotient (ptq)  in the weather model
* Add a new function MovingSum to calculate accumulated values in a giving window (The default value is set as zero).

byzheng:
resolves #1551

### DIFF
--- a/Models/Weather/Weather.cs
+++ b/Models/Weather/Weather.cs
@@ -425,6 +425,19 @@ namespace Models
         }
 
         /// <summary>
+        /// Gets the daily Photothermal Quotient (ptq).
+        /// </summary>
+        public double ptq
+        {
+            get
+            {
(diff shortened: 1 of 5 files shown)


Models/Weather/Weather.cs
```diff
@@ -425,6 +425,19 @@ public double Amp
         }
 
         /// <summary>
+        /// Gets the daily Photothermal Quotient (ptq).
+        /// </summary>
+        public double ptq
```
hol353: Bangyou, can you name this PhotothermalQuotient? We are trying to move away from acronyms.

PR #1561, opened 2017-03-15
Implemented mapping of LL, KL and XF on SoilCrop.cs

### LINKED ISSUE #1558: LayerStructure object not working properly with SoilWat
When using a LayerStructure object to set the fundamental layering of the soil model, The CropSoil Node gets its depth[] from the LayerStructure rather than from SoilWater so they don't align.

HamishBrownPFR:
resolves #1558 , Hopefully


HamishBrownPFR:
@hol353 Hi Dean.  I have made some changes to get SoilWat working with a LayerStructure node so we can use that to specify the fundamental layering of the water model and specify parameter values with different layering's that map onto this.  Most of the code needed was there for SoilWat and I did a bit of work last week to get Weirdo working with the LayerStructure in the same way.

Can you please review these changes and merge if you are happy.  I have someone waiting for this to be implemented so please get back to me it you think it will be a while before you can review it.

Thanks,

Hamish

Models/AgPasture/AgPasture.PastureSpecies.cs
```diff
@@ -2875,7 +2875,7 @@ public double[] LL
             get
             {
                 SoilCrop soilInfo = (SoilCrop)mySoil.Crop(Name);
-                return soilInfo.LL;
+                return soilInfo.LLMapped;
             }
```
hol353: I would prefer if crops didn't call the LLMapped properties. Instead there is some commented code in Soil.cs (line 467) that has methods for LL, KL and XF that calls the mapping code. This way crops can call:

return soil.LL(Name);

It also means there won't be any knowledge of mapping in all the crop models. This shouldn't be hard to get going.

HamishBrownPFR:
OK, I will take a look at that

HamishBrownPFR:
@hol353 I have re-implemented it as you suggested.  Merge if you are happy with this or get back to me if there is anything else you don't like.

Issue #1540, opened 2017-03-06
Creation of New SoilNutrient Model


hut104:
A new model will be created with the aim to replace the soilN model in the longer term.  It will provide a capacity to model C and N cycling, initially, with P and K cycling to be added later.  It will be developed using approaches developed for PMF.

jbrider:
refer to #3522

hut104:
@sarahcleary This now ready for review. I suggest Dr Valerie Snow (@sno036 ) as a potential reviewer.

sarahcleary:
@APSIMInitiative/reference-panel - please note this for discussion in next meeting.  If you have any comments or recommendations for reviewers, please provide these here prior to this meeting.

sarahcleary:
@hut104 - as part of the review process, can you please provide a few sentences which will be used as part of the communications (news article/ResearchGate update) when ready for release?

sarahcleary:
@hut104 - as per request above - can you please provide a few sentences which will be used as part of the communications (news article/ResearchGate update)?  Once I've received this, we can commence the review process. The review will (or likely to) be conducted by @sno036 and @Keith-Pembleton

sarahcleary:
@hol353 - please note action from last RP meeting - to discuss 'news release" paragraph with @hut104

hut104:
The Nutrient model replaces the Soil Nitrogen model within APSIM.  The model is developed using similar infrastructure to that used in the crop models so that soil nutrient pools and fluxes can be described using the APSIM user interface.  This should allow further development of the model by non-programmers and automation of soil nutrient model documentation.  
The initial implementation reproduces the science within the APSIM Soil Nitrogen model.  Enhancements, such as the incorporation of published approaches for BioChar, can be added in a relatively simple manner in later releases.
The model has also been designed to assist the addition of other soil nutrients to the same modelling framework.  For example, P cycling could be added to reproduce the SoilP model functionality with APSIM Classic.

sarahcleary:
Thanks @hut104.  
@Keith-Pembleton and @sno036 - as discussed, this model is now ready for review.  As a trial, we are using the proposed process - solely via GitHub.  Please refer to the Reviewer's Instructions and commence review: http://www.apsim.info/Documentation/TechnicalandDevelopment/ReviewersInstructions.aspx
Please let me know if these Instructions need to be amended in anyway.

sno036:
[...]

PR #1571, opened 2017-03-19
Change how solutes are advertised and removed soil temp from SoilNitr…

hol353:
…ogen.

Working on #1540


hol353:
retest this please jenkins

PR #1611, opened 2017-03-29
Use a better mechanism to fix drag-and-drop on OS X.

zur003:
Working on #1606


hol353:
Eric this stalled on Jenkins in the Installation project. I clicked the red 'x' to stop the Installation. Don't know why there is a problem.

zur003:
Possibly because the "Working on" issue was already closed. I wanted to revisit the issue rather than opening a new one, but wasn't quite sure whether this approach would work. Looks like it didn't.

Issue #1679, opened 2017-05-19
New way to simulate dynamic of stem development


byzheng:
using the apex number to simulate the dynamic of stem development.

PR #1727, opened 2017-06-13
Tiller feature

fai04d:
Working on #1679 

Some software engineering work to get Bangyou's changes to fit our workflow. Created an interface to handle multiple different apex models and added required XML changes to the file converter. Removed optional models.


fai04d:
Merge went the wrong way. I'll open another PR.

PR #1754, opened 2017-06-22
Convert.ToDouble needs System.Globalization.CultureInfo.InvariantCulture

oseledets:
Otherwise on non-standard locales (in my case, Russian) ApsimX can not open almost any model with "Number in wrong format" error.

Working on #1755


hol353:

hol353:
add to whitelist jenkins

hol353:
@zur003 Eric, can you pull down this pull request and run it? It looks OK to me but just want to check it at runtime. I'd do it but I'm travelling until end of next week. If Ok, let me know and I'll merge.

@oseledets Well done and thanks for submitting a pull request! I'm impressed!

hol353:
retest this please Jenkins

oseledets:
Will redo that - I added System.Xml.Xmlserializer to the dependencies since for some reason it did not work on my mac.

zur003:
The changes to Convert.ToDouble look fine, although I suppose it raises the question of whether we want to use the "Invariant" culture for everything, or only in the internal serializations. That is, should the GUIs be invariant, or culturally aware? Certainly we want to use a consistent representation in the .apsimx files.

This patch does need to have the changes affecting the generation of Models.XMLSerializers.dll removed. We need to have the serializer dll re-generated during builds, to ensure it reflects the current state of Models.exe. Unfortunately, the "sgen" stuff in Models.csproj doesn't play on Mono-based systems, and as a result, attempts to build everything on OS X and Linux don't work until this bit is removed. I should try to come up with a better way to do this. Mono does have an "sgen" utility, but it's not quite up to the task. The Microsoft "sgen" is a .net application and does run correctly under Mono, so it's possible to copy it from a Windows system and run it, although I'm not sure about the strict legality of using it that way.

hol353:
@zur003 Where are we at with this pull request?

hol353:
Opened a new pull request for this - #2151

Issue #860, opened 2016-05-26
Compile APSIMX under ubuntu


byzheng:
Following the instruction that @zur003 provided, I can compile `Models.exe` using Ubuntu 14.04 now. However I make a few changes of source codes including
- Copy all nuget packages from windows to ubuntu 
- Remove the target `AfterBuild` in the Models.csproj
- On line 59, change “bin” to “Bin” (otherwise the build will go into the wrong directory) in the Models.csproj
- On lines 45 and 63, change the setting for TreatWarningsAsError from “true” to “false” in the Models.csproj
- change the setting for TreatWarningsAsError from “true” to “false” in the ApServer.csproj

zur003:
The TreatWarningsAsError problem is interesting. The Mono C# compiler is reporting a number of warnings that the MS compiler is not (mostly dealing with unused variables). I'm not sure why the MS compiler is not reporting these; it probably should be.

zur003:
As mentioned above, the Mono C# compiler issues warnings for code that the MS compiler does not. In the interest of making compilation under Mono slightly easier, I will submit a series of pull requests which intended to reduce, if not fully eliminate, these warnings.

hol353:
@zur003 Can we close this one?

zur003:
Yes, close it.

PR #1756, opened 2017-06-23
Enable generation of XML serializer dll when building on Mono-based s…

zur003:
…ystems. Working on #860


zur003:
Retest this please, Jenkins

Issue #1775, opened 2017-07-06
GTK UI - Grid control focus


hut104:
Some of the grid controls are difficult to use because of issues with focus.

For example, when editing the XY pairs for a linear interpolation function
1) if the users clicks inside a cell it selects the row but not the number for editing.  It should select the number ready for editing. (ie put the cell in edit mode)
2) When the user changes a number and hits enter the control loses focus.

As a result, the users has to click twice between any edits.
It would be good if it worked like excel (ie edit and select the next cell for editing).

PR #1789, opened 2017-07-11
Try to make editing of cells a bit more Excel-like.

zur003:
Working on #1775 

This isn't perfect, but should make data entry a bit easier. The new behaviour is that if a user is editing a cell and presses either enter or tab, the current values are saved and the editing focus moves to the next editable cell. Using enter moves the focus down the column (shift-enter moves it up). Using tab moves the focus along the row (and although I'd like shift-tab to move left along the row, shift-tab turns out to be surprisingly hard to capture).

### DIFF
--- a/ApsimNG/Views/GridView.cs
+++ b/ApsimNG/Views/GridView.cs
@@ -80,6 +80,11 @@ namespace UserInterface.Views
         /// </summary>
         private string defaultNumericFormat = "F2";
 
+        /// <summary>
+        /// Flag to keep track of whether a cursor move was initiated internally
+        /// </summary>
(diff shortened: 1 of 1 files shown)


zur003:
How did a build number of -1 get in there? Retest this please Jenkins

zur003:
@hol353 - Not sure what's going on here with Jenkins. It appears that the issue number is somehow going is as -1 rather than 1775. Perhaps because I had the magic words that sound like "working onn" and "re-sovles" embedded elsewhere in the comment? I'm going to close this and try a new pull request (with the extra magic words omitted.)

Issue #1798, opened 2017-07-19
Urea hydrolysis is not using the correct value of OC


rcichota:
The calculation of urea hydrolysis in SoilNitrogen uses the value of OC , but currently this value is gathered at the beginning of the simulation and never updated. This is not an issue for short term simulations but it may cause considerable deviations for long runs.
Also there is a check for minimum urea content, if urea is lower than this, then all is hydrolysed.  The value is a bit too high and probably should be in ppm...

PR #1814, opened 2017-07-21
Enable hydrolysis to used actual OC

rcichota:
Working on #1798 
Changed how OC is defined in the calculation of urea hydrolysis, previously it used resetOC (which is the OC at initialisation), now it used current OC values

### DIFF
--- a/Models/Soils/SoilNitrogen.CNPatch.Processes.cs
+++ b/Models/Soils/SoilNitrogen.CNPatch.Processes.cs
@@ -608,7 +608,6 @@ namespace Models.Soils
                 {
                     // potential fraction of urea being hydrolysed
                     double totalC = (hum_c[layer] + biom_c[layer]) * g.convFactor[layer] / 10000;  // (100/1000000) = convert to ppm and then to %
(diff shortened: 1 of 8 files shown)


rcichota:
Resolves #1798 
I have changed the hydrolysis function so that it uses actual C instead of resetOC, also I change the test for minimum urea that is automatically hydrolised. The value in there was set by me a couple of years back and it may be too large for small layers.  The change in OC caused small diffs in wheat and SugarCane, while the change in the test resulted in small diffs in Plantain and SugarCane (the later basically coutered the diffs caused by change in OC).
As I got some time, I went through and upated a few variables so that they are consistent in ApsimX and Classic...

rcichota:
A this issue was still open I have added to it the few minor changes in code organisation/documentation so that both AspimX and classic are alike. 
I also removed a few conversion to float, these were needed in classic as the infrastructure used floats, but only create rounding errors in ApsimX as doubles are used. This created minor diffs in Plantain (improved) and SugarCane (one variable has R2 slightly worse than before), which I think are not cause for concern.

Issue #1799, opened 2017-07-19
SoilNitrogen is not updating soil water parameters


rcichota:
SoilNitrogen uses soil water limits (LL15, DUL, and SAT) plus current water content to compute the moisture factor for its processes. However these are gathered only at the beginning of the simulation, if the parameters are changed (via manager script, for example, to simulate soil compaction) the changes are not captured by SoilNitrogen.  This occurs in one set of simulations in WheatValidation.apsimx (Lincoln2015).

rcichota:
I just had a better look and noticed that the simulations that crash when I set links to the soil water parameters in SoilNitrogen are probably set up by you guys.  I run the simulation with and without the links and the differences are not large but not negligible (2-3% reduction in grain.Wt).  Hope this doesn't cause much grief, or that it even goes in the right direction. Let me know if you have concerns about me making the changes in SoilNitrogen effective
@HamishBrownPFR @RobZys

PR #1815, opened 2017-07-23
Change sw_dep for a direct link to Soil.SoilWater.SWmm

rcichota:
Working on #1799 
Step 01, change only sw_dep


rcichota:
Resolves #1799 
Changed the variables related to soil water (sw, SAT, DUL, and LL15) into links to Soil.SoilWater.
There is only one set of simulations in WheatValidation (Lincoln2015) that had significant diffs, as these were the only ones where soil properties changed during the simulation.  I have talked to Hamish (he and Rob have added those simulations) and he said that we could go on and accept the changes, that simulation set is still a work in progress and he's going to re-check them after this change...

PR #1834, opened 2017-08-02
Root demand function

### LINKED ISSUE #1832: Root demand function
For the moment, for calculating the root dry matter demand, it must include the partition fraction. It is not consistent with other organs and also provides less flexibility.

junqi108:
resolves #1832

### DIFF
--- a/Models/Plant/Organs/Root.cs
+++ b/Models/Plant/Organs/Root.cs
@@ -54,13 +54,14 @@ namespace Models.PMF.Organs
     public class Root : BaseOrgan, IWaterNitrogenUptake
     {
         #region Links
-        /// <summary>The arbitrator</summary>
-        [Link]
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
retest this please jenkins

hol353:
This is going to need a bit of work. Rather than change and commit a lot of .apsimx files, we now modify the change tool to change the XML automatically. This helps maintain compatibility for other users. So, can you only commit the source changes and do a push. Jenkins will run red which is OK for now. I'll then pull down your branch, modify the change tool for you and do a push. This will then run green and I'll merge. @HamishBrownPFR can help.

junqi108:
Yeah, thanks very much, Dean. I will submit a new pull request later.
Best regards.
Junqi

PR #1860, opened 2017-08-14
Wheat

HamishBrownPFR:
Working on #453.  @hut104  can you please review this before Dean merges it.  I have done some work on the Lincoln2015 simulations because it is a good water stress dataset.  Tidied up the data and the simulation set up and made some changes to the model to make it more sensitive to water stress so it predicted this data set better.  It appears the overall stats have improved and the results for APS26 appear to have improved also.  However, it would be good to get your perspective on this.


HamishBrownPFR:
@hut104 
Hi Neil, can you have a look at what I have done here

PR #1869, opened 2017-08-15
Test the beta function for thermal time

byzheng:
Test the thermal time equation of EW (Beta function) to see the results of wheat validation.

working on #1868


byzheng:
No need to test this feature

Issue #1413, opened 2016-12-12
Oat model creation


RobZys:
We require an oat model

PR #1936, opened 2017-08-25
Get Allens progress into the build

HamishBrownPFR:
working on #1413


HamishBrownPFR:
Will put this up under a more descriptive name because it is more than the original name inferred now

PR #1985, opened 2017-09-04
Leaf size age rank

### LINKED ISSUE #1984: Add new tillering model.
Add new tillering model from Wang, et al. New model is currently only active in LeafApex.apsimx.

fai04d:
Resolves #1984 
Implement new tillering model.


fai04d:
This PR changes a number of outputs and those changes will need to be reviewed by the model developers.

HamishBrownPFR:
Hi @fai04d and @hol353.  I am wondering if it would it be better to implement the apex tillering model into an alternative leaf class.  I have had a quick look at this pull request and it appears to be forcing parameters into the leaf class which are going to be redundant in a lot of models and the leaf class is already quite cluttered with parameters.  It seems to me the approach is sufficiently different to justify implementation in a separate class help reduce confusion when parameterising models.

fai04d:
Enli wants the apex model to be the default model for Apsimx which I think is why we didn't create a new class.

fai04d:
This PR has turned into a mess of conflicting files. I'm going to cancel this one and start on a clean branch.

PR #1995, opened 2017-09-08
Dm conversion efficiency

### LINKED ISSUE #1991: Return value of DMConversionEfficiency equal to 1
We have changed to DMConversionEfficiency into other values, but the DMConversionEfficiency always equal to 1 when I debug. 

Seems this line code force DMConversionEfficiency  to 1:

virtual public double DMConversionEfficiency { get { return 1; } set { } }

https://github.com/APSIMInitiative/ApsimX/blob/1b58df65adf9f680a50cce5f259afcf1fe4db096/Models/Plant/Organs/BaseOrgan.cs#L56

fai04d:
Resolves #1991 
DMConversionEfficiency is now editable in Generic and Reproductive organs.

### DIFF
--- a/Examples/Chicory.apsimx
+++ b/Examples/Chicory.apsimx
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Simulations xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="8">
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Simulations xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="14">
   <Name>Simulations</Name>
   <DataStore>
     <Name>DataStore</Name>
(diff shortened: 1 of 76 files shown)


fai04d:
Retest this please Jenkins.

fai04d:
Retest this please Jenkins.

hol353:
Justin 

Don't commit all the .apsimx files. You should let the APSIMFileConverter.cs do this.

Also, why are you renaming DMConversionEfficiency to DMConversionEfficiencyFunction? We're tying to move the other way i.e. get rid of the 'Function' suffix on links.

HamishBrownPFR:
Hi @fai04d and @hol353.  I am wondering if it would it be better to implement the apex tillering model into an alternative leaf class.  I have had a quick look at this pull request and it appears to be forcing parameters into the leaf class which are going to be redundant in a lot of models and the leaf class is already quite cluttered with parameters.  It seems to me the approach is sufficiently different to justify implementation in a separate class help reduce confusion when parameterising models.

fai04d:
@HamishBrownPFR suggested it in https://github.com/APSIMInitiative/ApsimX/issues/1991:
1 rename the DMConversionEfficiency function in Generic organ (and all xml files, use the change tool) to DMConversionEfficiencyFunction

I'll revert the apsimx files.

Enli wants the apex model to be the default model for Apsimx which I think is why we didn't create a new class.

Issue #1999, opened 2017-09-11
Reduce the amount of memory that APSIM uses


hol353:
There are two parts to this.

1. Considerable memory is being used by keeping all outputs in memory until all simulations have completed. Then they get written to the .db file.
2. All simulations are created up front. This consumes a lot of memory for very large jobs.

josharp22:
Hi Dean
Just tested your recent commit on a big simulation. 3675 of the 3675 simulations in a large factorial ran without any problems when saved locally (crashed at 2151 when saved on a network drive). Thank you very much for fixing this, will make a huge difference :-)
Cheers
Jo

zur003:
@hol353 - found a small bug arising from your database changes. If you have PredictedObserved components in your model, the very first time you select one of them in the GUI, you get the error message "no such table: _Simulations". Arises at line 570 in DataStore.cs, I think. Can't quite work out what's going on here.

hol353:
I'll look into it.

zur003:
Looks like a threading issue. FileName is null one moment and non-null the next...

hol353:
Hmmm How can I reliably reproduce the problem? Seems OK to me.

zur003:
Open Prototypes\Potato\Potato.apsimx. Open the DataStore node, select NodeAppearance. I then see the message appear at the bottom of screen. If running in the IDE, I can set a breakpoint on line 74. This line should be reached only in FileName is null, but when things stop at the breakpoint, FileName has been set (and the error message does not appear). That sounds like a threading issue, but I don't see any extra threads, unless maybe Gtk is doing something weird. Haven't tried this with the Windows.Forms GUI.

hol353:
I still can't see what you're seeing. If potato.db exists, I get no error at all. If potato.db does not exist I get a series of exceptions which are trapped and so get no error at bottom of screen in GUI. Are you sure it is line 74 of DataStore.cs?  Have you pulled from APSIM.Shared? Have you run potato to see if that fixes the problem?

zur003:
Potato.db exists, but does not contain a _Simulations table (but it does have a Simulations table). Is the problem that it's an "old" database? I also see the problem if I open the "Standard toolbox", and select PredictedObserved (under Data store and analysis models).

I see the same issue when using the Windows.Forms GUI.

I also see weirdness in the debugger that I don't understand. The FileName of the DataStore is being set, but I can't work out where, so perhaps it's during deserialisation?

[...]

PR #2002, opened 2017-09-11
New data store

hol353:
Working on #1999


hol353:
retest this please jenkins

HamishBrownPFR:
@hol353  Hi Dean.  Noting your 3rd to last commit where you have changed the data base to write a whole simulation at a time.  Does this mean the changes won't fix the memory problems Jo was having, where multiple zones were all within one very large simulation?

hol353:
It will still help a lot and uses a LOT less memory. The problem now is that the 'experimental run' menu item doesn't work yet. I'll work on that next week. Jo will have a wait for that I suspect. I wanted to get this in and merged because I changed so many files.

Issue #1352, opened 2016-11-08
Validate Plant models OrganArbitrator


HamishBrownPFR:
The biomass partitioning (N and Wt) is handeled by the OrganArbitrator and it needs a detailed validation.  There are two detailed N uptake datasets for Wheat and Maize where irrigation and N treatments have been applied in factorial.  These will be used to check the performance of the OrganArbitrator.

HamishBrownPFR:
I will be working on this over the next while.  Will involve making sure the N x I treatments in the Maize and Wheat validation sets are working properly, some tidying up of these test of needed, possibly some tidying up of models and observed data and probably some tidying up of the AutoDoc so the doco for these two models can be cited.

PR #2004, opened 2017-09-14
Arbitrator

HamishBrownPFR:
working on #1352.  Maize validation restructured to use a single observed file and verbose variable names.  Have had to accept a new set of stats because some variable names have changed but have checked the results the I get when run locally with the results in the Maize documentation for the last commit and they are identical so I am happy I haven't mucked anything up.


HamishBrownPFR:
@hol353 This is ready for merging if you are happy with it

Issue #2005, opened 2017-09-14
Should update glade format used in GUI


zur003:
Forms in the Gtk user interface currently make use of the old libglade format. This should be update to the GtkBuilder format. This will be a small step in moving from Gtk 2 to Gtk 3, and should allow a DLL to be removed from the installation set.

zur003:
This now seems to be working OK. Closing this issue.

PR #2006, opened 2017-09-15
Use GtkBuilder rather than libglade to construct Gtk user interfaces …

zur003:
…from glade.  Working on #2005


zur003:
Retest this please Jenkins

PR #2015, opened 2017-09-20
Leaf tiller model

### LINKED ISSUE #1984: Add new tillering model.

fai04d:
Resolves #1984 

Added new tillering model to Leaf.

### DIFF
--- a/Models/Core/APSIMFileConverter.cs
+++ b/Models/Core/APSIMFileConverter.cs
@@ -17,7 +17,7 @@ namespace Models.Core
     using System.Text.RegularExpressions;
     using PMF;
     using System.Data;
-	
+
     /// <summary>
     /// TODO: Update summary.
     /// </summary>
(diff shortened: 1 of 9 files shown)


HamishBrownPFR:
Hi @fai04d and @hol353. I am wondering if it would it be better to implement the apex tillering model into an alternative leaf class. I have had a quick look at this pull request and it appears to be forcing parameters into the leaf class which are going to be redundant in a lot of models and the leaf class is already quite cluttered with parameters. It seems to me the approach is sufficiently different to justify implementation in a separate class help reduce confusion when parameterising models.

hol353:
The real problem is that Leaf needs a complete redesign. It has become too large and monolithic. I'd like to see a leaf class that has functions (or sub models) for each of its processes, one of which might be tillering. Until this happens, we don't have many options. Yes, this tillering stuff could be incorporated into another Leaf class that has 90% of the same code as Leaf.cs but this isn't desirable either. For now let's go ahead with modifying Leaf.cs with a view to a redesign soon.

HamishBrownPFR:
An alternative leaf class could inherit the existing one so it is not duplicating code?
We have been talking about refactoring leaf for a long time now.  We really need to get on and do it.  How should we progress it?

PR #2026, opened 2017-10-02
Add consideration for generic SoilFertility when computing N demand

### LINKED ISSUE #2025: N uptake overstimated when using GLFSoilFertility
The generic factor GLFSoilFertility can be used in AgPasture to limit growth due to issues other than N supply. However, when this value is relativelly small N concentration in the plant can go beyong the maximum due to the fact that N uptake is not adjusted when GLFSoilFertility limits biomass growth.
A simple fix is to correct soil N demand, using GLFSoilFertility, limiting it to a value that would avoid large N uptake when the generic soil fertility factor is the most limiting factor.

rcichota:
Resolves #2025 
Added a correction for soil N demand so that N uptake is limited when using the generic GLFSoilFertility factor. This can still result in N concentration going above maximum, due to N fixation and remobilised from senesced tissues. This should be minor and rare, and a full fix would entail refactor these calcualtion which is a big job and may not be necessary...

### DIFF
--- a/Models/AgPasture/AgPasture.PastureSpecies.cs
+++ b/Models/AgPasture/AgPasture.PastureSpecies.cs
@@ -4758,7 +4758,7 @@ namespace Models.AgPasture
             {
                 // N demand is greater than fixation and remobilisation, N uptake is needed
                 senescedNRemobilised = RemobilisableSenescedN;
-                mySoilNDemand = demandLuxuryN - (fixedN + senescedNRemobilised);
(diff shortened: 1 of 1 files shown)


rcichota:
Hi Dean, can you tell me what is wrong here. When I click on 'details' I get a table which seems to say that 60% of the tests passed, thus 40% failed. But I cannot see which ones failed or what to look for.
@hol353

hol353:
The problem is that APSIM.PerformanceTests isn't working correctly yet. I'll merge this.

Issue #2041, opened 2017-10-06
Not properly working to show figures and edit report 


byzheng:
With the latest version of APSIMX (fresh clone), program is not properly working and/or crash in these conditions.

* Variables are not reportable, but exists in the Report. These variables will be ignored in previous version.
* Variables are not exists in the Report, but add into a figure.
* Click folder with a few figures. 
* Edit report after clicking any folders with figures, or a figure which has a missing variable. 

Don't have any problems with a version on 26/05/2017 (820510eb20464b9c7c6e0315bbba7cfdec0a2280).

hol430:
I'm unable to reproduce most of these, but clicking on a folder with a few figures (or even just a single figure) will freeze the UI for several seconds while the graph loads. Aside from annoying users, this can often cause nasty side effects if the unsuspecting user clicks attempts to click on another node(s) while the UI is frozen. 

I'm not sure what to do about this, short of loading new views in a separate thread, which is a solution I'm not keen on.

PR #2053, opened 2017-10-10
Avoid having null values for Live and Dead

zur003:
Working on #2041


hol353:
retest this please jenkins

PR #2082, opened 2017-10-17
Swamp Data and Properties panels in the report

### LINKED ISSUE #2081: Can we swamp "Data" and "Properties" panels in Report module?
In the report module of ApsimNG.exe, two panels are presented (i.e. "Data" and "Properties"). It is slow to load the whole table in the "Data" panel if there is a big table. However, we are interested to change the "Properties" in most of cases. So it will speed up to avoid load whole Data table before selecting Properties panel.

byzheng:
resolves #2081


zur003:
Retest this please Jenkins

PR #2093, opened 2017-10-20
Switch to new versions of Gtk win32 DLLs

### LINKED ISSUE #2080: White regions when clicking ApsimNG.exe
I am starting to use ApsimNG.exe, but found there is a white regions when clicking on GUI and cannot see the text underneath, then back to normal when clicking another regions.

image

zur003:
Resolves #2080 
@hol353 - There is a chance this change will cause problems on Hi-DPI displays, but my mock-up testing of this indicates that it should be OK. Can you test this on an actual high resolution device?


hol353:
Looks good to me on my Surface Pro. I'll merge this.

PR #2103, opened 2017-10-24
Dm conversion efficiency

### LINKED ISSUE #1991: Return value of DMConversionEfficiency equal to 1

fai04d:
Resolves #1991 
Removed changes to .apsimx files; all conversions done via APSIMFileConver.


hol353:
@fai04d The diffs look odd to me and not related to the issue.  Seems like most of the changes are other peoples changes from previous commits. Can you make sure this change is against the tip revision.

fai04d:
@hol353 Seems to be ok now. I've pulled APSIM/master into the branch which looks to have tidied it up.

fai04d:
Retest this please Jenkins.

Issue #2104, opened 2017-10-24
OrganArbitrator.FN has very big values


byzheng:
Wheat.Arbitrator.FN (OrganArbitrator class) has very big values (see figure below) which is caused by the small values of N.TotalPlantDemand.

![image](https://user-images.githubusercontent.com/1657381/31965692-7038d2fa-b94b-11e7-8961-ea165880fc75.png)

hol353:
@byzheng Is this still an issue or can we close it?

byzheng:
Can close it now

PR #2105, opened 2017-10-24
Check the value of N.TotalPlantDeman

byzheng:
Working on #2104 

FN  equals to zero if N.TotalPlantDeman has smaller value, but I am not sure this is the best method.

### DIFF
--- a/Models/Plant/Arbitrator/OrganArbitrator.cs
+++ b/Models/Plant/Arbitrator/OrganArbitrator.cs
@@ -74,7 +74,9 @@ namespace Models.PMF
         /// <summary>Gets the n supply relative to N demand.</summary>
         /// <value>The n supply.</value>
         [XmlIgnore]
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
Why do you want to set FN to zero of demand is small.  FN is an index of Nitrogen supply (relative to demand) and if demand is small and supply is large is should be a large number, not zero.

byzheng:
I think you are correct. We don't need to care about the extreme big value in this case as FN is just a ratio of demand and supply. The figure just looks ugly.

HamishBrownPFR:
Is it better to put an upper limit on FN,  Off the top of my head things that respond to FN use values up to about 3 so higher values don't have any meaning.  Could limit it to 5??

byzheng:
It will be better to check the reason for small N.PlantTotalDemand (E-14 in my tested case, See below for demand and supply). I would expect it should be equal or close to 0 at `Ripening` stage. 

PS: Double.Epsilon equal to 4.94065645841247e-324 in C#.

image

PR #2143, opened 2017-11-07
Issue/irrigation parameters#2140

### LINKED ISSUE #2140: Irrigation parameters could have default values settable outside Apply
The Apply method in Irrigation has one mandatory parameter (Amount) and several optional ones (depth = 0.0, efficiency = 1.0, willRunoff = false, duration = 1). This is simple and flexible in general, but it means that if any of the optional values were to change, they need to be given in the Apply command. This can make these a bit verbose (in case, for example, of a long list in a Schedulle where 'efficiency' is always 0.8), but also limits the ability to use a independent manager to change some of these (for example, to change the 'efficiency' once or twice during a simulation due to change in equipment).
A simple solution to this would be to add explicit default values for the optional parameters, and enable them to be set during the simulation, but not using the Apply method.

rcichota:
Resolves #2140 
I have made all the optional parameters for the Apply method to have explicit default values, these are defined outside the method. When user defined values are given, this are only used on that application. I also added a couple of new parameter: StartTime and WillIntercept. The first, and a new output (IrrigationTotal) mimic the behaviour of classic Apsim. The other parameter is a boolean and is meant to be used in MicroClimate. It is a bit tentative, but the idea is that when MicroClimate computes rainfall interception by canopy it could also consider the irrigation. However, I have not looked closely how easy it would be to implement that.
@hol353 
I have implemented a method for manually changing the default parameters. I opted for this instead of making the default values public. Any feedback is welcome.
I have shuffled a little the code, including moving the code for IIrrigation to its own file (which existed already, but was empty). I have also added a few check for the parameters and updated the documentation. You may want to check that this is ok too.
PS: do we really need the 'exception' tags on the header of the methods, they just repeat what is below...


rcichota:
Three simulations have failed due to the latest changes in irrigation:
 - Weirdo: in one simulation the duration was implied (old default was 1 hour, new value is 24hrs), fixed by adding duration as given parameter
- Wheat: in one simulation three of the irrigation parameters were given (amount depth and efficiency) but this relyed on order, which has changed. The depth and efficiency were defaults, so I delete them.
 - SoilWater: the test using SubSurface irrigation has two "erroneous" settings (one with depth -10 and other 3200, which is below the soil - total depth 3000), this reflects a test in Classic. The negative value is re-set to zero, as was in Classic, the problem was with the value below the soil. In ApsimX the value was not changed (and never used too, so no problems), in classic the value was re-set to the bottom layer. I think this is not ideal as the user may never know. I have implemented checks that will result in fatal error if the parameters are off bounds, so I had to fix the test simulation to avoid adding irrigaion below the soil.

rcichota:
Retest this please Jenkins

rcichota:
Retest this please Jenkins

hol353:
Hi @rcichota, does this resolve the issue?

hol353:
Looks good to me. Does this resolve the issue? If so can you change 'working' on to 'resolves' in the top comment box.

rcichota:
Resolves #2140 
I did what I needed to manipulate the variables and updated the behaviour to match the model in Classic Apsim. We'll need to have another go at this model to incorporate solutes within the irrigation water, but that sould be another issue, so let's close this one for now

Issue #2153, opened 2017-11-13
Nitrogen uptake depends on the root depth in each layer


byzheng:
Currently, nitrogen update is calculated the for all layers with Live.Wt greater than zero. A layer will provide the same amount of nitrogen even in which root only has a smaller proportion. Is it more logical to allow nitrogen update depending on the root depth in each layer?

PR #2154, opened 2017-11-13
Adjust root nitrogen uptake

byzheng:
working on #2153


byzheng:
I would expect lots of simulations will be failed as the new way for nitrogen uptake. @HamishBrownPFR Could you have a look at this issue and my implementation?

hol353:
The performance testing workshop seems to be working. If you click the Details link beside it, you can see what is different. Only maize has some diffs e.g. increased RMSE for above ground weight.

byzheng:
@hol353 I don't see any big difference for maize.apsimx. Looks this change doesn't have big impacts on nitrogen uptake.

HamishBrownPFR:
Your implementation looks correct to me.  I would have also expected this change to have an effect.  I will think on that some more

hol353:
Actually there are quite a few differences if you look at the testing site.

byzheng:
@hol353 Do I need to do anything for this pull request?

hol353:
@HamishBrownPFR 
This change makes the model fit worse for many models - see the performance testing website. Barley in particular is very different. The Reference Panel will want to know how you justify a worse fit. Can you provide a paragraph here for the Reference Panel on how you justify this.

HamishBrownPFR:
I don't think we can justify this without further investigation.  Some of the changes are too big.  For instance in Barley.Grain.Wt  
image.  Barley has not yet been released but Rob and Edith are actively working on it at the moment and having this sort of degradation in model performance dropped on them would be a bit off putting.  It is possible that barley is set up with a slow root front velocity or simulations have very thick layers which makes it very sensitive to this change.

HamishBrownPFR:
@hol353  I have also noticed some things that look odd in the stats.  For example Oats HarvestPO Grain.Wt.  There have been some quite large changes in the stats but the predicted numbers that have non-zero observations along-side them have not changed (apart from one very small change).

hol353:
Yes I had noticed this as well. Loretta thinks she has found the problem and is working on a fix.

hol353:
OK fixed now - stats have been updated.

HamishBrownPFR:
For what ever reason, the stats for barley are looking much better now.  There are some big diffs but these are all on stats that are already poor so can be disregarded as noise.  The code changes here give a refined estimation of N available, taking account of the proportion of the deepest rooted layer explored by roots.  There are some small effects on stats of complete (released) models but these are tolerable and may be addressed by refitting some model coefficients now there is a more complete mechanism in-place.
Recommend merging

PR #2165, opened 2017-11-16
Fix problem with leaf biomass being zeroed on harvest

### LINKED ISSUE #2164: Harvest causes zeroing of leaf biomass
Harvest should not cause all leaf biomass to be removed.  e.g for maize the default biomass removals for leaf biomass are:
image
But currently all dead is removed following harvest.  This is because the Structure class sets the Germinated property to false in the onHarvest event and the following day the Structure class then thinks the crop has not germinated and zeros all of the cohorts.  There is actually quite alot of stuff that happens in Structure.cs that should only happen on EndCrop and is duplicated there.
image

I will try removing the functionality from OnHarvest and see if that fixes the problem.

HamishBrownPFR:
Resolves #2164 

Remove remove crop ending actions from Structure OnHarvest which are duplicated in OnCropEnding

### DIFF
--- a/Models/Plant/Structure/Structure.cs
+++ b/Models/Plant/Structure/Structure.cs
@@ -466,7 +466,7 @@ namespace Models.PMF.Struct
             CohortToInitialise = 0;
             TipToAppear = 0;
             PotLeafTipsAppeared = 0;
-            ResetStemPopn();            
+            ResetStemPopn(); 
         }
 
         /// <summary>Called when crop is ending</summary>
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
Failing on a single variable in the chick pea test.  I suspect this is coming from differences in stem number calculations but i am not sure what is correct, the current or accepted values.  I will do some more investigation

HamishBrownPFR:
@hol353 This can be merged as the change is due to the fixing of a bug.  The difference in the Chickpea test is in Shell.PodNumber.  In the model shell number is the product of StemPopulation and FloweringNodes.  FloweringNodes is a constant of 5.  In the accepted tests Shell.PodNumber is always 5 times the sown plant population which means stem number per plant was always 1 for the accepted tests.  This Chickpea model is paramaterised in a way that stem number should always exceed one but the model was not doing this.  I am not sure how the change I have made has fixed this but what the model is doing following my change is correct.  
PS. shell number didn't affect yield predictions because the partitioning is done with a harvest index function

Issue #1281, opened 2016-10-26
Code needs some tidying up


zur003:
Once upon a time, we had a dream of writing code that complied with (most of) the StyleCop recommendations. I know I've drifted a bit from that standard, and I think everyone else has too, to some extent. I'm putting this issue here as something we can say we are "working on" when committing changes that don't change logic, but just give us tidier code.

zur003:
Running Code Analysis from the IDE reveals a number of other problems that StyleCop doesn't catch, and that don't show up as compiler warnings. Many of these also need to be fixed.

PR #2170, opened 2017-11-17
Style fixes

her123:
working on #1281


hol353:
I personally don't like adding extra curly brackets when there is no need for them. They just make the method longer and harder to fit on a screen. This isn't in the Microsoft guidelines is it?

her123:
By default in StyleCop this is the format. The M$ guide https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions doesn't seem mention {} ! We should have an agreed style for APSIM NG and adjust the StyleCop settings as needed. {} for code blocks is important but as you say not always necessary.

hol353:
I think we just adopt the Microsoft standards and adjust StyleCop to that.

PR #2179, opened 2017-11-20
Adjust stem population and cohort population

### LINKED ISSUE #2178: Stem and cohort population in the ApexTiller

byzheng:
resolves #2178

### DIFF
--- a/Models/Plant/Organs/LeafCohort.cs
+++ b/Models/Plant/Organs/LeafCohort.cs
@@ -788,16 +788,14 @@ namespace Models.PMF.Organs
         {
             //Reduce leaf Population in Cohort due to plant mortality
             double startPopulation = CohortPopulation;
-            if (!(Apex is ApexTiller))
-            {
-                if (Structure.ProportionPlantMortality > 0)
-                    CohortPopulation -= CohortPopulation * Structure.ProportionPlantMortality;
 
-                //Reduce leaf Population in Cohort  due to branch mortality
-                if ((Structure.ProportionBranchMortality > 0) && (CohortPopulation > Structure.MainStemPopn))
-                    //Ensure we there are some branches.
-                    CohortPopulation -= CohortPopulation * Structure.ProportionBranchMortality;
-            }
+            if (Structure.ProportionPlantMortality > 0)
+                CohortPopulation -= CohortPopulation * Structure.ProportionPlantMortality;
+
+            //Reduce leaf Population in Cohort  due to branch mortality
+            if ((Structure.ProportionBranchMortality > 0) && (CohortPopulation > Structure.MainStemPopn))
+                //Ensure we there are some branches.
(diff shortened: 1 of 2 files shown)


byzheng:
The difference comes from the way to reduce the cohort population. In the master, the CohortPopulation directly reduced by a proportion.
```
CohortPopulation -= CohortPopulation * Structure.ProportionBranchMortality;
```
In the pull request, I used the same way in the Structure to reduce cohort population to keep consistence with total stem population.

```
CohortPopulation -= Math.Min(Structure.ProportionBranchMortality * (CohortPopulation - Structure.MainStemPopn), CohortPopulation - Plant.Population);
```

HamishBrownPFR:
@hol353, lets use this as an example to work out a better way for reviewing diffs.  It appears @byzheng has made an improvement to the code, there are lots of diffs and most appear to be minor improvements or degradation in the stats.  However, without going through the first table that collates the PO table results and clicking each one that has diff and then going through each of the resultant tables and checking the stats it is not possible to know for sure that all the diffs are trivial.  
I find this first table is not that useful and the % of tests passed can be a small number of small diffs making things look bad when they are not.  It could also be a high % passed the the few that failed are large diffs and are important.
What would be more useful for quickly reviewing stats is if there was a table that combined the stats from all PO tables and then sorted based on the size of diffs, highest to lowest, so we can go straight to the biggest diff and review it along side the other stats calculated for the same variable.  
Another table of the same setup but ranked in the opposite order so the size of improvements can also be quickly assessed would be helpful in deciding if diffs are acceptable or not.

byzheng:
@HamishBrownPFR Agree with your point. A table merged from all PO tables will be really useful. From a technical point view, I would suggest to use js library [DataTables](https://datatables.net/) which is a powerful library to show table in the website including filter, sort, etc.

hol353:
@cla473 Having a single table for all stats for all pull request tables would be great. Something that is sortable by different things would also be good. This could replace the front table at the bottom. Would this be easy or hard?

hol353:
Note, I'm not in the office today. Working from home with no mobile reception either.

byzheng:
Filter is also useful to view a submit of results.

HamishBrownPFR:
We don't want to loose the ability to review stats at the POTable level either (the way it is currently done) as there are cases where you do want to look at all the stats for a specific POTable

byzheng:
@hol353 Can be merged? I am trying to clean up my working branch as a few un-merged commits mess up my branch and It is hard to add new changes now.

hol353:
retest this please jenkins

hol353:
retest this please jenkins

hol353:
@HamishBrownPFR I've rerun this. Can you have a look at the stats to see if anything pops up?

HamishBrownPFR:
This looks OK to me.  Code change tidies things up.  No diffs in any files we would not expect them.  The biggest diffs are in stats that were already pretty poor fits so I interpret this as being largely noise.

Recommend merging

PR #2180, opened 2017-11-20
Tidy up bits in test simulations

HamishBrownPFR:
working on #1352.  Cosmetic changes to simulation set ups to make sure I can get all the variables needed for graphs.  I have deconvoluted some of the variable names in maize (eg changed Stem.Live.Wt to Stem.Wt because there is no dead fraction).  This may effect stats which were using the other variable


HamishBrownPFR:
Diffs in maize test are all OK, for a couple of variables it is picking up some observations and doing tests that were previously missing.  For the other, the diffs are due to change variable names.

Wheat diffs are all very small but need to work out why they are there at all

HamishBrownPFR:
Wheat diffs are OK as well.  I have removed the EndCrop from the Lincoln2015 simulation (so biomass remains until the final observed date) which has changed the stats on DailyObsPred.  This is expected and the performance of the model has not changed so this can be merged.

Also noticed some diffs coming from a data duplication/order issue.  It appears there are duplicate instances of data rows on some dates, one which has a value for   Structure.TotalStemPopulation and one which has a zero.  The order of these is changing which is causing false diffs. 

image

hol353:
@cla473 The duplicates problem looks like a bug to me.

HamishBrownPFR:
@hol353.  I am happy for this to be merged now if you are.

hol353:
Jenkins didn't run clean. Maize auto-doc fail by the look of it.

HamishBrownPFR:
all good to merge now?

PR #2191, opened 2017-11-22
Altered moving average function and wheat.xml

### LINKED ISSUE #2189: Moving average function should not require an initial value
At the moment the moving average functions initial value is a parameter.  Instead it should just have the value of the first day on the first day, adding more days until it gets to the number specified and then dropping the oldest off the end.

HamishBrownPFR:
resolves #2189 

It seems the existing moving average function was causing some problems.  
- Initial value was assigned on the day of simulation commencing to all members in the array and then they were replaced 1 by 1 each day following.  In functions that wheat was using it for this could be placing zero values into the array before sowing.  I have modified the code so that it now has a StartMovingAverageStage on which it adds a single member to the array with the value for that day and keeps adding members each day until the array is at the size specified in the NumberOfDays parameter and then drops the last one off each day.
- Initial value for phyllochron should change with cultivar but it has not been set to do so.  Removing initialValue and having the function pick up the first days value should fix this.
- Array not reset between consecutive crops.  The array is now cleared on sow.

I expect some effect on stats and need to review and see if other changes are needed to compensate

### DIFF
--- a/Models/Core/APSIMFileConverter.cs
+++ b/Models/Core/APSIMFileConverter.cs
@@ -24,7 +24,7 @@ namespace Models.Core
     public class APSIMFileConverter
     {
         /// <summary>Gets the lastest .apsimx file format version.</summary>
-        public static int LastestVersion { get { return 21; } }
+        public static int LastestVersion { get { return 22; } }
 
         /// <summary>Converts to file to the latest version.</summary>
         /// <param name="fileName">Name of the file.</param>
@@ -180,7 +180,7 @@ namespace Models.Core
 
                 try
                 {
-                    double area = Convert.ToDouble(areaString, 
+                    double area = Convert.ToDouble(areaString,
                                                    System.Globalization.CultureInfo.InvariantCulture);
                     if (area <= 0)
                         XmlUtilities.SetValue(zoneNode, "Area", "1");
@@ -569,7 +569,7 @@ namespace Models.Core
             foreach (XmlNode n in XmlUtilities.FindAllRecursivelyByType(node, "Leaf"))
             {
                 XmlNode dmFunction = APSIMFileConverterUtilities.FindModelNode(n, "DMConversionEfficiencyFunction");
-                if (dmFunction!=null)
+                if (dmFunction != null)
                 {
                     XmlUtilities.SetValue(dmFunction, "Name", "DMConversionEfficiency");
                 }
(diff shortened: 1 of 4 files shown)


HamishBrownPFR:
As suspected, the fix to Moving Average has effected phenology,  I will look into how to correct for this

image

HamishBrownPFR:
@hol353 I noticed that there are some massive diffs again for Barley which am I not sure makes sense, it has phenology set up in a similar way to wheat and the phenology variables are showing similar sorts of diffs to wheat so I am not sure how to explain such large diffs in the biomass variables for barley.  We noticed the same thing with @byzheng 's pull request #2153 the other day.  I wonder if there is a bug in the test for Barley some how?  I will investigate the source of the errors specific to this pull request some more

hol353:
I grabbed the latest barley.db file from Jenkins. I ran the GUI and put a tests node on HarvestObsPred and compared the stats in the GUI to the stats on the testing web site. They are the same!

hol353:
Could the problem be that you haven't added:

<StageToStartMovingAverage>Emergence</StageToStartMovingAverage>

to the file converter and so Barley (and other crops?) won't have this variable?

HamishBrownPFR:
Yes, that is certainly a problem.  Just trying to work out how to get this change into the converter.  I need to  remove the node called "InitialValue" from Functions of type "MovingAverage" then add a node called "StageToStartMovingAverage" and give it a value of the "StartStage" parameter for the EmergingPhase

hol353:
Don't worry about removing InitialValue. It will go away by itself whenever the user does a save.

HamishBrownPFR:
I am not sure why the Chicory stats are failing on stats.  It is not using the moving average function.  There appear to be no accepted stats in the table

hol353:
Don't worry about Chicory. Looks like they weren't included in the test suite (for whatever reason) when I accepted the stats on my pull request yesterday.

HamishBrownPFR:
@hol353 the changes I have in this pull request resolve the problems with the moving average function and the fix to the wheat phenology model that was needed to compensate for having fixed the error in the moving average model.  I have checked all the diffs and, given the context of this being an important bug fix, I can see no diffs that I would consider more than trivial.  There are some diffs to the CanopyPhotosynthesis test so it would be good of @byzheng could also review the diffs and agree he is happy to have this merged.

byzheng:
@HamishBrownPFR @hol353 I don't setup up CanopyPhotosynthesis.apsimx, but I think it should be fine to merge it as there are no significantly changes except grain number.

HamishBrownPFR:
Thanks @byzheng It is OK you didn't set up the simulations.  I was just wanting an opinion that the changes are acceptable so that we are overlooking important stuff when merging code changes that result in changes to model performance.

PR #2201, opened 2017-12-03
Maize model improvements

HamishBrownPFR:
Working on #1352 
The analysis of the arbitrators performance for the maize model highlighted a number of problems in the maize model.  I have used the Lincoln2012 trial to improve a number of aspects of the model, in particular partitioning of N and Wt to stem, husk and rachis and the response of LAI senescence to drought.  These changes have improved the overall stats for the maize model as well.


HamishBrownPFR:
Some statistics have decreased but the decreases are small for statistics that were not all ready poor.  Most of the statistics have improved and there are some large improvements giving better all round stats.  I recommend merging but it would be a good idea to get @hut104 or someone else to review.

hut104:
?I think you can ignore the Gliricidia results.  I'll talk to Philip today about getting updated files for these.  They have lots of predicted and observed zero values.  The stats won't make much sense with these.

Neil.


HamishBrownPFR:
Yeah, I wasn't worrying about the Gliricidia results initially.  I have made some changes to the maize model and was wanting an opinion on if they represent and improvement (and should be merged into master) or if you think there are some problems with them.  I am guessing maize is in the Gliricida simulations and so changes to the maize model impact those simulations also???

hut104:
Gliricidia is a leguminous agroforestry tree species. It is used as a green manure crop and so the validation for this species includes tests that it provides appropriate N supply to companion crops.  So, in this case, the gliricidia test checks that the companion maize crops are suitably simulated.


So, it needs to be included, but don't worry just now as I don't think the files for gliricidia are 100% right.


HamishBrownPFR:
I'm not worrying about the Gliricidia, its a check of the changes I have made to maize and and 2nd opinion that the overall the stats have improved.

HamishBrownPFR:
I have got to much stuff in here.  I will close this and put changes up in smaller lots

PR #2141, opened 2017-11-06
Rewrite codes for nitrogen redistribution

byzheng:
resolves #2138

PR #2203, opened 2017-12-05
Implement @BYZheng fix to N supplies

### LINKED ISSUE #2138: In PLANT calculate reallocation before retranslocation?
@HamishBrownPFR In the nitrogen supply function, is it more logical to calculate reallocation before retranslocation? As `nitrogenSupply.Retranslocation` removes `nitrogenSupply.Reallocation` although there are no impact for current wheat model ( `senescenceRate` equals to zero).
 
https://github.com/APSIMInitiative/ApsimX/blob/871830c04d39f368ac450d33aa55237d54e2653f/Models/Plant/Organs/GenericOrgan.cs#L298-L304

HamishBrownPFR:
Resolves #2138 

@byzheng I have put your changes into to generic organ into a seperate pull request to assess their impact seperate from the changes you made to leaf in #2141

### DIFF
--- a/Models/Plant/Organs/GenericOrgan.cs
+++ b/Models/Plant/Organs/GenericOrgan.cs
@@ -305,6 +305,9 @@ namespace Models.PMF.Organs
             nitrogenSupply.Retranslocation = Math.Max(0, startLive.StorageN * (1 - senescenceRate.Value()) * nRetranslocationFactor.Value());
             if (nitrogenSupply.Retranslocation < -biomassToleranceValue)
                 throw new Exception("Negative N retranslocation value computed for " + Name);
+
+            nitrogenSupply.Fixation = 0;
+            nitrogenSupply.Uptake = 0;
             return nitrogenSupply;
         }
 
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
OK the big diffs are coming from the changes to generic organ.  In theory changing the order that Retranslocation and Reallocation are calculated should not cause any diffs because we never have senescence in generic organs so this will always be zero and won't matter if it is first or last.  So the diffs must be coming from the different way that retranslocation is calculated.

HamishBrownPFR:
I have done some careful checking of the changes to the stats for things that were worrying me.  For example NSE for Maize.Rachis.Wt dropped from 0.68 to 0.58 which looks like quite a big shift.  However there are only 6 observations for this variable, only one of them has moved and only a small amount but it seems to affect the stats alot.  Similarly NSE for Wheat.Stem.N dropped from 0.54 to 0.45.  The points that have shifted are the ones that were already well under predicted (Screen shot below) suggesting this bug fix is exposing some other problem in the model.

image

HamishBrownPFR:
I am going to close this pull request and recommend merging of  #2141

PR #2208, opened 2017-12-06
Wheat

### LINKED ISSUE #2207: Improvements to wheat model
I am working on a paper describing the PMF arbitrator and have found a number of issues in the wheat model in doing so.

HamishBrownPFR:
Resolves #2207.  Have improved LAI predictions under drought and biomass partition.

@byzheng these changes may effect what you are working on so have a look.
@hut104 this follows on from a pull request I closed a couple of months ago where I was trying to resolve this issue but caused problems in some of the regions.  I think I have resolved this now.


HamishBrownPFR:
@hol353 I have reviewed all the diffs in this pull request and recommend that over all the pull request give a considerable improvement in some statistics and represents a number of processes more accurately.   However, it would be good if @hut104 and or @byzheng could review my interpretation here.

Nearly all of the stats that have sown a reduction were; already poor, are small and trivial or are on datasets with a small n.  When balanced against a number of improvements in some important stats I think overall this represents an improvement.  The only diff I could find that appeared show a meaningful reduction in model performance was for the soil organic N in the Pendalton set.  I have changed a number of parameters that effect partitioning so there will be less N in straw added to the soil.  As yield is substantially under predicted in these simulations because of unresolved issues with rainfall data so predicted inputs will be low and when this can be resolved I expect it will correct the apparent reduction in performance seen here,

@hut104 and @byzheng I have made a number of changes to the model so it will simulate the Lincoln2015 data set well so this forms a useful basis for the arbitrator paper I am working on.  These changes improved performance on a number of other test sets but worsened them on others.  Most notably the change to drougth stress effects on leaf senescence improved predictions on APS26, Lincoln2014 and Lincoln1991/92 simulations also.  I have also made a number of changes to give better partitioning of biomass and N into and out of stem, leaf and spike.  Where the model changes had worsened predictions I interpreted this as improvements to the model exposing problems we had previously fitted around.  The main problem simulations were the Turkish data and Cunderdin.  For the former, I have spent some time fitting the cultivar better and the model is predicting this set well now.  For the later there was a clear under prediction for the irrigated treatments because the crop was running out of N.  I interpreted this as incorrect estimates of deep N in the soil set up and increased initial N to improve predictions here.  There were also a few trials with systematic over predictions which were due to over prediction of grain number so I reduced the grainsPerGramOfStem parameter for these cultivars to give better predictions.

I know I have made a lot of changes here with the intention of improving the model.  I think the stats back this up but I am interested in your opinion of the adequacy of my explanation of what changes I have made

HamishBrownPFR:
Hi @hut104 are you able to have a look at this

hut104:
Hamish, A few suggestions.
1) We now have a thing called [Leaf].ExtinctionCoeff.VegetativeExtinctionCoeff.FixedValue
Could it be [Leaf].ExtinctionCoeff.VegetativePhase.FixedValue or something similar? e.g. to avoid repeating ExtinctionCoeff.
2) The diff tool says the following was added
+    <Live>
+      <Name>Biomass</Name>
+      <IncludeInDocumentation>true</IncludeInDocumentation>
+    </Live>
+    <Dead>
+      <Name>Biomass</Name>
+      <IncludeInDocumentation>true</IncludeInDocumentation>
+    </Dead>
This doesn't look right.  Should they be of type biomass with names of live and dead? 
3) The diff tool seems to suggest that the following has been added
+    <MaintenanceRespiration>0</MaintenanceRespiration>
Not sure where, but not sure why.

4) I notice that you use linear interpolations with the x value as
+        <XProperty>[Phenology].Stage</XProperty>
I find these problematic because if you change the phases, all these numbers get broken.  It works better, and doco is clearer if you can use phase names rather than stage numbers.  I've been burned so many times trying to get things to work after changing phases.  I also see that vegetative phases are using different X lookups for different interpolations (ie vegetative phase has different meanings).  Is it possible to make any of this cleaner?
5) you have some abbreviated names that could be changed
StructuralFractionGF and StructuralFractionEG
Could be called ValueDuringGrainFilling and ValueDuringEarlyGrowth or something.
6) There seems to be something added called
<ExpansionStressValue>0</ExpansionStressValue>
7) There seems to be the following added
+      <ApexGroupSize />
 +      <ApexGroupAge />

HamishBrownPFR:
@hut104 Thanks for these comments.  

1) I have changed the name of ExtinctionCoeff.VegetativeExtinctionCoeff.FixedValue as you suggest

2) The Biomass was added by the change tool when I had the wheat model in the .apsimx file under replacements.  You are right about the biomass objects, they appear to have the name and type fields inverted.  Doesn't seem to cause any problems with the model running so not sure what they are doing there.
 
3) Not sure what this one is

4) I have also had problems using Phenology.Stage instead of StageNames.  However, the function that uses StageNames applies constant values across stages whereas the linint using Phenolgy.Stage as the xproperty allows gradual change in parameter values as the crop develops which was needed where I have used it.  Maybe we need to develop a linint function that works of StageNames and doesn't break if the names of stages are changed.

5) I have changed these names as you suggest
6) Not sure what this one is either, has you comment rendered properly in github?
7) This was included by the change tool.

Issue #2199, opened 2017-11-29
Move to 64bit APSIM Next Generation


hol353:
We need this urgently so that large simulations with many zones (e.g. #2195) can run. Val will also probably run into this problem when she explores larger urine patch modelling.

The problem is going to be the third party packages we use in particular the GTK and SQLite .dll files.

@zur003 Do you have time to look into this?

zur003:
Hmmm. 64-bit SQLite shouldn't be a problem, but 64-bit Gtk# may be a bit harder. Of course we're already running 64-bit on Linux and OS X - the problem is getting a suitable set of native DLLs for 64-bit Windows.

HamishBrownPFR:
Hi @zur003.  We are needing to do some project planning and plans will depend alot on if APSIMX can go 64bit and solve the memory problems we are having.  Can you give an estimate on how hard or easy you think this will be and how long you think it may take.

josharp22:
Hi @zur003 and @hol353 
Is there any update on whether this is likely to happen and what the timescale might be please? As indicated we have a large project that needs to run multiple zones and if this isn't likely to happen need to amend the plan asap.
Cheers.

hol353:
Hi josharp22
This is a job for Eric who has been on long service leave. I'm not sure if he is back this week or next. I'm also not sure what else he has on his todo list. We'll have to wait until he responds to this.

josharp22:
Thanks for the update @hol353 will wait until @zur003 returns.
Cheers.

zur003:
Hi @josharp22 . The problem is the lack of support for 64-bit Gtk# on Windows, so we can't readily run the GUI as 64-bit. This isn't likely to change anytime soon. You have a couple of short-term options: (1) Run the model without the GUI (this will require the 64-bit version of Sqlite, but that's readily available) or (2) work on a Linux or Macintosh system instead of Windows; on those platforms we ought to be able to run everything as 64-bit (though I now realise that I haven't actually tested this).

I kind of suspect the only way we're going to get everything running as 64-bit on Windows is to rebuild all the Gtk# support from source code, and that's not a trivial exercise - and even we try, we're likely to find there's some technical difficulty that would account for why this hasn't already been done by the good folks at the Mono project.

I suppose another possibility is to run the GUI as 32-bit, and find some magic that would allow us to run the model as 64-bit. I don't really know how feasible this might be.

josharp22:
[...]

PR #2255, opened 2018-01-29
Make both 32-bit and 64-bit versions of sqlite3 available on Windows.…

zur003:
… Allow Models.exe to run as as 64-bit assembly. Working on #2199


zur003:
Retest this please Jenkins

zur003:
@hol353 - for this to work as intended, a pull request in APSIM.Shared must also be applied (although I don't think that accounts for the problem with the SoilNitrogenPatch performance test).

zur003:
Retest this please Jenkins

Issue #1137, opened 2016-09-07
Add "Run On Cluster"


mcauliffer:
Add in an option to run simulations on a cluster (current sim(s) or directory of sims, as per classic). Initially this will use the Microsoft Azure platform that we are testing with Microsoft but will be intended to be extensible for other cluster types (Amazon, google, condor?)
At this stage it is planned to target the GTK version with the assumption it will become the standard interface.

sno036:
@hol430 Bug crept into the run on cluster.

Before pointing to the licence files:
**Tools Directory not found: C:\Program Files\APSIM2018.03.02.1632\Bin\tools" before pointing it to the licence file**







The argument must not be empty string.
                      Parameter name: accountName
                         at Microsoft.WindowsAzure.Storage.Core.Util.CommonUtility.AssertNotNullOrEmpty(String paramName, String value)
                         at Microsoft.WindowsAzure.Storage.Auth.StorageCredentials..ctor(String accountName, String keyValue, String keyName)
                         at UserInterface.Presenters.NewAzureJobPresenter.GetCredentials(Object sender, EventArgs e)
                         at UserInterface.Presenters.NewAzureJobPresenter.Attach(Object model, Object view, ExplorerPresenter explorerPresenter)
                         at UserInterface.Presenters.ExplorerPresenter.ShowInRightHandPanel(Object model, String viewName, String presenterName)

hol430:
Is this the first time that you've tried using this via an installed version of Apsim?

sno036:
Yes.  The version run from VS crashed so I tried that.  Just thought – I didn’t update Shared.  Just done that now and rebuilt everything and got the same error message.



The argument must not be empty string.
                      Parameter name: accountName
                         at Microsoft.WindowsAzure.Storage.Core.Util.CommonUtility.AssertNotNullOrEmpty(String paramName, String value) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\Common\Core\Util\CommonUtility.cs:line 142
                         at Microsoft.WindowsAzure.Storage.Auth.StorageCredentials..ctor(String accountName, String keyValue) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\Common\Auth\StorageCredentials.cs:line 139
[...]

Issue #806, opened 2016-04-25
Select which Factorial combinations to run


HamishBrownPFR:
I have just added ~1500 simulations to the wheat validation.  I am only interested in the results of ~500 of them but is was easier to set them up as a full factorial with redundant simulations.  However this will cause issues with the run speed of the tests.

In cases where experiments are not balanced we don't want to run all of the factorial combinations.  Is is possible to have a check box for each of the simulations that are listed on the Experiment view and have the model only run those which are checked?

HamishBrownPFR:
I have slimmed this down now to just 13 (not 60) cultivars (
![image](https://cloud.githubusercontent.com/assets/7295858/14805013/650496f0-0bbc-11e6-818c-19d64540eb52.png) and 40 sowing dates (
![image](https://cloud.githubusercontent.com/assets/7295858/14805025/77ecb64e-0bbc-11e6-9331-ad7586e2ffe5.png)).

In full combination this is 520 simulations but only 193 of the potential combinations actually occurred (the ones listed below). 

Perhaps there could be an option for validation experiments where the simulation name has to be present in an observation file for the combination to be created and run?

CPTCultClaireSOW2000-03-03	Claire	3/03/2000
CPTCultClaireSOW2000-05-15	Claire	15/05/2000
CPTCultOtaneSOW2000-05-15	Otane	15/05/2000
CPTCultOtaneSOW2000-09-20	Otane	20/09/2000
CPTCultAmarokSOW2001-04-09	Amarok	9/04/2001
CPTCultClaireSOW2001-04-09	Claire	9/04/2001
CPTCultAmarokSOW2001-06-15	Amarok	15/06/2001
CPTCultClaireSOW2001-06-15	Claire	15/06/2001
CPTCultClaireSOW2001-08-09	Claire	9/08/2001
CPTCultAmarokSOW2002-03-07	Amarok	7/03/2002
CPTCultClaireSOW2002-03-07	Claire	7/03/2002
CPTCultOtaneSOW2002-03-07	Otane	7/03/2002
CPTCultAmarokSOW2002-05-15	Amarok	15/05/2002
CPTCultClaireSOW2002-05-15	Claire	15/05/2002
CPTCultOtaneSOW2002-05-15	Otane	15/05/2002
CPTCultClaireSOW2002-09-09	Claire	9/09/2002
CPTCultOtaneSOW2002-09-09	Otane	9/09/2002
[...]

PR #2256, opened 2018-01-30
Factor control and cloud job submission

holzworthd-agr:
Working on #1137 and #806.


hol353:
@holzworthd-agr This didn't compile Drew. Click 'Details' above on 'Jenkins' line. Then click 'Console Output' link to see compile errors.

holzworthd-agr:
Yeah, I see that. Seems like a problem with the System.Data.SQLite reference, which was used in the MARS implementation of the downloader. I see there's a SQLite class in APSIM.Shared though, which I'm currently using to rewrite that part of the code.

hol353:
@zur003 Any thoughts on why this didn't compile? Wrong version of msbuild on Jenkins perhaps?

hol353:
@zur003  Also lots of warnings about ToolsVersions on Jenkins. Need to get that cleaned up as well.

holzworthd-agr:
@hol353 Should I revert my change to the tool version then?

hol353:
Give it a go and see what happens.

zur003:
I don't think the ToolsVersion stuff really matters all that much. See Microsoft's description of this issue at https://msdn.microsoft.com/en-us/library/bb383796(v=vs.120).aspx. To some extent, though, it does raise the issue of which version(s) of Visual Studio we want to "officially" target. 

It appears to me that the main reason the build failed was the absence of System.Data.SQLite. I suggest adding it as a NuGet package, so that fetching it happens automatically. 

There also error messages for code in AzureJobDisplayView.cs, but the problem there is not obvious to me.

holzworthd-agr:
I've added System.Data.SQLite as a NuGet package, but after looking at it some more I'm still not sure what's wrong with AzureJobDisplayView.cs - it compiles locally just fine.

hol353:
OK I'll pull your branch and have a look on my computer. It might take me a hour to get around to it though.

hol353:
I've had a look at this and it compiles fine on my computer as well. Eric is looking into it now. I'm guessing Jenkins is using an old msbuild version

zur003:
Retest this please Jenkins

zur003:
Retest this please Jenkins

zur003:
Retest this please Jenkins

zur003:
For reasons that remain a mystery to me, getting scoop to use the latest MSBuild.exe seems to have fixed the compile problem. However, there's an error in the UITests phase that won't be obvious at first glance: The .cs files in Tests\UserInterfaceTests will require explicit conversions from IPresenter to ExplorerPresenter.  Similarly, a cast is needed in Documentation\CreateModelDocumentation.cs.

hol353:
@zur003 Can you give Drew the modified UserInterfaceTests files that he needs - or push to his branch? I'm not around tomorrow and I've run out of time today.

holzworthd-agr:
@zur003 I added the changes you mentioned, but I'm not sure that this is going to work - I used the IEnumerable.OfType() method which requires System.Linq. To use this I manually added System.Core.dll to the referenced assemblies in the CompileTextToAssembly() method of the APSIM.Shared.Utilities.ReflectionUtilities class. 

If I get this change merged into APSIM.Shared, will it affect this build?

zur003:
Keep it simple. Couldn't you just use an (ExplorererPresenter) cast, or else "as ExplorerPresenter"?
For example:
    ExplorerPresenter presenter = tabbedExplorerPresenter.Presenters[0] as ExplorerPresenter;

That should work, and doesn't require pulling in the  Linq baggage.

holzworthd-agr:
You're right - that's definitely easier - I've made the changes you suggested. I ran the generate documentation batch script, and it appears to work for Chicory but fails on the subsequent tests (it couldn't find the PO table in the apsimx file). I suspect this is because I have run the chicory simulation in Apsim but none of the others.

I'm not sure how to test the 2 files under \Tests\UserInterfaceTests, so I didn't.

hol353:
retest this please jenkins

hol353:
Drew, you've committed potato.apsimx which has affected the stats. Can you remove it by going to SourceTree, clicking on the specific commit that has potato.apsimx in it, right click on it and select Log Selected. The move to the previous commit in the list, right click and select 'Reset to commit..'. If you need help, set up a meeting with me next week (Mon 10-3pm QLD time is OK for me).

holzworthd-agr:
@hol353 It seems to have failed the performance test again, even after I reset potato.apsimx to the commit at which it was last modified.

PR #2257, opened 2018-01-30
Change build settings for Release builds as well as Debug builds.

zur003:
Working on #2199.


zur003:
Retest this please Jenkins

zur003:
Retest this please Jenkins

zur003:
@hol353 - The differences showing up in the performance testing appears to be real, since with this change, the test models will be run using 64-bit code instead of 32-bit (in Jenkins "Run Tests and Prototypes", which runs Models.exe from the command line). Apparently this is resulting in some small differences in model outputs. What do you suggest I do?

hol353:
I've looked at the diffs and they are only minor so I've 'accepted' the new stats and merged this.

PR #2264, opened 2018-01-31
Grapevine clean version

junqi108:
working on #1292


junqi108:
Hi there: the test for grapevine is newly added, and it is under developing so the results would keep changing. I do not know why it affects the results of Maize. I only changed the files in grapevine model through the user interface. best regards. junqi

HamishBrownPFR:
@hol353 can you see or suggest any reason why the maize tests are failing with @junqi108 's pull request?

zur003:
@hol353 - A lot of the recent performance tests seem to be "failing" on Maize : PredictedObserved : Maize.Stem.N : NSE. It is possible the "accepted" stats are wrong. It would be even more worrying if the various runs aren't giving consistent results. Any clues?

hol353:
Yes, something has gone badly wrong with the performance site with my last 'accepted' stats on Wednesday. I'll have to look into it.

PR #2265, opened 2018-02-01
Add code to enable partittioning delta N among CNPatches

### LINKED ISSUE #2258: Add partition logic to new setNxx methods in SoilNitrogen (for SoilCNPatches)
Dean has added methods for setting the N forms (SetUrea, SetNh4, and SetNo3) of SoilNitrogen. In these methods, the module making the change is identified so it can be used to control the logic for partitoning that change into each SoilCNPatch.  The logic was sitting only on OnNitrogenChanged, it must be added to the new methods so the patch capability of SoilNitrogen can be used appropriately

rcichota:
Resolves #2258 
Added code to newly instated methods: SetUrea, SetNH4, and SetNO3 in SoilNitrogen. Now these will call the PartitionDelta method, which aportions the values passed (delta N) to the SoilCNPatches.  This is only relevant when using the SoilCNPatches capability.

### DIFF
--- a/Models/Soils/SoilNitrogen.Variables.cs
+++ b/Models/Soils/SoilNitrogen.Variables.cs
@@ -2446,17 +2446,44 @@ namespace Models.Soils
             }
         }
         /// <summary>Setter for urea</summary>
+        /// <remarks>
+        /// This is necessary to allow the use of the SoilCNPatch capability
+        /// The values passed, or in fact the deltas, need to be partitioned appropriately when there is more than one CNPatch
+        /// </remarks>
         /// <param name="callingModelType">Type of calling model</param>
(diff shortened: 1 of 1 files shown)


rcichota:
Retest this please Jenkins

rcichota:
Retest this please Jenkins

hol353:
I'm happy with the minor change in R2 for the sugarcane validation. Where you expecting changes?

rcichota:
I was hoping for no changes, but because SoilNitrogen now calculates the deltas itself, it could create rounding errors... I think that's what's happening with SugarCane, it only made worse one measure in one variable (which was already pretty poor). So, if you happy, this can be merged.

PR #2278, opened 2018-02-06
Oil palm et

### LINKED ISSUE #2277: Add AnnualET output for oil palm
oil palm simulations in validation set and example would benefit from an annual ET figure for people to look at simple water balance.

hut104:
resolves #2277


hol353:
retest this please jenkins

PR #2285, opened 2018-02-07
Don't copy gtk-sharp dlls into Bin during builds. Also, revert ToolsV…

### LINKED ISSUE #2284: Setting "Copy Local" for gtk-sharp DLLs has broken non-Windows installs
Recent changes altered the setting for "Copy Local" for gtk-sharp, gdk-sharp and glib-sharp from "false" to "true". This causes problems with the Linux builds, as the installer will attempt to install copies of these (it installs everything in the Bin folder), but doesn't configure them properly.

zur003:
…ersion number as it breaks MonoDevelop builds on Linux.

Resolves #2284


zur003:
Retest this please Jenkins

zur003:
Closing while I fix up git merge problems.

PR #2291, opened 2018-02-09
Update most of the NuGet packages

### LINKED ISSUE #2127: The NuGet packages need to be updated once in a while.
The various NuGet package should probably be updated every once in a while. Currently, the OxyPlot.Core package is sufficiently out-of-date that it is no longer available from its repository.

zur003:
Resolves #2127.

When merging, note that there is a parallel pull request in APSIM.Shared.

### DIFF
--- a/ApsimNG/App.config
+++ b/ApsimNG/App.config
@@ -1,132 +1,132 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <configSections>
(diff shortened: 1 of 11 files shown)


hol353:
Is this now resolved? Can you change the 'Working on' to 'Resolves'?

zur003:
We will want to update the NuGet packages from time to time. Would you prefer to log those changes against a single on-going Issue (one that is never "resolved"), or create a new Issue for each set of updates?

zur003:
Ok. Then this is resolved.

PR #2303, opened 2018-02-13
Display the inner exceptions of parsing errors.

### LINKED ISSUE #2300: XML parsing exceptions don't show correctly
The XML parsing exceptions need to be shown in the status window with the exception message from the parser.

zur003:
@hol353 - note that there is an accompanying change in APSIM.Shared.

Resolves #2300


hol353:
retest this please jenkins

Issue #1822, opened 2017-07-28
Grape Vine Model


HamishBrownPFR:
Perennial fruit crops are less adaptable to climate than annual crops 
because they are kept for many
years after establishment leaving fewer opportunities to alter genetics 
and management. However, no
formal perennial fruit model has been developed in APSIM to help make 
strategic decisions when
establishing and managing these. To this end, we used grapevine Vitis 
Vinifera L. as a model plant to
develop a perennial fruit crop model in APSIM next generation.

A series of developments was undertaken to adapt the APSIM source code 
to simulate perennial woody
plants by Junqi Zhu, @HamishBrownPFR, @hut104 and @hol353. The model was 
built
to simulate the phenology, (strip) light interception, canopy 
development, carbon allocation, carbon
storage and remobilisation, and yield formation of grapevine as a model 
perennial crop. It
includes the following features:

1. A vine and inter-row strip configuration to represent the vineyard 
setup. Currently inter-row
planting was represented by a generic plant model ‘Slurp’. The ‘Slurp’ 
model simply takes up
water and nitrogen, but does not predict crop growth or yields. We 
forced its leaf area index to
follow a sensible growth pattern under the vineyard management in 
Marlborough New Zealand
(NZ) with periodic mowing events.

2. An adapted strip light interception model triggered automatically by 
the strip setup.

3. A new carbon allocation method to allocate the carbohydrate 
simultaneously based on sink
strength as well as sink priority.

4. A yield module that uses the pre- and post-flowering weather 
conditions in previous season along
with current season weather and carbon status to determine bunch number 
per shoot, berry
number per bunch and potential berry size.

5. An adapted phenology model that starts from endodormancy of the bud 
instead of emergence
and resets to endodormancy by a management method (pruning) in winter to 
capture the
perennial behaviour.

6. Flexibility in adapting to a different training system by controlling 
the bud number per vine,
canopy height, canopy width and depth as inputs.

sarahcleary:
@HamishBrownPFR - just following up the concept note.  For background, please refer to https://github.com/APSIMInitiative/ReferencePanel/issues/57#issue-595681585

junqi108:
@APSIMInitiative/reference-panel 
[...]

PR #2308, opened 2018-02-14
Grapevine

junqi108:
working on #1822 
maybe there is an easy way to adding it to the date function, but I am now well versed in that region. So I set up a separated function for it which is also simple. This function is useful when i want to validate the grapevine flowering time which sometimes jumps from December to January.

### DIFF
--- a/Models/Core/Model.cs
+++ b/Models/Core/Model.cs
@@ -171,6 +171,7 @@ namespace Models.Core
         [XmlElement(typeof(Models.PMF.Functions.Constant))]
         [XmlElement(typeof(Models.PMF.Functions.DeltaFunction))]
         [XmlElement(typeof(Models.PMF.Functions.DivideFunction))]
+        [XmlElement(typeof(Models.PMF.Functions.DOYFromJuly1st))]
         [XmlElement(typeof(Models.PMF.Functions.ExponentialFunction))]
         [XmlElement(typeof(Models.PMF.Functions.ExpressionFunction))]
         [XmlElement(typeof(Models.PMF.Functions.ExternalVariable))]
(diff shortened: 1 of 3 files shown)


junqi108:
retest this please Jenkins

hol353:
@HamishBrownPFR  What is the need for this function? If we have to have it can we make it more generic. Maybe a property that takes a date string like '1-jul'. The code can then do a DateUtilities.GetDate call to turn this into a DateTime. The function could then be called something more generic e.g. DaysSince.

Also don't derive from ICustomDocumentation. Then you can get rid of the Document method.

hol353:
I'd also like a separate issue created for this new function. The pull request can then resolve the issue.

HamishBrownPFR:
Hi @hol353 and @junqi108.  This is needed because some aspects of perennial physiology are analysed relative to a fixed data rather than an identifiable event like sowing that we have in annuals.  
I am not sure the best way to do this.  We could modify the accumulate function to do the job but it would get messy so I suspect a separate function that increments day number between specific dates and resets on specific dates would be the best start.  
@junqi108  use the Accumulate function as a template but instead of incrementing on PostPhenology do it on a clock event so it is not then limited to plant relevance,  something like StartOfDay would be appropriate.  Also instead of specifying StartStageName, EndStageName and ResetStage name change these to dates.  Doing it this way will mean to accumulate days it needs a child constant of 1 but the child function could be other things which will make the function more generic.
Perhaps we call this function AccumulateByDate and we call the current Accumulate Function AccumulateByPhase to make their distinction clear.

HamishBrownPFR:
@junqi108 give me a call or skype if you need some help turning the Start End and Reset stages to dates.

PR #2323, opened 2018-02-20
added extra output to perrenial leaf and fixed light interception error

### LINKED ISSUE #2322: Leaf classes do not use the light interception data from micromet
Leaf organ classes currently only use their own calculation of light interception.  As a result, they do not compete with eachother properly.
They should use the light interception data from micromet.

hut104:
resolves #2322


hol353:
retest this please jenkins

PR #2335, opened 2018-02-21
Use a different approach to export bitmaps of maps on Windows.

### LINKED ISSUE #2261: Maps not showing in Autogenerated documentation
Certainly an issue for the new nutrient model. May be an issue for other models.

zur003:
Resolves #2261


zur003:
@hol353 - Any clue as to why the performance tests failed?

hol353:
Yes. A previous pull request (2313) changed the name of the predicted/observed table which we're using a test case for being able to rename tables in the performance site. It isn't fully complete yet and so every pull request will fail for the moment. Shortly I'll get you to do a pull and then repush to your pull request. This will fix the problem. I'll let you know when.

hol353:
OK, can you pull and then push back to this pull request

PR #2341, opened 2018-02-26
calculate fraction of bud burst at the budding phase, updating the me…

junqi108:
working on #1822


junqi108:
I updated the data set for grapevine. The stats are still under changing for the moment. So please accept the current stats. Thanks very much.

PR #2345, opened 2018-02-28
tidy up the model graph

junqi108:
working on #1822 
change the thermal time function for bud burst, change the target value from 51 to 150 for bud burst to improve prediction. tidy up the name and graph, add grapevine in report title for including other crops


junqi108:
retest this please jenkins

junqi108:
I improved the fit in the graph, so the stats need updates as well.

PR #2346, opened 2018-02-28
Merged Allan and Rob's Oat model development work

hut104:
working on #1413


hut104:
retest this please Jenkins

hut104:
The stats for this are mostly all improvements.  The stats for grain number are only slightly "worse" because I removed the observed values which were zero as it makes no sense to predict grain number prior to flowering.  The soil water stats were slightly different because allan added some extra soil layers.  Allan will need to revise these simulations at some stage now that all is merged.
I recommend merging this version.

PR #2359, opened 2018-03-01
Wheat Validation Leaf Graph Error

### LINKED ISSUE #2357: Wheat Leaf Number Graph has zero values
The leaf number graph in the wheat model has zero value predictions because the Leeston experiment is harvested when the crop is mature rather than on the real harvest date.  This results in the model providing zero leaf numbers late in the season.
The harvest rule could be removed and an operations file added to harvest after the final measurement date.

hut104:
resolves #2357


hut104:
retest this please jenkins

hut104:
This appears to now be fixed.  I'd suggest merging this.  Almost all changes are improvements with one small negative impact on one statistic.

Issue #2298, opened 2018-02-13
Grid view could use a description panel


her123:
The panel for the property presenter/grid view would be more functional if it included a html panel in the lower section to contain some doco for people entering in property values.

zur003:
Some of the recent modifications for CLEM now enable this sort of thing. If the Model being presented is given a Description attribute, a heading with the Model name and its description is placed above the grid. There's also provision for added a URI to obtain more information.
@hol353 - have you had a chance to look at this? If so, are you happy enough with the "look and feel" that we should start using it for models outside CLEM?

hol353:
I think the look and feel is good. Yes we could start using these outside of CLEM.

hol353:
I can't figure out how to make this work. Can someone enlighten me?

hol430:
@hol353 we now have a description panel in the property presenter. Can this be closed?

PR #2363, opened 2018-03-02
Enable dynamic resizing of CLEM-style description labels, and allow t…

zur003:
…hem to use Pango markup. Working on #2298


zur003:
Why are there differences in the Performance Tests?

hol353:
Yes I saw that. I'm not sure. Can you do a pull and then repush to see if that fixes it. If not, then ask Loretta.

PR #2377, opened 2018-03-05
Add units to graphs again and fix placement of labels on Event Graphs

### LINKED ISSUE #2376: There are no units on graphs.
Also the placement of labels on event graphs is not close to the event lines.

hol353:
Resolves #2376


peter-devoil:
(2018.4.9.2516 - OSX) graph Y axis label (units) is not updated when Y variable is changed.

PR #2409, opened 2018-03-06
chang root activity from negative to positive

### LINKED ISSUE #2391: errors in root activity calculation
in RootZoonState.cs, the calculation of RAw is not correct. 
Uptake of each layer is negative, so the value of RAw is negative, however after the calculation, a max was taken. So the value of Raw was always 1e-20

RAw[layer] = Uptake[layer] / LayerLive[layer].Wt
                                   * soil.Thickness[layer]
                                   * Soil.ProportionThroughLayer(layer, Depth, soil.Thickness);
                        RAw[layer] = Math.Max(RAw[layer], 1e-20);

junqi108:
resolves #2391 working on #1822


hol353:
retest this please jenkins

Issue #2370, opened 2018-03-04
Strip Crop Light Model


hut104:
The simple model of Fang Gou et al (2017) should be implemented into MicroClimate to allow strip crop simulations, but also to provide a light model for simple agroforestry systems.

sno036:
Sounds interesting!  Can you send the publication around?

HamishBrownPFR:
@hut104  can we close this issue now

PR #2434, opened 2018-03-07
Strip crop

hut104:
working on #2370


hut104:
The height model in wheat was changed to stop canopy growth late in the crop.  This has caused "improved" stats in the wheat model (very very minor changes).  Some long term simulations have decreased stats, but these are still under development.  
The height model has been adjusted to fit real data.  It gives better validation stats.  I suggest merging these changes.

PR #2446, opened 2018-03-08
clean up lots of zero O and P data that impacted on daily stats

### LINKED ISSUE #2380: Data incorrectly added to wheat observed data file
The Yucheng experiment data includes soil water data in mm stored in the columns for volumetric.  As a result, the soil water stats look terrible because it includes mixed data.

hut104:
resolves #2380


hut104:
The results are much improved for the stats targeted by this effort.  There are a few stats that have been adversely impacted by now including more late season observations in the analysis (ie we have delayed harvest to the observed final harvest date, thus including more "noisy" late season data in the PvsO tables).  However, stats are still OK and the results are better overall.
I suggest merging this change.

PR #2488, opened 2018-03-21
Return to the previous cursor and scroll position when returning to e…

### LINKED ISSUE #2440: Return to existing script line when returning to Manager script
When selecting between Manager components it would be nice to return to the last position in the script for each manager as APSIM Classic does.

zur003:
…dited Manger scripts. Working on #2440


zur003:
Retest this please Jenkins

zur003:
This mostly fixes the problem, but I'm not marking it as resolved just yet. This pull request restores the editing position correctly, but returning to a Manager node always restarts the user on the "Properties" tab. It might seem more intuitive if, when after editing the script, they were returned to the "Script" tab. (However, if this was done, it probably ought to be done with other tabbed views, such as Weather, to be consistent.)

PR #2501, opened 2018-03-26
Potato validation expansion

### LINKED ISSUE #2500: Potato validation tidy up
We have a collaborator who whats to expand the potato validation and we have got our hands on the dataset that Senthold used to validate Substore.  Do a clean up of the current validation to make sure it is a good base do build off.

HamishBrownPFR:
resolves #2500 .  I expect the stats will change because I found a problem with the set up of one simulation which had entered plant spacing in cm when it should be mm.  I have also tidied up the naming of experiments and factors and combined all of the data into a single file.  This has increased N in some cases.


HamishBrownPFR:
This failed the performance test because stats are showing up for node number and cover that were previously not showing up.

PR #2504, opened 2018-03-27
fixing infinite loop issue in leaf

### LINKED ISSUE #2503: Leaf class can get stuck in an infinite loop
The code in leaf has a while loop that does not necessarily exit.  It is part of the Apex model work. @byzheng

hut104:
resolves #2503


hol353:
retest this please jenkins

PR #2533, opened 2018-04-17
Maintenance respiration

### LINKED ISSUE #2531: Subtract maintenance respiration from daily fixation.
In the current version, the maintenance respiration is subtracted from `Live` component. After discussing with Enli, we suggest to subtract maintenance respiration from daily fixation before biomass allocation.

byzheng:
resolved #2531. 

The logic in this pull request is to subtract maintenance respiration after `SetupSupplies`, before `SetupDemands` in the `OrganArbitrator`. A new function is added to subtract `FixationSupply` according to total `Maintenance Respiration`.

* Remove all codes to change live weight caused by maintenance respiration, but keep the calculation of maintenance respiration in the same place.
* Add a new function to subtract fixation according to the total maintenance respiration according the the ratio of fixation in each organ. 
* Call the new function immediately after `SetupSupplies`.


junqi108:
Thanks very much, Bangyou. I will change this.

Yours Junqi

byzheng:
Note: This pull request depends on #2535 to fix the bug in the grapevine.apsimx.

byzheng:
The new changes include the maintenance respiration subtracts from daily fixation (photosynthesis), then metabolic and storage biomass in the live component if there are not enough carbon in the fixation.

hol353:
Is there any reason to have the 2 .apsimx files in this pull request?

byzheng:
GrapeVine.apsimx should not be in the pull request. I have reverted into the master version.
CanopyPhotosynthesis.apsimx included in the pull request to update the parameter values of maintenance respiration.

hol353:
Looks good. Is this now 'resolved' rather than 'working on'?

byzheng:
It is resolved. Have edited description/.

PR #2551, opened 2018-04-29
Use https when accessing resources on apsim.info site.

### LINKED ISSUE #2550: Upgrade functionality is broken
The upgrade form is currently unable to connect to the web service used to generate the upgrade list, most likely due to the recent move to https on the apsim.info site.

hol430:
Resolves #2550


hol430:
Although this will fix the issue in the long run, the problem remains that users will be unable to upgrade to this version due to their upgrade form being broken.

zur003:
Can we have the web site do a re-direct?

hol353:
Yep. I'm looking at it now.

hol353:
OK fixed. http works as well as https

PR #2607, opened 2018-05-15
Feature/plantain model#1468

### LINKED ISSUE #1468: Develop a forage plantain model
Plantain (Plantago lanceolata) is an herbaceous perennial plant with a broad distribution in grasslands of the temperate world. It consists of a rosette of basal narrow leaves from which one or more flowering stalks sprout in summer.  Since the 1990's it has been selected and used as a forage crop in New Zealand and its use is increasing worldwide. Plantain can be grown in pure or mixed swards, growing in a wide range of soil types, and it does not requires high fertility.
Given increasing use of forage plantain in farming systems, there is demand for an APSIM model that is able to simulate its growth and regrowth after harvest.  It is one of the forage crops considered in a New Zealand study with the objective of reducing N losses from grazing systems by manipulating the mix of forages grown, thus influencing N and water usage as well as animal diet.

rcichota:
Resolves #1468
Setting  up Plantain to go into release


rcichota:
@hol353 , the Plantain simulation has been moved and is running OK (thanks for your help), only the tests need to be accepted as there were none before.  This should conclude the steps to have Plantain released, and resolves this PullRequest.

PR #2613, opened 2018-05-21
Changed DoApexCalculation method to stop infinite loop

### LINKED ISSUE #2612: Infinite loop in DoApexCalculations
There is a For loop which reduces the value of apex number in Leaf.cs (whenever this value has been reduced in Structure). The apex number is trecked in Leaf as a list (**apexGroupSize**); the problem seem to be that the loop ignores the first item of taht list, so when trying to reduce the value and only the first item contains something, then the loop is never exited...

rcichota:
Resolves #2612 
Changed the for loop in the DoApexCalculations method to avoid infinite loop. Conter now goes to zero (instead of 1).


rcichota:
Hi @byzheng 
The changes I made to DoApexCalculations seem to have caused no harm to most simulations, except in one of the TarleeRotationTrial (fairly small). I do not know what the method I changed really does, but it was causing the simulations to stop in an infinite loop (on the second while loop). Allowing the reduction of apexGroupSize down to the index zero fixed the issue but you may see repercussions I cannot. Can you please check that the changes are not bracking anything and thus whether we canaccept the diffs in the Tarlee simulation. Thanks

byzheng:
@rcichota I have looked at your codes and make some modifications. It seems working for me. Could you have a look?

rcichota:
@byzheng 
I've checked what you changed and I can see two issues. One it the fact that the counter (i) is not allowed to go below 1, so the firts item of the list is never reduced. This is what was causing some simulation to go on an infinite loop. In some cases the value of Structure.ApexNum is set to a number smaller than one (something like 0.97). I did not test the code with your changes (was waiting to see if it passed the tests in Toowoomba), but but I can see that your change could make the original issue come back. I am trying to test this now.  Maybe the problem is that both Structure.ApexNum and apexGroupSize are not integer, as you commented, they were supposed to be...

byzheng:
@rcichota The counter is not allowed below 1 because the branching mortality is not caused the main stem death, only branching. I have added two Exception to check.  

I am not sure when Structure.ApexNum is less than 1. If you could send me an example file, I have have a look.

rcichota:
@byzheng The exception you put do catch the issue, I am getting the second one. This one stops when oats are growing, but I had the same problem arising whn fodderbeer was growing. FooderBeet doesn't even have stem. This highligts the fact that we should be carefull with methods we implement as the organs can be used in many different ways by different plants...
Here goes a simulation that crashes in sep-1972, this is when ApexNum of oats go below one. Hope this helps.
[RainshelterRotationSim(TestApex).zip](https://github.com/APSIMInitiative/ApsimX/files/2024541/RainshelterRotationSim.TestApex.zip)

byzheng:
@rcichota Thanks for your example. This problem should be fixed now. It seems I have misunderstood the unit of branching mortality. I assume it is the number of branch death in daily, but it actually is the ratio of branch death. It would be worth to clarify in the codes/documentation.

rcichota:
@byzheng We still have isues here. There are times when ApexNum goes below one and the only place to remove is what you call the main stem (item zero of the apexGroupSize). This is causing the error catch you added to be triggered several times (especially with oats). Is this crash suppose to happen (that is ApexNum should never go below one) or can it be ignored (revisited in the future)?  If reducing the first item of apexGroupSize is not recommended, can we modify the test to ignore any left deltaApex?
Thanks for helping with this, my team needs to be running several simulations for sensitivity analyses with oats and this issue is hindering us...

byzheng:
@rcichota I cannot understand why ApexNum can go below 1. ApexNum in Structure is initialised as 1 and increased by branching rate (which should be positive), reduce by branching mortality (which should be a ratio of branching (ApexNum - 1)). In this case, ApexNum should be always more than 1.   If you can send me a example file, I can have a look.

rcichota:
@byzheng I really don't have an idea of what is going on, but several tests with oats (found in the prototype folder) are crashing for that particular problem.
Would @HamishBrownPFR or @RobZys have any idea why, or whether it is a problem if ApexNum goes below one?

PR #2632, opened 2018-05-29
Fixed the potential area growth 

### LINKED ISSUE #2631: Actual leaf area expansion
@hol353 and @HamishBrownPFR  

I am trying to figure out why the actual leaf area is not equal to max leaf area in the cohort parameters  under no stress (cell division, expansion, carbon). All stresses have been disabled in a test simulation. However, the leaf size (actual live leaf size) is much bigger the parameter MaxArea  for Leaf 1 (400 vs 346), but smaller for leaf 2 (716 vs 770).  

For leaf 2, I have traced into Structure.cs ([L549-L522](https://github.com/APSIMInitiative/ApsimX/blob/3b4441a0514487c51281c1d02cf20cb3105052d7/Models/Plant/Structure.cs#L549)) which set the cohort Age into 9.13 Cd at appearance. Consequently, the leaf growth duration is reduced 9.13 Cd, then potential leaf size is reduced too. I don't understand why the CohortAge is set as a function of phyllochron at leaf cohort appearance. Could you explain the logic of these codes?

For leaf 1, I guess it is related with PotentialAreaGrowthFunction in [LeafCohort.cs](https://github.com/APSIMInitiative/ApsimX/blob/3b4441a0514487c51281c1d02cf20cb3105052d7/Models/Plant/Organs/LeafCohort.cs#L1015). The `Age + tt` should be capped by `GrowthDuration`, although `SizeFunction` has same slight different to reach MaxArea. I can submit a pull request for this change.

byzheng:
resolves #2631

### DIFF
--- a/Models/Plant/Organs/LeafCohort.cs
+++ b/Models/Plant/Organs/LeafCohort.cs
@@ -1033,8 +1033,14 @@ namespace Models.PMF.Organs
         /// <returns>(mm2 leaf/cohort position/m2 soil/day)</returns>
         public double PotentialAreaGrowthFunction(double tt)
         {
-            double leafSizeDelta = SizeFunction(Age + tt) - SizeFunction(Age);
-                //mm2 of leaf expanded in one day at this cohort (Today's minus yesterday's Area/cohort)
+            if (Age > GrowthDuration)
+            {
+                return 0;
+            }
+            double ageToday = Math.Min(Age + tt, GrowthDuration);
+
+            double leafSizeDelta = SizeFunction(ageToday) - SizeFunction(Age);
+            //mm2 of leaf expanded in one day at this cohort (Today's minus yesterday's Area/cohort)
             double growth = CohortPopulation*leafSizeDelta;
                 // Daily increase in leaf area for that cohort position in a per m2 basis (mm2/m2/day)
             if (growth < 0)


byzheng:
@hol353 The difference are caused by reducing the expansion growth of each leaf cohort (capped to leaf expansion duration in the size function). Can merge it if you think the difference can be ignored. 

I will separate the problem of cohort age on appearance into another issue.

hol353:
@HamishBrownPFR Can you please check this?

byzheng:
@hol353 Any problems about these changes to close this pull request?

HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
@byzheng can you provide me a specific example where this problem is occurring so I can debug it.  
With LeafSizeShape = 0.01; the The SizeFunction() should not be returning values that are any more that 1% greater than the max so I think the bug you have found is in SizeFunction()

HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
Hi @byzheng.

Dean has done some refactoring in Leaf/Stem/Apex which means this pull request is confilcted now.  Are you able to pull master, resolve conflicts and push this up again.

byzheng:
@HamishBrownPFR I have merged and pushed the master branch, but didn't notice any conflict.

HamishBrownPFR:
Sorry, conflict was the wrong word.  You needed to pull, merge any conflicts (if any) and then push again to get it to run on Jenkins.  Looks like this is what you have done, I can take a look at it now.

byzheng:
@HamishBrownPFR I have merged my changes into master branch. Could you have a look at performance test?

HamishBrownPFR:
@byzheng sorry for the hold up on this at my end.  I have had a couple of goes at looking into this but keep getting side tracked.  My concern is that the change you have made is only necessary because of a bug in the leaf size function code and I if this is the case I would prefer to fix the problem at the root.  I will look into it now

HamishBrownPFR:
@byzheng I have emailed you a spreadsheet which contains a simple prototype of the leaf size function.  If the shape parameter is kept at 0.01 LeafSize calculated by the model will always exceeded the value of the MaximumLeafSize parameter by 1% when Age > GrowthDuration.   As this was a small and constant deviation we decided it was acceptable and left the code like this.  
However, when you introduce your fix the differences in MaximumLeafSize in the performance tests are far greater than 1%.  This is why I suspect there is some bug in the SizeFunction code.
I will investigate further

byzheng:
@HamishBrownPFR I noticed the difference but it is only appearing after growth duration. See the picture for replotted parameter in your excel file.

When TT equals to Growth Duration (i.e. 100Cd), the leaf size equals to potential area (i.e. 10 mm2). So I don't think there is a bug in the size function.

image

byzheng:
@HamishBrownPFR I also tested several values of oneLessShape parameter which have the same trend.

image

byzheng:
@HamishBrownPFR Thanks.

Issue #2617, opened 2018-05-23
Issues with intellisense


hol430:
There are a few issues remaining with intellisense in the operations presenter. An unhandled exception under certain conditions, an incomplete completion list, etc.

hol430:
Another issue is that it takes a long time to initialize the intellisense presenter, due to the need to load in the assemblies used in one of the underlying data structures. In some cases, this initialization is actually being performed twice.

hol430:
Another issue is that if multiple nodes in the simulations tree have the same name, there's no way to tell which one the user is referring to. The wheat example is a good example of this. If the user types `[Wheat]` into the report, they could be referring to the Wheat simulation or the Wheat crop. Currently the intellisense will generate options for the first node it finds with the name Wheat - in this case, the simulation.

I'm not sure what to do about this - it's an easy trap to fall into, especially since it exists in at least one of our examples.

hol430:
An issue with intellisense in report/property/cultivar settings is that no completion options will be generated for members of an array element. 

I suspect that this issue will also exist with any other data structures using square bracket syntax (such as dictionaries), but we will deal with such cases if and when we encounter them.

rcichota:
I've got a couple of issues to add/clarify:
- the Intelisense is being raised any time a 'dot' is typed, this means that typing a number with decimals becomes quite annoying...
- currently the intelisense shows two lines, the full 'description' (incluing whether it is public, as gest and sets etc) and the description given in the xml tag 'Summary'.  Isn't only the later that is meant to be shown, in general? This is not an actual problem, only make the pop up thingy quite verbose...

hol430:
Yeah, I'm currently working on how the description is displayed - and as part of that, I'm going to add the type column back in, and move the type info out of description. 

The bigger problem with description is that many classes have a very long description (>10 lines). Currently, I'm only displaying the first two lines of the description, but I'm open to suggestions about what to do here. 


The point about attempting to generate completion options when you're trying to type a number is a good catch - I'll stop it from doing this.

rcichota:
I was thinking about those very long description too. They arose from we not having sorted out in time how the auto-documentation was going to be made.  In my opinion (and following the general recommendation from microsoft) these description should be revised and made one-liners. Additional tags should have been used for a more explanatory description (I use the 'remarks' tag when coding). However I am afraid of the size of the task or revising the tags...

rcichota:
[...]

PR #2637, opened 2018-05-30
Avoid use of the "??" operator, as it is not supported in VS 2015.

zur003:
Working on #2617

### DIFF
--- a/ApsimNG/Presenters/IntellisensePresenter.cs
+++ b/ApsimNG/Presenters/IntellisensePresenter.cs
@@ -56,7 +56,9 @@
         /// <param name="textEditor">Reference to the view holding the text editor. Cannot be null.</param>
         public IntellisensePresenter(ViewBase textEditor)
         {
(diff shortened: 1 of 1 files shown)


hol430:
Why does Visual Studio 2015 not support this feature? Is it because the compiler does not support C#6 features?

I only ask because I've been using the null-conditional operator `?.` which I believe as also added in C# 6. If `??` is not supported, I would not expect `?.` to be supported either.

zur003:
I'm not really sure about the details of the various C# versions, but it appears the compiler that comes with VS 2015 can cope with ?., but not with ??.
This raises the question of just which versions of C# and .Net we should be targetting and supporting. My own inclination is to continue to support VS 2015 stuff for a while longer.

hol353:
We need to continue to support VS 2015 for a while longer. If we need to avoid C#6 stuff then that's what we'll have to do.

PR #2692, opened 2018-06-13
Allow models to be disabled

### LINKED ISSUE #743: Allow components to be disabled (and enabled!)
You know what I mean!

hol353:
Resolves #743


zur003:
I wonder whether strike-through of the text might be better than a colour change. It may be less problematic for users who (eventually) use different coloured themes.

hol430:
In order to make the coloured text stand out, I also changed the colour of selected nodes in the explorer view's tree, which is something I'm having second thoughts about. If we used strikethrough, we wouldn't have to modify the colours at all, meaning that any active gnome/gtk themes should still work well with apsim.

hol430:
@hol353 See `OnSetCellData(...)` in ExplorerView, line 152. This is where changes are applied to the cell. To actually control what changes we want based on the state of the node, see `GetNodeDescription(...)` in ExplorerPresenter, line 971.

ApsimNG/Menus/ContextMenu.cs
```diff
@@ -635,6 +635,8 @@ public void Enabled(object sender, EventArgs e)
         {
             IModel model = Apsim.Get(explorerPresenter.ApsimXFile, explorerPresenter.CurrentNodePath) as IModel;
             model.Enabled = !model.Enabled;
+            foreach (IModel child in Apsim.ChildrenRecursively(model))
+                child.Enabled = !child.Enabled;
```
hol430: Do we want to toggle the child nodes, or actually set their enabled status to the same as the node which the user has toggled?

e.g. If the user disables a manager script, and then goes and disables the paddock, the paddock and all of its children will now be disabled, except for that manager script, which will now be enabled.

PR #2704, opened 2018-06-14
Don't generate intellisense for numbers in manager scripts.

hol430:
Working on #2617


her123:
Please avoid using poor (VB based) syntax such as "out _)". It doesn't build in VS2015.

PR #2714, opened 2018-06-18
Improve table formatting in Markdown HTML to PDF conversion.

### LINKED ISSUE #2655: Could use better Markdown documentation
The Apsim Next Gen "Memo" documentation at https://www.apsim.info/Documentation/APSIM(nextgeneration)/Memo.aspx does not accurately reflect the Markdown implementation we are using. For example, strike-through is not supported, and code blocks should be delimited with ~~~ rather than ```. This should be modified to better reflect what we actually provide.

It might be even better if we had a help dialog or something similar accessible from within the application

zur003:
Resolves #2655


zur003:
Retest this please Jenkins

PR #2735, opened 2018-07-02
Show model documentation in right hand panel even when IncludeInDocum…

### LINKED ISSUE #2729: Component descriptions not always appearing in GUI
When a component is selected in the explorer tree, but has no values to present, the GUI normally shows an HTML description of the component. However, if the component has been flagged as not to be included in the documentation, that description no longer appears. I think it would be better if the description appeared in the GUI, even when the component is excluded from the auto-documentation.

hol430:
…entation is false.

Resolves #2729


hol353:
Does this mean that if the user clicks on Simulations in the tree it provides doc in the right hand view for the whole simulation? This will be slow and not what we want.

hol430:
No, documentation for child nodes is not generated.

Issue #2557, opened 2018-05-01
R component in APSIM


hol430:
We need the ability to run arbitrary R code from APSIM. Users without an existing R installation will be able to run APSIM as usual, but will not be able to use this feature.

PR #2751, opened 2018-07-06
Allow Linux users to use the R component.

hol430:
Working on #2557 

As far as I know, this should work on macOS as well, but I haven't tested it yet.


hol353:
This is now conflicted.

PR #2769, opened 2018-07-20
Patched to allow Ctrl-Delete to work correctly.

### LINKED ISSUE #2750: Ctrl + Delete in text editor
When attempting to delete a word via ctrl + delete, only the first character appears to be deleted. Then, after moving the caret (via one of the arrow keys for example), the word is correctly removed. 

This appears to be an issue on Ubuntu as well as Windows, and occurs in all of instances of the mono text editor which I have tested - the manager view, report view, and operations view.

Ctrl + backspace appears to work fine.

zur003:
Resolves #2750


hol430:
Was this a bug with the monodevelop text editor?

hol430:
retest this now jenkins

zur003:
Yes, I had to tinker with the Mono.TextEditor.dll source to fix it. Basically it was modifying the text buffer as expected, but was not refreshing the display.

I sometimes wonder whether I ought to add a copy of the Mono.TextEditor source code to the repository. The main Mono repository for MonoDevelop, where this originated, is now at version 7.6, but I'm working off the 3-year-old 5.9.3 branch, because after that point, the editor became a lot less readily "embeddable". 

What would really be nice is to also be able to incorporate some of the MonoDevelop intellisense logic, but I don't think that would be easy to do.

I keep wondering whether there are better alternatives somewhere, but haven't found anything yet.

PR #2803, opened 2018-07-25
Don't immediately save file when version updates are applied.

### LINKED ISSUE #2461: Simulations are saved even when user says don't
If a simulations is open (say from the Prototypes folder) and the user expand a few of the nodes, perhaps copies a manager spcript, etc. (no changes actually made to the simulation) then choses to close the simulation, a window will pop up asking whether to save any changes in the simulation. No matter what answer is given, any 'changes' will be saved.  These changes are basically inconsequent: the version of the xmlSchema (line 2) and few minors things like replacing tabs with spaces, but also incluse model parameters that have been added after the simualtion was initially saved (I noticed CarbonConcentration in plant organs recently).  This is not a bug per se, but it is annoying as it makes SourceTree see uncommitted changes when in reallity you only wanted to have a look at a given simulation...

zur003:
Resolves #2461


hol353:
add to whitelist

PR #2807, opened 2018-07-25
Allow multiple irrigation operations to occur within a single day.

### LINKED ISSUE #2706: duplicate lines of Operations code being ignored for same operation
One of the sugar validation sims inadvertently had an identical operation line repeated three times (an irrigation event). NextGen would ignore the repeats and only apply one of the irrigations, whereas APSIM7.10 applied all three. Not even sure this is a fault - could be advantageous to leave as is? 

When I changed one of the irrigation numbers it applied only the final operation out of the three - so it is only applying one irrigation per date which could be a problem in some instances.

zur003:
Resolves #2706


zur003:
retest this please jenkins

PR #2808, opened 2018-07-25
Phenology refactor

### LINKED ISSUE #1946: Add Phase interface
Currently plant finds phenology phases by making a list of all classes of type phase so all phase classes must inherit for Phase.cs to be of the right type.  This means all phases must implement stuff that is in Phase.cs that they are not necessarily using.  For example leaf appearance and leaf death phase do not need ThermalTime in their source code but have to have a ThermalTime child so the link in Phase can be resolved.  As a short term fix I have made Tt optional with some code to handle things if it is present.  
A better solution would be to write a IPhase interface that all phase classes implement so the plant can collect phases of type IPhase but they don't have to have links they are not using

HamishBrownPFR:
resolves #1946

Don't merge yet.  IPhase implemented but I still need to refactor phase classes to tidy up


zur003:
retest this please, Jenkins

PR #2810, opened 2018-07-25
Remove redundant dates from Operations in base simulation that are ov…

### LINKED ISSUE #2619: Incorrect dates in Maize validation set
Operations for two of the experiments in the African section of the Maize validation set appear to be broken. If you look at the NRate operations of JMW2Katumani and MJW2Kiboko, you will find that the fertilizer is set to be applied in 1985, even though the simulation doesn’t actually start until 1988.

HamishBrownPFR:
…erwritten by replacements

resolves #2619


hol353:
add to whitelist jenkins

PR #2848, opened 2018-07-30
Swim is not releases so example should not be there

### LINKED ISSUE #2847: SWIM example should be removed
SWIM is not released

HamishBrownPFR:
resolves #2847


hol353:
retest this please jenkins

PR #2869, opened 2018-08-02
Feature/fodder beet oats rotation#2827

### LINKED ISSUE #2827: Add fodderbeet-oats rotation from PFR rainshelter
We will add the simulation to ApsimX\Tests\Validation\System. This is from a rotation experiment involving fodderbeet followed by oats performed at the rainshelter facility of PFR.

rcichota:
Resolves #2827 
Updating the simulation, tidying up reports and graphs


rcichota:
@hol353 , I need some help understanding the problem with this pull request. I have only changed the report and graphs of this test simulation. But the fail seem to stem from another simulation, it is missing a variable. I ran the simulation in my PC and indeed that variable is not available, though the simulation runs fine. It seems that the solute model is not being recognized. But (to my knowledge) I have all up-to-date... Can you see whether I am missing something? Thanks
image

rcichota:
This Resolves #2827 , the tests can be accepted as the only diff is the inclusion of a few more variables

PR #2871, opened 2018-08-02
Fertiliser refactor

### LINKED ISSUE #2835: Refactor Fertiliser Model
The fertiliser model has some redundant code that can be removed to improve maintenance.

hut104:
resolves #2835


hut104:
This commit has implications for Prawn modelling (Go Figure!) so let's not commit this work.

PR #2932, opened 2018-08-14
Fix display of EventNamesOnGraph

### LINKED ISSUE #2868: Bug in EventNamesOnGraph
when a series graphing a plant variable has an EventNamesOnGraph child it should show a verticle line at on the day each phase appears.  Currently is it putting a grid across the whole graph
image

hol430:
Resolves #2868 

This should also hopefully reduce the overlap in the event names.


hol430:
retest this please jenkins

Issue #2916, opened 2018-08-09
Create Installation system is currently offline


hol430:
The 'create installation' process on hasn't been rebuilt after migrating to the new Jenkins instance. This is needed fairly urgently as we have several weeks' worth of bugfixes which are unavailable to users.

hol430:
Re-opening, because the documentation is not being uploaded to apsim site.

PR #2940, opened 2018-08-16
Bugfix for create macOS installation

hol430:
Working on #2916 

I also reduced chocolatey's verbosity - the log file for the previous run had almost 40 000 lines of output.


hol430:
@hol353 can you merge this please?

Issue #2942, opened 2018-08-16
Crop.xml converter


HamishBrownPFR:
It would be good to have a tool that we can run locally, when refactoring crop model code, that runs each crop.xml file through the converter and saves it so these changes can be comitted.  Currently we have to copy each model into a replacement node in a .apsimx file, save it, open it so the converter is applied to the xml and then copy this back to the crop.xml in the source code.

hol430:
Would this be a standalone tool, or a feature built into Apsim? I'm guessing most regular users won't want or need something like this.

HamishBrownPFR:
This might not be straight forward.

I have just gone through doing this manually and the converter put a whole lot of stuff in that doesn't need to be there such as <enabled> tags.  This is a problem because it obscures the intended changes in version control.  I am now repeating the process and changing the version number in the .apsimx file I am using to be only one prior so only the intended changes are applied.

Conclusion.   Any converter tool would need to specify which version changes to apply to avoid putting unnecessary xml into the crop files

HamishBrownPFR:
After repeating the exercise I can further conclude, it is not the converter but the GUI that adds all the <activated> tags and reorders things as well.  This provides greater justification for producing a converter tool for crop xml files and the current procedure, putting them throuh the GUI to convert them, creates lots of cruft.

HamishBrownPFR:
@hol430 this would be a standalone tool that would only be used by people who are making changes to plant code

HamishBrownPFR:
I wonder if there are any parallels here with #2486.  A little GUI tool that opened a crop.xml and applied a converter on opening would be awsome

hol430:
Yeah it would be handy. We could have this accessible through Apsim, but only available to people building from source, by using the `#if DEBUG........#endif` syntax.

hol430:
So this tool just needs to open an XML file, run the converter to get this file up to the latest version, and then save the XML file?

hol353:
- The crop XML files don't have a version number in them. Would have to ask user what version to convert from / to. 
- The crop XML files don't exist in the released version of APSIM. Could use that fact to determine whether the tool is made available to user as a menu item in GUI.
- Rather than a separate tool, could be integrated into the GUI as a button on the home screen that is only available if crop XML files exist. The GUI wouldn't load the file like a normal .apsimx file so the 'cruft' shouldn't get added.

hol430:
Will the user ever need to take an xml file from a new version to an older one?

[...]

PR #2944, opened 2018-08-17
Thermal time re name

### LINKED ISSUE #2943: Further tidying of phenology code
Need to rename Thermal time to Progression and remove budding phase because it duplicates most of the functionality of generic phase

HamishBrownPFR:
resolves #2943.  I have depricated the budding phase becasuse its functionality duplicated that of generic phase and moved the few specific bits into a budNumberFunction class.  I have also renamed ThermalTime on phases to Progression because this is often not ThermalTime.

@hol353 I am not that happy with the crop.xml files I am submitting as the conversion process has added some cruft.  The alternatives are to resolve issue #2942 then redo this pull request of go through each phase in each crop model and change them manually.


HamishBrownPFR:
@hol353 don't merge this if it runs green (like thats going to happen), I have comitted a .apsimx file that I should not have.  I will wait for this pull request to see what other problems need fixing then submit another pull request

HamishBrownPFR:
@hol353 can I get your advise on what do do with this pull request as I am reaching the end of my tether with it.  It is failing the maize tests with some very minor diffs that could be accepted without to much issue.  However I would quite like to know where these diffs are coming from and use the checkpointing system to diagnose it.
There issue with the extra cruft that the converter put into the crop.xml files still persists.  I think I could remove most of the cruft with find and replace.  The tool we have been discussing in #2942 would help me remedy this problem but not sure if it is close or not.

hol353:
I am trying to work on the checkpoint code to do private members as well as public. I'm having some troubles getting it going. Most of my problems are finding time to work on it. I'm in Sydney this week and Brisbane next week. I'm working on it this afternoon and may get it going. If not it will probably be Friday or next Monday before I get back to it. For now just go onto something else and wait for me to get checkpointing going.

HamishBrownPFR:
OK, thanks @hol353.  I will be good to use this as a test case for the checkpointing and crop converter so I will hang fire until these are both operational.

hol353:
@HamishBrownPFR I've had a go at fixing the checkpoint mechanism. Can you give it a go. It should now write a .json file in the same directory as your .apsimx file. It should also now have private fields. It seems to work for phenology but may not work for other organs / classes.

HamishBrownPFR:
@hol353 I have got some checkpointing working well enough to debug this.  The differences in Maize stats all track back to a change in the way thermal time is accumulated in the germination phase where this pull request fixes a bug so thermal time is accumulated properly.  This has subtle effects on the timing of husk growth which feeds through to trivial differences in the stats.  Happy for these stats to be accepted and the pull request to be merged if the current build test runs OK.

PR #2961, opened 2018-08-21
Crop XML converter tool

### LINKED ISSUE #2942: Crop.xml converter

hol430:
resolves #2942 

This will only be available in debug versions, so #2956 must be merged before this.


hol353:
This is now conflicted.

PR #2970, opened 2018-08-22
Moved tests for models that are currently under review into Validatio…

### LINKED ISSUE #2969: Move tests for models under review into Validataion folder
Currently we are not checking stats of Oat, Barley, White clover, fodder beet or E Grandis models which are all under reference pannel review.  We really should be testing these because the models are nearly complete and we need to know they have not been changed between when they were submitted and when the review comes back.

HamishBrownPFR:
…n folder so their stats are considered in the performance testing

resolves #2969


HamishBrownPFR:
@hol353 this is failing the performance tests because barley, oat and fodder beet don't have baselines accepted yet.  Please accept and merge.

HamishBrownPFR:
These need to go to a different location

PR #3003, opened 2018-08-29
Refactor generic organ

### LINKED ISSUE #3000: Refactor Organ/Arbitrator interface
Currently DM and N supplies and demands are methods that do a whole lot of calcualtions.  They should really be properties with private sets in each organ on a specific event to reduce redundant calculation

HamishBrownPFR:
resolves #3000


jbrider:
@HamishBrownPFR you've subscribed to the DoPotentialPlantGrowth event in a number of the Organs - was that meant to be the new events setNDemand etc...

HamishBrownPFR:
@jbrider The SetNSupply, SetDMSupply and SetDMDemands are called on the DoPotentialPlantGrowth Event which is invoked by the clock.  The SetNDemand method subscribes to a [SetNDemands] event which is invoked by the arbitrator after potential allocations of DM have been done.  Does that answer your question?

jbrider:
@HamishBrownPFR Yes it does. Will this break the Priority ArbitrationMethod though as the setNDemands is event based?

jbrider:
@HamishBrownPFR Sorry, the last question was wrong.  I have a dependence between DMDemands which I think events would break... I thought it was doing it in the ArbitrationMethod, but it's doing it in the calculation of DM Demands, So it wasn't N specifically, but more a question on what these events will do to existing calculations that are reliant on the order of the Organs.

HamishBrownPFR:
The order that organs does things in comes from their order in the tree.  When the arbitrator class makes a list of organs if follows the order the organs are specified in the xml and then each time the arbitrator applies an action to a list of organs it is with top down progression through the list.  I am pretty sure that when an event is broadcast the order of handling also follows the order that models are specified in the tree.  Up to now we have done everything with no dependencies between organs so that the number and type of organs in a crop model can be arbitrary, it is one of the design philosophies of PMF.   Let me know if you need more explanation or give me some specifics on how you are doing your arbitration so I can think through the implications of what I am doing with regard to what you are doing

jbrider:
I have a DMDemand function that is reliant on the order in the tree, and I'm still in the middle of the Nitrogen arbitration, so I'm not sure on it just yet - as I mentioned previously, I have had a couple of false starts.  If the events fire in the same order it shouldn't be a breaking issue though.  My first reaction is not to trust event ordering, but I think that is influenced by UI events rather than the pub/sub pattern used inside the model.  
In simple terms, the N arbitration in sorghum calculates the Structural Demand first, then pulls any it needs from other organs, then partitions any remaining among all organs, then calculates grain's requirements, and again pull's from other organs.  PMF calculates all the pools first and the allocates. There is ordering happening within the requests to other organs for available N, and again in the Retranslocation.  Most of this ordering should be handled through the implementation of Phase functions within the Demand/Retranslocation which is what I am in the middle of implementing.  I was looking to implement a new Arbitrator to encapsulate all of that without affecting the existing PMF code too much and then refactor later.

HamishBrownPFR:
this one has failed because there are no baseline stats committed for the models I moved into tests.  Please check, merge and accept stats

HamishBrownPFR:
@hol353 this should be good to go now.  @jbrider  @hut104 @byzheng @junqi108 @Xiumei2017 fyi. 
 This all relates to another issue I am working on to externalise the parameterisation of the DM and N demands into the xml.  However, I am having to repent a number of other sins to achieve this.  Specifically I ran into problems with circularity in demand calculations so needed to fix their calculation to specific events.   To achieve this I also had to remove the inheritance of GenericOrgan from SimpleLeaf and Nodule.  These organs are now badly in need of refactoring to tidy them up but we will deal with that in a separate issue.

HamishBrownPFR:
p.s. @hol353 I am not sure why it is showing up 121 files changed.  This is wrong.  when viewed from the forked repo is correctly shows 10 files changed (see screen shot)

image

hol353:
There are a bunch of file moves from prototypes to under review folder in this pull request. They shouldn't be there. You must have done this refactor on the wrong branch. I can't really merge this.

hol430:
Those files were moved in #3004. I don't think Hamish has actually moved them in the pull request.

HamishBrownPFR:
No, none of those file moves are on my branch and I don't know why they are showing up in the pull request.  If I view the pull request from the origin repository where I made if from it only shows up 10 files changed

PR #3004, opened 2018-08-29
Move tests for models under review from prototypes into tests

### LINKED ISSUE #2969: Move tests for models under review into Validataion folder

HamishBrownPFR:
resolves #2969


HamishBrownPFR:
@hol353 this one failed because there are not baseline stats.  Please check, merge and accept stats

Issue #3001, opened 2018-08-29
Not all models displaying doco on website


HamishBrownPFR:
Wheat, SCRUM, Slurp and others are missing
![image](https://user-images.githubusercontent.com/7295858/44816985-124d4500-ac38-11e8-933a-5b24a8eadf8a.png)

hol430:
Yes, for some reason, it's failing to generate the plantain documentation with error message 

```
System.Exception: System.IO.IOException: The file '..\Documentation\PDF\Plantain.pdf' already exists.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalCopy(String sourceFileName, String destFileName, Boolean overwrite, Boolean checkHost)
   at Script.Execute(MainPresenter mainPresenter) in c:\Users\hol430\AppData\Local\Temp\ApsimX\53z5n5wc.0.cs:line 46
   at UserInterface.Presenters.MainPresenter.ProcessCommandLineArguments(String[] commandLineArguments) in C:\Users\hol430\Documents\ApsimX\ApsimNG\Presenters\MainPresenter.cs:line 754
   at UserInterface.Presenters.MainPresenter.Attach(Object view, String[] commandLineArguments) in C:\Users\hol430\Documents\ApsimX\ApsimNG\Presenters\MainPresenter.cs:line 93
   at UserInterface.UserInterface.Main(String[] args) in C:\Users\hol430\Documents\ApsimX\ApsimNG\Main.cs:line 44
```

Event though I am able to successfully generate plantain documentation locally. As soon as the script encounters an error, it exits without attempting to create documentation for the rest of the models. I will look into it.

PR #3005, opened 2018-08-30
Don't upload conference papers alongside model documentation

hol430:
Working on #3001

Do these conference papers even belong in source control?

```
ApsimX\Tests\Validation\Maize\Ciampitti Maize Population N Dynamics.pdf
ApsimX\Tests\Validation\System\Hudson\SR08104.pdf
ApsimX\Tests\Validation\System\Tarlee Rotation Trial\EA9950865.pdf
ApsimX\Tests\Validation\System\Wagga\Chan et al 2002.pdf
ApsimX\Tests\Validation\System\Wagga\Heenan et al 1994.pdf
ApsimX\Tests\Validation\System\Wagga\Heenan et al 1995.pdf
```

### DIFF
--- a/Docker/Documentation/doco.bat
+++ b/Docker/Documentation/doco.bat
@@ -20,8 +20,9 @@ if exist %apsimx%\results.7z (
 		echo Error unzipping %apsimx%\results.7z
 	)
 )
-
+dir %apsimx%\Bin
 cd %apsimx%\Documentation
+set FC_DEBUG=8191
 call GenerateDocumentation.bat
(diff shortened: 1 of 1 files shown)


hol353:
These papers are put in there by the model owner. I think we want to keep them. @HamishBrownPFR is this right?

HamishBrownPFR:
I think it was @hut104  who put them there.  Source control is probably not the right place to be storing references but I suspect Neil put them their because they are not easy to track down electronically.

Issue #2979, opened 2018-08-23
Ubuntu installs not working


zur003:
After doing an "upgrade" on Ubuntu, apsim is failing with a message that the DLL glibsharpglue-2 could not be found.

zur003:
It comes from elsewhere. The ApsimNG installer indicates that it is dependent on the gtk-sharp2 package (among others). The "glue" files are installed as part of that package, and should be installed in the global assembly cache by the gtk-sharp2 installer.
Things get a bit messier, though, with the "native" DLLs (as opposed to .Net assembly DLLs), as the names are different on different platforms. libglib-2.0-0.dll on Windows has counterparts named libglib-2.0.so.0 on Linux, and libglib-2.0.0.dylib on OS X. There need to be mapping specified in the .config files of the applications - see Mono.TextEditor.dll.config for an example. Files loaded into the GAC by the gtk-sharp2 installer should already have .config files set up for them.
Getting all this right can be a bit tricky.

zur003:
The OS X installer is "broken" in the same way.

zur003:
Part (all?) of the problem seems to that Windows DLLs are being included in the distribution. Instead of pulling in libglib-2.0.so.0, mono sees libglib-2.0-0.dll is sitting there and loads it, but since it's a Windows native binary, it chokes on it.

hol430:
I've just noticed an interesting issue in the Windows upgrade form. Whenever I click on one of the buttons, checkboxes, or textboxes, the main Apsim window moves in front of the upgrade form.

Is anyone else experiencing this?

hol430:
@zur003 I've manually created and uploaded a new macOS installer for Apsim version 2018.08.24.2916. Does this work for you?

zur003:
Almost. The OS X version also requires the file MonoMac.dll, which is in the repository in the ApsimNG\Assemblies directory. For the Mac version, this needs to be copied into the Bin directory.

hol430:
Re-opening this so I can add MonoMac.dll to the installer.

hol430:
I've just uploaded another macOS installer for Apsim version 2018.08.24.2916, which includes MonoMac.dll. Does this work for you?

zur003:
Yes, it looks like the right files are there now.

zur003:
The Mac OS X stuff now seems to by mostly working OK, although I get an error when the updater is trying to start up the new copy. That's a fairly minor problem; it runs just fine from Launchpad.

[...]

Issue #2629, opened 2018-05-29
Upgrade functionality is broken


hol430:
Upgrade functionality appears to be broken on Windows builds. Attempts to write to the registration database return a HTTP 404 not found error code. This causes the upgrade operation to be aborted.

zur003:
Even worse on Ubuntu Linux. Kills the application.

zur003:
This appears to be related to the switch from http to https about a month ago. I find that using http to call the Registration service appears to still work OK, but using https fails with the 404 error. I haven't looked into this too thoroughly, but https://serverfault.com/questions/776390/iis-7-wcf-webservices-giving-404-error-over-https suggests that it may be a problem with web.config on the service. Could someone with access to www.apsim.info look into this?

hol430:
Upgrading is now working once again, but I would like to fix the crash on Linux. Seems to occur when we try to display the error message dialog. I will look into it when I get the chance.

hol430:
Closing now, as this should be fixed by #3007

hol430:
Opening, as this seems to be broken again. A bug in the HTMLView causes this to fail, as the license can't be displayed.

PR #3007, opened 2018-08-30
Don't let download errors during update kill the application.

zur003:
Working on #2979


hol430:
#2629 should be closed when this is merged

Issue #2695, opened 2018-06-13
More functionality in Summary


sno036:
Nice to have in Summary:
 - search for a text string
 - standard keyboard controls for select all and copy
 - filter to show messages only from selected components (some models are overly verbose (e.g. a grazing generates four seperate messages a day from PMF)

PR #3019, opened 2018-09-03
Select all, copy, and text search for HTMLView.

hol430:
Working on #2695 

I've only added these features to the web browser used on Windows. Select all and copy (but not search) already works on the webkit browser used on Linux. Haven't tested on a Mac.


zur003:
On Mac, you can select a range and copy, but there's no obvious way to "select all" or "find".

hol430:
That's probably a feature you'll have to add yourself, as I don't have a Mac handy.

zur003:
I had a quick look at it, but put it in the "too hard for now" basket. There's not a lot of documentation to be found about working with Safari embedded in a Gtk# application.

Issue #3022, opened 2018-09-04
Expand checkpointing to make it more useful for debugging


HamishBrownPFR:
The check pointing concept is potentially very powerful for debugging.  Current limitations.

The SaveStateOfObject method writes the full simulation structure if the object argument is an IModel.  This means you can't checkpoint subsets of a simulation

It is not possible locate checkpointed chunks in simulation time.

The WriteToFile method creates a new file, overwriting the previous output each time it is called so you can't checkpoint to the same file more than once in a simulation

The JsonConvert.SerializeObject method serializes the whole object including its parents so even if you do specify a specific part of a model to checkpoint, it will ends up checkpointing the whole simulation, making a massive file.

hol353:
The SaveStateOfObject method has to write the whole simulation structure. This is because each model has a 'Parent' field that will be serialized. Lots of models also have [Links] to the simulations object or DataStore or Summary. All these will be seen by the serialisation engine. I tried nulling all parents and links but this didn't work because many properties rely on the links. I suspect writing the whole of simulation is the only option for the foreseeable future so we should probably get rid of the second argument. The json checkpoint file is around 10 Mb on my computer. They diff quite well.

The remaining points you highlight can all be achieved through the first argument to SaveStateOfObject e.g.

checkpoint.SaveStateOfObject(clock.Today.ToString("yyyy-MM-dd"), Chicory.Phenology);

The first argument is a file name. If you put the clock date as the filename you'll get multiple files that are timestamped and diffable. No overwriting will occur.

HamishBrownPFR:
I would be better if we could create a single file with date stamps in it rather than having to diff files one by one for each date until the problem is found.  

There seem to be some approaches for excluding certain properties from serialization.  I am trying to get my head around that at the moment but I fear it will be beyond my programming skills.

HamishBrownPFR:
Some stuff here that may work  https://www.newtonsoft.com/json/help/html/ReducingSerializedJSONSize.htm

HamishBrownPFR:
@hol353 what I have done below seems to work (See below).  I will do a bit more testing to confirm.  I note there is also some stuff in JsonSerializerSettings to ignore properties that are null rather than throwing an error.  Would this also be something that would be useful to implement?

PR #3023, opened 2018-09-04
Check point fixes

HamishBrownPFR:
working on #3022 

Hi @hol353.  I have made the changes you suggested but still found a number of limitations using the checkpointing to debug.  I have added some extra code to try to add the functionality I am wanting (addressing 3 of the 4 points raised in #3022 .  However, the idea still fails.  Esentially, if you try to checkpoint a subset of a simulation (I am trying to checkpoint Maize.Phenology daily so I can see what changes first) the Serialization includes the parent so you end up with unworkabley large files.

I am not sure if this can be resolved but am putting the pull request up so you can review and we can work it out.

### DIFF
--- a/Models/Core/Checkpoints.cs
+++ b/Models/Core/Checkpoints.cs
@@ -40,6 +40,49 @@ namespace Models.Core
                 WriteToFile(name, o);
         }
 
+        /// <summary>
+        /// Write a message line into checkpoint file
+        /// </summary>
+        /// <param name="message"></param>
+        /// <param name="name"></param>
+        public void WriteMessageLine(string name, string message)
+        {
+            if (CheckPointFile == null)
+                MakeCheckPointFile(name);
+            CheckPointFile.WriteLine(message);
+        }
+
+        /// The checkpoint file to write to
+        public StreamWriter CheckPointFile = null;
+
+        /// <summary>
+        /// Adds the status of the model to the CheckPointFile
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="o"></param>
+        public void AddToCheckpointFile(string name, object o)
+        {
+            if (CheckPointFile == null)
+                MakeCheckPointFile(name);
+            AppendToFile(o);
+        }
+
+        private void MakeCheckPointFile(string name)
+        {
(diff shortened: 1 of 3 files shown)


HamishBrownPFR:
@hol353 I think I have found a solution to the final problem so hold of merging this for a while

HamishBrownPFR:
@hol353.  What I have pushed up here enables debug checkpointing as outlined in #3022.

Rather than changing the existing checkpointing functionality I have added some additional methods to enable this.  All good to merge and close #3022 if you are happy with the way I have implemented this.  If not let me know.

Models/Core/Checkpoints.cs
```diff
@@ -53,5 +55,68 @@ private void WriteToFile(string name, object o)
...
+        /// <param name="name"></param>
+        public void WriteMessageLine(string name, string message)
+        {
+            if (CheckPointFile == null)
+                MakeCheckPointFile(name);
+            CheckPointFile.WriteLine(message);
+        }
+
+        /// The checkpoint file to write to
+        public StreamWriter CheckPointFile = null;
```
hol430: Should this be a property (e.g. have a getter and a setter)? I see it's also missing a summary tag, as are a few of the methods below.

HamishBrownPFR: Not sure why you would make that a property.  Not sure why not either mind you.

All the public methods have summaries, private ones don't need them

hol353: You shouldn't have public fields. Either make it a private field or a public property. Even private methods should have a summary comment.

HamishBrownPFR: I will change it to private.  Why is it that fields shouldn't be public?

hol353: It is part of the Microsoft C# standard that fields should not be public. If you Google it you'll see lots of discussions on this.

hol353:
This pull request appears to be an combination of multiple branches. e.g. there are a bunch of .xml files included that shouldn't be.

HamishBrownPFR:
Closing this pull request, don't have time to fix it now

PR #3029, opened 2018-09-05
small changes to get SimpleLeaf working for sorghum implementation

jbrider:
Working on #572

### DIFF
--- a/Models/Plant/Organs/SimpleLeaf.cs
+++ b/Models/Plant/Organs/SimpleLeaf.cs
@@ -80,7 +80,28 @@ namespace Models.PMF.Organs
         /// <summary>
         /// 
         /// </summary>
-        public int TipsAtEmergence { get; set; }
+        public int TipsAtEmergence
+        {
+            get
+            {
+                if (TipsAtEmergenceFunction != null)
+                {
+                    return (int)TipsAtEmergenceFunction.Value();
+                }
+                return _tipsAtEmergence;
+            }
+            set
+            {
+            }
+        }
+        private int _tipsAtEmergence = 0;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        [Link(IsOptional = true)]
+        public IFunction TipsAtEmergenceFunction { get; set; }
+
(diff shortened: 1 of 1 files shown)


hol353:
@jbrider You've added 2 optional functions. We're moving away from optional functions, instead making all functions mandatory. To modify all .apsimx files that use SimpleLeaf, create a file converter method in:

    ApsimX\Models\Core\ApsimFile\Converter.cs

If you have a look at the UpgradeToVersion10 method in this file you'll see an example of a change method for adding several constant functions to an organ.

I also notice that you an empty setting for TipsAtEmergence. This means that it can't be set. If this is intended then get rid of the empty setter.

jbrider:
Why the move away from optional? 
It seems like a better solution than writing a converter every time?

I must have missed bringing the setter back in from a snapshot - will fix that.

jbrider:
So the existing optionals that are in there, just haven't been removed yet?

jbrider:
TipsAtEmergence is in the ILeaf interface as an int - is it preferred to change TipsAtEmergence to an IFunction and change the interface to a double, or have the getter for TipsAtEmergence return the result of an IFunction as an int?  
I'm only interested in defaulting this value to a 1 so that Structure will start counting leaves - is it acceptable to save the TipsAtEmergence in the xml file as 1? It is completely hidden from the interface at that stage. I haven't checked yet if the documentation picks it up.

hol353:
We are working at getting rid of optional functions. I removed most (all?) from GenericOrgan a few months ago.

I would leave the interface returning an int then write a getter in SimpleLeaf that returns the int result of a 'TipsAtEmergenceFunction'. Then for any SimpleLeaf organ in all crops have the converter insert a constant function called 'TipsAtEmergenceFunction' that returns a zero so that existing behaviour is retained. Then for your sorghum crop model have it return 1.

I'm unsure what you mean when you say "... is it acceptable to save the TipsAtEmergence in the xml file as 1? It is completely hidden from the interface at that stage".

PR #3031, opened 2018-09-06
Check point fixes

### LINKED ISSUE #3022: Expand checkpointing to make it more useful for debugging

HamishBrownPFR:
resolves #3022 .  See also #3023 for discussion relating to this.  Had to kill that pull request because phantom files were showing up where weren't actually there.  Closing and opening again has dealt to them but I don't know why they showed up in the first place.


hol353:
@HamishBrownPFR This doesn't really resolve issue 3015. Issue number wrong? Also can you get rid of the CohortArrayLive in the other model resource xml files? It think they are also in wheat.

HamishBrownPFR:
Changed to correct issue (sorry about that), removed CohortArrayLive and CohortArrayDead from Wheat.xml and Potato.xml.  Removed full model xml from Microclimate test.  Added converter method to remove these from other models.

hol353:
One final thing sorry. You have microclimate.apsimx in there. Any reason for that?

HamishBrownPFR:
Yep, I did a search for CohortArrayLive and Dead and found some in that file.  Turns out it had full xml structure for the wheat and maize models so I changed this so it is using the released model

hol430:
@HamishBrownPFR MicroClimate.apsimx still appears to have two instances of the full XML for the wheat model, starting on lines 2673 and 8277. Is this intentional?

HamishBrownPFR:
@hol430  this was not intentional, I should have deleted them

PR #3032, opened 2018-09-06
Handle Command-A and Command-C in Mac web browser for "select all" an…

zur003:
…d "copy". Working on #3019 
Note that this replaces the MonoMac.dll assembly


hol430:
@zur003 How difficult do you think it's going to be to implement search functionality in the Safari/webkit browsers? It was super easy on Windows, all I had to do was send a ctrl + f keystroke event to the browser, but the same trick doesn't seem to work on the webkit browser. 

Guess I might have to modify the webkit assembly like you've done here with the MonoMac.dll.

PR #3039, opened 2018-09-07
Method to visualise released models

### LINKED ISSUE #2486: Method to visualize released models
The UI is really good for visualizing the structure and settings of models when they are in the replacements node.  However, when the model is released it is necessary to copy it back to a replacements node (which is a bad thing to do) to see the set up easily.  Otherwise we have to look at the xml which is slow going.

Possible solutions:
- Have model nodes display their full xml structure.  There was a bug recently that meant each crop model in the tree was showing all of the crop model structure below it.  This was quite handy but very dangerous as changes could be made to the model here.  If this could be done but with the content below the model read only and displayed by the GUI but not saved in the .apsimx file this would work.
- Have a button on the home page (something like "View released Models") which brings opens a simulation file that has the full structure of each released model.  Sort of like the Crops folder in the standard tool box but with each model expandable to show full model set up.
- Other ideas???

hol430:
Resolves #2486 

Right clicking any model which is built from a resource gives an option to 'view model structure'. This will show all children of the model in the tree, but they are read-only. The read-only mechanism works via the command system, so any commands added in the future will have to check if the model is read-only before executing. 

Any changes to the model from the UI which are not made via the command system will be possible, but such cases should be treated as a bug. 

It would be great if the command system was the only way to modify a model from ApsimNG.


HamishBrownPFR:
@hol430 this sounds really good.

Not sure what you mean by making changes via the command system.  If the model is read only it should not be possible to make changes in the GUI.  We still need to be able to change it if it is under the replacements node in the GUI.

hol430:
To make an undo-able change to a model from the GUI, it must use what we call a command. If you just go setting model properties as you would normally, e.g.

```
Manager.Script = null
```

this change won't be undoable. That said, I'm sure there are some presenters out there which do exactly this, but the problem is it's hard to track them all down. So if you ever find a change that you can't undo, let me know.

More info on the command system [here](http://apsimnextgeneration.netlify.com/development/userinterface/1-overview/) if you're interested. 

Good point about the replacements node though - I don't think I tested on Friday if this still works properly. 

@hol353 please don't merge this just yet

PR #3045, opened 2018-09-10
Fix display of experiments when some of the factors have only a single level.

### LINKED ISSUE #3043: Factor with one level
In a simulation, I have set two factors 1) Cultivar with two levels, 2) Management with one level

image

However, the Management factor is stored as the column name "Factors" and value as "Management". 

image

I would expect to store as the first Factor, e.g.

| Cultivar | Management |
| ------------- | ------------- |
| P100 | IHN |
| P150 | IHN |

In this way, I can mixed experiment with multiple levels and only one levels when plotting figures.

hol430:
Resolves #3043


byzheng:
@hol430 How are you going with this issue?

PR #3075, opened 2018-09-18
Add MonoMac.dll to Windows installer.

### LINKED ISSUE #3044: Error loading MonoMac
This is a bug that manifested itself when I clicked on a soil water node, although I suspect that any call to `APSIM.Shared.Utilities.ReflectionUtilities.GetTypeFromUnqualifiedName()` should produce this problem.

While attempting to enumerate the types in the ApsimNG assembly via `assembly.GetTypes()`, I get an exception `Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.`

I re-added MonoMac.dll as a reference to the solution and the problem went away. Unfortunately, git is not showing any files in my solution as being changed, so I'm not sure that I can push a 'fix'. Furthermore, I'm not sure I can even reproduce this problem anymore, now that I've 'fixed' it.

Is anyone else able to reproduce this problem?

hol430:
Resolves #3044


hol430:
retest this please jenkins

PR #3076, opened 2018-09-18
Add folder name to simulations table

### LINKED ISSUE #3062: Include Simulations in Factor table
The "SimulationName" field is not included in PO tables which means the only way of identifying a simulation if read outside of the GUI (in python for example) is by "SimulationID"  To overcome this problem APSIM puts a Factors table (example of a factor table read into python below) in the data store which provides an extremely useful index for matching SimulationID to simulation name and other useful data in the Factors table.

However, simulations that are not part of an experiment don't appear in the Factors table so it is not possible to match simulation name to simulation ID outside of the GUI. 
Would it be possible to have single simulations included in the Factors table.  They would not have values in the ExperimentName, FactorName, or FactorValue columns but getting the other information is necessary.

image

hol430:
Resolves #3062


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
Retest this please jenkins

hol430:
Retest this please jenkins

hol430:
Retest this please jenkins

hol430:
@hol353 There is something seriously wrong with the oil palm validation set. If you checkout your master branch and have a look at the oil palm validation set, you'll see that there are 3 experiments. Each experiment has a single factor with 3 levels. Therefore, you would expect each experiment to generate 3 simulations, and thus, you should end up with 9 simulations total.

For some reason however, Apsim reports only 7 simulations being run, and you can confirm this by looking around in the oil palm .db file. This problem seems to be fixed by this pull request (not sure why), hence the difference in stats.

PR #3078, opened 2018-09-19
Restructure the apex model

### LINKED ISSUE #3077: Apex model is not working as expected
In the Apex model, the group size and age are not working as expected especially when calculating the live stem number (the one we observed in the field)

byzheng:
resolves #3077

### DIFF
--- a/Models/Plant/Structure/ApexBase.cs
+++ b/Models/Plant/Structure/ApexBase.cs
@@ -215,7 +215,7 @@
         /// <param name="sender">The sender.</param>
         /// <param name="e">The <see cref="EventArgs"/> instance containing the event data.</param>
         [EventSubscribe("DoPotentialPlantGrowth")]
-        private void OnDoPotentialPlantGrowth(object sender, EventArgs e)
+        protected void OnDoPotentialPlantGrowth(object sender, EventArgs e)
         {
             if (phenology.Stage > 4 & !SenescenceByAge)
             {
--- a/Models/Plant/Structure/ApexBase.cs
+++ b/Models/Plant/Structure/ApexBase.cs
@@ -96,7 +96,8 @@
 
         /// <summary>Total apex number in plant.</summary>
         [Description("Total apex number in plant")]
-        public double[] GroupSize {
+        public double[] GroupSize
+        {
             get
             {
                 return apexGroupSize.ToArray();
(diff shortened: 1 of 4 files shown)


Models/Plant/Structure/Structure.cs
```diff
@@ -80,6 +81,10 @@ public class Structure : Model
         [Link]
         public IFunction branchMortality = null;
 
+        /// <summary>The leaf apex model</summary>
+        [Link(IsOptional = true)]
+        public IApex Apex = null;
+
```
hol353: The only reason you're including apex into structure is so that you can call DoCalculations. This doesn't seem right. Instead, can ApexBase subscribe to event 

[EventSubscribe("DoPotentialPlantGrowth")]
private void OnDoPotentialPlantGrowth(object sender, EventArgs e)

Also we don't want any optional functions.

Models/Plant/Structure/ApexBase.cs
```diff
@@ -28,8 +79,8 @@ public abstract class ApexBase : Model, IApex
         [ChildLink]
         private IFunction stemSenescenceAge = null;
 
-        /// <summary>The number of apex in each age group.</summary>
-        private List<double> apexGroupSize = new List<double>();
+        /// <summary>The apex group.</summary>
+        public List<double> apexGroupSize = new List<double>();
```
hol353: public fields aren't allowed. Turn into a public property starting with an uppercase letter.

byzheng:
The state variables apexGroupSize and apexGroupAge is updated every day in the structure or others, but not in the LeafCohort. In the LeafCohort, apexGroupSize and apexGroupAge is kept when leaf appearance (no update after appearance).

Not sure how to structure these codes. 

DoPotentialPlantGrowth can be used to call DoCalculation, but still need to find a place to store the Apex interface.

hol430:
retest this please jenkins

Models/Plant/Structure/Structure.cs
hol353: Bangyou, can you now remove this link in structure.cs. I don't think it is needed. The apex model will still be loaded into memory and will run without this link. If you keep the IsOptional link in LeafCohort then it should be able to find the apex model even though it is a child of the Structure class. Try it and see if it works.

byzheng: @hol353 I have removed the link in Structure and keep IsOptional in the LeafCohort as you suggested.

PR #3079, opened 2018-09-19
N demand refactor

### LINKED ISSUE #2202: Organ biomass demand and supply calculations need to be externalised
Currently each organ has a biomass demand function which determines total biomass demand and the partitioning of this into structural, metabolic and non-structural fractions is done in the code.  This makes it difficult to have periods where the organ may be demanding storage DM but not structural.  Similarly for supply from re-translocation and reallocation these are parameterised with a coefficient and the amounts to retranslocate are calculated in the code.  Also for N, demands are parameterised with different N concentration parameters and the actual amounts of N demand are calculated in the code.  

This all reduces the flexibility we have in parameterising DM and N supplies and demands without changing code, something that is becoming harder to do now so many models are using the same code.  I think the solution is to externalize the calculation of Demands and Supplies to and from each biomass component (Structural, Metabolic and Storage) into functions so these can be changed for different crop  models without effecting other models.  
So each day the organ code would ask a series of functions what their supplies and demands were and the functions would return these in g/m2.

HamishBrownPFR:
resolves #2202


hol353:
The only way it can happen is if the pull request has been run more than once. The fail then comes from the previous run.

HamishBrownPFR:
My pull requests normally fail at least once

hol430:
It was flagged as a failure within seconds of the pull request being created.

HamishBrownPFR:
@hol353 this is good to merge if you are happy with it.  I have made some minor changes to N demands in pushing them out into functions so expect some small changes.  All the red crosses are trivial changes on stats that were mostly poor for a start anyway.  This is balanced by a greater number of green ticks where trivial improvement in stats have been given.  Overall I conclude there is not meaningful change in the performance of the model and this is OK to merge

PR #3081, opened 2018-09-19
Add another edge case to factor code.

### LINKED ISSUE #3080: Factor with one level
Continue on #3043, this bug is still not fixed.

In the attached apsimx and screenshot. The table in the right should show the value of each level, not the factor name. 

image

[factors.zip](https://github.com/APSIMInitiative/ApsimX/files/2395300/factors.zip)

hol430:
Resolves #3080


hol430:
@hol353 This fixes one of the experiments (Wellcamp) in the Barley validation set, hence the changed stats.

PR #3083, opened 2018-09-20
removing slnDemandFunction as it has been made unnecessary.

jbrider:
working on #572

Is it preferred to write an undo function in the converter, or manually update the resource files and remove the previous converter function - it added a constant function that nobody should have used in the couple of days it was in the code base.


jbrider:
retest this please jenkins

hol430:
Yeah, I was testing some changes to the build system yesterday, which may have caused some problems.

retest this please jenkins

PR #3089, opened 2018-09-24
Fixes to morris to get it working with changes in R

### LINKED ISSUE #2984: Morris Elementary effects do not look right
The value of Mu star should approach a near constant value with the number of paths explored.  It now seems to diverge.

hol353:
Resolves #2984


hol430:
retest this please jenkins

PR #3091, opened 2018-09-25
Reduce the cohort population instead of max leaf area

### LINKED ISSUE #3090: Cohort population vs max area in the flag leaf
@HamishBrownPFR  In the leaf cohort model, FinalFraction is applied to reduce the max leaf area of flag leaf. Consequently, the maximum tiller number will be really sensitive to the time of terminal spikelet. 

I would suggest to reduce the cohort population of flag leaf, instead of max leaf area. The dynamic of flag leaf should be the same in the per square meter.

byzheng:
resolves #3090


hol430:
Retest this please jenkins

hol353:
@HamishBrownPFR Can you have a look at this small change? The diffs in performance stats are only very minor.

HamishBrownPFR:
@byzheng has run me through this.  He had implemented a good solution to the problem here and this should be merged

hol353:
Ok. I'll rerun it first as it is quite old.

Retest this please jenkins

Issue #2196, opened 2017-11-24
Error in simulation counter?


josharp22:
Whilst running a large factorial discovered that the number of simulations run can exceed the number of simulations in the file.

![image](https://user-images.githubusercontent.com/12078610/33191324-09cc843c-d11d-11e7-9b36-9a36c047bb6b.png)

PR #3096, opened 2018-09-26
Progressbar

### LINKED ISSUE #3093: Progress bar doesn't work properly when running sims
On computers with lots of cores (e.g. 32), running lots of sims, the progress bar doesn't accurately reflect progress. An error has also been seen in RunCommand.cs line 171:
                foreach(Models.IClock clock in jobManager.SimClocks)
                {
                    if (clock != null)
                        numberComplete += clock.FractionComplete;
                }
The error was saying that the enumeration had changed between iterations of the loop. That is because jobManager.SimClocks will be changed on a different thread.

hol353:
Resolves #3093

### DIFF
--- a/ApsimNG/Commands/RunCommand.cs
+++ b/ApsimNG/Commands/RunCommand.cs
@@ -168,10 +168,17 @@
             double numberComplete = 0.0;
             if (jobManager.SimClocks != null)
             {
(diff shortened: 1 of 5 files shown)


hol430:
Does this also fix #2196?

hol353:
Yes - probably. Close it and we'll see what happens.

PR #3102, opened 2018-09-27
Water and N arbitration events are triggered after growth is simulated

### LINKED ISSUE #3101: Water arbitration better to be triggered after growth is simulated
Currently 'DoWaterArbitration' is triggered before phenology and growth are simulated. It means that the model is using water demand calculated with the PM equation to estimate soil water uptake. If it is triggered after the growth is simulated, the water demand (i.e. actual water uptake) can be related to biomass assimilation (for example in a transpiration use efficiency-based model) and water uptake can be then calculated accordingly. Currently, water is extracted from soil before growth is simulated.

behnamxa:
Resolves #3101

### DIFF
--- a/ApsimNG/Views/UpgradeView.cs
+++ b/ApsimNG/Views/UpgradeView.cs
@@ -74,7 +74,7 @@ namespace UserInterface.Views
         /// </summary>
         public UpgradeView(ViewBase owner) : base(owner)
         {
-            Builder builder = ViewBase.MasterView.BuilderFromResource("ApsimNG.Resources.Glade.UpgradeForm.glade");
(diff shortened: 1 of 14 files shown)


behnamxa:
Water and N arbitration are done before and after phenology and growth are simulated. First time, it does not remove the demand from soil profile. After growth is simulated, arbitration is repeated and this time water and N demand/allocated are removed from soil profile. This way, modeller have the option to update 'actual' water/N demand inside the 'Photosynthesis' model. This change does not change the model outputs if an 'RUEModel' is being used.

hol430:
Retest this please jenkins

PR #3107, opened 2018-09-28
 Remove released models' structure from XML when saving a file.

### LINKED ISSUE #3040: Converter to remove full xml structure from released models
I have come across a few cases recently where simulations have the crops have full model configuration below them instead of using the released configuration.  There was a bug at one stage that caused this.  Should we write something into the converter so it strips out full configurations for released models and replaces them with the pointer to the released model

hol430:
Resolves #3040 

Small change to make all `ModelCollectionFromResource` instances serialise to their released version, unless they are under a replacements node. Users can get around this by renaming the model in the simulations tree.

### DIFF
--- a/Models/Core/Apsim.cs
+++ b/Models/Core/Apsim.cs
@@ -98,6 +98,23 @@ namespace Models.Core
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


hol430:
Retest this please jenkins

Issue #2069, opened 2017-10-15
Develop a white clover model (PMF)


rcichota:
White clover (Trifolium repens) is herbaceous perennial plant of the Fabaceae family.  It is native of Europe and central Asia, but now is spread worldwide, being recognised as the most important forage legume in temperate zones. White clover provides high quality feed for animals and is tolerant of grazing and trampling.  It is typically used in mixed swards, alongside grasses, in which it improves feed quality and it can supply N to the system through biological N fixation. The plants are short, with stems functioning as stolons from which petioles with three leaves and flower stalks grow.  Roots are generally shallow and also grow from the stolon nodes.
The objective here is to develop a basic white clover model using the plant modelling framework (PMF). In this project the focus in on describing the general growth pattern of monoculture, as well as the re-growth after cut/grazing and including N balance.  As PMF uses MicroClimate and SoilArbitrator to manage resource access, it should be readily possible to develop simulations with multiple species.  While developing the White Clover model, the need for it to be used in mixed swards simulations will be taken in consideration as much as possible.  However for fully testing this, a grass model is needed (none is currently available in PMF, although AgPasture may be an alternative).

rcichota:
@sarahcleary 
I am finally ready to cal the review of WhiteClover to a finish. I have finally got around to re-trace some to the steps taken during the review, eliminate some of the junk and align the simulation with new json format. All this was completed this week and Dean has merged into the main branch.
A lot of small modifications where done along the way to comply with the reviewers comments, which centred mostly on the documentation itself. Several of the changes are/will be useful for other models too.  The flow of the text, the ranking of items, the exposure of information about organs and parameters have been all improved, some of the graphs are still too cluttered, but this is something Dean at al, are looking into. I have also added a few more comments about the outcomes of the validation, but kept this to  minimum, as the amount of data was fairly limited it is difficult to point out why some of the results are not that good or more definite directions in which the model could be improved.  There is some work on going for red clover and diverse pastures that may help improving some of the biomass/N allocation which I think is not quite right in the white clover model. This will include some sensitivity analyses too.
I hope this is sufficient an answer to the review, but let me know if I need to expand further on any point.

sarahcleary:
thanks @rcichota 
@APSIMInitiative/reference-panel - please note @rcichota's comments above.  Please review and provide any feedback prior to the RP meeting on the 21/5,  
thanks
Sarah.

sarahcleary:
[...]

PR #3113, opened 2018-09-29
Feature/white clover model#2069

rcichota:
Working on #2069 
Updating model and documentation following review by AI.
No need to merge just yet


hol430:
Retest this please jenkins

hol430:
For some reason there was an error in the performance tests service. I've encountered the same issue in #3076. I will look into it.

rcichota:
Retest this please Jenkins

rcichota:
@hol353 
I think you can you merge this now.  Hope you don't have anything against the changes done to auto-documentation. There is still a couple of fairly minor issues that would be good to solve, but they are not crucial. As a lot of classes were modified, I fear conflicts will creep up if we don't merge this soon.
I also found an error in the script harvesting WhiteClover, so I need to fix and test that, and thus things are not related to documentation anymore...

hol353:
@hol430 There are some auto-doc changes in here that make me nervous given how flaky the auto-doc stuff is. Can you have a look at those changes and see if they are OK?

ApsimNG/Commands/ExportNodeCommand.cs
```diff
@@ -373,7 +373,14 @@ private void CreateGraphPDF(Section section, AutoDocumentation.GraphAndTable gra
...
                                               graphAndTable.xyPairs.Parent.Parent.Name + graphAndTable.xyPairs.Parent.Name + ".png");
-
+            int count = 0;
+            // If there are multiple graphs with the same name, they may overwrite each other.
+            // Therefore, we attempt to generate a unique name. After 20 attempts, we give up.
+            while (File.Exists(PNGFileName) && count < 20)
+            {
+                count++;
+                PNGFileName = Path.Combine(graphDirectory, graphAndTable.xyPairs.Parent.Parent.Name + graphAndTable.xyPairs.Parent.Name + Guid.NewGuid() + ".png");
+            }
```
hol430: Seems like this is the only change to the autodocs' infrastructure. This is actually one of my changes. It fixes a bug with some graphs in the autodocs using the same image. which resulted in some interesting graphs.

autodocsgraphbug

Models/Functions/MaximumFunction.cs
```diff
@@ -63,7 +63,7 @@ public void Document(List<AutoDocumentation.ITag> tags, int headingLevel, int in
 
                 // write children.
                 foreach (IModel child in Apsim.Children(this, typeof(IFunction)))
-                    AutoDocumentation.DocumentModel(child, tags, -1, indent + 1);
+                    AutoDocumentation.DocumentModel(child, tags, headingLevel + 1, indent + 1);
```
hol430: @hol353 This is also one of my changes. I believe that passing in -1 as the heading level was was a bug, but can't be certain. It's incredible that no one has noticed this causing problems before, so maybe it was intentional.

rcichota: I did some changes to the documentation of organs (adding few items there were not being documented, but should: photosynthesis, N fixation; and I played with the heading level to make the documentation looks nicer). I did also change a bit the documentation of Phenology and tried to modify Demands (the later is still not working fully, it is related to heading level. In these I've added the Document Method to few classes and this may need checking.

hol430:
@hol353 It seems that the autodocs will not display any heading which are more than 6 levels deep - see line 584 of `ApsimNG/ExportNodeCommand.cs`:

```
if (heading.headingLevel > 0 && heading.headingLevel <= 6)
{
    ...
}
```

As a result, there are several headings missing in the white clover documentation. Is there a good reason why we do this?

rcichota:
Retest this please Jenkins

hol353:
Nope other than the heading number gets silly e.g. 2.4.6.2.6.1

hol430:
Ah of course - there is no html \<h7\> tag. I guess if we wanted to allow for headings past 6 levels deep, we could just display these as bold/strong text.

hol430:
retest this please jenkins

rcichota:
Hmmm, it should be reasonable to think that more than six levels of headings should be unnecessary, hence the html limitation. But the way we can describe models in APSIM, with the possibility of using cascading functions, it is not difficult to get to level seven (at least in did not found difficult). In these cases (or at least in most of them) the numbering is not required, or even counter-productive. I was wondering if it is possible to identify whether a function being documented is within another function, in which case it would not need to be documented as a numbered item (or could be one level but not more than that, something like that).  This has to be done by the class that put all the documentation together, playing with the heading number in the Documentation method does not help because there is nothing in there setting  whether it should be numbered or not.  Besides a function could be in so many different levels...
I see also that if a  function has children it adds a "where:" line, so after this the items should/would not be numbered. They could be highlighted in bold of italics...

rcichota:
Hey @hol353 and @hol430 
Just want to confirm whether this can be merged or you still checking the changes I've made. I am more than happy to clarify and/or undo anything that you find ify or needs more thought into it. The tests ran green so it is a good opportunity to merge as the number of files with modifications is large. And I want to start committing some changes to the white clover model that will actually change its tests (I found a couple of errors in the parameterisation), so I don't want to complicate this merge even further...

PR #3123, opened 2018-10-04
Make BiomassRemoval models document all child memos.

### LINKED ISSUE #3114: memos not being printed in documentation from 'biomassremoval' nodes - Oats model
Doesn't matter which organ, memo's in the Oats model aren't being printed when in the 'biomass removal' nodes

### LINKED ISSUE #3115: unified table for biomass removal constants would be handy one day
It's not urgent, but when setting biomass removal defaults, it is currently necessary to enter values in 5 separate nodes (harvest, cut, prune, graze, thin). It would be great if all of these could be viewed/altered in a single table within each organ.

Additionally, it might also be useful to have biomass removal sitting outside replacements as a management action (e.g. same as 'harvesting'), so it doesn't need to get specified for each organ - all 5 organs could be collated in one place for this.....not sure what others think.

### LINKED ISSUE #3116: orphaned headings in documentation (minor)
At some point when the documentation generator is being improved, it would be good to make sure that headings can't appear at the bottom of the page when no text or graph from that section is going to  appear on the same page.

hol430:
Resolves #3114, Resolves #3115, Resolves #3116 

These memos will be documented after the table generated by the BiomassRemoval object, unless the memo contains a heading.

### DIFF
--- a/ApsimNG/ApsimNG.csproj
+++ b/ApsimNG/ApsimNG.csproj
@@ -211,6 +211,7 @@
     <Compile Include="EventArguments\LegendClickArgs.cs" />
     <Compile Include="EventArguments\NeedContextItemsArgs.cs" />
     <Compile Include="Interfaces\IMethodCompletionView.cs" />
+    <Compile Include="Presenters\BiomassRemovalPresenter.cs" />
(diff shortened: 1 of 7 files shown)


hol430:
retest this please jenkins

PR #3129, opened 2018-10-08
Check whether initial leaf area is more than max leaf area

### LINKED ISSUE #3128: Check whether initial leaf area is more than max leaf area.

byzheng:
resolves #3128


byzheng:
@HamishBrownPFR  Could you have a look at this pull request? In the source codes, the age of leaf cohort at appearance is depending on the initial area which make sense. However, the codes don't check whether the initial area is more than max area. I think it is not reasonable if initial area is more than max area. 

Actually, the biomass could be reduced to zero if initial leaf area is much bigger than max leaf area. 

For example, the test simulations with different initial leaf area of Leaf 1 with max area 260.
image

byzheng:
@HamishBrownPFR I don't track into details. But the major problem is Age at appearance could be more than growth duration if initial area is more than max area. 

Alternatively, we can set Age equals to GrowthDuration if initial area is more than max area

hol353:
@HamishBrownPFR Can you have a look at this please?

HamishBrownPFR:
@byzheng and I have organized a skype meeting to discuss this next week

byzheng:
@HamishBrownPFR As we discussed in this morning, these files have setup up the initial area is more than max leaf area for leaf 1. It will be better to let major developers of these models to fix them as it might be change the performance and need to reparameterize.

C:\ApsimX\Prototypes\Agroforestry\Odhiambo.apsimx
C:\ApsimX\Prototypes\Gliricidia\Gliricidia Validation.apsimx
C:\ApsimX\Prototypes\Agroforestry\AgroforestrySystem.apsimx
C:\ApsimX\Prototypes\Teff\Teff.apsimx
C:\ApsimX\Tests\Validation\MicroClimate\MicroClimate.apsimx
C:\ApsimX\Tests\Validation\Maize\Maize.apsimx
C:\ApsimX\Tests\Validation\Wheat\Wheat.apsimx
C:\ApsimX\Examples\Factorial.apsimx
C:\ApsimX\Examples\SCRUM.apsimx
C:\ApsimX\Examples\Maize.apsimx

byzheng:
@HamishBrownPFR I have checked Teff. The initial area is 380 but max area is about 267.

hut104:
Hi @byzheng 
Are these still required?  It may be worth closing this pull request.

byzheng:
@HamishBrownPFR Could you have a look at the pull request?

HamishBrownPFR:
@byzheng I had a look at this one quite some time ago but couldn't decide what to do with it and didn't get back to it.  Apologies for this.  I will re-run the performance tests and have another look

Retest this please Jenkins

HamishBrownPFR:
@byzheng to be able to test this we are going to have to change the models that are failing otherwise the error message you added stops the build and the testing wont progress.  Suggest we increase the parameters for MaxLeafArea so the first leaf is not initialised larger than its Max area.  Alternatively we could reduce the initial size of the first leaf to not exceed maximum but I am pretty sure these were set at the higher values to get correct early LAI predictions.  Perhaps try the changing the initial area first (that will be quicker) and if that causes big problems try the other approach.

HamishBrownPFR:
@hol353 there a few minor changes to performance in wheat and maize but I think we can live with them because they are trivial and it will be good to get this bug fix in

hol353:
retest this please jenkins

hol353:
@hol430 Why does this pull request keep getting stuck on Jenkins? It has been run several times now. Same result. Can you log onto Muck and have a look?

PR #3147, opened 2018-10-11
 Add row filter to data store view.

### LINKED ISSUE #3143: can we install a row filter for 'report'?
Would be handy to isolate a single experiment (or row/date even) - a row filter operating the same way as the column filter would do the trick

hol430:
Resolves #3143


hol430:
retest this please jenkins

hol430:
retest this please jenkins

PR #3154, opened 2018-10-14
Fix for event names on graphs not always working

### LINKED ISSUE #3153: Event names on graphs don't always work.
Seems to be a problem with empty series.

hol353:
Resolves #3153


hol430:
retest this please jenkins

PR #3170, opened 2018-10-17
'MicroClimate' potential transpiration should be able to be updated/replaced by other models

### LINKED ISSUE #3169: 'MicroClimate' potential transpiration should be able to be updated/replaced by other models
Currently, potential water demand is dependent on potential transpiration calculated by 'MicroClimate' and Penman-Monteith equation. If one needs to replace this value by another value calculated by, for example, the 'Photosynthesis' model (for instance, if one needs to calculate transpiration and photosynthesis with a TE-based model, as I do), it can be done before "DoWaterArbitration' is triggered. It needs a bit of change in 'ICanopy' and 'Leaf' as currently 'CalculateWaterDemand' gives 'PotentialEP' as the organ's water demand. These changes allow using any model except Penman-Monteith to calculate water demand whereas one can still use other state variables calculated by MicroClimate.

behnamxa:
Resolves #3169

- In **ICanopy.cs**, a new property **WaterDemand** was added to hold actual water demand. It is needed if one needs to replace/update crop potential transpiration calculated by **MicroClimate**.

- In **Clock.cs**, a new event handler **DoUpdateWaterDemand** was added right before **DoWaterArbitration** which allows one to update/replace potential transpiration calculated by **MicroClimate** in other models like the **Photosynthesis** model. It does **NOT** affect model performance if the event is not triggered.

- In **Leaf** and **SimpleLeaf** (in **PrennicalLeaf** it is done this way already), **Fw** is now calculated as WaterAllocation/PotentialEP instead of WaterAllocation/CalculateWaterDemand() (in which, **CalculateWaterDemand()** used to give **PotentialEP**, but now gives **WaterDemand** which is by default equal to **PotentialEP**). In **OrganArbitrator**, crop water demand is calculated as the sum of organs' **WaterDemand**s. Finally, **WaterAllocation** is divided by **PotentialEP** to calculate **Fw** (water stress factor). This way, one can use updated water demand (**WaterDemand**) instead of that directly calculated by **MicroClimate** to calculate **WaterAllocation** and then **WaterAllocation** is compared to **PotentialEP** (which can be updated when **DoUpdateWaterDemand** is triggered) to calculate water stress factor.

- In **Root.cs**, a new property **TotalExtractableWater** was added.


hol430:
You've got a merge conflict in the gitignore file

hol430:
Looks like the conflict to the gitignore file was introduced in 856e0ecfbb07b2502c7df80bd3ff2476d685de92. Actually, I notice that the old Morris example - which has since been deleted - was also added in this commit. I wonder if there are any other changes in this commit which should not be there. As a side note, the addition of the old Morris example explains why this pull request is failing on Jenkins. 

It should be possible to reset files to their state before this commit - if you let me know which git client you are using (e.g. sourcetree, github desktop), I can provide more specific advice. 

The gitignore file was also touched in commits 9d620b20ba4bf25ce3d1e73ade731a9e2d3d199e and 7c14c22be109ae21e041ffe8be3cebcefcfa2ae1, so it may be worth resetting this file to before these commits. Actually, the changes to `StorageDemandFunction` in 9d620b20ba4bf25ce3d1e73ade731a9e2d3d199e are also a bit odd, there may be a problem with this commit as well.

behnamxa:
Drew, Thanks for the explanation. I am using SourceTree. The problem is that this new branch was created by merging and old branch with the current master branch. So I am not sure how I can reset each individual file to their previous states. Moreover, I did not touch StorageDemandFunction. If anything is wrong, I think it happened during the merge.

hol430:
Looks like that fixed it. Not sure why the StorageDemandFunction shows up as changed in 9d620b2, but I guess since it's not showing up in the list of the pull request's changed files it's not a problem.

hol430:
retest this please jenkins

behnamxa:
I built the branch from scratch. But there is still problems.

hol430:
retest this please jenkins

behnamxa:
After meeting with Karine and @byzheng, we decided to discard the changes in 'CanopyPhotosynthesis'. Now the Photosynthesis model calculates daily assimilate under optimum water conditions.

behnamxa:
retest this please jenkins

PR #3191, opened 2018-10-22
Issue/sward re sow#3188

### LINKED ISSUE #3188: AgPasture sward does not grow after sowing
This only happens when using sward. After the plant(s) is killed by raising EndCrop(), and then sown, it will not grow. Single species do grow. It is probably caused by Sward not passing all the right info to species..

rcichota:
Resolves #3188 
I have fixed the issue by adding the code to check for 'pseudo-germination' in the sward, and fixed a resulting bug. whereby the biomass allocation at the day of emergence was throwing an exception (division by zero).
I also consolidated the two variables holding root information, with some tidying up of the code to make sure all root processes are executed in all zones.


rcichota:
Retest this please Jenkins

Issue #3181, opened 2018-10-19
No metabolic weight added to Live during SetDryMatterAllocation (GenericOrgan)


RobZys:
The code for SetdryMatterAllocation in GenericOrgan  can  have allocated metabolic weight but there is no line of code assigning this new allocation to the Live biomass

RobZys:
Issue resolved in pull request 3192

PR #3192, opened 2018-10-22
Metablic wt fix

RobZys:
Working on #3181


RobZys:
@hol353  Dean can you merge this as it resolved a bug in the Simple organ not having metabolic biomass

PR #3195, opened 2018-10-23
Report functions should attempt to find dates stored in other models.

### LINKED ISSUE #3194: Report functions which use dates stored in other models
According to the report documentation, the to/from dates in an aggregate function can be:

> Any APSIM event (e.g. [Wheat].Sowing) or date (e.g. ‘31-Dec’, ’15-Jan-2001’)

However, it would be useful to be able to use a date stored in another object. e.g. `[Clock].Today`

For example, if we have a manager script which has a public `DateTime` property, we should be able to use the value of this property in a sum in the report.

```
Sum of [Weather].Rain from [Field].Manager.Script.CustomDate to [Clock].EndDate as TotalRainfall
```

hol430:
Resolves #3194


hol430:
retest this please jenkins

hol353:
Does this actually work? When ReportColumn constructor is called it will be at the start of the simulation. The assumption is that the fromValue and toValue have already been calculated by this time. This may not be the case. Does this need to get the values every day of the simulation?

hol430:
My use case was fetching dates stored as properties in a manager script. This seemed to work in my testing a few days ago if I recall correctly. I believe the script's properties get set immediately after it is compiled.

PR #3225, opened 2018-10-26
Leaf.FRGRFunction should not be dependent on/linked to Photosynthesis model

### LINKED ISSUE #3224: Leaf.FRGRFunction should not be dependent on/linked to Photosynthesis model
Currently, **Leaf.FRGRFunction** is linked to **Photosynthesi** model and reads **FT**, **FN** and **FVPD** from that model. If a user accidentally/intentionally replaces the photosynthesis model with a different model and forgets to take care of **FRGRFunction**, **CanopyConductance** will not be calculated correctly. It is the case with **CanopyPhotosynthesis** model which does not calculate these three stress factors (unlike the default RUE model). 

I suggest calculating all three stress factors locally, instead of reading them from the **Photosynthesis** model.

_It's best to remove VPD and MeanT from everywhere in the code and read them at run-time. That way, user can decide how to calculate these daily mean variables. But many of the models need to be changed if that is to be done._

behnamxa:
Resolves #3224

- I am working on **Wheat**.

- These suggested changes are **NOT** scientific. Just the way a few state variables are calculated/read. This changes will make sure if one changes/replaces the default **Photosynthesis** model, **Leaf.FRGRFunctionis** is still correctly calculated,

- **Leaf.FRGRFunction** is now independent of the **Photosynthesis** model. All three stress factors (**FT**, **FN** and **FVPD**) are now calculated locally. Previously they were read from the (RUE) **Photosynthesis** model and if that model was changed, **FRGRFunction** would be always zero.

- **DailyMeanVPD** (similar to that of temperature) was added to the code and used in **Leaf.FRGRFunction** and RUE **Photosynthesis** model. One can set the weight of VPD at maximum temperature while calculating daily mean **VPD**. The default weight is 0.66, like in the **RUEModel**.

- **VPD** in the **RUEModel** is now read from **IWeather** or is set zero if **IWeather** is null (_I would remove **VPD** from **RUEModel** and just read the **FVPD** from the model tree, as I did for Wheat. But some other crops still need it_).

- **VPD** in hPa was added to **IWeather** (i.e. **Weather**, **ControlledEnvironment**, **MockWeather** and **LocalMicroClimate**), based on the same equation used in the RUE model (0.66 VPDmax + 0.34 VPDmin). It is useful if one does not want/need to use a different **VPD** formula and needs a quick access to its value in the code.

- In **OilPalm**, the unit of **VPD** was corrected. The redundant **svp** function was replaced with the same function in the utilities.

- It's best to remove **VPD** and **MeanT** everywhere in the code and read them at run-time. That way, one can decide how to calculate these daily mean variables. But many of the models need to be changed if that is to be done.


hol353:
I'm unsure why you have both DailyMeanVPDFunction and DailyMeanVPD with duplicated code. I can't see any reason to have both. It looks like DailyMeanVPDFunction exists to do the same calculation as DailyMeanVPD and then do a linear interpolation on some X/Y pairs. We already have a function to do linear interpolation where you can point to a model for the X variable. Couldn't you use this?

Also, I'd like @HamishBrownPFR  or @hut104 to have a look at the science changes in this.

hut104:
There are similar issues with temperature functions.  We embodied very simple science (e.g. how we average two numbers) into code.  It could be removed and we could add the averaging in a simple function.

behnamxa:
@hol353 hi, I just wanted to follow previous leads on temperature. We definitely don't need the 'Function'. Should I remove it and recommit? And the science is the same. Just the place stress factors are calculated/read from.

behnamxa:
@hol353 If I remove the function, then it is needed to always have **VPD** (calculated by **DailyMeanVPD**) besides the **FVPD**. Is it OK? My idea was to avoid making the tree too busy.

image

behnamxa:
OK. I came up with this: I removed **DailyMeanVPD** and made the **XYPairs** in **DailyMeanVPDFunction** optional. So, user has the option to use VPD value or the interpolated value from a linear function. This way, no need to add **VPD** separately everywhere it is needed beside **DailyMeanVPDFunction**. Do you approve this approach?

hol353:
We don't like the idea of optionals. We want just one way of doing things. I would do it the other way around. Just get rid of DailyMeanVPDFunction. It isn't needed at all.

hol353:
Your screenshot is OK to me. You could make VPD a child of FVPD.

behnamxa:
I just removed the function and used the daily mean VPD in **Photosynthesis** model and in **FRGRFunction** along with a linear interpolation function. Sorry for multiple commits.

PR #3231, opened 2018-10-29
Managers should not remove all children after rebuilding script model.

### LINKED ISSUE #3230: APSIM crash if a memo is attached to a manager
I wanted to add a more thorough description of a script manager that controls defoliation of PMF plants.  This would include rationale and usage instructions, so a bit too long to simply add to the code, so I wanted to use a memo. However, APSIM crashes badly, and/or does not save the added memo.  I tried this when the manager was in the 'Replacements' as well as within a simulations and APSIM always ended up crashing (though not always in the first try) and loosing anything added to it.

hol430:
Resolves #3230 

This fixes the memo not being saved, but not the crash, which I have not been able to reproduce.

### DIFF
--- a/Models/Manager.cs
+++ b/Models/Manager.cs
@@ -240,7 +240,7 @@ namespace Models
                             SetParametersInObject(Script, parameters);
 
                             // Add the new script model to our models collection.
-                            Children.Clear();
+                            Children.RemoveAll(x => x.GetType().Name == "Script");
                             Children.Add(Script);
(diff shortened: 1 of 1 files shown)


zur003:
This probably fixes the crash as well. I could produce a crash if I added a memo, went back to the script to edit it (to force recompilation), then tried to return to the memo.

hol430:
Yeah that did the trick. It seems to be fixed by this change.

PR #3266, opened 2018-11-08
Added /m command line switch.

### LINKED ISSUE #3262: Add a multiprocess switch to models.exe
Would be handy if we could use the experimental multi-process job runner from the command line.

hol430:
Resolves #3262


hol353:
Do you want this merged?

hol430:
The /m switch doesn't respect the /verbose switch if it's provided, so I'd like to fix that first.

PR #3275, opened 2018-11-13
Adding private variables to store water demand

### LINKED ISSUE #3276: Problem with factorial simulations when private fields were not used
When I did not use private fields to store the values of the public properties, I got strange results when factorial simulations were run simultaneously.

behnamxa:
Resolves #3276

I seems necessary (I am not quite sure) to have private fields to store water demand, just like **PotentialEP**, especially when multiple simulations (e.g. factorial experiments) are being run simultaneously. If it is, it must be done in other models as well. I just made changes to those models that call **WaterDemand** (as I added this variable to **ICanopy** weeks ago).

### DIFF
--- a/Models/Plant/Organs/Leaf.cs
+++ b/Models/Plant/Organs/Leaf.cs
@@ -178,9 +178,14 @@ namespace Models.PMF.Organs
             }
         }
 
+        private double _WaterDemand;
         /// <summary>Sets the actual water demand.</summary>
         [Units("mm")]
-        public double WaterDemand { get; set; }
+        public double WaterDemand
+        {
+            get { return _WaterDemand; }
+            set { _WaterDemand = value; }
+        }
 
         /// <summary>Sets the light profile. Set by MICROCLIMATE.</summary>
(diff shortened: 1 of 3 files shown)


Models/Plant/Organs/Leaf.cs
```diff
@@ -178,9 +178,14 @@ public double PotentialEP
             }
         }
 
+        private double _WaterDemand;
```
hol430: Suggest checking out the [code style guidelines](https://apsimnextgeneration.netlify.com/development/model/2-codingstyle/). Specifically:

> The camelCasing convention, used only for argument names to methods and private field names, capitalizes the first character of each word except the first word

> Do not use underscores to differentiate words, or for that matter, anywhere in identifiers.

hol353:
I can't see the point of this PR. What problem are you trying to solve? There shouldn't be any need to have a private field.

behnamxa:
Thanks Dean. I knew the styling rules. But I followed previous leads, the way you had named PotentialEP. I will fix and recommit. Thanks.

behnamxa:
Oh sorry I didn't see the second comment. OK, no recommit then.

PR #3288, opened 2018-11-15
Add new simulations into the SoilArbitrator test/validation and delet…

sno036:
Add new simulations into the SoilArbitrator test/validation and delete the old v2 validation simulation

Resolves #3288


sno036:
Dean - I think I've done this right.  Forgot to add 

Resolves #3288

PR #3293, opened 2018-11-16
Allow changes to memos to actually have an effect.

### LINKED ISSUE #3292: Changes to Memos have no effect
Changes to a memo in the UI have no effect - after changing to another node and back to the memo again you will find that the change was not saved.

hol430:
Resolves #3292


hol430:
@hol353 As far as I can tell, all other `ChangeProperty` calls which I haven't modified are still functional...for now.

PR #3298, opened 2018-11-19
Renamed obs to match sims in RS1718

### LINKED ISSUE #3297: some oats validation names don't match
The 2017-18 rainshelter trial experiment got renamed but the obs data didn't

RobZys:
Resolves #3297


RobZys:
@hol353 Dean can you accept these stats as no change to model just aligned the naming of the observed

PR #3301, opened 2018-11-19
Tests for simulations lacking predicted/observed data

### LINKED ISSUE #3289: Jenkins testing needed for simulations that don't have predicted / observed data
For tests that have predicted / observed data, Jenkins makes sure that the goodness of fit doesn't change. For other tests that don't have predicted / observed data, Jenkins simply ensures that they don't throw exceptions. These tests could fail (without throwing an exception) and we wouldn't know. This needs to change.

What we need is way of specifying a yes / no test that runs post simulation. If this test fails then Jenkins should flag it as a fail. One way we could do this is to allow a manager component to be a PostSimulationTool. The user could then write a script method that returns a boolean e.g.

[Link]
IStorageReader db;

bool DoTest()
{
    DataTable data = db.GetData("Sim1");
    double[] yields = DataTableUtilities.GetColumnAsDoubles(data, "Wheat.Grain.Wt");
    if (MathUtilities.Average(yields) > 1.5)
        return true;
    else
        return false;
}

hol430:
Resolves #3289 

@sno036 Once this has been merged, you can create a manager script to test your results. A test script should look something like this:

```csharp
using System;
using Models.Core;
namespace Models
{
    public class Script : Model, ITest
    {
        public void Run()
        {
            bool testHasFailed = true;
            if (testHasFailed)
                throw new Exception("oh dear");
        }
    }
}
```

Note that, for now, Apsim will run any such scripts, no matter which simulation they are in. I've also added some documentation to the website for this feature.

### DIFF
--- a/ApsimNG/Commands/RunCommand.cs
+++ b/ApsimNG/Commands/RunCommand.cs
@@ -52,7 +52,7 @@
             this.jobName = model.Name;
             this.explorerPresenter = presenter;
             this.explorerPresenter.MainPresenter.AddStopHandler(OnStopSimulation);
-            jobManager = Runner.ForSimulations(explorerPresenter.ApsimXFile, model, false);
+            jobManager = Runner.ForSimulations(explorerPresenter.ApsimXFile, model, true);
 
             if (multiProcess)
                 jobRunner = new JobRunnerMultiProcess(false);
--- a/Docs/content/Usage/Tests.md
+++ b/Docs/content/Usage/Tests.md
@@ -1,0 +1,31 @@
+---
+title: "Test Results"
+draft: false
+---
+
+To test your data, you can add something like this to a manager script:
+
+```csharp
+using System;
+using Models.Core;
+namespace Models
+{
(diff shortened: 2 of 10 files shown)


hol430:
This is failing because we previously were not using the `/RunTests` command line switch on Jenkins, and so any `Test` nodes were not run. The weirdo prototype is now reporting significant differences when regression testing SoilWaterContents. 

The slurp example is also failing, although I wonder if this ever worked. We are using `Clock.Today` to match the predicted and observed data, but the dates in the observed file `Tests\Validation\Slurp\ShallowSoilTest\SoilWaterContent.xlsx` are not even close to the dates in the predicted table.

hol430:
The performance tests were still run - so the released models were being validated. But yes, it is concerning. Fortunately, only two simulations seem to be failing their tests. It is going to be difficult to track down the source of the problem in the weirdo prototype.

sno036:
WEIRDO and PMF/Slurp are both @HamishBrownPFR I think.

Issue #1312, opened 2016-10-31
Tropical Pasture Model for PMF


hut104:
Projects currently require pasture models for Brachiaria and Bambatsi Panic.  Will start reproducing the Bambatsi model from APSIM 7.8 and will bring in data from Brazil for Brachiaria.

sarahcleary:
@hut104 - in an RP session this morning it was suggested that this model should go in review.  It was noted that it needed some additional work/assistance but also noted that some publications were referencing this model.  Are you comfortable for this to be reviewed?

hut104:
Am following up with collaborators.

________________________________

hut104:
... and the test isn't running.  Someone's broken it - so don't put it out for review.

________________________________

hol353:
When you get it running you'll need to create a PR to move the files into the Tests\UnderReview directory. Make sure there is an example in there and the docs look good.

pablobarbera1:
Hello. We have started to calibrate the tropical pasture prototipe for Setaria sphacelata in northern Argentina conditions. We have 3 years of observed data (aboveground biomass: leaf, stem, live, dead) of a rainfed cutting experiment with 2 N levels. First summer was particularly dry and the next summer was quite wet. Soil water seems to work well for obs-pre data (0-80 cm deep). The leaf.fw values are lower for the first summer, but aboveground growth does not respond well. Water stress didn´t affect the plant in the simulation as in observed data. Would anyone suggest calibration of FW or FVPD? Thanks!

sarahcleary:
@hut104 and/or @hol353 ??

pablobarbera1:
Here I attach 2 plots for a better understanding. 0 and 345 are N levels. Last growth period of 2017 and first of 2018 were very dry, as seen in the Leaf_fw plot. Thanks!
![AbovegroundWt](https://user-images.githubusercontent.com/70026808/121511857-798aad00-c9bf-11eb-9d78-ebffb5e9d668.png)
![Leaf_fw](https://user-images.githubusercontent.com/70026808/121511968-9921d580-c9bf-11eb-8312-b98656a561f5.png)

PR #3315, opened 2018-11-22
Tropical pasture

hut104:
working on #1312


JJguri:
Is there some example of Tropical Pasture in NG? 
Some colleagues and I want to calibrate this model using a big dataset of Setaria sphacelata from Argentina. Thanks, @hut104 @hol353

PR #3322, opened 2018-11-23
Updated the performance tests collector to work with json apsimx files.

### LINKED ISSUE #3314: The simulations "Under Review" do not appear in the performance tests
It would appear that results for tests for models "Under Review" are not included in the performance test results.

hol430:
Resolves #3314 

@hol353 I haven't tested this, but if the Eucalyptus validation set appears in the performance tests portal, this will have worked.


hol430:
Please don't merge this just yet. If it works, I will remove all of these binaries from ApsimX, and get Jenkins to manually clone and compile the performance tests collector.

PR #3333, opened 2018-11-27
Ensure few variables are zeroed after an EndCrop is raised

### LINKED ISSUE #3332: Minor discrepancy in Agpasture growth after sowing
I noticed that white clover seem to have an advantage start just after sowing a mixed sward. This seems to be related to the variable fixedN not being zeroed after a EndCrop.

rcichota:
Resolves #3332


hol353:
retest this please jenkins

PR #3364, opened 2018-12-06
Sorghum devt

jbrider:
Working on #572


Models/Functions/AccumulateAtEvent.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>The accumulated value</summary>
+        private double AccumulatedValue = 0;
+
+        /// <summary>The child functions</summary>
+        private List<IModel> ChildFunctions;
+
+        /// <summary>The phenology</summary>
+        [Link]
+        Phenology Phenology = null;
```
hol430: These private fields should start with lower case letters.

hol353:
I'll merge this on the proviso that the phenology flags and the optional functions are removed at some point in the near future.

PR #3377, opened 2018-12-10
When InitialResidueMass=0 and Organic Matter is to be reset, an exception is thrown

### LINKED ISSUE #3376: Is crop always on the ground? (Exception when OrganicMatter is reset)
When I run a continuous simulation and try to reset the organic matter at a specified date, an exception is thrown. I see that even after harvest, there is a crop on the ground and it is in the germination phase (sowing date is 20-Apr)! Is this normal? Should I add a model to totally kill the crop and simulate fallow? I am not sure whether these two things (having a crop on the ground and getting and error while resetting OM) are related.

image

image

BenAbabaei:
Resolves #3376 

When **InitialResidueMass**=0 and organic matter is to be reset, an exception is thrown. The reason is that if **InitialResidueMass**=0, the subroutine **ReadParam** does not fill the object **SurfOM**.

Can someone have a look at this change?

### DIFF
--- a/Models/Surface/SurfaceOrganicMatter.cs
+++ b/Models/Surface/SurfaceOrganicMatter.cs
@@ -438,7 +438,7 @@
             double totN = 0;  // total N in residue;
             double totP = 0;  // total P in residue;
 
-            if (InitialResidueMass > 0.0)
+            if (InitialResidueMass >= 0.0)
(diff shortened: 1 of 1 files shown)


hol353:
retest this please jenkins

Issue #3382, opened 2018-12-11
CO2 effect on Gsmax from Leaf Component


hut104:
There should be an effect of CO2 on the gsmax value provided to micromet.  The modifier should perhaps adjust gsmax relative to a reference value of approximately 350 given the timeframe of the underlying work on the gsmax model and parameters used (ie from Kelliher et al) and the assumptions that are already in the RUE model (which is itself relative to 350 ppm).

ric394:
This is very old. 
Closing. 
If this is still an issue, please feel free to create another issue with a reference to this one.

Issue #3381, opened 2018-12-11
Minimum acceptable CO2 is too high for simulations starting before 1990


BenAbabaei:
According to [Cape Grim Greenhouse Gas data](https://www.csiro.au/en/Research/OandA/Areas/Assessing-our-climate/Latest-greenhouse-gas-data) reported by CSIRO, atmospheric CO2 concentration in Australia was way below **350 ppm** in the 1980s. But the function **RUECO2Function** does not accept values below 350 and throws an exception. It seems that the underlying formula works fine with lower CO2 concentrations (leading to <1 values). Could we change this?

```
                if (MetData.CO2 < 350)
                    throw new Exception("CO2 concentration too low for RUE CO2 Function");
                else if (MetData.CO2 == 350)
                    return 1.0;
                else
                {
                    double CP;      //co2 compensation point (ppm)
                    double first;
                    double second;
                    CP = (163.0 - temp) / (5.0 - 0.1 * temp);
                    first = (MetData.CO2 - CP) * (350.0 + 2.0 * CP);
                    second = (MetData.CO2 + 2.0 * CP) * (350.0 - CP);
                    return first / second;
                }
```

![image](https://user-images.githubusercontent.com/39573622/49778017-c5a4fa00-fd4e-11e8-9e0f-96899f4afc1c.png)

BenAbabaei:
@hut104 Hi Neil. By any chance, have you thought about this change? Shouldn't there be a way to run APSIM for [CO2]<350, e.g. for the first half of the 20th century or pre-industrial era? Could we make the threshold a parameter which can be changed by the user?

PR #3384, opened 2018-12-12
Initial work for CO2 impact on Gsmax to simply reproduce current results

hut104:
working on #3382 
Do not merge as yet.  Need converter to rename gsmax to gsmax350.  Once implemented, the results should reproduce the previous output.  Once this is shown to be working correctly, we then need to add co2 impact on gsmax.  This may impact results if CO2 is specified in files.


BenAbabaei:
Neil, working on CO2, any comment on this (https://github.com/APSIMInitiative/ApsimX/issues/3381)? I have to remove the line in my branch to run ApsimX for before 1990.

hol430:
Neil doesn't get github notifications, you'll need to send him an email

hut104:
?Hi,

Thanks for the message.  I saw your issue on Github.  I may end up fixing that code in my work.

Thanks again,

Neil.

PR #3407, opened 2019-01-08
Minor changes for VS2015 compatibility

### LINKED ISSUE #2276: VS2015 build broken
New language features in AzureJobDisplayView.cs have broken the Visual Studio 2015 build.

her123:
resolves #2276

### DIFF
--- a/ApsimNG/Views/CLEM/CLEMView.cs
+++ b/ApsimNG/Views/CLEM/CLEMView.cs
@@ -191,7 +191,7 @@ namespace UserInterface.Views
             }
 
             // check if tab has been added
-            if(nbook.GetTabLabelText(tab) is null)
+            if(nbook.GetTabLabelText(tab) == null)
             {
                 nbook.AppendPage(tab, tablab);
             }
@@ -201,8 +201,9 @@ namespace UserInterface.Views
                 tab.Remove(child);
                 child.Destroy();
             }
-            if (control is ViewBase view)
+            if (control.GetType() == typeof(ViewBase))
             {
+                ViewBase view = (ViewBase)control;
                 tab.Add(view.MainWidget);
                 tab.ShowAll();
             }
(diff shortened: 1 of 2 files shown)


ApsimNG/Views/CLEM/CLEMView.cs
```diff
@@ -201,8 +201,9 @@ public void AddTabView(string tabName, object control)
                 tab.Remove(child);
                 child.Destroy();
             }
-            if (control is ViewBase view)
+            if (control.GetType() == typeof(ViewBase))
```
hol430: This will evaluate as false if control's type is derived from `ViewBase`. Will VS2015 compile if we have

```csharp
if (control is ViewBase)
```

her123: It doesn't support the is operator

her123: Maybe using 
if(typeof(YourBaseObject).IsAssignableFrom(type))

or IsInstanceOfType()

hol430: But what about, for example, ExplorerPresenter around line 700:

```csharp
catch (Exception err)
{
    if (err is TargetInvocationException)
    {
            err = (err as TargetInvocationException).InnerException;
    }

    string message = err.Message;
    message += "\r\n" + err.StackTrace;
    MainPresenter.ShowError(err);
}
```

Does this compile with VS2015?

her123: Good point.

her123: From what I can work out this test
if (x is null)
is pattern matching (C#7). C#6 (VS2015(.net4.6)) can perform the tests 
if (err is TargetInvocationException)

Issue #3408, opened 2019-01-08
Need to remove incorrect microclimate assumptions


hut104:
Some soil models assume that the fraction of light reaching the surface is simply related to the crop cover in their zone.  however, in multi-point simulations, plants in neighbouring zones can shade the surface.  The soil models should use a value of trasmitted radiation from micromet to avoid this incorrect assumption.

PR #3409, opened 2019-01-09
Surface Radiation Assumptions Removed

hut104:
working on #3408


hut104:
Will close this and approach this a new way given developments on microclimate

Issue #3316, opened 2018-11-22
Calculating cumulative value of traits in daily reports


BenAbabaei:
I have been trying to use a script like this (just an example):

`Sum of [Wheat].Structure.ThermalTime from [Clock].StartOfSimulation to [Clock].Today as Test
`

to calculate the cumulative values of a daily output variable, but I failed. Is there any way to do this without needing to use a manager script, add that variable to the code or add a new arithmetic function to **ApplyAggregation** in **APSIM.Shared**?

hol430:
This should work - looks like there is a bug here though. I'll look into it.

sno036:
Just bumping this up the priority list. Most things seem to be working well but the aggregation using "Today" is not and it would also be really handy to add something that did an aggregation (Sum, Diff, Ave - as in the documentation) from whenever the last report was done (only of this particular Report model remembering that there can be multiple Reports in a single simulation) to Today. Having these would allow dumping a bunchy of Managers that are only there to do these sorts of calculations.  

Sum of [Weather]Rain from [Clock].StartOfSimulation to [Clock].Today as CumRain
Sum of [Weather]Rain from _LastReport_ to [Clock].Today as SumRain

PR #3410, opened 2019-01-09
Testing some changes to report aggregation

### LINKED ISSUE #3305: In 'Reporting Variables' panel, events cannot be chosen from a pop-up list
It would be handy if we could choose **events** like **[Wheat].Sowing** from a pop-up list (like other objects or traits) when we need to calculate an aggregated value using arithmetic functions. Users are not usually familiar with all the **events** available. Currently **events** can only be chosen that way in the **Reporting frequency** panel in **Report** objects, not in the **Reporting variables** panel.

image

hol430:
Working on #3316 
Resolves #3305

### DIFF
--- a/ApsimNG/Presenters/ReportPresenter.cs
+++ b/ApsimNG/Presenters/ReportPresenter.cs
@@ -131,7 +131,7 @@ namespace UserInterface.Presenters
         /// <param name="e">The argument values</param>
         private void OnNeedVariableNames(object sender, NeedContextItemsArgs e)
         {
-            GetCompletionOptions(sender, e, true, false, false);
+            GetCompletionOptions(sender, e, true, false, true);
         }
 
         /// <summary>The view is asking for event names.</summary>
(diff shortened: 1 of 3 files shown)


hol430:
Depends on APSIMInitiative/APSIM.Shared#77

sno036:
@hol430 - as you are working on this .... is it possible for the array aggregation to put something in that sums (for example) the whole array?  e.g. sum(nh4n[]) or sum(nh4n[:])? At present the from/to indices need to be inserted and this causes issues with using the same report across multiple soil descriptions etc.

Also suggest adding in the documentation for clarity that if indices are used, they are 1 (not 0) based.

hol430:
retest this please jenkins

hol430:
I thought you could just put `sum(nh4n)` into report...does this not work?

sno036:
Duh!

sno036:
Hi Drew - thanks for this - the " to [Clock].Today" is now working correctly.  Is there any possibility of having something like 
    sum of [Weather].Rain from *TheReportsLastReport* to [Clock].Today as sumRainBetweenReports
where *TheReportsLastReport*  is some sort of keyword that knows when this particular Report (there can be several Report models in a single simulation) last did an output?

hol430:
Off the top of my head I don't think this will be as simple as adding a LastReportDate property to the report - I'll look into it.

PR #3423, opened 2019-01-15
Added [Serializable]

### LINKED ISSUE #3396: ITest example on website doesn't work.
The ITest example on the website when the script is not directly under a folder due to serialization issues. The script class needs to be given the `[Serializable]` attribute.

MikeStower:
Resolve #3396


hol353:
add to whitelist

Issue #3425, opened 2019-01-15
Need to tidy up Detached and Removed


rcichota:
Each organ has variables to store the info about biomass removal. *Detached* keeps track of biomass being transferred to surfaceOM or soil FOM due to senescence, harvest/grazing, or PlantEnding; *Removed* holds the info about biomass being totally removed from the system due to harvest/grazing.  There are a few inconsistencies between different organs and few errors when PlantEnding is raised:
 - GenericOrgan zeroes the variables before they can be used in the report
 - Leaf doesn't zero them and accounts for Detached twice
 - Root has two active PlantEnding methods and also accounts for Detached twice
 - ReproductiveOrgan and HIReproductiveOrgan don't zero the variables
 - PerennialLeaf doesn't zero Removed
The errors seem to be affecting reporting only, not the actual amounts returned to SurfaceOM

rcichota:
Upon closer look, I realize that the error does affect residue return in Leaf and Root. This occur when Harvest and PlantEnding happens in the same day (which is typical). In this case the biomass amount returned is greater than the existing value, this is due to Harvest returning some fraction, but the Live and Dead biomasses are not updated, then these are fully returned on PlantEnding...

PR #3426, opened 2019-01-15
Remove one of the PlantEnding methods in Root

rcichota:
Working on #3425

### DIFF
--- a/Models/Plant/Organs/GenericOrgan.cs
+++ b/Models/Plant/Organs/GenericOrgan.cs
@@ -517,7 +517,7 @@ namespace Models.PMF.Organs
         /// <param name="sender">The sender.</param>
         /// <param name="e">The <see cref="EventArgs"/> instance containing the event data.</param>
         [EventSubscribe("PlantEnding")]
-        protected void DoPlantEnding(object sender, EventArgs e)
+        protected void OnPlantEnding(object sender, EventArgs e)
         {
             if (Wt > 0.0)
             {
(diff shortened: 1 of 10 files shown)


rcichota:
This Resolves #3425 
I have tidied up and standardized few methods that handle the two variables I was interested in tracking (Removed and Detached biomass). They were reporting some wrong values or not being zeroed correctly upon CropEnd. The changes fix the reporting (though I had to add extra variables to plant)
I traced the wrong values to a lack of updating to the Live and Dead variables (these are the sum of all live and dead biomass, over leaf cohorts or soil layers). This problem only happens for Root and Leaf and when harvesting and ending the crop in the same day. The plant was returning some biomass at RemoveBiomass and then again at PlantEnding without updating itself, thus returning more biomass and N than it actually had...
Only three simulations show considerable diffs. I believe they can be accepted, though it would be good if someone can check them out.
The one from the Rainshelter here in Lincoln I am a bit familiar with, it showed only minor deviations, and a few are improvements;
The changes to the SCRUM simulation are also small, though for a couple of points they are large...
The Tarlee simulation showed somewhat greater diffs but only for the top layer (which is where it was expected as the residue return now changed)

rcichota:
@hol353 , this can be merged now I believe.
The Root organ had two PlantEnding methods, which I merged into one. I've just noticed the command to return biomass in the one I deleted was the correct one, the other returned all biomass to Surface OM. Last commit fixes it... This also removes the diffs in the validation simulations

hol353:
Looks Ok to me but should 'working on' be 'resolves'?

lie112:
There are still a few fixes I am working on, then it will be resolved and I can move on. Should I keep creating Issues as I go through and keep checking and fixing CLEM or can they all go through the one “getting it working” issue?

PR #3484, opened 2019-02-06
Factors with multiple specifications should make only a single replacement per spec.

### LINKED ISSUE #3378: Nested factors
I have a large set of simulations to run and I need to efficiently use factors. I am interested in creating nested factors. In the following example, I have three factors; **Site**, **ISM** (initial soil moisture) and **Sowing** (sowing date and cultivar). In each location, sowing date and agronomic management are different. When I put **Sowing** under the **Management** folder, everything works fine. But I need to use multiple **genotypes** and that's why I want to have **Sowing** as a factor (Sowing = combination of sowing date and cultivar). 

- **Is there any way to have all the combinations of factors in each site, but different combinations in different sites?** 
- **Is there any way to link a property in a manager script (e.g. cultivar name) to a factor**? 
- **Is there any way to use Sowing Date and Cultivar as two separate factors**? 

image

hol430:
(I think this) resolves #3378


hol430:
retest this please jenkins

hol430:
Ok, I've fixed this issue in the wheat, oats and fodderbeet validation sets. Each of these files contained experiments with factors which had specifications (text) which did nothing. 

This pull request will still run red until I fix the canola prototype.

hol353:
Actually, go ahead and change canola.

hol430:
OK - I'll just delete the problematic weather nodes; the actual met files will still be there so we can easily add the weather nodes back in if need be.

PR #3492, opened 2019-02-10
Sorghum leaf

jbrider:
Working on #572


jbrider:
I have used isOptional on a couple of functions in the short term. 
Just wanted to check that changes so far didn't affect existing models.
Plan is to change them using the converter process once I have it working fully.
You can hold off on merging if you would prefer.

Issue #3494, opened 2019-02-10
Addition of new sensitivity analysis model


FlorjanCamlek:
I will add in a new model that facilitates a factorial ANOVA analysis. Model collects data at the end of an experiment run, sends it to R, and returns the results.

sno036:
This is in addition to the existing Sobol and Morris suite of methods. It requires a complete factorial design but handles combinations of categorical and continuous variables well and gives first order and interaction effects. The complete factorial design means that it is not suitable for large numbers of factors or levels (there consider using Morris or Sobol). Florjan has been adding documentation for all three methods in tutorial style.

PR #3499, opened 2019-02-11
Addition of a model that facilitates a factorial ANOVA sensitivity an…

FlorjanCamlek:
…alysis. Commit includes a tutorial-style example.

working on #3494


Models/Sensitivity/FactorialAnova.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            string sql = "SELECT * FROM Report";
+            DataTable predictedData = dataStore.RunQuery(sql);
+            if (predictedData != null)
+            {
+                IndexedDataTable predictedDataIndexed = new IndexedDataTable(predictedData, null);
+
+                string outputNames = StringUtilities.Build(Outputs, ",", "\"", "\"");
+                string inputNames = StringUtilities.Build(Inputs, ",", "\"", "\"");
+                string anovaVariableValuesFileName = Path.Combine(Path.GetTempPath(), "anovavariableValues.csv");
```
hol430: If someone runs (from the command line) a whole directory of apsimx files at once, several files get run in parallel. If several of these files contain Anova analyses, then you could have two such analyses running at the same time. And when they get to this line here, they will both attempt to write to the same .csv file. 

You can have a look at the latest version of Morris.cs to see how I fixed it there. I only fixed it recently (last week or so?) which is probably why you didn't notice it when you were writing this - which seems to be based on the Morris code.

FlorjanCamlek: Thanks for pointing that out - I have added the relevant bits of code.

PR #3500, opened 2019-02-12
Allow for file upgrader to upgrade files which don't contain a Simulations object.

### LINKED ISSUE #3475: We need a tool to convert soils from xml to json
We all have libraries with soil data developed over time, these currently are all in xml format (either from classic Apsim/Apsoil or earlier versions of ApsimX). It would be great if we had a tool to convert this soil information into json, especially when using batch runs.  @hol353 commented that it should be possible to "add a command line switch to models.exe that converts a apsimx file to the latest version". This could then be used to convert a soils library.

hol430:
Resolves #3475 

@rcichota This should allow you to upgrade your file. However, there are some syntax errors in the xml - namely, there are two folders which are called folder instead of Folder. I haven't written a converter to deal with this (maybe I should) but for now you will need to do a find and replace of the xml, and replace folder with Folder.

### DIFF
--- a/Models/Core/ApsimFile/Converter.cs
+++ b/Models/Core/ApsimFile/Converter.cs
@@ -15,7 +15,7 @@
     public class Converter
     {
         /// <summary>Gets the latest .apsimx file format version.</summary>
-        public static int LatestVersion { get { return 52; } }
+        public static int LatestVersion { get { return 51; } }
 
         /// <summary>Converts a .apsimx string to the latest version.</summary>
         /// <param name="st">XML or JSON string to convert.</param>
@@ -72,12 +72,7 @@
                         throw new Exception("Cannot find converter to go to version " + versionFunction);
 
                     // Found converter method so call it.
-                    object[] parameters = new object[] { returnData.Root, fileName };
-                    method.Invoke(null, parameters);
-
-                    // If any of the arguments were passed by reference, apply the changes now.
-                    returnData.Root = parameters[0] as JObject;
-                    fileName = parameters[1] as string;
+                    method.Invoke(null, new object[] { returnData.Root, fileName });
 
                     fileVersion++;
                 }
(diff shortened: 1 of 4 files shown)


rcichota:
That's great, thanks @hol430 
Just to confirm, I need to have all tags with "Folder", instead of "folder"? Simply case sensitivity. That's fine, especially as it work either way in xml...
Looking forward to give it a go

hol430:
Yep, that's right - it's just a case sensitivity thing. Folder in ApsimX starts with a capital F

hol430:
Hey Rogerio - this should get Jenkins over the line so this can get merged and you can use the upgrader. Unfortunately, I had to undo a change which would have been helpful for you but wasn't very well thought out on my part. Basically, your file gets successfully upgraded but it can't be opened due to the fact that there is no top-level simulations node. I would have liked to automatically add a Simulations node if it's missing, but this caused a bunch of other problems. 

It's easy enough to add a Simulations node manually though - just edit the xml or json and change the top-level node from Folder to Simulations. 

You can edit the xml to look something like this (don't forget to edit the closing tag as well!):
```xml
<Simulations version="37" creator="Apsim 7.10-r4186" name="AllSoils">
  <Folder name="New Zealand">
    <Folder name="Canterbury">
```

Or if you've already run the upgrader, you can edit the json to look something like this (just edit the $type line):
```json
{
  "$type": "Models.Core.Simulations, Models",
  "ExplorerWidth": 300,
  "Version": 51,
  "Name": "AllSoils",
  "Children": [
```

This should allow you to open the file. You'll then need to add a DataStore, which appears to be missing. Here's an upgraded version of the file you sent me if that helps: [soils.apsimx.txt](https://github.com/APSIMInitiative/ApsimX/files/2854280/soils.apsimx.txt)


This should get you going for now...I'll see if I can get another pull request up tomorrow which automates some of this stuff.

rcichota:
Many, many thanks @hol430 
I just tested here and it works like a charm. I take note of the point you raised above, but it is not important for us for now. We have a file with an extensive library of soil data (GIS linked), we use an R script to go and pick the right soil and change it in the simulation, so we need the library in the right format, but do not need to open it directly...

Issue #3498, opened 2019-02-11
Models visible to UI via Add model


lie112:
It seems the old approach of manually adding models in models.cs has been removed and any model (implementing IModel) in the code is now visible in the UI via Add Model.
Can we add an Attribute (NoDisplay or such) that leaves a model hidden from the user (e.g. for base classes and models the user is not suppose to use directly)?
This attribute could be checked in Apsim.IsChildAllowable or Apsim.GetAllowableChildModels.

It might also be useful to add NotChildOf or InvalidParent (opposite of ValidParent) attribute to ensure some models do not appear under particular models in the tree. This would mean we can remove CLEM models from appearing under a Zone for non-CLEM users and simplify the add model window.

hol430:
There is a property `IModel.IsHidden`. Setting this to true should hide the model from the user. Does this do what you want?

lie112:
The strangest thing has happened. I can no longer repeat the problem today and the Add model dialog looks ok now (I should have done a screen shot). Am I correct that the manual adding of available models has been removed, or was it moved somewhere else?

The option you provided only works if the model has been created where the Add model dialog has to work with all models in the system so needs a reflection approach. I'll see if the problem happens again. 

I would like the negative (Not) parent of option to get CLEM components out of other users selection lists.

hol430:
Your unable to reproduce issue could be related to which model you tried to add models to. I think you should be able to right click on a folder or a factor and go add model, and have all of the CLEM Models show up.

> Am I correct that the manual adding of available models has been removed, or was it moved somewhere else?

By "manual adding of available models" do you mean right clicking on a model and clicking "add model"? If so, the code you're interested in is probably in AddModelPresenter.cs. Specifically, the call to Apsim.GetAllowableChildModels().

> I would like the negative (Not) parent of option to get CLEM components out of other users selection lists.

This sounds like a great idea - the add model dialog gets quite cluttered, especially with all of the CLEM models which I'm (usually) not interested in. I'm not sure how you've setup the inheritance hierarchy of your CLEM classes, but would it be logical to make any model which inherits from CLEMModel unable to be added to most "normal" models? e.g. 

```csharp
public static bool IsChildAllowable(object parent, Type childType)
        {
            if (childType == typeof(Simulations))
                return false;

            // CLEMFolder can be added to Simulations object.
            if (parent.GetType() == typeof(Simulations) &&
                childType == typeof(CLEM.CLEMFolder))
                return true;

[...]

PR #3506, opened 2019-02-13
Added a filter to the add model/function view.

hol430:
Working on #3498 

This is a case-sensitive search - is that ok?


hol353:
Probably better if it was case in-sensitive. Otherwise @hut104 will whinge.

hol430:
retest this please jenkins

hol430:
@hol353 have you guys done something to break the performance tests? This pull request only touches ApsimNG, so I don't see how this could possibly impact the stats.

hol353:
The stats were changed this morning because of a new barley model. When I accepted the stats this morning it doesn't seem to have worked. Done it again now.

PR #3511, opened 2019-02-13
Merged data from official release with Allan's work

### LINKED ISSUE #3505: Observed Data missing from Oats Validation Set
Graphs have missing OBS data.  Need to get missing data from model developer.

hut104:
resolves #3505


hut104:
The changes to the observed data has resulted in slight decrease in goodness of fit on some data, but a better fit overall.  I recommend that we accept these changes from the model developer.

PR #3519, opened 2019-02-14
Wheat Stem Structural Fraction

### LINKED ISSUE #3518: Stem Structural Fraction in Wheat
The structural fraction for stem in wheat is incorrect.  The value needs to be changed to reflect timing of storage of reserves and early growth of stem structure.

hut104:
resolves #3518

### DIFF
--- a/Models/Resources/Wheat.json
+++ b/Models/Resources/Wheat.json
@@ -4885,8 +4885,8 @@
                     },
                     {
                       "$type": "Models.Functions.PhaseLookupValue, Models",
-                      "Start": "StartGrainFill",
-                      "End": "FlagLeaf",
+                      "Start": "FlagLeaf",
+                      "End": "EndGrainFill",
                       "Name": "ReproductiveGrowth",
                       "Children": [
                         {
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@hut104 I have taken a look at this and there is a problem with the predictions of spike weight now.  The problem seems to be similar to what was in stem.  The structural fraction is only set to 0.3 for the duration of growth.  Should be something like 0.9 up to some time prior to start of grain fill and then decrease to 0.01 the same as stem is doing now.  Do you want me to pick this one up and try some stuff

hut104:
Yep,
I’m snowed under here this week.  I also think we need to check the impact on any processes that may have been simulated using stem reserves (e.g. stress factors or C supply factors).  I noticed that there were changes in various plant outputs.

One of the big issues this week was getting the structural/storage side of things right for soybean.  I’ve started putting structural mass on a lot of the biomass plots to understand the dynamics.

Neil.

HamishBrownPFR:
Closing because has gone stale.  Replaced by #4523

PR #3520, opened 2019-02-15
Soil nutrient increment

### LINKED ISSUE #3522: Initialise FOM pools by layer
The new soil Nutrient model requires FOM to be input on a layered basis (Called RootWt in Soil Nitrogen).  Many soil carbon modelling efforts would also benefit directly from an ability to specify the amount of FOM in each layer rather than the current application of an arbitrary exponential distribution with depth.
This will impact SoilNitrogen as well but will allow interoperability between soil nutrient and soil nitrogen.

hut104:
resolves #3522

### DIFF
--- a/Models/Core/ApsimFile/JsonUtilities.cs
+++ b/Models/Core/ApsimFile/JsonUtilities.cs
@@ -219,6 +219,47 @@ namespace Models.Core.ApsimFile
         }
 
         /// <summary>
+        /// Gets a list of property values.
+        /// </summary>
+        /// <param name="node">The model node to look under.</param>
+        /// <param name="propertyName">The property name to return.</param>
+        /// <returns>The values or null if not found.</returns>
+        public static List<string> Values(JObject node, string propertyName)
+        {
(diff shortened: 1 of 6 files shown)


Models/Core/ApsimFile/JsonUtilities.cs
```diff
@@ -218,6 +218,47 @@ public static void RenameProperty(JToken modelToken, string propertyName, string
...
+            return null;
+        }
+
+        /// <summary>
+        /// Sets a list of property values.
+        /// </summary>
+        /// <param name="node">The model node to look under.</param>
+        /// <param name="propertyName">The property name to return.</param>
+        /// <param name="values">New values</param>
+        /// <returns>The values or null if not found.</returns>
```
hol430: Did you forget to delete this <returns> comment tag? Also not sure if the propertyName desciption is correct

hut104: Dean wrote that code - I don't know anything about it.

Issue #3509, opened 2019-02-13
Profiling and refactoring needed to increase execution speed


hut104:
Profiling and refactoring may assist in increasing execution time.

hol353:
Nutrient is very slow. Almost 30% of run time is in calculating carbon and nitrogen flows:
![image](https://user-images.githubusercontent.com/1943827/88493035-51d26680-cff2-11ea-8ee5-f1fb2d3c7bd9.png)

hut104:
Could approach this in a similar manner to previous speed-up work in Nutrient - develop functions that encapsulate a particular approach for N flows.

hol353:
Some timings using 100 year wheat example.

APSIM Classic: 41 sec
APSIM Next Gen with nutrient: 11.5 sec  (3.5 times quicker than classic)
APSIM Next Gen with soil nitrogen: 7.5 sec (5.5 times quicker than classic)

hol430:
Is this after your changes? The wheat example normally takes around 40 seconds for me on my laptop.

hut104:
Dean, Try these again with the latest changes in the "Wheat" pull request currently on jenkins.

________________________________

hol353:
Good job Neil. New timings:

APSIM Next Gen with nutrient: **8 sec**

NFlow is now well down the list of slow methods. Now 27% of runtime is spent in CarbonFlow:

![image](https://user-images.githubusercontent.com/1943827/88854450-a2390680-d234-11ea-9ed9-bf53418900ff.png)

hol353:
Run time is largely unchanged at around the 8sec mark:
![image](https://user-images.githubusercontent.com/1943827/88997100-38068b80-d332-11ea-9afc-b7e91ce54ccb.png)

hut104:
Those two lines are now insignificant components of overall execution time.  I'll now try the other bottlenecks.

hol353:
No real difference. Maybe slightly quicker. I think you're being beaten by the fact that these functions are called inside of soil layer for loops. The hit counts are very high.

![image](https://user-images.githubusercontent.com/1943827/89134850-c7e34a00-d56b-11ea-8dad-531153ceb392.png)

hol430:
How much speed do you gain if you remove the 'read tomorrow's met data' hack which causes an exception to be thrown at the end of most simulations?

hol353:
Runtime now 7.9sec
[...]

PR #3531, opened 2019-02-18
Making Between function index based

HamishBrownPFR:
working on #3509


HamishBrownPFR:
Going to try a different approach

PR #3541, opened 2019-02-21
Reduce outputs

HamishBrownPFR:
working on #3509


HamishBrownPFR:
This means the Between function is now index based.  Finally managed to Iron out all the subtle interacting bugs.

PR #3546, opened 2019-02-24
Sorghum debug

jbrider:
Working on #572


jbrider:
Is the conflicting file causing the random changes?
I believe that I can't fix that?
It ran green last commit, but failed this one - there is nothing in the commit that should have affected Maize or Sugarcane.

hol353:
I doubt it would be the conflicted file.

@hol430 I wonder if this is because the performance collector now ignores null predicted and observed values. I wonder what it was doing to with them before. Funny how it only changed maize and sugarcane. The diffs are only tiny so I can accept them - so long as they aren't random.

hol430:
I guess the conflicting file could be causing the change in stats. Either way, the pull request can't be merged until you fix the conflicting file. Just pull down the latest changes, fix the conflict and push back up to the remote branch. Will be interesting to see if that fixes the problem.

There's a bug in the performance tests sometimes, where the stats don't get accepted properly when we click the accept stats button. That could be the cause of the apparent failure, but my first guess would be the conflicted file.

jbrider:
I just need to get back up to date then... damn, thought I was avoiding that bit ;)
Will see what it does when I have fixed the conflict, because it has been a bit random.

jbrider:
@hol353, Can you merge this please.

PR #3547, opened 2019-02-24
Add stomatal CO2 response to Oats

HamishBrownPFR:
working on #3382

### DIFF
--- a/Models/Resources/Maize.json
+++ b/Models/Resources/Maize.json
@@ -2226,16 +2226,42 @@
       "FractionDied": 0.0,
       "Name": "Leaf",
       "Children": [
+    {
+      "$type": "Models.Functions.ExpressionFunction, Models",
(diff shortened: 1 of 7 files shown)


HamishBrownPFR:
@hol353 please merge.  Performance test fail is coming from small diff in Maize file which is acceptable and a number of other metrics have improved.  I have investigated the cause of this as I was not expecting any diffs because all simulations in the maize test set have 350 ppm CO2.  The diffs come about becasue the stomatal conductance equation references FCO2 which is calculated by the RUECO2Function class.  This class returns a value of 1 for C3 crops at 350ppm CO2 so the change implemented for Oat, Fodder beet and Potato has no effect.  However, for C4 crops (maize) it returns a number that is slightly different to 1 and so we get diffs.  @hut104 are you happy with this diagnosis and to have this merged??

Issue #3555, opened 2019-02-27
Enable Photosyntehesis and transpiration in generic organ


RobZys:
For plants where substantial gas exchange occurs in non leaf organs

HamishBrownPFR:
This is complete with the two merges incorporated

PR #3556, opened 2019-02-27
Aded new energy balance class

RobZys:
Working on #3555

@hol353 @hut104 working with @HamishBrownPFR  to prototype implementation of radiation interception and transpiration and photosynthesis in non leaf organs.
We have made a new class (EnergyBalance.cs) that implements icanopy and can be dropped on a generic organ, along with a photosynthesis function, to give the desired functionality.
We have implemented this on Fieldpeas.Stem in prototypes and done some simple testing.
It appears this will work and is quite tidy but we would like some feedback on weather this is an appropriate route to take. if yes we will continue testing and getting it to work properly.

added missing XML ignores to simple leaf
testing on field peas


hol353:
Looks like a good solution to me. Does this mean the we can use this mechanism the the various leaf models e.g. Leaf, SimpleLeaf, PerrenialLeaf.

HamishBrownPFR:
Yes, I am pretty sure we will be able to replace simple leaf with this construction we are creating.  It would work for the perennial leaf and leaf also which would mean a bit more grouping of parameters under organs which would be nice.

jbrider:
I think this a good idea in concept - but didn't GenericOrgan get removed fro the list of interfaces for the leaf objects not long ago - I didn't quite understand why at the time - maybe this deals with that?  
From a code perspective, I found that there are a lot of unused variables in some of these interfaces - particularly used in SimpleLeaf.  Is this a step towards minimising some of those or is it just putting them out of sight? It would certainly remove a lot of the repeated properties that currently exist across the different Leaf classes.

jbrider:
@hol353 What are we supposed to be doing with <XMLIgnore> now that we have moved to json?

Also - I was confused for a second when I went to comment on this as a new issue had been created to deal with it - is that something to do with Jenkins or just a better name?  I'm not sure what the process is supposed to be?

HamishBrownPFR:
@jbrider we removed simple leafs inheritance of generic organ because inheritance is discouraged.  I think that is what you are recalling.  It means stuff is now duplicated across simple leaf and generic organ.  I think ultimately we should be able to do away with simple leaf and just have a generic organ which we drop objects onto that implement other interfaces to make it into a leaf.  This is a step in that direction.
Once you have got the sorgham model complete it would be really good to sit down together and work through a tidy up of some of this stuff.

jbrider:
@HamishBrownPFR I don't necessarily agree that inheritance should be discouraged - but that is a longer discussion - probably when we do get together to look how we can better integrate sorghum.  I'm not saying this needs inheritance either - this drop in component will certainly address some of the duplicated properties that currently exist, which is good.  
Is Microclimate going to be required for this? - there is an exception being thrown in one of the OnGrowth events if it is not present.

HamishBrownPFR:
Yes microclimate will be a dependency for this.  The purpose of it is to enable a plant to have multiple organs implementing ICanopie which means these organs can intercept radiation and use water.  Microclimate is needed to work out the partitioning of radiation between the two (or more organs)

hol430:
> What are we supposed to be doing with XMLIgnore now that we have moved to json?

I think the json serialisation engine still respects `[XmlIgnore]`, but it's probably better to be explicit and use `[JsonIgnore]`, which lives in the `Newtonsoft.Json` namespace.

HamishBrownPFR:
@hol353 @hut104 @jbrider @RobZys @rcichota 
An issue we have with this implementation is that BaseArbitrator.cs looks for water demands on IOrgan if they implement IHasWaterDemand and gets and allocates demands to the corresponding organ.  e.g:
```csharp
foreach (IArbitration o in Organs)
                if (o is IHasWaterDemand)
                    WDemand += (o as IHasWaterDemand).CalculateWaterDemand() * Plant.Zone.Area;
```
If we move the energy balance stuff out of the organ into a specific class then the arbitrator doesn't find them so the soil arbitrator does not register their demand or allocate them any water.

The solution we propose is to have the arbitrator create a list of ICanopy objects (the same as it does for IOrgan currently) and iterate through these for getting water demands and setting water allocations.  This also means the IHasWaterDemand interface becomes unnecessary because every organ that implements ICanopy also implements IHasWaterDemand.  Thus we also propose the IHasWaterDemand interface is deprecated and its contents moved into ICanopy.

Any issues with this??

hut104:
?IHasWaterDemand was a bit of a kludge anyway.

This sounds better to me.


jbrider:
@hol353 @hut104 @RobZys @rcichota @HamishBrownPFR 
I am a bit concerned about implementing ICanopy - as sorghum isn't currently using MicroClimate, and the description for the interface is "This interface describes MicroClimate".  At present the interface only describes a handful of properties of which Sorghum uses most, so it's not an issue as it is - my concern is how many extra properties will it gain.  Happy to comment on them as they come up though.

HamishBrownPFR:
@jbrider if sorgham is not using Microclimate than it should not need to implement ICanopy.  I will check this statement as I am not sure if there is anything else than communicates through ICanopy.

HamishBrownPFR:
Sorry, I was wrong, ICanopy is used all over the place.

HamishBrownPFR:
IHasWaterDemand contains three properties:

        /// <summary>Gets or sets the water demand.</summary>
        double CalculateWaterDemand();

        /// <summary>Sets the organs water allocation.</summary>
        double WaterAllocation { get; set; }

        /// <summary> Flag to test is Microclimate is setting PotentialEP value </summary>
        bool MicroClimatePresent {get; set;}

I am guessing that anything that implements ICanopy also Implements IHasWaterDemand so will have all of these properties so moving them from IHasWaterDemand into ICanopy will be a zero sum game??

hut104:
In any case, Sorghum will need to talk with microclimate to get inter-crop and spatial interactions calculated.   We will need this for agroforestry in africa.  Microclimate will also help with strip crop (ie skip row) simulations of light etc.



jbrider:
@hut104 I wasn't aware that was what MicroClimate would do - Val sent me the documentation after the meeting last week but I haven't had a chance to get into it.  First iteration of Sorghum is a straight port that is supposed to give the same results as Classic. After that I will need some assistance to introduce some science changes.

hut104:
The doco may be simply for the penman monteith part of the model written over a decade ago.  It has moved on a lot, and is adding capabilities with current projects.

hol353:
retest this please jenkins

hol353:
Is this ready to be merged?

RobZys:
yes it can be merged if you want to add the energy balance code for other to examine
In order for generic organ to have photosynthesis to work other things need to be sorted

PR #3567, opened 2019-03-04
Fixed calculations and clarified units in outputs and equations

### LINKED ISSUE #3565: Treeproxy N uptake errors
There are two errors in the tree proxy N uptake equations.  The 10000 should be removed from the following two lines

double NDemandkg = GetNDemandToday()*10 * treeZone.Area * 10000;

 PotNO3Supply += Uptake.NO3N[i] * ZI.Area * 10000;

Some of the unit conversions should also be documented more explictly to assist interpretation within the code

hut104:
resolves #3565


hol353:
retest this please jenkins

PR #3576, opened 2019-03-07
Fixed a bug where y-axis settings are incorrectly displayed as dates.

### LINKED ISSUE #3573: Date format for the range in the numeric y axis if x-axis is date
When I change the option of y axis, the minimum and maximum values will show as date if x-axis is a date.

See the red circle below. 
image

hol430:
Resolves #3573


hol430:
retest this please jenkins

Issue #2821, opened 2018-07-26
Improve Datastore database performance


her123:
Add Firebird support.
Generalise database interface.

zur003:
Firebird will require some refactoring of the database structure. Firebird has a 31 character limit on the length of field names, so column names like Wheat.Phenology.CurrentStageName aren't going to work. We will probably need a new table that associates a field identifier with a field name.

hol430:
@her123 Can this issue be closed?

zur003:
The Firebird support has not been maintained, and bits no longer work (for example, PagedDataProvider uses "rowid",which is specific to SQLite, and is not available in Firebird). Firebird support is probably worth fixing, and some parts of the existing code can be simplified. For example, the previous 31 character limit on field names is now 63 characters (or perhaps that's 63 UTF-8 encoded bytes). That's long enough to cope with things like "Wheat.Phenology.CurrentPhase.Name" without all the hoop-jumping I introduced in #3780, which should greatly simplify things.

PR #3578, opened 2019-03-07
Data store

hol353:
Working on #2821. This is a major refactoring of DataStore in an attempt to clean the code and make it more consistent. The DataStore now works with tables of data from report and other models rather than row by row. This should be more efficient.


jbrider:
It's probably too late for this refactor seeing it's nearly finished, but in the future I would like to look at being able to choose the final storage mechanism.  For instance: being able to run large factorials that push the data through to a web server rather than trying to collate database files for post processing.  It's a common practice in Domain Driven Design implemented through a repository pattern.

hol430:
I think you should be able to do this if you create a model which implements `IStorageWriter` and `IStorageReader`. You could then delete the datastore from your .apsimx file, and replace it with the new storage mechanism. Not that I've tested this.

jbrider:
Good point... was looking too in the table code when the interface is right
there at the top.
😬

hol353:
I think it's easier than that. The code talks to a IDatabaseConnection to read/write to the database. @zur003 has created a Firebird implementation of IDatabaseConnection. No reason why we can't implement another for any database.

jbrider:
I was looking at avoiding a database at all - similar to some of the server
logging systems where it just pumps it out to a web api.
Something to look at in the future.

zur003:
Credit for the Firebird work should go to @her123 rather than myself, although I've been working (slowly) on a few remaining problems. One nice thing about Firebird is that it should be able to switch readily between a local file and a remote database for the data store.

zur003:
The _Units table seems to have disappeared. 

Also, the previously "hidden" _InitialConditions, _Messages, _Simulations, and _Checkpoints tables are now visible in the general reporting view. Is this a deliberate change?

jbrider:
The sorghumleaf file I am using just doubled the run time for a single year simulation.  
Already have issues with this machine taking along time to run the simulation which just got ridiculous. Win7 machine was taking 18 seconds to run - is now taking 45+ to run.  It seems to finish the model quickly as the progress bar is quick - but takes forever to generate the report - or whatever it's doing at that point.  The file has a lot in the report, and a lot of charts.  Home laptop (Win10) previously ran the file in 2 seconds, so I haven't discounted something strange with this machine.

hol353:
_Units table disappearing sounds like a bug. On my computer I'm getting one.

I hadn't noticed the hidden tables suddenly appearing. Not intentional. Raise an issue.

I haven't seen a slow down in runtime speed. If anything I was observing a slight improvement yesterday. Probably depends on how many variables are being outputted. Yesterday I was running 100 year wheat continuous simulation in 17 sec. Not many outputs though. I'll do some profiling.

hol353:
@jbrider Can you send me your slow simulation with weather file so that I can profile it?

jbrider:
It's in the prototypes/sorghum directory - SorghumLeaf.apsimx.

jbrider:
Home machine is doubled as well - at 5.8 seconds.
It's completely overloaded with unneeded data outputs at present... probably makes it a good test case for this though.

zur003:
I'm running the Wheat validation and am waiting and waiting for the database writing to complete at the end of the runs.

zur003:
Re the _Units table - with the Wheat validation, I'm getting the table with entries for the various ObsPred tables, but not for the report tables.

hol353:
So I found a major slow down in one of the event connection methods. This significantly improves the wheat validation dataset. On my computer it has halved the run time down to 11 minutes. @zur003 I don't know why you're seeing the situation of waiting for the database to finish writing. It doesn't happen for me. There is a small pause at the end of the sim while it reads the observed data and creates the predicted/observed tables. Eric, can you retest running the wheat validation?

@jbrider The new refactored DataStore has been optimised for lots of concurrent simulations like the validation datasets. Essentially instead of writing each individual row from report as the Report.DoOutput() method is called, the rows are stored in a datatable and written as a group of rows. This speeds the up the writing of the .db file and simplifies the code. The downside seems to be that single simulations run slower. I will keep looking at it to see what I can do.

jbrider:
I merged these changes and it made a small relative difference to a single simulation (48->42 seconds).
Mine is doing the same as Eric's I assume - the progress indicator goes across in a second or 2, then it is busy for the next 40 seconds.

hol430:
Have you deleted the .db file since pulling down the datastore changes? I've come across strange problems which have been solved that way, often at times when I've just pulled down changes to the datastore code.

jbrider:
Deleting it didn't do much.  I did notice it is creating a .journal file and deleting it multiple times - the desktop has one of those old fashioned mechanical hdd's in it which would probably explains the significant slow down.

hol430:
My surface has an internal ssd and an external hdd. I've just run your sorghumleaf.apsimx from my compiled version of apsim on the ssd (and the .apsimx file also stored on the ssd). This took 4.94 seconds.

I then installed apsim to the external disk, copied your .apsimx and met files to the external disk, ran the file again. Result was 5.85 seconds. Not a big difference. Haven't tried the wheat validation set yet, as I don't have the patience. I'll set it running over lunch.

jbrider:
Well it was a theory... doubt it would be a Win7 thing either?

zur003:
@hol353 - I'm still seeing the very long delay with the Wheat Validation, even after deleting the .db file. As @jbrider described, I'm seeing Wheat.db-journal coming and going during the wait.

hol353:
Yes Wheat.db-journal will come and go. This is due to the begin/end transaction calls. One thing Neil and I noticed this morning is that my 8Gb Windows 10 surface pro was running low on memory and APSIM was memory constrained. Closing visual studio and a bunch of other apps changed my wheat validation runtime from 649 sec to 423 seconds.

zur003:
@hol353 - I've just tried building and running the wheat validation on Linux. The simulations themselves were relatively slow to run on Linux (that's a Mono problem), but writing to the datastore at the end was pretty quick (maybe a few seconds). So perhaps something is a bit odd with my Windows setup, but I don't have a clue what it might be.

jbrider:
@hol353  So you're not using google chrome on the Surface Pro then. 
32 Gb on the desktop should have Apsim covered... and a few tabs in chrome.
How often should it be doing a transaction - is that a once a day event?

hol353:
Currently each 100 rows (days) of output is a single transaction. This is done to improve the write speed to the .db.

jbrider:
Did you get a chance to profile a file with the large amount of report variables?
Sorghumleaf does have a fair few at the moment.

hol353:
Yes I've profiled SorghumLeaf but there isn't anything that stands out as being slow. I'm still looking at it although there have been lots of interruptions today.

jbrider:
In that case it would likely be machine dependent. 
@zur003 your windows isn't Win7 is it?

zur003:
No, @jbrider, I'm running Windows 10.

rcichota:
I am running windows 7 and start having the same delay described above. It started this morning after I updated APSIM...

zur003:
My Linux results make slightly more sense now - JobRunnerAsync (along with JobRunnerMulitProcess) tries to determine the number of available processor by querying the NUMBER_OF_PROCESSORS environment variable, which is normally set on most Windows machines, but not on Linux. Hence on Linux it defaulted to running on a single core, which was a large part of why the runs themselves seemed so slow. However, that also seems to tie in with the DataStore issues. With only a single execution thread, the database stuff at the end was very quick to complete; if I modify things to use all 12 cores at hand for execution, then I get a substantial wait at the end while the DataStore work completes.

That makes a bit of sense, I suppose, in that the DataStore Writer might be "squeezed out" to some extent when all cores are busy running simulations, but I'm not seeing anything similar on Windows. On Windows, I'm always seeing a long delay at the end. During this time, the job runner appears to be working its way through a queue of around 7500 jobs (this is with Wheat Validation), all manipulating the database in some way. Nothing I've tried seems to change that.

hol353:
Thanks for this @zur003. When running APSIM on Neil's 32 core computer using the 'experimental run option' (which runs 32 sims using multiple runner processes) runs the sims **VERY** quickly. It then spends ages at the end to write all the tables to the DB. Perhaps you're seeing something similar. It appears that the writing can keep up with 4 sims running in parallel but not much more.

I'm not sure what to do about it. Go back to writing row by row to the .db?

hol353:
Just a thought Eric. What happens if you set the NUMBER_OF_PROCESSORS environment variable on Windows to one less than the number of cores to give the writing thread some time to do it's stuff?

zur003:
@hol353 - I'm going to try a series of runs to see how overall time varies with NUMBER_OF_PROCESSORS. It may make sense to leave two "free" (leaving one for db access, the other for GUI).
I've also found that with an old-style spinning hard disk, you can speed up the writing a lot by using "PRAGMA journal_mode=WAL". See the Sqlite documentation for the pros and cons of this approach.

zur003:
OK. I think I've got this thing figured out. By doing a series of test runs using the wheat validation set, I found that disk access to the database is the limiting factor, but only when one is using a physical disk. With SSD drives, this isn't showing up as a problem, which probably explains why Dean isn't seeing a slowdown, but Jason, Rogerio, and myself are. Anyway, on my system with a physical disk, the runs were taking around 350 seconds regardless of the number of processors. (This was after changing journal mode to WAL; before that the times were more like 1500 seconds.) Changing processors only affected whether that time was consumed during vs. at the end of the runs.

Fortunately, there is a fix! I've found that using Sqlite pragmas to set both journal mode to WAL, and synchronous to NORMAL massively reduces the amount of disk access required. I will open a new issue on this and commit the changes shortly.

hol353:
Good detective work Eric!

hol430:
Does the synchronous setting affect runtime or thread safety when using the multi-process runner?

zur003:
It shouldn't be a problem. From the SQLite documentation:

When synchronous is NORMAL (1), the SQLite database engine will still sync at the most critical moments, but less often than in FULL mode. There is a very small (though non-zero) chance that a power failure at just the wrong time could corrupt the database in journal_mode=DELETE on an older filesystem. WAL mode is safe from corruption with synchronous=NORMAL, and probably DELETE mode is safe too on modern filesystems. WAL mode is always consistent with synchronous=NORMAL, but WAL mode does lose durability. A transaction committed in WAL mode with synchronous=NORMAL might roll back following a power loss or system crash. Transactions are durable across application crashes regardless of the synchronous setting or journal mode. The synchronous=NORMAL setting is a good choice for most applications running in WAL mode.

Issue #2645, opened 2018-06-01
Magic calculation for water demand in the RUE model 


byzheng:
There are some magic calculations in the RUE model. To reproduce my problem, I have created three simulations with three different version of RUE model. The base simulation is set up with water limited conditions.

* RUE: The raw RUE model in the release
* RUENoStress: All stress factors are changed to constant with value 1
* CustomRUE: The same algorithm is created with multiple and minimum functions in the GUI

If you look at the second panel of WaterDemand, the water demand CustomRUE model is really tiny compared with other two models, even when the fixation supply is the same between RUE and CustomRUE before 60 DAS. 

The only difference between RUE and CustomRUE are all codes in RUEModel.cs are not called. It seems there are some magic calculation about water demand in the RUE model. 

However, I cannot find any codes related with water demand in the RUEModel.cs. Any tips will be really helpful.

![image](https://user-images.githubusercontent.com/1657381/40824819-d9fef3b2-65b8-11e8-9dde-b5059ffcaafc.png)

The apsimx and met files are attached to reproduce my problem. 

[photosynthesis-waterdemand.zip](https://github.com/APSIMInitiative/ApsimX/files/2061294/photosynthesis-waterdemand.zip)

byzheng:
I have narrowed down this problem to `VariableReference`. Model references to three variables in Photosynthesis, i.e.  `[Leaf].Photosynthesis.FT.Value()`, `[Leaf].Photosynthesis.FN.Value()` and `[Leaf].Photosynthesis.FVPD.Value()`. In the `CustomRUE` model, the three variables are missing, but program doesn't give any error messages. I guess the C# return a small value (close to zero), then cause the tiny water demand. 

It might be a bug in the `VariableReference` class which should throw out an error message if the variable reference is not found. It might be only related with Leaf.Photosynthesis in the test simulation.

ric394:
Unable to test example file. If this still an issue, please make another issue.

Issue #3554, opened 2019-02-27
Null doesn't throw error in variablereference function


hut104:
The following code allows a variable reference to an incorrect link to simply return zero rather than throwing an error.
If the link is incorrect, the highlighted line will return o as null but then go on to return a value of zero.  It should test for o as null and throw.

        /// <summary>Gets the value.</summary>
        /// <value>The value.</value>
        public double Value(int arrayIndex = -1)
        {
            **object o = locator.Get(VariableName.Trim());**
            if (o is IFunction)
                return (o as IFunction).Value(arrayIndex);
            else if (o is Array)
                return Convert.ToDouble((o as Array).GetValue(arrayIndex), 
                                        System.Globalization.CultureInfo.InvariantCulture);
            else
            {
                double doubleValue = Convert.ToDouble(o, System.Globalization.CultureInfo.InvariantCulture);
                if (double.IsNaN(doubleValue))
                    throw new Exception("NaN (not a number) found when getting variable: " + VariableName);
                return doubleValue;
            }
        }

hol430:
via @byzheng:

>Step to reproduce this problem. See example below
>
>    Open wheat example, create replacement and copy wheat.json into replacement.
>    Copy "Photosynthesis" under "Leaf" into "Photosynthesis_RUE"
>    Create "VariableReference" as "Photosynthesis" and add value as "[Photosynthesis_RUE"
>
>The value of "VariableReference" is missing right bracket. APSIM doesn't give any error.
>
>However, the biomass is close to zero.
>
>[test.zip](https://github.com/APSIMInitiative/ApsimX/files/3092074/test.zip)

PR #3591, opened 2019-03-11
Variable reference

hut104:
working on #3554


hol430:
Will this also fix #2645?

hut104:
Will close pull request and work incrementally.

PR #3594, opened 2019-03-11
Fixed bug in variable property when setting a single element in an array.

### LINKED ISSUE #3592: Apsim.Set() method does not work for setting a single value of an array.
If we try to set a single value of an array object (e.g. `[Phenology].ThermalTime.XYPairs.X[0]`) via `Apsim.Set()`, we end up with a new array with only a single object in it - ie all other elements in the array are erased.

hol430:
Resolves #3592


HamishBrownPFR:
@Xiumei2017 When this is merged it should fix the problem we found today

PR #3595, opened 2019-03-12
Event connection speed up

### LINKED ISSUE #3509: Profiling and refactoring needed to increase execution speed

hol353:
Resolves #3509

### DIFF
--- a/Models/Core/Events.cs
+++ b/Models/Core/Events.cs
@@ -1,14 +1,9 @@
-﻿// -----------------------------------------------------------------------
-// <copyright file="Events.cs" company="APSIM Initiative">
-//     Copyright (c) APSIM Initiative
-// </copyright>
-// -----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jbrider:
Surprised that this would have a large effect - I assumed it would just run this process at the beginning of the simulation run? What sort of speed improvements did it give?

Related Q: Do we record processing time in the stats for comparison?

hol353:
The wheat validation creates over 600 simulations once the run button is clicked. Each of these needs to have it's events connected up. It always amazes me what profiling turns up. Profiling showed that APSIM was spending around 37% of it's time in this method (and the child methods that it calls).

No we don't record processing times in the stats. The problem is that Jenkins run times seem to vary quite a lot. Not sure why. One build and run can take 16 minutes whereas the next run can take 25 minutes.

PR #3597, opened 2019-03-12
Ensure units are obtained early when setting up columns for the datas…

### LINKED ISSUE #3596: Units are not always being written to the DataStore
Recent changes to the DataStore have resulted in the _Units table not being written as expected. This can be seen by running the Wheat example simulation (be sure there is not an already-existing .db file). When the Report is viewed, no units appear.

zur003:
…tore so they will be saved correctly. Resolves #3596


hol430:
retest this please jenkins

PR #3601, opened 2019-03-12
CLEM Add a basic pivot table model to allow easier interaction with resource ledgers.

### LINKED ISSUE #3600: CLEM Add basic pivot table for interacting with ResourceLedgers.
A new feature for CLEM, designed to make pivoting data in ledgers easier.

MikeStower:
Resolves #3600.


ApsimNG/Presenters/CLEM/PivotTablePresenter.cs
```diff
@@ -0,0 +1,229 @@
...
+using System.Reflection;
+using UserInterface.Commands;
+using UserInterface.Presenters;
+
+namespace ApsimNG.Presenters.CLEM
+{
+    class PivotTablePresenter : IPresenter
+    {
+        /// <summary>
+        /// The CustomQuery object
```
hol430: Outdated comment? There are also a few empty summary tags for methods further down in this class

ApsimNG/Views/CLEM/PivotTableView.cs
```diff
@@ -0,0 +1,364 @@
...
+    /// <summary>
+    /// Displays the result of a Custom SQL Query on a DataTable
+    /// </summary>
+    class PivotTableView : ViewBase
+    {
+        // Components of the interface that are interactable
+        // Taken from PivotTableView.glade
+        private VBox vbox1 = null;
+
+        //private ComboBox ledgerbox = null;
```
hol430: Unused/commented code should probably be removed. We can always get it back from version control if need be

ApsimNG/Views/CLEM/PivotTableView.cs
```diff
@@ -0,0 +1,364 @@
...
+            }
+
+            public TrackChangesArgs(ViewBox box)
+            {
+                Name = box.Name;
+                Value = box.ID;
+            }
+        }
+
+        public class ChangePivotArgs : EventArgs
```
hol430: (I think) The convention is that separate classes (even those as small as this one) should have their own file. Pretty sure there's an event arguments folder somewhere in ApsimNG where you could put this.

hol353: In this case it looks like the classes are nested inside an outer class which I guess provides context so perhaps you can leave it as it is. If you do want to put them in separate files, I wouldn't put them into a general event arguments folder. Better to keep the CLEM files together as much as possible.

ApsimNG/Views/CLEM/PivotTableView.cs
```diff
@@ -0,0 +1,364 @@
...
+            /// 
+            /// </summary>
+            public void Detach()
+            {
+                Changed -= parent.OnInvokeUpdate;
+                box.Changed -= OnChanged;                
+            }
+        }
+
+        public ViewBox Ledger;
```
hol430: To be compliant with our style guidelines, these public members should be properties - ie

```csharp
public ViewBox Ledger { get; set; }
```

Or if they don't need to be public, just change them to private.

hol353:
I can't merge this pull request as it looks like a bad merge has happened. For example, the changes to the DataStore that I made recently have been undone by this pull request. For that reason I can't imagine that a Jenkins compile will run green.

Models/Core/IStorageWriter.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
hol430: This (along with the changes to CustomQueryView and CustomQueryPresenter) look to be the result of a bad merge. There was a recent refactor of DataStore which moved this class to Models/Storage/IStorageWriter.cs. Unsure what the best way to fix this would be. Probably just revert the merge commit where you pulled from origin (or whatever you call the upstream repo).

ApsimNG/Presenters/CLEM/PivotTablePresenter.cs
MikeStower: Fixed all the documentation mistakes/commented code.

ApsimNG/Views/CLEM/PivotTableView.cs
MikeStower: Changed them to properties.

ApsimNG/Views/CLEM/PivotTableView.cs
MikeStower: I was unaware of any convention and deliberately didn't put them in separate files because they are so small and don't have utility outside the context of the parent class. My preference would be to leave them where they are to reduce clutter.

Models/Core/IStorageWriter.cs
MikeStower: Sorry, still learning how to correctly manage merging/resolving conflicts and utilising separate branches. I think I fixed it.

hol430:
There are a lot of changes showing up in this pull request - might have to do a rebase against master or close this pull request and create a new one.

PR #3606, opened 2019-03-13
Grain nitrogen concentrogen

### LINKED ISSUE #3605: Minimum nitrogen concentration in the grain
Arbitration counts the minimum nitrogen concentration during biomass allocation. During grain filling stage, there are enough nitrogen to allocate into new biomass of grain, even there are more nitrogen in the whole grain. 
TODO: Whille add some examples here to demonstrate this problem.

I propose to change to recalculate minimum nitrogen requirements according to whole grain, NOT only new biomass.

byzheng:
resolves #3605

### DIFF
--- a/Models/Plant/Organs/ReproductiveOrgan.cs
+++ b/Models/Plant/Organs/ReproductiveOrgan.cs
@@ -395,12 +395,19 @@ namespace Models.PMF.Organs
         {
             get
             {
+                // Biomass allocation today
                 double dmAllocation = potentialDMAllocation.Storage +
                     potentialDMAllocation.Structural +
                     potentialDMAllocation.Metabolic;
-                // Required the nitrogen content
+                // Required the minimum nitrogen content according to biomass at the end of day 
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@byzheng this change represents a different assumption for how how biomass is deposited in the grain.  The assumption we were working on is that once biomass is fixed into the grain it is immobile and its N content locked away.  So if the N content of the grain is above minN early but there is not enough N to meet demands later the grain will deposit biomass at the minN conc.  
I am not certain which assumption is correct (what you propose or what is there currently) but the more I think about it the more it seems the one you propose is correct.
We have got some issues with diffs with this change but still waiting on the performance testing site fix to review these

byzheng:
@HamishBrownPFR I have gone through performance test. The big difference comes from TarleeRotationTrial with higher grain yield as I expected. Could you have a look this pull request again?

byzheng:
We find another way to temporary solve this problem through allowing retranslocation of grain nitrogen by itself.

PR #3617, opened 2019-03-14
Use a technique to get the number of available processors that works …

### LINKED ISSUE #3614: Writing to the SQLite datastore is slow on physical disk drives
A bit of testing has shown that writing to the DataStore can be slow when using a physical disk drive; writing to SSD drives tends to be suitably quick. This typically manifests as a very long wait at the end of a set of simulations (the wheat validation set provides a good example of this).

zur003:
…cross-platfom.

Use SQLite pragmas to greatly improve disk access speed.
Allow the DataStore to be emptied in cases where the _Checkpoints table has not yet been written.

Resolves #3614

### DIFF
--- a/Models/Core/Runners/JobRunnerMultiProcess.cs
+++ b/Models/Core/Runners/JobRunnerMultiProcess.cs
@@ -61,12 +61,7 @@ namespace Models.Core.Runners
 
             // Determine number of threads to use
             if (numberOfProcessors == -1)
-            {
-                string numOfProcessorsString = Environment.GetEnvironmentVariable("NUMBER_OF_PROCESSORS");
-                if (numOfProcessorsString != null)
-                    numberOfProcessors = Convert.ToInt32(numOfProcessorsString);
-                numberOfProcessors = System.Math.Max(numberOfProcessors, 1);
-            }
+                numberOfProcessors = System.Math.Max(Environment.ProcessorCount - 1, 1);
 
             cancelToken = new CancellationTokenSource();
 
--- a/Models/Core/Runners/JobRunnerMultiProcess.cs
+++ b/Models/Core/Runners/JobRunnerMultiProcess.cs
@@ -61,7 +61,14 @@ namespace Models.Core.Runners
 
             // Determine number of threads to use
             if (numberOfProcessors == -1)
-                numberOfProcessors = System.Math.Max(Environment.ProcessorCount - 1, 1);
+            {
+                int number;
+                string numOfProcessorsString = Environment.GetEnvironmentVariable("NUMBER_OF_PROCESSORS");
+                if (numOfProcessorsString != null && Int32.TryParse(numOfProcessorsString, out number))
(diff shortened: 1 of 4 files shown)


peter-devoil:
Using Environment.ProcessorCount instead of the environment variable "number_of_processors" will cause grief when a) it hasn't been allocated all CPUs, and b) if there's a job monitor that kills off rogues that exceed nominated CPU usage thresholds. b) is unavoidable on PBS schedulers (ie most universities), and a) occurs when the user is trying to target partially allocated machines in a pool running typically long jobs.

zur003:
@peter-devoil - what do you suggest? The problem I was seeing was that the NUMBER_OF_PROCESSORS environment variable is normally only defined on Windows, so Linux and OS X users were, by default, only getting the use of a single processor.

zur003:
Would it be OK to use the NUMBER_OF_PROCESSORS environment variable if it is defined, then fall back to Environment.ProcessorCount if it is not? That would have the advantage of providing the user with control over processor use, even if the mechanism is a bit obscure.

peter-devoil:
Yes, I think that's the best approach. Most of the time, you want unfettered usage, but sometimes you need to throttle back.

zur003:
Retest this please Jenkins

hol353:
What's the reasoning behind ExcelInput implementing IComputationalyTimeConsuming?

zur003:
@hol353 - The Jenkins builds were failing to run Examples\Slurp.apsimx. I eventually tracked this down to being a thread-related problem. This is a very small, quick, simulation and my changes had sped up the writing of the Report and other tables to the point that they were completing before the threads reading the Excel Observed data and writing that data to the database had time to do their work. I get rather lost in the multi-threading logic, but marking the ExcelInput with IComputationalyTimeConsuming [sic] apparently ensures that those jobs have actually completed.

zur003:
This is now dependent on a pull request in APSIM.Shared. Could that be merged?

zur003:
Retest this please Jenkins

hol353:
Hmm. RunOrganiser is supposed to wait until all jobs have finished before it runs any post simulation tools. Sounds like that isn't happening.

zur003:
Yeah, threading issues are difficult to debug. I think what's happening is that the reading of the Excel files in ExcelInput is sufficiently slow that those threads haven't submitted the tables to the StorageWriter before the StorageWriter thread has completed writing the other tables. There may be a better solution.

hol353:
@zur003 I managed to reproduce the problem with Slurp and I've come up with a better fix. As you say it wasn't waiting for the ExcelInput model to complete. The fix in Apsim.Shared should work. Can you remove your ExcelInput change, pull from Apsim.Shared and see if that works?

hol430:
I have an SSD, but I just experienced the long pause when the progress bar is on 100%, so perhaps the problem is not related to your storage medium.

That being said, the two sqlite pragmas appear to fix the problem, so I guess it's a bit of a moot point.

PR #3626, opened 2019-03-18
Fixed minor problem with cleaning up memory on destruction of a TreeView

### LINKED ISSUE #3618: Tab display is corrupted when simulations are opened
If you are not running in a build environment, with source code at hand, the tab for the first simulation opened hides the "Home" tab.

zur003:
Resolves #3618

### DIFF
--- a/ApsimNG/Views/MainView.cs
+++ b/ApsimNG/Views/MainView.cs
@@ -314,12 +314,13 @@
             eventbox.TooltipText = text;
             eventbox.ButtonPressEvent += on_eventbox1_button_press_event;
             eventbox.Add(headerBox);
-            eventbox.ShowAll();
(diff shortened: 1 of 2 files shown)


hol430:
Does this fix #3618?

zur003:
@hol430 - Yes, it should fix #3618. I accidentally did a git commit while there was already a pull request open, and so combined to different patches,.

hol430:
Should probably edit the initial comment then, to add "Resolves #3618". Weird how the bug doesn't show up in debug mode - maybe it's the result of some optimisation performed by the compiler in release mode.

zur003:
I changed the initial comment. Do I need to have Jenkins run it again?

hol430:
No need - Jenkins doesn't go looking for a the version number until after this is merged, when the CreateInstallation job gets run.

PR #3631, opened 2019-03-18
Prevent git from tracking .db-shm and .db-wal files.

hol430:
Working on #3614

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,5 @@ Documentation/errors.txt
 .vscode/
 *.db
 *.db-journal
+*.db-shm
+*.db-wal


zur003:
One question here is why there are any .db-shm and .db-wal files in the first place. These should be deleted automatically when a database connection is closed. Their presence would suggest that something is not being cleaned up when it ought to be.
Of course developers often kill a program in mid-execution when debugging, which will leave these things lying around, so it does make sense to filter them out of git.

hol430:
Yeah I'm pretty sure I killed apsim from visual studio while running the factorial example. This was the only file which generated the .db-shm and .db-wal files. 

The other .apsimx files which I've run and closed properly seem to have correctly cleaned up after themselves.

PR #3632, opened 2019-03-19
 Allow users to save results to a different directory/filename. 

### LINKED ISSUE #3615: New feature - output .db files to another directory
Hi Team, 

Follow the discussion with @HamishBrownPFR about outputting the simulation results to a different directory to the one the `.apsimx` files in. 
It would be great to have this feature that allows users to define an output directory if it is not too tricky. 

Thank you.

hol430:
Resolves #3615 

@hol353 I have added `Open()` and `Close()` methods to IDataStore. Is this ok?


Models/Storage/DataStore.cs
```diff
@@ -39,7 +39,6 @@ public class DataStore : Model, IDataStore, IDisposable
         public bool useFirebird { get; set; } = false;
 
         /// <summary>Returns the file name of the .db file</summary>
-        [XmlIgnore]
```
hol353: By removing this XmlIgnore, the filename will always be written to the .apsimx file. This means that if the user moves their apsimx file to another directory it will still write the .db file in the original directory because it will use the path as stored in the .apsimx. Need to find a way around this.

hol353:
Having Open() and Close() methods is OK.

Models/Storage/DataStore.cs
hol430: Ah that's true - that would be a problem. Although moving the file will usually result in an error anyway unless you move the .met files as well.

The problem is that if a custom file name is given, we want to continue using that filename, even if the file is moved. On the other hand, the default filename should be changed if the file is moved.

To marry these two different use cases, I've added a new property - CustomFileName. The DataStore will use the custom file name if it is given, and will fallback to the default file name when a custom file name is not provided.

hol353:
This now has a conflict.

Issue #3659, opened 2019-03-26
Decreasing total biomass (above ground) in the later grain filling period


byzheng:
## Problem description

In the `CalculatedNutrientConstrainedDMAllocation` of `OrganArbitrator.cs`, the actual biomass allocation is constrained by minimum nitrogen concentration of each organ. The extra biomass is thrown out from the system including fixation (photosynthesis) and retranslocation. It might make sense to fixation, but doesn't for retranslocation. 

Consequently, the total above ground biomass of wheat might decrease during later grain filling period if there are not enough nitrogen for grain filling. For example, the figure below shows the results in one trial. 

![image](https://user-images.githubusercontent.com/1657381/54975927-6c529500-4fe4-11e9-8bbf-32ea71de11fe.png)

## Proposed solutions

I believe the extra biomass should not throw out the system, at least for retranslocation.

* Only apply the nutrient constrained to fixation (photosynthesis). If there are not enough nitrogen for allocated biomass, the retranslocation biomass will be restored into original organs. I guess we need to store the sources of allocated biomass for each organ and will require lots of coding. 
* Extra biomass stores into a specific organ as storage (e.g. stem in wheat). In this case, we only need to look through all organs to find the one with biggest storage demand and add extra biomass into that organ. And there are no constrains for storage and metabolic.

@HamishBrownPFR Do you have any suggestion about this issue?

HamishBrownPFR:
I am not certain you you diagnosed that correctly.  Going of memory and how it should work.  
 Translocation is only taken out of an organ if it is allocated, it is not taken out of the organ then discarded if not allocated.  The reduction in above ground biomass could be detachment of dead biomass?

byzheng:
Hi @HamishBrownPFR  I don't think there are any detachment for wheat before harvest. See the figure below. The above ground biomass is decreased since 121 DAS. 

![image](https://user-images.githubusercontent.com/1657381/55200057-2c6bf780-5208-11e9-874e-d4e93b6273e0.png)

BenAbabaei:
It would help if you show senescence and growth (allocation) on the same graph.

byzheng:
@BenAbabaei Thanks for your looking. This issue is not related with the senescence and allocation. We should not expect total above ground biomass decreasing during the growth season if there is no detachment and no retranslocation when there are not enough nitrogen to satisfy the requirement of minimum nitrogen concentration.

PR #3660, opened 2019-03-26
Allocate extra biomass to organ with biggest storage.

byzheng:
Working on #3659

Two major changes about science in this pull request.

* Nutrition constrain only applies for structural, not metabolic and storage
* Nutrition constrain DM is allocated to the organ with biggest storage demand


byzheng:
This issue is addressed by @HamishBrownPFR  at #3834

PR #3676, opened 2019-03-31
Added MathNet.Numerics to ApsimNG/Intellisense

### LINKED ISSUE #3673: Adding Accord.Net Statistics package into ApsimX
Accord (http://accord-framework.net/) is a whole bunch of useful libraries for many things (don't be fooled by the Machine Learning up front and centre in the website) including the generation of statistical distributions. I would like the ability to generate some bi-variate log-normal distributions in a manager script.  @FlorjanCamlek has tested using this in a standalone application getting Accord from https://www.nuget.org/packages/Accord.Statistics/3.8.2-alpha. He has also populated a simple test that might work and is attached. @hol353 would you mind seeing if this is OK/suitable to include with APSIM and adding it to the solution if it is?  Thanks, Val

[TestAccord.zip](https://github.com/APSIMInitiative/ApsimX/files/3020956/TestAccord.zip)

hol430:
Resolves #3673 

Unfortunately, because the intellisense lives in ApsimNG, I've had to add a MathNet.Numerics reference to ApsimNG. Not Ideal, but it should keep Val happy for now. Might be worth moving a lot of the intellisense code into Models one day...and maybe making the intellisense/manager compilation code figure out for itself which assemblies it needs.

### DIFF
--- a/ApsimNG/ApsimNG.csproj
+++ b/ApsimNG/ApsimNG.csproj
@@ -930,6 +930,10 @@
       <HintPath>..\packages\MarkdownDeep.NET.1.5\lib\.NetFramework 3.5\MarkdownDeep.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="MathNet.Numerics, Version=4.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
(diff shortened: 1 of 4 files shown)


hol430:
APSIMInitiative/APSIM.Shared#86 should be merged after this is merged

PR #3677, opened 2019-03-31
 PropertyPresenter should show units in description column, if possible. 

### LINKED ISSUE #3675: Confusing description of attributes when they are showing in a table
For example, in the InitialLeaves[1], there are two attributes Rank and Area which are showing in a table. It is confusing for the users for `Area mm2` when users replace values in the cultivars. 

image

It will be more clear if the name and unit can be separated into two columns.

|  Name |  Unit  |  Value  |
|---|---|---|
|  Rank |   | 1  |
|  Area |  mm2  |  200  |

hol430:
Resolves #3675


hol430:
retest this please jenkins

hol430:
retest this please jenkins

PR #3717, opened 2019-04-09
Refactor experiment/factor and graph series code.

### LINKED ISSUE #3716: Refactor of graph series and experiment / factor code
The code in experiments / factor / factorvalue is very convoluted. This needs a tidy up. Also, the Morris and Sobol models have 'factors' as well. Need to broaden factors beyond experiments so that they can be produced by and model. This has implications for graph series as well.

hol353:
Resolves #3716


hol353:
retest this please jenkins

PR #3721, opened 2019-04-10
2 Bugfixes for file format

### LINKED ISSUE #3719: Exceptions thrown when opening file are ignored
When opening a file via a call to `FileFormat.ReadFromFile<T>(string, out List<Exception>)`, we often seem to ignore any exceptions returned via the out parameter. As a result, Models.exe can sometimes exit with exit code 0 after running into an error.

### LINKED ISSUE #3720: Manager scripts cannot override the OnCreated() method
When we open an .apsimx file, we loop through all models in the tree, and call their `OnCreated()` method. However, a manager's script model does not exist in the tree until the manager's `OnCreated()` method is run. This means that although a script may override the `OnCreated()` method, this method will (currently) never be run.

hol430:
Resolves #3720 and resolves #3719


hol430:
retest this please jenkins

PR #3725, opened 2019-04-11
Add an empty row at top of intellisense popup

### LINKED ISSUE #3724: Add an empty row when automatically list variables
When i try to modify an variable in the `report` or `VariableReference`, ApsimVersion is added when I press `dot` to show the variable list and then `Enter` to finish editing. See example below 
changevalue

Then I have to manually delete `ApsimVersion` or press `Esc` after pressing dot. I know it is not a big issue, but would be better to add an empty item at the top of variable list. In this case, no variable is selected in default.

hol430:
Resolves #3724 

This resolves the issue at hand, but I wonder if it would be better to allow the user to continue typing into the primary input field (report, grid cell, manager, or whatever), and simply narrow down the list of options shown based on what the user types. This would be similar to how visual studio handles it. That way, the user doesn't need to press anything to close the popup  - they can just continue typing and ignore the intellisense if they want.

### DIFF
--- a/ApsimNG/Views/IntellisenseView.cs
+++ b/ApsimNG/Views/IntellisenseView.cs
@@ -241,8 +241,6 @@
                 completionForm.Move(x, y);
                 completionForm.Resize(completionForm.WidthRequest, completionForm.HeightRequest);
                 completionView.SetCursor(new TreePath("0"), null, false);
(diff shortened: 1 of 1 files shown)


jbrider:
I would prefer it to work similar to VisualStudio.

Issue #3689, opened 2019-04-02
Get AgPasture into release


sno036:
Past due that AgPasture was officially in release. Here starting the process. This is what Mariana is doing.

sno036:
@hol353 can you add "mpandreucci" to the project please? (is that 'whitelisting'?)

hol353:
@sno036 @mpandreucci Is this now ready for review by the Reference Panel?

mpandreucci:
I don't know if @sno036 will want to add some managers to it. If she doesn't want to add anything else to it, then I think it's ready.
The only thing is that the graph still has some black points, instead of all of them being coloured. I'm not sure if it's something that needs to be fixed by you @hol353

hol353:
Which graph is it?

mpandreucci:
The predicted x observed from the Radcliffe validation test

sarahcleary:
@APSIMInitiative/reference-panel  - to be discussed in Friday's RP meeting. In the meantime, any suggestions for reviewers?

sno036:
Matt Harrison suggested as a reviewer

sarahcleary:
RP agreed to see if Matt Harrison was available and willing to review (AI RP May Meeting).  @sarahcleary to contact Matt Harrison

sarahcleary:
@sno036 - can you please provide the instructions for the reviewer - where can he find the validation example and documentation?

sarahcleary:
Sent request to Matt Harrison to see if he is available to review model

sarahcleary:
[AgPasture review ApsimX Matthew Harrison.docx](https://github.com/APSIMInitiative/ApsimX/files/3284212/AgPasture.review.ApsimX.Matthew.Harrison.docx)
[AgPastureTest3929MH1.pdf](https://github.com/APSIMInitiative/ApsimX/files/3284213/AgPastureTest3929MH1.pdf)
[AgPasture3929MH2.pdf](https://github.com/APSIMInitiative/ApsimX/files/3284214/AgPasture3929MH2.pdf)
Review by Matthew Harrison.  Minor comments attached.

sarahcleary:
@APSIMInitiative/reference-panel - Review docs from Matthew Harrison above.  Please read prior to July RP meeting

sarahcleary:
@sno036 - please note comment in RP Meeting

sno036:
[...]

PR #3727, opened 2019-04-11
Ag pasture dev

mpandreucci:
working on #3689


hol353:
Good work @mpandreucci. It ran green. But...
Can you remove the PDF files? We don't commit them.

mpandreucci:
Yup, I’ll do that. I ended up staging everything at one point and that’s why they ended up there. Sorry

mpandreucci:
@hol353 how do I take the pdf file out of it? I asked Laura but she didn't know how to do it through source tree. Do I need to delete it from my prototypes folder?

jbrider:
@mpandreucci in Sourcetree, there should be a 'stop tracking' option when right clicking on the file in the staging window - that would be the easiest way.
Alternatively - if you remove the file from the folder it should show up in Sourcetree as removed - you can then commit that file as being removed.

mpandreucci:
Hi @jbrider . I'll have a go at removing it and then committing the change because I couldn't find out how to 'Stop tracking' it.
Thanks for that.

hol353:
@mpandreucci :There are still 2 pdf files in here. Look at the 'Files changed' tab in the pull request.

mpandreucci:
I'll delete them too (not sure how they ended up there since they have nothing to do with agpasture). Sorry about that

PR #3728, opened 2019-04-11
Fix: field not found in pred/obs table with mix of expts and simulations

hol353:
Working on #3716


jbrider:
Would this account for the drop in performance?  Clicking on a folder with multiple graphs is noticeably slower.

hol353:
Nope. I've noticed it is a bit slower on my computer as well. I'm still looking at this stuff.

PR #3743, opened 2019-04-15
Issue/expose micro climate outputs

### LINKED ISSUE #3740: Need some outputs from MicroClimate
Currently MicroClimate has no true outputs that can be used in the report (the only values available are inputs). We need to enable a few in order to understand water balance calculations

rcichota:
Resolves #3740 for now
I have exposed a few of the outputs from MicroClimate, basically those that were exposed in Micromet: interception, gc, ga, petr, peta, pet_total, net_radn, net_rs, net_rl, soil_heat, and dryleaffraction, which I have renamed as: PrecipitationInterception, CanopyConductance, AerodynamicConductance, PetRadiationTerm, PetAerodynamicTerm, PetTotal, NetRadiation, NetLongWaveRadiation, NetShortWaveRadiation, SoilHeatFlux, and DryLeafTimeFraction. I also added outputs for RadiationInterception and CanopyCover (which I wanted for multi-species simulations). 
An important note is that all of these only return the values for the first zone; I am not sure if it is dangerously incorrect or is acceptable for now. It would be good if @hol353 and/or @hut104 would have a quick look. My guess is that it is ok for now, if we need to change to account for all zones this can be done when MicroClimate is re-factored (on when actually needed).
In the end I have commented out the outputs for canopy and aerodynamic conductance (gc and ga) as I am not sure these was done correctly in Apsim classic (there is even a comment in there). The issue is that only the first value was being reported (which I interpreted as being from the top layer of the first crop). This indeed seems odd. In the Report written about Micromet there are formulas to compute the aggregate values, but it seems they were never implemented. Maybe @sno036 and @hut104 can comment on that, but they are only outputs and I don't need them right now, so will leave to another opportunity.


rcichota:
Retest this please Jenkins

PR #3747, opened 2019-04-17
Fixed syntax error in jenkins script.

hol430:
Working on #1281


hol430:
@hol353 Not sure why this failed the performance tests. The code that I've changed does not even get run until Jenkins' CreateInstallation job.

hol353:
The performance tests are showing that we suddenly have fodderbeet stats when we didn't have previously. Looks like the fodderbeet stats went missing when I accepted the new stats and merged #3746 this morning. Will accept the stats again now that they are back. Don't know why this happened.

PR #3749, opened 2019-04-17
Prototype dark theme

### LINKED ISSUE #820: Graphics in high contrast mode
The background of graphics is black in the Windows high contrast mode (See figure below). It will be east to read through  fixing the background colour as white.

image

### LINKED ISSUE #3380: Mechanism for applying Gtk style to HTML elements.
As mentioned in the discussions [here](https://github.com/APSIMInitiative/ApsimX/pull/3345#discussion_r238860905) and [here](https://github.com/APSIMInitiative/ApsimX/pull/3345#discussion_r240465453), we need some mechanism which can apply Gtk style to elements in the HTMLView. In my work on the dark theme, I've implemented a crude hack which basically sets the background and foreground colours to the Gtk background/foreground colours of widgets in an insensitive state. While this works, it doesn't allow for more advanced Gtk style settings to be applied to the HTMLView.

I'm not sure what will be the best way to solve this.

### LINKED ISSUE #3748: Implement a dark theme
It would be handy to have a dark theme for apsim. This would reduce eye strain when switching between apsim and other dark programs such as a terminal, IDE, etc.

It would also be good if the default (light) theme were compatible with any system-wide gtk themes which the user may have installed.

hol430:
Resolves #3748 
Resolves #820
Resolves #3380


hol353:
Is this one ready for merging?

hol430:
Yeah I guess so. This will probably introduce some issues, but we'll just have to sort them out as they show up.

PR #3754, opened 2019-04-17
 Don't abort early when running post sim tools 

### LINKED ISSUE #3751: Post simulation tools don't always run
If a post-simulation tool fails, no other post-simulation tools will be run. Apsim will immediately abort the run and post an error message. One problem with this behaviour shows up when a user has a PredictedObserved model and a probability model in the same file. If the predicted data comes from simulation x, but the user only runs simulation y, the PredictedObserved model will throw an error, and the probability model will not run.

hol430:
Resolves #3751


hol430:
retest this please jenkins

PR #3771, opened 2019-04-18
If graph annotation has no colour, use default.

### LINKED ISSUE #3762: Problems with graphs after recent changes
- When I copy the graph into Paint, the background is black.
image

- Regression equation and statistics are not show,
image

- The way the legend items are sorted still doesn't make sense. It seems it still depends on the order of the simulations.
image

### LINKED ISSUE #3779: Reinstate stats on graphs
Recent releases of apsimx have OvP graphs with stats omitted - please reinstate.

Thanks,
Philip

hol430:
Resolves #3762
Resolves #3779
This will fix the bug where regression equation doesn't show up.


hol430:
retest this please jenkins

PR #3780, opened 2019-04-29
Modifications to support Firebird databases.

zur003:
Working on #2821.

This patch allows the use of Firebird databases for the datatore, although to do so requires the user to manually modify the .apsimx file. Firebird support should still be regarded as "test" only. There is no good reason to switch to it, as performance is a bit poorer than that of the existing SQLite implementation. However, I think it's still worth having it present, as it could easily be adjusted to allow the database to reside on a remote server rather than the local machine, and this might be useful for bulk Condor or Docker runs.

### DIFF
--- a/ApsimNG/Presenters/CLEM/PropertyTablePresenter.cs
+++ b/ApsimNG/Presenters/CLEM/PropertyTablePresenter.cs
@@ -564,7 +564,10 @@ namespace UserInterface.Presenters
                         DataTable data = null;
                         if (storage.Reader.TableNames.Contains(tableName))
                         {
-                            data = this.storage.Reader.GetDataUsingSql("SELECT * FROM " + tableName + " LIMIT 1");
+                            if ((storage.Reader is DataStoreReader) && (storage.Reader as DataStoreReader).connection is Firebird)
+                                data = this.storage.Reader.GetDataUsingSql("SELECT FIRST 1 * FROM [" + tableName + "]");
+                            else
+                                data = this.storage.Reader.GetDataUsingSql("SELECT * FROM [" + tableName + "] LIMIT 1");
                         }
 
                         if (data != null)
--- a/ApsimNG/Presenters/CLEM/PropertyTablePresenter.cs
+++ b/ApsimNG/Presenters/CLEM/PropertyTablePresenter.cs
@@ -561,19 +561,8 @@ namespace UserInterface.Presenters
                     if (cell.Value != null && cell.Value.ToString() != string.Empty)
                     {
                         string tableName = cell.Value.ToString();
-                        DataTable data = null;
                         if (storage.Reader.TableNames.Contains(tableName))
-                        {
-                            if ((storage.Reader is DataStoreReader) && (storage.Reader as DataStoreReader).connection is Firebird)
-                                data = this.storage.Reader.GetDataUsingSql("SELECT FIRST 1 * FROM [" + tableName + "]");
-                            else
-                                data = this.storage.Reader.GetDataUsingSql("SELECT * FROM [" + tableName + "] LIMIT 1");
-                        }
-
-                        if (data != null)
-                        {
-                            fieldNames = DataTableUtilities.GetColumnNames(data);
-                        }
+                            fieldNames = storage.Reader.ColumnNames(tableName).ToArray<string>();
                     }
                 }
             }
--- a/ApsimNG/Presenters/DataStorePresenter.cs
+++ b/ApsimNG/Presenters/DataStorePresenter.cs
@@ -181,25 +181,7 @@ namespace UserInterface.Presenters
                         var names = ExperimentFilter.GenerateSimulationDescriptions().Select(s => s.Name);
                         string filter = "S.[Name] IN " + "(" + StringUtilities.Build(names, delimiter: ",", prefix: "'", suffix: "'") + ")";
                         if (!string.IsNullOrEmpty(view.RowFilter.Value))
-                        {
-                            // For Firebird, we need to convert column names to their short form to perform the query
-                            if (dataStore.Reader is DataStoreReader && (dataStore.Reader as DataStoreReader).connection is Firebird)
-                            {
-                                string rowFilter = view.RowFilter.Value;
-                                List<string> output = rowFilter.Split('[', ']').Where((item, index) => index % 2 != 0).ToList();
-                                foreach (string field in output)
-                                {
-                                    string shortName = ((dataStore.Reader as DataStoreReader).connection as Firebird).GetShortColumnName(view.TableList.SelectedValue, field);
-                                    if (!string.IsNullOrEmpty(shortName))
-                                    {
-                                        rowFilter = rowFilter.Replace("[" + field + "]", "[" + shortName + "]");
-                                    }
-                                }
-                                filter += " AND " + rowFilter;
-                            }
-                            else
-                                filter += " AND " + view.RowFilter.Value;
-                        }
+                            filter += " AND " + view.RowFilter.Value;
                         data = dataStore.Reader.GetData(tableName: view.TableList.SelectedValue, filter: filter, from: start, count: count);
                     }
                     else if (SimulationFilter != null)
--- a/ApsimNG/Presenters/DataStorePresenter.cs
+++ b/ApsimNG/Presenters/DataStorePresenter.cs
@@ -3,7 +3,6 @@
 namespace UserInterface.Presenters
 {
     using System;
-    using System.Collections.Generic;
     using System.Data;
     using System.Linq;
     using APSIM.Shared.Utilities;
--- a/ApsimNG/Presenters/DataStorePresenter.cs
+++ b/ApsimNG/Presenters/DataStorePresenter.cs
@@ -3,6 +3,7 @@
 namespace UserInterface.Presenters
 {
     using System;
+    using System.Collections.Generic;
     using System.Data;
     using System.Linq;
     using APSIM.Shared.Utilities;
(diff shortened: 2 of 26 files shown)


ApsimNG/Presenters/CLEM/PropertyTablePresenter.cs
```diff
@@ -564,7 +564,10 @@ private string[] GetFieldNames(int propListIndex)
                         DataTable data = null;
                         if (storage.Reader.TableNames.Contains(tableName))
                         {
-                            data = this.storage.Reader.GetDataUsingSql("SELECT * FROM " + tableName + " LIMIT 1");
+                            if ((storage.Reader is DataStoreReader) && (storage.Reader as DataStoreReader).connection is Firebird)
+                                data = this.storage.Reader.GetDataUsingSql("SELECT FIRST 1 * FROM [" + tableName + "]");
+                            else
+                                data = this.storage.Reader.GetDataUsingSql("SELECT * FROM [" + tableName + "] LIMIT 1");
                         }
```
hol353: Can we abstract away the check for Firebird? If we call Reader.GetData instead of Reader.GetDataUsingSql then we should be able to get rid of this test.

ApsimNG/Presenters/DataStorePresenter.cs
```diff
@@ -178,9 +179,27 @@ private DataTable GetData()
...
                     {
                         var names = ExperimentFilter.GenerateSimulationDescriptions().Select(s => s.Name);
-                        string filter = "S.NAME IN " + "(" + StringUtilities.Build(names, delimiter: ",", prefix: "'", suffix: "'") + ")";
+                        string filter = "S.[Name] IN " + "(" + StringUtilities.Build(names, delimiter: ",", prefix: "'", suffix: "'") + ")";
                         if (!string.IsNullOrEmpty(view.RowFilter.Value))
-                            filter += " AND " + view.RowFilter.Value;
+                        {
+                            // For Firebird, we need to convert column names to their short form to perform the query
+                            if (dataStore.Reader is DataStoreReader && (dataStore.Reader as DataStoreReader).connection is Firebird)
+                            {
```
hol353: Any way we can get rid of the test for Firebird?

ApsimNG/Presenters/PropertyPresenter.cs
```diff
@@ -593,7 +593,12 @@ private List<string> GetFieldNames()
                         string tableName = cell.Value.ToString();
                         DataTable data = null;
                         if (storage.Reader.TableNames.Contains(tableName))
-                            data = storage.Reader.GetDataUsingSql("SELECT * FROM " + tableName + " LIMIT 1");
+                        {
+                            if ((storage.Reader is DataStoreReader) && (storage.Reader as DataStoreReader).connection is Firebird)
+                                data = storage.Reader.GetDataUsingSql("SELECT FIRST 1 * FROM [" + tableName + "]");
+                            else
+                                data = storage.Reader.GetDataUsingSql("SELECT * FROM [" + tableName + "] LIMIT 1");
+                        }
```
hol353: Call Reader.GetData instead.

Models/Graph/Series.cs
```diff
@@ -726,7 +726,25 @@ private DataTable GetDataInScope(IStorageReader reader, List<SimulationDescripti
                                    StringUtilities.Build(simulationNames, ",", "'", "'") +
                                    ")";
             if (Filter != null && Filter != string.Empty)
-                filterToUse += " AND " + Filter;
+            {
+                // For Firebird, we need to convert long column names into their short form.
+                if (reader is DataStoreReader && (reader as DataStoreReader).connection is Firebird)
+                {
```
hol353: What is this code doing? Can we get rid of the test for Firebird?

Models/PostSimulationTools/PredictedObserved.cs
```diff
@@ -77,60 +77,165 @@ public void Run(IDataStore dataStore)
...
                 StringBuilder query = new StringBuilder("SELECT ");
-                foreach (string s in commonCols)
+
+                if (dataStore.Reader is DataStoreReader && (dataStore.Reader as DataStoreReader).connection is Firebird)
                 {
-                    if (s == FieldNameUsedForMatch || s == FieldName2UsedForMatch || s == FieldName3UsedForMatch)
-                        query.Append("O.'@field', ");
-                    else
-                        query.Append("O.'@field' AS 'Observed.@field', P.'@field' AS 'Predicted.@field', ");
+                    Firebird fb = (dataStore.Reader as DataStoreReader).connection as Firebird;
```
hol353: And again. I really don't want tests for Firebird in models.

Models/Storage/DataStoreReader.cs
```diff
@@ -14,7 +14,7 @@
     public class DataStoreReader : IStorageReader
     {
         /// <summary>A database connection</summary>
-        private IDatabaseConnection connection = null;
+        public IDatabaseConnection connection { get; private set; } = null;
```
hol353: Public properties should start with uppercase.

Models/Storage/DataStoreReader.cs
```diff
@@ -177,10 +177,32 @@ public void Refresh()
                 fieldList = fieldNames.ToList();
 
             bool hasSimulationName = fieldList.Contains("SimulationID") || fieldList.Contains("SimulationName") || simulationName != null;
+            bool hasCheckpointName = fieldNamesInTable.Contains("CheckpointID") || fieldNamesInTable.Contains("CheckpointName") || checkpointName != null;
 
-            sql.Append("SELECT C.[Name] AS CheckpointName, C.[ID] AS CheckpointID");
+            sql.Append("SELECT ");
+
+            if (count > 0 && connection is Firebird)
+            {
```
hol353: Having tests for Firebird here is not so bad but the ultimate would be to abstract everything dealing with database access to the SQLite.cs and Firebird.cs classes. Is this possible?

Models/PostSimulationTools/PredictedObserved.cs
```diff
@@ -77,60 +77,165 @@ public void Run(IDataStore dataStore)
...
+
+                        query.Append(" AND O.[SimulationID] in (");
+                        foreach (string simulationName in simulationNames)
+                        {
+                            if (simulationName != simulationNames[0])
+                                query.Append(',');
+                            query.Append(dataStore.Writer.GetSimulationID(simulationName));
+                        }
+                        query.Append(")");
+                    }
```
hol353: Wow. What is all this code doing?

Models/Storage/EmptyCommand.cs
```diff
@@ -56,7 +59,11 @@ public void Run(CancellationTokenSource cancelToken)
             {
                 if (!tableName.StartsWith("_"))
                 {
-                    var data = database.ExecuteQuery(string.Format("SELECT * FROM {0} LIMIT 1", tableName));
+                    DataTable data = null;
+                    if (database is Firebird)
+                        data = database.ExecuteQuery(string.Format("SELECT FIRST 1 * FROM [{0}]", tableName));
+                    else
+                        data = database.ExecuteQuery(string.Format("SELECT * FROM [{0}] LIMIT 1", tableName));
```
hol353: Do we need a SelectQuery method in IDatabaseConnection to abstract away limit / first?

Models/Storage/InsertQuery.cs
```diff
@@ -29,42 +32,33 @@ public InsertQuery(string tableNameForQuery)
...
-                var queryHandle = GetPreparedQuery(sqlite, columnNames);
-                sqlite.BindParametersAndRunQuery(queryHandle, rowValues);
-            }
-            else
-            {
-                List<object[]> values = new List<object[]>() { rowValues.ToArray() };
-                database.InsertRows(TableName, columnNames.ToList(), values);
-            }
+            var queryHandle = GetPreparedQuery(database);
+            database.RunBindableQuery(queryHandle, rowValues);
```
hol353: Yes, this is nice. Gets rid of a test for database type.

ApsimNG/Presenters/PropertyPresenter.cs
zur003: Using GetData would be overkill. What is needed here, and also with CLEM/PropertyTablePresenter.cs and Storage/EmptyCommand.cs, is a function to quickly test whether or not a table is empty. It's currently doing that by testing whether it can read 1 record. Using GetData would cause the entire table (which could be large) to be read. That's overhead these calls are trying to avoid. Probably we need an IsTableEmpty(tableName) function in IDatabaseConnection.

Ideally, I'd prefer having no SQL at all in ApsimNG or Models - it should all reside deep in the bowels of APSIM.Shared.

Models/PostSimulationTools/PredictedObserved.cs
zur003: The biggest hassle in dealing with Firebird databases is that field (column) names are limited to 31 characters. ApsimX has stuff like "Wheat.Phenology.FloweringDAS.Value" and "Wheat.Structure.FinalLeafNumber.Value" where the labels get too long to be used as field names. I've tried to get around this by adding an additional table in Firebird to map these "long" table names to short ones (the short ones being things like, "COL_7" or "COL_9"). The trick in trying to automate conversion between the long and short names as much as possible. This gets very tricky with the PredictedObserved tables, since the SQL has to use only the "short" names, but we're trying to join predicted and observed tables on the basis of "long" names, and the mappings between short and long are different in the predicted and observed tables.

So what's happening here is we look at the long names of the predicted and observed tables, find any matches, convert those long names back to short ones to do a query with joins, then convert the names in the result back to long names again.

It's ugly, but I couldn't think of a better way. 
If you can think of a simple way to do this, I'd welcome it.

zur003:
Retest this please Jenkins

ApsimNG/Presenters/PropertyPresenter.cs
hol353: Yes, I'd prefer to have no SQL anywhere except SQLite.cs and Firebird.cs. The GetData method actually has a count argument that can be used to only read one record. Your TableIsEmpty method works well though.

Models/PostSimulationTools/PredictedObserved.cs
hol353: If we have a QueryJoin method of IDatabaseConnection that Predicted/Observed called would that hide this code in Firebird.cs?

Models/Storage/DataStore.cs
```diff
@@ -222,5 +222,19 @@ public void Close()
...
+        /// <returns></returns>
+        public IEnumerable<string> ColumnNames(string tableName)
+        {
+            if (connection != null)
+                return connection.GetColumnNames(tableName);
+            else
+                return null;
+        }
+
     }
```
hol353: Why is this needed. Doesn't DataStore.Reader have a ColumnNames method?

zur003: Good call. It's not needed.

hol353:
@zur003 Are you wanting this merged?

zur003:
@hol353 - Yes, please. The PredictedObserved stuff is still ugly, but may have to do for now.

Issue #3795, opened 2019-05-02
Filter in graphs is not working


rcichota:
Graphs that have a filter set up are not showing up and cannot be edited, this seems to be worse for observed values. This may be related to #3716 
I have a set of graphs for WhiteClover validation where I used a filter to show graphs for each NRate (I had set something like:```NRate = 100```), this seem to no longer work. No graph is produced and the Series window is all greyed out and one cannot change anything.  Creating a graph from scratch allow me to plot predicted values (but only if I use:```NRate LIKE 100```), attempt to filter using the "=" sign freezes the Series window. If I try to plot only the observed data, no option works, giving the error message: "A column named 'NRate' already belongs to this DataTable".

zur003:
Perhaps that should that be: [NRate] = 100 ? The on-line help indicates that all column names need to be in square brackets.

rcichota:
Thanks @zur003 , adding square brackets does work. However, going without used to work, so now old graphs have to be re-done (maybe, I will check after Dean's fix on XY graphs).
An extension of this issue, maybe, is that if I add an extra Factor (say a parameter I am optimizing), then the graphs show no observed values, even if I try to plot them separately and without any filter...

rcichota:
Ok, graph are now working (with or without square brackets).
However, I still think that there is something wrong with graphs showing observed values. If I disable some of the Factors or add a new one, no observed values are shown. even if on their own and with no filter. Surely this is not the expected behaviour.?
Being a bit pedantic here, but it was a nice thing that the when using filters, the colour of series remained unaltered (that is they followed the same scheme as the graphs without filter). Now all colours (I presume lines and markers too) are reset.  @hol353 Is this intended or a side effect that can be fixed?
It used to be like:
![image](https://user-images.githubusercontent.com/12175986/57057603-2d5be180-6cfe-11e9-80f1-9580362e2ae8.png)
Now it looks like this:
![image](https://user-images.githubusercontent.com/12175986/57057627-68f6ab80-6cfe-11e9-8f75-66248581bfa0.png)

hol353:
Not intended. Will try and fix tomorrow. Where can I find the graphs of your screen shot?

hol353:
Another Q: How are you disabling factors?

rcichota:
Thanks @hol353 , 
I am using the WhiteClover simulation (on /Tests/UnderReview). In either Experiment there is a Folder with graphs titled "ObservedResults".

[...]

PR #3800, opened 2019-05-02
Fix: Filter on series presenter not working due to duplicate field names

hol353:
Working on #3795


hol353:
@rcichota: Does this patch fix your problems?

PR #3822, opened 2019-05-06
Converted simulation to JSon

rcichota:
Working on #2069 
Re-tracing some early steps of the review, tracking what went awry
Not to be merged


rcichota:
@hol353 
This should resolves #2069 for now. I have addressed the points raised on the model's review so it should be enough to get it through the RP. Can you merge this please. There is only one fairly minor deviation in the stats from what was there before and it should be ok to accept it.
I am not really happy with the performance, especially the N content, but I could not find a better way to do it, not helped by the limited amount of data to test against. There is always future work that can hopefully improve on this.

PR #3826, opened 2019-05-06
Initial work on a sorghum validation set

hol430:
Working on #572


hut104:
note spelling mistake in error message
  throw new Exception("Imvalid demand property");

hol430:
Yeah there's more than a few spelling mistakes to be found around the place (seem to recall a 'wether' which should be 'whether' among others). I didn't want to touch anything I didn't have to in this pull request though.

Maybe we need some more converters like this one:

```csharp
/// <summary> Upgrades to version 6. Make sure all KLModifier, KNO3, KNH4 nodes have value XProperty values. </summary>
private static void UpgradeToVersion6(XmlNode node, string fileName)
{
	foreach (XmlNode n in XmlUtilities.FindAllRecursivelyByType(node, "XProperty"))
	{
		if (n.InnerText == "[Root].RootLengthDensity" ||
			n.InnerText == "[Root].RootLengthDenisty" ||
			n.InnerText == "[Root].LengthDenisty")
			n.InnerText = "[Root].LengthDensity";
	}
}
```

zur003:
In the Apsim context, "wether" (or maybe "weather") seems every bit as likely to be the intended word as "whether".

hol430:
`//wether the retranslocation is added or removed is confusing`

PR #3828, opened 2019-05-07
Removed an optional link from generic organ. 

hol430:
Working on #572

### DIFF
--- a/Models/Plant/Organs/GenericOrgan.cs
+++ b/Models/Plant/Organs/GenericOrgan.cs
@@ -60,7 +60,7 @@ namespace Models.PMF.Organs
         /// <summary>The DM retranslocation factor</summary>
         [ChildLinkByName]
         [Units("/d")]
-        private IFunction dmRetranslocationFactor = null;
+        public IFunction DMRetranslocationFactor = null;
 
         /// <summary>The DM reallocation factor</summary>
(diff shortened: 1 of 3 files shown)


Models/Plant/Organs/GenericOrgan.cs
```diff
@@ -60,7 +60,7 @@ public class GenericOrgan : Model, IOrgan, IArbitration, ICustomDocumentation, I
         /// <summary>The DM retranslocation factor</summary>
         [ChildLinkByName]
         [Units("/d")]
-        private IFunction dmRetranslocationFactor = null;
+        public IFunction DMRetranslocationFactor = null;
```
hol353: Why have a public link? Also public fields are never a good thing.

hol430: I made it public because it now needs to be accessible from outside the class. I left it as a field because the ChildLinkByName attribute cannot be applied to properties. There also seem to be a number of other public field links in this class.

hol353: Accessible from outside the class? By another model or by report? If report, then it will see private links.

hol430: It needs to be accessible from another class - `RetranslocateNonStructural`.

PR #3834, opened 2019-05-08
Changing arbitrator to deal with nutrient constrained biomass allocat…

### LINKED ISSUE #3659: Decreasing total biomass (above ground) in the later grain filling period

HamishBrownPFR:
…ion properly

resolves #3659


HamishBrownPFR:
This pull request is a resionably important bug fix as the biomass allocation was not working as we expected in some cases where there is N deficiency.
@hol353 I have reviewed the performance tests.  There is alot of red but most of it is trivial in my interpretation.  Major impacts of this change are an increase in stem wt in Maize and wheat.  However it has had small effects on a number of other models so it would be good to get a few other comments (@byzheng @hut104 @jbrider @hol430 @rcichota

jbrider:
This is probably a good time to reiterate @hut104  comment that it would be good for prototypes to be included in the stats... 

From the first commit it looks like it shouldn't affect sorghum too much - were the issues with sorghum after that first commit just bad timing when merging back to the master branch? I know @hol430 put up a pull request that made significant changes to GenericOrgan and added tests around those changes.

I'm curious why you added a test for Type ="N" in Retranslocation and Allocation - I could only see calls where N is passed in specifically. If it's there to guard against someone adding a call to those functions with DM shouldn't it fail rather than silently doing nothing.

HamishBrownPFR:
@jbrider the first commit failed for sorghum because the the arbitrator was double counting allocations because there were assigned as += rather than just +.  I haven't investigated this further but it means sorghum is allocating DM at another place in the code also.  It the testing I did it was doubling everything so I assume by change it to just set (=) rather than add (+=) it should be correct.  You may want to investigate this further.  
If the Allocation routine is not being called with biomass than we could remove the DM type test.

jbrider:
@HamishBrownPFR I will check the sorghum part out, but don't hold up this commit because of that - the code here shouldn't affect it, and if it does then I need to fix it anyway.  
I missed a call to Retranslocation with DM as a parameter while checking it out before.  So the if == "N" will need to stay - which is not great... I know cause I have used it a few times for the sorghum cod and it hurts every time.  
Inside of GenericOrgan I created an object to handle retranslocation logic similar in behaviour to the IArbitrationMethod classes used to do the actual allocation.  
It might be worth thinking about using more of these type of methods inside the arbitrator itself to encapsulate some of the common logic into external classes - reallocation, retranslocation, maintainence?  I feel like the Arbitrator itself is trying to do way too much at present.  It's probably a larger refactoring though, so I can't commit to it right now, but maybe in August?

HamishBrownPFR:
@jbrider I agree the arbitrator is in need of a jolly good refactoring.  Some day...

rcichota:
I had a look Plantain and WhiteClover, the differences are really small and I  don't see any problem in accepting them.

PR #3845, opened 2019-05-10
Minor fixes to database queries used in graphs.

zur003:
Working on #2821


zur003:
Looks like this isn't going to work the way I thought...

PR #3855, opened 2019-05-16
attempting to fix the canopy base for trees, which does not have a ba…

junqi108:
working on #3856 
attempting to fix the canopy base for trees, which does not have a base of 0

### DIFF
--- a/Models/MicroClimate/ZoneMicroClimate.cs
+++ b/Models/MicroClimate/ZoneMicroClimate.cs
@@ -80,16 +80,13 @@ namespace Models
                     double HeightMetres = Math.Round(Canopies[compNo].Canopy.Height, 5) / 1000.0; // Round off a bit and convert mm to m } }
                     double DepthMetres = Math.Round(Canopies[compNo].Canopy.Depth, 5) / 1000.0; // Round off a bit and convert mm to m } }
                     double canopyBase = HeightMetres - DepthMetres;
-                    if (Array.IndexOf(nodes, HeightMetres) == -1)
+                    if (Array.IndexOf(nodes, HeightMetres) == -1 | Array.IndexOf(nodes, canopyBase) == -1)
                     {
-                        nodes[numNodes] = HeightMetres;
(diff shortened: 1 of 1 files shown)


Models/MicroClimate/ZoneMicroClimate.cs
```diff
@@ -80,16 +80,13 @@ private void DefineLayers()
                     double HeightMetres = Math.Round(Canopies[compNo].Canopy.Height, 5) / 1000.0; // Round off a bit and convert mm to m } }
                     double DepthMetres = Math.Round(Canopies[compNo].Canopy.Depth, 5) / 1000.0; // Round off a bit and convert mm to m } }
                     double canopyBase = HeightMetres - DepthMetres;
-                    if (Array.IndexOf(nodes, HeightMetres) == -1)
+                    if (Array.IndexOf(nodes, HeightMetres) == -1 | Array.IndexOf(nodes, canopyBase) == -1)
```
hol430: This should be logical OR rather than the bitwise OR (note the second '|' character)

```suggestion
                    if (Array.IndexOf(nodes, HeightMetres) == -1 || Array.IndexOf(nodes, canopyBase) == -1)
```

junqi108:
Do not need to merge it now. The microclimate code cannot deal with two layers for one canopy, so i give up in changing this.  
 if (tallest.Canopies.Count>1)
                throw (new Exception("Strip crop light interception model must only have one canopy in zone called "+tallest.zone.Name));
            if (shortest.Canopies.Count > 1)
                throw (new Exception("Strip crop light interception model must only have one canopy in zone called " + shortest.zone.Name));
            if (tallest.DeltaZ.Length > 1)
                throw (new Exception("Strip crop light interception model must only have one canopy layer in zone called " + tallest.zone.Name));
            if (shortest.DeltaZ.Length > 1)
                throw (new Exception("Strip crop light interception model must only have one canopy layer in zone called " + shortest.zone.Name));

PR #3866, opened 2019-05-19
Fix error bars on graphs

### LINKED ISSUE #3865: Error bars on graphs not working
When graphing observed data that has error columns, an exception will be thrown: "collection was modified, enumeration operation may not execute".

hol353:
Resolves #3865

### DIFF
--- a/Models/Graph/SeriesDefinition.cs
+++ b/Models/Graph/SeriesDefinition.cs
@@ -198,7 +198,7 @@
         public IEnumerable<string> SimulationNamesForEachPoint { get; private set; }
 
         /// <summary>Gets the error values</summary>
-        public IEnumerable Error { get; private set; }
+        public IEnumerable<double> Error { get; private set; }
 
         /// <summary>Add a clause to the filter.</summary>
         /// <param name="filter">The filter to add to.</param>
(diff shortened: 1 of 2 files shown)


JJguri:
@hol353 could you please specify which module and where I need to use to see the error bars in a graph? How do these values need to be added to the obs file?

hol430:
@JJguri If the y variable on your graph is called "Yield" then apsim will look for a column called "YieldError" and show the data in that column as error bars (if such a column exists). So you just need to rename your column containing the error data to the same name as the observed column with the word Error at the end.

You also need to make sure your marker type is set to one of the "Filled" marker types - e.g. FilledCircle. This probably shouldn't be necessary and might be a bug so I'll look into it.

As a (silly) example:

image

image

JJguri:
Thanks @hol430 

I got the bars for my time-series graphs. Now I would like to put it in the x-axis of an XY graph. Please see example below. The difference here is that the data source is a TimeSeriesData table created to match pre and obs. How should I include the AboveGroundWtError in that table?
Untitled

hol430:
Looking at it a bit closer, I think we currently only have support for error data on the y series. I'll fix this as part of #5341

JJguri:
> Looking at it a bit closer, I think we currently only have support for error data on the y series. I'll fix this as part of #5341

Please keep me posted about the development of this functionality. I am exporting outputs and plotting these externally because I cannot add error bars to the p-o graphs in the UI.

PR #3868, opened 2019-05-20
Cover code analaysis tool

### LINKED ISSUE #3867: Add code coverage tool.
CodeCover + ReportGenerator (both open source) will bring code coverage analysis to APSIM.

hol353:
Resolves #3867


hol353:
Note that I can't get ReportGenerator from NuGet for .NET 4.6. It only seems to support .NET 4.7.

hol430:
We don't normally track files under ApsimX/packages/ do we? These files will be downloaded automatically whenever the developer does a nuget restore.

hol353:
We don't normally but I can't get the package from NuGet because we target .NET 4.6.

jbrider:
Is there a reason to stay on 4.6?
For that matter, how much is stopping us from moving to dotnet core?
The gui is cross platform compatible - would be nice  if the code was too.

hol353:
No real reason to stay on 4.6. Rather than move to 4.7, I'd like to change to dotnet core. Need to make sure all the nuget packages we use and the GTK stuff works on core.

hol430:
I've never tried to compile/run with dotnet core. Would be very nice, as mono can be unreliable at times. That said, if it turns out to be a lot of work to get apsim running under dotnet core then it may not be worth it, as .NET 5 is supposed to be released November 2020 anyway.

jbrider:
I remember there being a tool you could run to tell you what wasn't
supported by dotnet core... think it was a MS release so shouldn't be hard
to track down.

I've been using dotnet core 2 on linux server back ends for 18+ months -
it's worked well.
It will be the imported libraries that will be the issue not the C# code.

zur003:
Moving to .Net core might be feasible, but would probably take quite a bit of work. There appears to be a version of Gtk# for it now (see https://github.com/GtkSharp/GtkSharp), but it's Gtk 3.22 so the Gtk stuff would need porting from version 2 to 3. (This could be rather interesting, if it works well. Windows.Forms will be added to Core 3 as "Windows platforms only", leaving Gtk# or QML for cross-platform GUI development.) 
There is now also a version of PDFSharp (https://github.com/ststeiger/PdfSharpNetStandard). 
The fact that we're already cross-platform should make moving to dotnet core easier (since we're already avoiding Windows-specific P/Invoke calls), but I doubt if it would be completely smooth sailing.

jbrider:
I was looking for the compatibility tool and realised just how many nuget
packages we are using... definitely not an overnight job.

PR #3883, opened 2019-05-23
Font selection

### LINKED ISSUE #3876: Font selection
Would be handy if we could choose a font face, and bonus points if there's an easy way to change the font size.

hol430:
Resolves #3876 

As part of this, I've moved the theme and font selection into a settings drop-down on the home screen. The drop-down looks somewhat odd on mouseover for certain gtk themes on Linux, because I suspect I've used the wrong widget for the job here. Looks ok on Windows though - will revisit this when I get the chance.

I've also removed the ctrl + + and ctrl + scroll shortcuts for font size selection, as font size can be changed from a dialog now.


hol430:
Don't merge this just yet

PR #3904, opened 2019-05-28
Sorghum stress

jbrider:
Working on #572

Fixing root senescence.


hol430:
@jbrider I've just merged the grain heat stress function into this branch. It seems to fix the negative DM demand from grain.

Models/Plant/Organs/Root.cs
```diff
@@ -1311,9 +1311,49 @@ private void OnDoActualPlantGrowth(object sender, EventArgs e)
 
 
                     Biomass Loss = Live * senescedFrac;
-                    Live.Subtract(Loss);
-                    Dead.Add(Loss);
+                    //Live.Subtract(Loss);
+                    //Dead.Add(Loss);
                     Senesced.Add(Loss);
```
hol353: Won't this result in the senesced biomass not being moved from live to dead?

hol430: Looks like `Live` and `Dead` are just wrappers around the private `liveBiomass` and `deadBiomass` fields. Setting `needToRecalculateLiveDead` to true causes these two fields to be recalculated the next time we access them via `Live` or `Dead`. The calculation is based on the contents of `PlantZone`, which we modify here. Seems a bit convoluted to me but it appears to work. I'm not sure on why we need this change or what it achieves but I think it's ok.

Issue #3835, opened 2019-05-08
Add multi-species simulations


rcichota:
As part of developments of forage models, there is a need to test whether Apsim can reproduce situation with various plant species in the same field. I will be adding a couple of examples with comparison against experimental data and sensitivity analyses. This will not be a comprehensive test, but a start to study how the model is handling this very important issue.

PR #3919, opened 2019-06-04
Feature/multi species pasture#3835

rcichota:
Working on #3835 
Updated the original simulation and added new data and simulations for experiments from Denmark


hol430:
retest this please jenkins

Issue #3830, opened 2019-05-07
Allow methods with arguments to be used in reports


lie112:
I would like to be able to use a method with arguments (only one string to start with) in the report variables. At present I believe there is limited code to access to methods, but not including arguments. I don't want to create public properties as I don't know the full suite that the users will require in the dynamic model building environment. 
Is this something others would use and something that could easily be added? I can have a look at adding this functionality as it will really assist some reporting I am needing to do in CLEM if no one else can do it.

hol430:
I haven't really looked into it, but I imagine this should be possible. Seems like the simplest approach might be to add functionality to the Locator, allowing it to properly detect methods. I think at the moment it will look for the .Value() method of an IFunction, but no other methods.

lie112:
I have implemented this functionality. Reports will now accept a method with any number of arguments of the types string, int32 and double (more can easily be added to a switch statement). I was able to add a small section of code to the Locator and add an arguments array to the VariableMethod to achieve this.

I have a couple questions/problems

1. What is the difference between Locater.cs and Locator.cs? I have added the code to Loactor.cs as this is the class operating in reports, but Locater is used extensively throughout code and looks very similar.

2. I need a bit of smart code to determine when an equation has been entered or whether it is a method with arguments inside brackets. I noted a comment in the code (Locator.GetInternal) "// expression - need a better way of detecting an expression". I have removed the "(" from the equation identifying search characters to allow the code to move on to identifying properties, models and methods. I figured an equation would have a +/* (note there is no -, I assume because people use - in model names and it will break too many runs. This may be why my minus '-' in an equation never worked.). I still need to add some code that removes all arguments from the string before breaking the string into components based on split('.") for namePathBits in Locator.GetInternal as the user may have added a "." to a double value in the arguments which is mistakenly taken as part of the model name structure.

I am not really familiar with equations people use. Is there another clever means of detecting an equation. I was thinking of looking between any  "(" and ")" and if there was no +/*- char it would be considered arguments and removed from the determination of namePathBits and replaced to determine method arguments.

At present using methods in reports works as long as the user doesn't add a "." in a double value in the arguments. I have not pulled this code until I get the equation identifier and passing doubles (including a ".") as arguments working.

hol430:
[...]

Issue #3678, opened 2019-04-01
Race condition in Morris/R component


hol430:
This popped up in one of the recent jenkins builds:

```
Commencing simulations...
ERRORS FOUND!!
Error in file: C:\Jenkins\workspace\PullRequest_v2\ApsimX\Examples\Tutorials\Sensitivity_MorrisMethod.apsimx /RunTests
RScript executable: C:\Program Files\R\R-3.5.1\bin\x64\Rscript.exe
[1] "Installing package sensitivity to location C:/Users/scoop/AppData/Local/Temp/dabedbf2-e5c3-4af7-80cf-3492a466712e"
package 'sensitivity' successfully unpacked and MD5 sums checked
The downloaded binary packages are in
	C:\Users\scoop\AppData\Local\Temp\RtmpS6FhwW\downloaded_packages

Sensitivity_MorrisMethod, Simulation MorrisMethodExampleSimulation1 has commenced.
...
Sensitivity_MorrisMethod, Simulation MorrisMethodExampleSimulation96 complete. Time: 20.09 sec
RScript executable: C:\Program Files\R\R-3.5.1\bin\x64\Rscript.exe
[1] "Installing package sensitivity to location C:/Users/scoop/AppData/Local/Temp/c3d64511-1963-46ce-9ccd-fab04a1d3c17"
package 'sensitivity' successfully unpacked and MD5 sums checked
The downloaded binary packages are in
	C:\Users\scoop\AppData\Local\Temp\RtmpEvqdkG\downloaded_packages

ERRORS FOUND!!
System.Exception: Error from R: 
Warning message:
package 'sensitivity' was built under R version 3.5.3 
Error in `$<-.data.frame`(`*tmp*`, "path", value = 1:25) : 
  replacement has 25 rows, data has 24
Calls: $<- -> $<-.data.frame
Execution halted

Script:
library('sensitivity')
params <- c("SOC","Load","DUL")
apsimMorris<-morris(model=NULL
 ,params #string vector of parameter names
 ,25 #no of paths within the total parameter space
 ,design=list(type="oat",levels=11,grid.jump=5)
 ,binf=c(1,100,0.2) #min for each parameter
 ,bsup=c(10,1000,0.4) #max for each parameter
 ,scale=T
 )
apsimMorris$X <- read.csv("C:/Users/scoop/AppData/Local/Temp/parameters9e683d36-ec74-4c73-b526-54055a407ded.csv")
[...]

PR #3926, opened 2019-06-05
Added ability to use Method() with constant args in Report. 

lie112:
Working on #3830
This works with any number of string, int32 and double* args (*without including a decimal place value).


lie112:
retest this please jenkins

lie112:
Found a small issue with this functionality. This submitted version only works for Methods directly related to a specified model in the variable string. I needed to move the code in the if statements of Locator.GetInternal so that it would accept a Method(arg) of a property returning a IModel.
This change was made in another branch I have been working on that I will have to submit separately after this pull is accepted. 
The code submitted here should still work and will not influence others work.

I am not sure why Jenkins is hanging while performing tests. I don't get errors or anything to address.

hol430:
There is a race condition somewhere in apsim (#3678) which occasionally manifests when running the unit tests on Jenkins. I think @hol353 is working on refactoring the way simulations are run, which I'm hoping will fix this issue. For now I'm just killing any stalled builds manually.

lie112:
I have encountered some issues while trying to test my new report allow method() code with some unforeseen entry values and a level of complexity I did not foresee.

(a) I don't understand the setup of the CanopyPhotoSynthesis .apsimx in Prototypes/LeafApex. It seems that this simulation runs 23 simulations and there appears to be threading taking place so debugging the Locator is a nightmare. I am trying to simplify it to one simulation with a single HarvestReport that I can check to get the following offending line fixed.

(b) What does the following entry in CanopyPhotoSynthesis .apsimx .Replacements.Wheat.HarvestReports mean:
**[Wheat].Structure.HaunStage.Value()TerminalSpikelet.Value()**
I don't understand this format. Should there be a period after the first Value()?
Is the TerminalSpikelet a property of the HaunStage.Value()?
Is it that the old code accepted this format by mistake?

I think if Value is picked up as a Method and there is a "." before TerminalSpikelet my new code should accept this, but this forces a lot of apsimx files to be updated to the new . format, or some code needs to add a . after Value() which is easy as I can replace all "Value()" with "Value()." and ignore empties in the split statement.

lie112:
It was also necessary to shift the MethodInfo calculation in the if statements of GetInternal as the initial version only worked for Methods of an identified IModel and not Methods of a model returned by a Property as could happen in the variable string. This is fixed, but I need to address the previous issue before any pull will be successful.

hol353:
1. APSIM Next Gen has always been multi-threaded. APSIM detects how many cores and computer has and will run simulations asynchronously. From the GUI you can right click on a simulation and select Run. You can also select an experiment, right click on the a simulation in the right hand view and select Run. Both of these will only run a single simulation. Easier debugging.
2. [Wheat].Structure.HaunStage.Value()TerminalSpikelet.Value() is an error. It is not valid syntax for a report variable. I'm guessing that it would have produced a blank column in the report table. It seems to also be in Validation\Wheat.apsimx.

hol430:
It would be very useful for debugging if we had an 'f4 - run single threaded' option in the GUI.

lie112:
Thanks. 
My question is how do I select a single simulation to run in a simulation with replacements and factors or is that not possible as I am inclined to believe? I tried disabling all but one of each factor but that gave me an error that it could not find a waterNode.Thickness in Soil.
Can I fix the offending lines in the apsimx files with reports that cause all the errors preventing my pull request?
What is the correct line?
I am hoping I haven't caused .Value() to stop working, but I will make sure a simulation with .Value() works before doing any more commits. I wasn't really sure what .Value() provided.
I believe I can remove the search for .Value as it will now be picked up as a method.

hol430:
Re running a single sim - I think you should be able to click on the experiment, then you should see a list of simulations in the right hand view. Just right click on the one you want to run and click run from the context menu. This used to be pretty buggy - can't remember if I ever got around to fixing it.

lie112:
Is there a way I can correct the invalid entries ([Wheat].Structure.HaunStage.Value()TerminalSpikelet.Value()) so I can get a successful pull in Jenkins? (a) just delete the TerminalSpikelet part? Will this influence data checking? I also need these entries fixed so I can test that Values() works with the new Methods code. I need to get this pull working so I can move to pulling other CLEM changes and new functionality.

hol430:
I don't know what this code is trying to accomplish, but I would suggest option a - just delete the terminal spikelet part. If it breaks something, Jenkins or the performance tests should tell us.

Is this line of code in a report? If so, I would imagine that it's just reporting 0 anyway - it's definitely not legitimate syntax.

lie112:
Is Grapevine.apsimx a very slow simulation? I have been running it for 15 minutes and only at 1/3 of 0 of 16 simulations. I need to get to the lines producing the error to work out what it wrong. This one doesn't appear to be a report variable problem and the error does not happen in all simulations, but is related to Locator.cs. This explains why Jenkins takes so long to do prototype tests.

hol430:
No, I was able to run the grapevine prototype in about a minute, so something has gone wrong for you. Maybe it's a race condition or some other temporary bug - in this case, you could try stopping the sims, deleting the .db file and restarting the sims. Let me know if that doesn't work, and I'll pull down your branch and see if I can run it on my pc.

Also, there are 21 simulations in my grapevine.apsimx - you may need to pull down latest changes.

lie112:
Just the Optimisation of ChillA (single run) takes 183 sec on my computer with a delete of db file.
I have pushed all current needed changes to my WorkingCLEM branch that is being pulled. 
I am trying to work out why the grapevine prototype is breaking and preventing my pull request.

lie112:
I seem to have opened a can of worms with multiple prototype runs now reporting errors:
_Variable is not of a reportable type. Perhaps  it is a PMF Function that needs a .Value appended to the name. ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.Exception: Cannot report variable [Teff].Structure.MainStemFinalNodeNo. Variable is not of a reportable type. Perhaps  it is a PMF Function that needs a .Value appended to the name._
Maybe these were previously ignored and I have added a bit more error trapping by allowing methods to be included. I need to be able to get one of these simulations running locally to debug this problem, but they all seems to be multi factor runs and I haven't been able to reduce them to a single run with the offending setup.

hol430:
Are you able to click on an experiment node? I get an error whenever I click on an experiment.

hol430:
I see there are some errors coming from the teff prototype. I opened this file, copied the simulation out from under one of the experiments, and deleted everything else. When I run this simulation on your branch I get a serialization exception (which I suspects hints at another problem) but if I skip past that I get the 'variable is not of a reportable type' error which you mentioned. I've run out of time to look into it today, but here's the file if you're interested:

[teff.zip](https://github.com/APSIMInitiative/ApsimX/files/3274959/teff.zip)

lie112:
Thanks. I'm working on the the issue. I had a !=null if statement around the wrong way, but fixing it now causes some of my reporting to fail (blank value). The model now will go back to the previous means of not reporting any issue with invalid fields. This has highlighted the number of invalid entries throughout the prototype reports. Adding any further error checking will require all prototypes and people other apsimx reports to be fixed, which is not necessarily a bad thing.

lie112:
My current problem is in eaxmples/oilpalm where the following report variable is supplied
_sum of sum([Soil].SoilNitrogen.Denitrification) from 1-Jan to 31-Dec as AnnualDenitrification &
sum of sum([Soil].SoilNitrogen.n2o_atm) from 1-Jan to 31-Dec as AnnualN2O_
I am not sure how this is handled in APSIMx. It does get applied to an aggregation column, but when the term is broken up the variable name is  sum([Soil].SoilNitrogen.Denitrification) which does not meet the criteria for an expression or the period delimited model and property/method format. This then returns a list of null that throw an error when we attempt to sum them as int or doubles. 
Is this an error in reporting or can you have a sum of sum()?

hol430:
Pretty sure this is legitimate syntax. Soil.SoilNitrogen.Denitrification is an array of doubles, so sum([Soil].SoilNitrogen.n2o_atm) is just a number - it's the sum of all elements in the array. So we're summing this number from 1-Jan to 31-Dec, which is legal.

lie112:
Yes. Been thinking about it. It is because I took the "(" out of characters to identify an expression, but this one is an expression without a +* or /. If the ( is in expression search criteria, then we can't pick up methods with arguments as opposed to expressions. Is there a specific list of words that can precede a ( in an expression. We could search for them e.g. "sum(" "average("......

lie112:
Does every report variable with a property of a model have to start with [model] in square brackets before further period delimited models, functions or methods? If this is the case I could check whether there is a [, ] or . before the fist (, otherwise treat it as an equation.

hol430:
There are a list of built-in function names (sum, avg, exp, sin, etc) hard coded into the expression evaluator and yes you could check them but hard-coding them in two different places would be a maintenance headache. If you wanted to extract them from the expression evaluator somehow then that might work - perhaps make an interface for built in methods? Maybe more work than it's worth.

Alternatively, if the expression contains a parenthesis, you attempt to resolve it as a method of an apsim object/model. If the locator returns null (as would happen in the case of sum() or avg()) you could assume it's a built in method and just let the expression evaluator handle it. Haven't looked at how you've implemented all this so maybe that would be difficult, I'm not sure.

Issue #3801, opened 2019-05-02
Simple Fruit Tree model


HamishBrownPFR:
Develop a simple fruit tree model that is parameterised for multiple crop types (Apple, Kiwifruit, Grapes, Peaches etc) to do water and N balance of orchards.

sarahcleary:
**ACTION from RP Meeting** - @Keith-Pembleton to link Hamish with a local consultant who developed a banana model which may have relevance to the Simple Fruit Tree Model.  @HamishBrownPFR might be able to help include banana model into NextGen.

sarahcleary:
@HamishBrownPFR - is this "Ready for Review"?

HamishBrownPFR:
superseded by #8175

PR #3927, opened 2019-06-05
Fruit trees

HamishBrownPFR:
working on #3801 

@hut104 can you please review what I have done here.  I have modified the strip crop radiation model to:
1. Allow canopies to have two layers so the taller canopy can all be above the shorter canopy as we see in orchards
2. Allow the canopy of the taller crop to overlap the shorter crop zone so it will capture more radiation from the short zone.
Let me know if you see any issues with what I have done.


HamishBrownPFR:
@junqi108  if you are interested in the changes I have made to the row crop radiation capture model, feel free to review and make comments

HamishBrownPFR:
@hol353 I think this one is good to merge now.  I have been discussing with @hut104  and we decided to implement tree row radiation interception code sepperte to strip crop for a start to be sure we have them right.  @junqi108 you may be interested to see how I have changed the strip crop model to work for row crops with alley overlaps.

Issue #3934, opened 2019-06-12
HTML presenter slight font change when Run performed.


lie112:
I have noticed that there is a slight change in the font displaying the Simulation details in the HTML presenter view when you click run. There is obviously a value changed when run is selected. I have noticed this happens on other HTML pages as well. The look it changes to is best.

hol430:
The html view also seems to use the wrong font for me.

hol430:
@lie112 I'm having trouble reproducing your problem. For me, apsim uses the wrong font in the html views and although running the simulations causes the right-hand view does flicker, the font doesn't seem to change for me. #3942 should get apsim to use the correct font. When it is merged, can you tell me if it has an impact on your problem?

lie112:
I don't understand the implications of the changes made to HTMLView (changing browser font allocation) but this is clearly the reason for my observations.

My version now repeats your observation. I get a flicker when Run is selected, but the font remains the same size while previously it increased slightly.

The problem is, I now only get the default Times Roman font for all my HTML views rather than the clean sans font previously used.

Exploring the changes made to HTMLView.cs the new font allocation provides "Sans 10" as the value as this is what is provided in (MasterView as ViewBase).MainWidget.Style.FontDescription while the old line now deleted provided "Segeo UI 9" as that was the value from this.MainWidget.Style.FontDescription . I would suggest that "Sans 10" does not exists and my browser is using the default Times New Roman font. The font of the MasterView probably needs to be set, or I have set a font somewhere that was overriding your functionality and is now ignored.

Also the default white background I have built all my HTML summaries on has changed to the default window background (type of grey) and doesn't look as good for things like the summaryfile view, but this is likely the theme background. I also occasionally get an "Object not found" error while trying to set the BackgroundColour in HTMLView.cs when I click on a model in the tree that uses a HTML view. Is this all related to applying a dark theme?

I have not yet implemented any dark theme for CLEM components.
I can add a background-color setting to my html pages.

lie112:
Once I changed the font to Segoe UI in the change font section, everything worked great. Is it possible to set something other than Sans as the default? The HTMLView does not like Sans and reverts to default TimeNewRoman which looks awful.  When I change the system font to Segoe UI all works and it looks very professional through all my summary screens.
I suggest the default font should work out of the box.

hol430:
[...]

PR #3942, opened 2019-06-13
Fixed browser font on Windows

hol430:
Working on #3934


lie112:
I have found that the default times new roman font is still being used in HTML summaries. I can't remember if this was fixed and reverted or hasn't changed. If I set the font in the settings dialog all the html summary fonts are upgraded, but this is a per ASPIMx file solution. It would be nice to have a Segoe UI as a default font for users that have not changed their APSIM font as it makes more professional looking HTML displays.

PR #3948, opened 2019-06-17
Make Factor and CompositFactor valid parents for Manager and Opperations

### LINKED ISSUE #3947: Allow Manager and Opperations drops on factor and composite factor
Add as valid parents

HamishBrownPFR:
Resolves #3947 

Not sure this is the best way to resolve this problem.  It resolves the specific issue and > 90% of experiment set ups are overwriting a manager or operations model so this should facilitate setting up of experiments from the GUI in most cases.  However there are cases where a soil, or clock or water or weather or something else is overwritten in experiment structures.  Should we put valid parent tags in all the models that are occasionally dropped on factors as well?  or can we change factor and composite factor so they will accept all drops?


hol353:
Yes we should change factor and composite factor to accept all drops but for now I'll merge this to keep you going.

PR #3957, opened 2019-06-20
When upgrading, update rego DB before downloading installer.

### LINKED ISSUE #3956: Upgrade form doesn't check if mandatory fields are filled
The upgrade form currently does not check if mandatory fields (name, email, etc.) are filled in before starting the download. This leads to an error being thrown when we try to write to the registrations DB after the download finishes, which results in the upgrade being aborted anyway.

hol430:
Also retry updating the registrations DB if it fails the first time.

Resolves #3956


hol430:
retest this please jenkins

PR #3960, opened 2019-06-21
Intellisense bugfixes and one-time speedup

### LINKED ISSUE #3819: Intellisense in factor view
Intellisense does not seem to work/appear when typing into the factor view.

hol430:
Resolves #3819 

I've also fixed the issue where some views (those which use intellisense) take a long time to load the first time you click on them. This is due to a one-time initialisation process which takes a long time. This initialisation now gets run in a background thread when we first start the program.


hol430:
retest this please jenkins

PR #3961, opened 2019-06-21
Added some null checks to HTMLView.

### LINKED ISSUE #3934: HTML presenter slight font change when Run performed.

hol430:
Resolves #3934


hol430:
retest this please jenkins

PR #3968, opened 2019-06-25
Remove citation from home scren.

### LINKED ISSUE #3962: Remove Citation text on home screen
A text box appears on the opening screen of the APSIM UI to tell users to cite the APSIM paper.  This goes against modern software interface trends of simplicity.  I'm also not sure that this is required for every time a users opens APSIM, and most users of APSIM NG would probably be aware of the paper in any case.  I think the bigger issue of old citations is with users of APSIM classic.
I suggest removing the text, and if really required, place it into an about box or somewhere less obtrusive.

hol430:
resolves #3962


hol430:
retest this please jenkins

hol353:
The text can't be removed. Has to be put somewhere. Turn help button into dropdown, create a help about menu in the drop down and show a dialog box with this text?

lie112:
Maybe it could go in the warning/error zone at the bottom of the screen so it is there every time someone starts APSIM, but will then be replaced by simulation status when you run a simulation, even if this box was made a little bigger. It could be a more concise block of text with a link to the website where the full acknowledgement is provided. Otherwise, an "About" popup off the menu with copyright, contact and web links would be the obvious place for this. This is more an education campaign about correct citation and acknowledgement by users publishing with APSIM Next Gen than a reason to use a portion of the simulation real estate continuously. My 2c worth.

PR #3969, opened 2019-06-25
Mru list size on home screen should never be 0.

### LINKED ISSUE #3966: Only most recent simulation icon shown on apsimx GUI
We used to have many previous simulation icons shown on the GUI that could be user-deleted if required, but now only the latest is shown. Can this be reset to the previous view, please.

Regards,
Philip

hol430:
Resolves #3966


hol430:
retest this please jenkins

Issue #3843, opened 2019-05-09
Area series doesn't show


her123:
The scatter series shows and the bar series also. When changing to an area series there is nothing shown. Am I missing something?
![Screen Shot 2019-05-10 at 9 04 34 am](https://user-images.githubusercontent.com/11879671/57492090-d5773900-7302-11e9-99f7-7899f41fb6e8.png)

hol430:
I think you need to specify x2/y2 coordinates. Drop-down boxes for these fields don't seem to appear though until you click away to another node and then return to the series again.

her123:
It may be good to call these series something else like Region series and then implement a true area series.

hol430:
What exactly do you mean by a true area series? Something like this?

![image](https://user-images.githubusercontent.com/36427516/57674987-0ab5bb00-7664-11e9-889f-3202c0b7b1d3.png)

her123:
Yes

her123:
Stacked area could be good also

her123:
Just checking that the Include series name in legend works. It doesn't appear to do anything.
![Screen Shot 2019-05-15 at 9 32 34 am](https://user-images.githubusercontent.com/11879671/57738847-85c6b200-76f4-11e9-9eb7-1a45e6fd0df2.png)

hol430:
The "Include series name in legend" will do just that - the name of the series in the tree (asdf in my screenshot) will be appended to the name of each series in the legend of the graph:

![image](https://user-images.githubusercontent.com/36427516/58390543-139b8900-8075-11e9-9c38-634e0d6e5bbb.png)

her123:
The area series looks good @hol430 .
What make areas series really useful is having the option for stacked area to show total volume of pasture that is made up of living and dead.

zur003:
I agree that a stacked area option would be good to have.

Area charts do not currently work if either of the axes is DateTime rather than double - this will raise an exception with the message "Specified cast is not valid".

PR #3975, opened 2019-06-26
Changed approach to casting of axis values of area charts.

zur003:
Working on #3843


zur003:
It looks like this will need to be re-run after APSIM.Shared is fixed. Builds appear to be broken at the moment.

hol430:
retest this please jenkins

PR #3952, opened 2019-06-19
Run refactor

hol353:
Working on #3951

PR #3982, opened 2019-06-29
Check position of SoilArbitrator

### LINKED ISSUE #3981: Add check and limit drops so soil arbitrator can only be directly under Simulation

HamishBrownPFR:
Remove zone as valid parent and add test to throw if arbitrator is anywhere other than under simulation.  Resolves #3981


hol353:
retest this please jenkins

HamishBrownPFR:
@hol353  @hol430 can anyone shed some light on this error message?

hol430:
Yeah that's a race condition which shows up occasionally on Jenkins. Hopefully it's fixed by #3952. The solution (or workaround) is to just re-run the tests which it looks like @hol353 has already done.

Issue #1721, opened 2017-06-12
Lucerne Model for APSIM Next Gen


HamishBrownPFR:
Various attempts to tidy up lucerne models in the past have not been completed.  A Ph.D. student working with Derrick Moot, Edmar Teixiera and Hamish Brown is focusing on lucerne modelling in APSIM (no experimental work so won't get side tracked).  More details to follow.

sarahcleary:
@HamishBrownPFR - as per the meeting today.  Is this model ready for review?

sarahcleary:
@HamishBrownPFR - is this model ready for review?

HamishBrownPFR:
Not yet

sno036:
A quick summary of my comments:
- a plot of the cumulative harvest across all the experiments so that users get a sense of how the model is performing on that basis. Maybe colour by country? At least annual values but consider seasonal and total.
- A test that livestock are removing biomass in the simulation 
- A great example simulation for users - with some decent management settings and outputting sensible/useable/popular outputs. I'd like to see an 'straight' lucerne simulation but also one in a rotation (with grazing even better!)
- The sensibility analysis was quite weak (as with many of the models) and could/should be improved. A good use of sensibility testing for a crop would be along a climate transect testing growth (or not) into more challenging climates

Well done! Major effort!

hut104:
1. Many of the soil water graphs are not quite correct with either inconsistent units between Obs and Pre values, or errors in APSIM output, or Obs and Pre data referring to different depths (not sure which).  These need to be corrected and ensure that the stats within Jenkins are providing an accurate estimate of model fit.
2. The Hudson simulations could potentially  be moved into the "systems" tests now.  We have wheat, chickpea, sorghum, mungbean and lucerne models now and so we we could simulate all rotations.
3. All observed data should use APSIM naming convention and units.  This would make it easier for the reader (or someone looking at  Jenkins stats) to know what they are evaluating.
4. Many of the Predicted values compared to Obs values are live fractions only.  I'd imagine that many measures (e.g. leaf and stem) include dead material as well.
5. The canopy simulations should be looked at closely.  The model description says that LAIdead = 0, DetachmentRate = 0, Kdead is zero, and yet senescence has positive values.  This means that lots of plant material could be not picked up in the stats, and senesced material may be staying on the plant but not showing up in LAI etc.  
[...]

PR #3985, opened 2019-07-01
Lucerne 1721

Xiumei2017:
Working on #1721


HamishBrownPFR:
@hol353 this is failing on red clover because there are no baseline stats committed for Nconc but it is producing number now.  At least that is what I think is happening.

rcichota:
There was an issue with Nconc, but that was (or at should have been) fixed in the merged commit. The issue was that there were two variables re- Nconc and a few missing values, I deleted one and fixed the other, and so there should be values in there now...

image

hol353:
@HamishBrownPFR I assume you're Ok with the changes to Plant in this PR. I notice a PopulationFactor has been added to the Plant class. What is that?

HamishBrownPFR:
Hi @hol353.  Yep I made those changes but pushed them up through Xuimei's repo so we could test things on her system first.  Population factor has a default of 1 so doesn't do anything in most cases.  For lucerne it reduces LAI and biomass demands when the plant density is low and individual plants can't grown any bigger to fill in gaps in the field.

hol353:
retest this please jenkins

hol353:
retest this please jenkins

PR #3986, opened 2019-07-01
Fruit trees

HamishBrownPFR:
working on #3801


Models/Functions/SupplyFunctions/RUEModel.cs
```diff
@@ -92,7 +92,7 @@ public double Value(int arrayIndex = -1)
             double radiationInterception = RadnInt.Value(arrayIndex);
             if (Double.IsNaN(radiationInterception))
                 throw new Exception("NaN Radiation interception value supplied to RUE model");
-            if (radiationInterception < 0)
+            if (radiationInterception < -0.000000000001)
```
hol353: Can you call MathUtilities.IsNegative(radiationInterception) instead? The tolerance is a lot less (0.00001) than the constant you have in the if statement which might be a problem.

Issue #2694, opened 2018-06-13
Develop a Red Clover model (PMF)


rcichota:
Red clover (Trifolium pratense) is a large-leaved herbaceous perennial plant of the Fabaceae family. It is native of native to Europe, Western Asia, and northwest Africa; it can now be found in many parts of the world. Red clover is widely grown as a fodder crop or green manure, although it can be used for herbal tea and also grows wild in swards in sub-tropical and temperature climates.  It produced high quality feed for animals and can also fix N through symbiotic relation with N-fixing bacteria.  The plants vary in height from 20 to 80cm with semi-erect stems, the leaves are alternate trifoliate and the petioles are 1-4cm long.  The flowers heads are dark pink or red and are pollinated especially by bumble bees.  Red clover has deep taproot which makes it tolerant to drought, it is also more tolerant to root weevil than white clover. 
The objective here is to develop a basic red clover model using the plant modelling framework (PMF). In this project the focus in on describing the general growth pattern of monoculture, as well as the re-growth after cut/grazing and including N balance. As PMF uses MicroClimate and SoilArbitrator to manage resource access, it should be readily possible to develop simulations with multiple species. While developing the model, the need for it to be used in mixed swards simulations will be taken in consideration as much as possible, however this is not a priority initially.

JJguri:
I cannot create the documentation file for the entire simulation. The following error appeared:

System.IndexOutOfRangeException: There is no row at position 88.
   at System.Data.RBTree`1.GetNodeByIndex(Int32 userIndex)
   at Models.Tests.Test(Boolean accept, Boolean GUIrun)
   at Models.Tests.Document(List`1 tags, Int32 headingLevel, Int32 indent)
   at Models.Core.AutoDocumentation.ParseTextForTags(String stringToParse, IModel model, List`1 tags, Int32 headingLevel, Int32 indent, Boolean documentAllChildren, Boolean doNotTrim)
   at Models.Core.AutoDocumentation.DocumentModelSummary(IModel model, List`1 tags, Int32 headingLevel, Int32 indent, Boolean documentAllChildren)
   at Models.Core.AutoDocumentation.DocumentModel(IModel model, List`1 tags, Int32 headingLevel, Int32 indent, Boolean documentAllChildren, Boolean force)
[...]

PR #3987, opened 2019-07-01
Remove reference to variables not observed (population and LAI)

rcichota:
Working on #2694
Hot fix to enable documentation to be built. This was cause by a reference to observed variables that had no values... (resulting from using an old table)


rcichota:
@hol353 can you accept these changes and merge, please. While solving the issue I've added a few lines to one of the datasets (there was one day with missing data...)

Issue #78, opened 2015-05-04
Fodder Beet Model


HamishBrownPFR:
Fodder Beet model needed

HamishBrownPFR:
@RobZys if you put up a pull request now that:
1. Removes the fodder beet model from replacements (it is in a number of .apsimx files so will need to search to make sure you get them all)
2. Adds a fodderbeet.json file into the resources
3. Includes a simple example in the examples folder
4. tidies up the auto doc as requested by @hol353 in #1845 
5. deal to any other needed so Dean can merge the pull request
Once it is all running clean I think this is ready for release.
@sarahcleary are you happy that all reference pannel requirements have been met and this can be merged

sarahcleary:
thanks @HamishBrownPFR.  Yes, as per the RP meeting yesterday https://github.com/APSIMInitiative/ReferencePanel/issues/27 - I understand if the above is done, the RP is happy for Fodder Beet to be put in release.

sarahcleary:
In release

PR #3983, opened 2019-06-30
Remove Link to replacement in Test Sims containing FodderBeet

RobZys:
Working on #78

PR #3993, opened 2019-07-04
Corrected case, so Linux builds will still work.

zur003:
Working on #78


hol353:
retest this please jenkins

zur003:
Retest this please Jenkins

zur003:
Can someone explain to me how just changing the case of a single image file name for the GUI can possibly cause the performance tests to fail?

hol353:
Yes, I've seen this a lot in the last few days. Seems to always be the same stat in Fodderbeet. Rob noticed a problem last week: https://github.com/APSIMInitiative/ApsimX/pull/3983. I'll investigate.

hol353:
@rcichota I think I've found a problem. The Fodderbeet simulation WaikatoNit0Splits2 always seems to fail the performance tests randomly. The performance site says it is a problem with the observed data being swapped around each run. When I look at the observed file is see this:

image

Notice how each date is duplicated i.e. there are two 2017-01-11. This is a problem. Can you fix this ASAP as it is holding up all builds?

rcichota:
Sorry @hol353 I haven't seen this earlier. The FodderBeet model is being handled by @RobZys and I will talk to him tomorrow and this should explain the weird results he was commenting on. I guess, in the mean time you can delete those lines as you've started...

hol353:
retest this please jenkins

hol353:
This has failed in PlantainGrazing.apsimx, specifically in grazparam.cs:line line 1013. I've seen this happen randomly several times over the last week or two. @zur003, @her123 do you have time to look into it? I've tried reproducing on my computer without success. I don't understand what grazparam is doing at this line.

her123:
I'll look into it Dean.

hol430:
retest this please jenkins

PR #4001, opened 2019-07-08
Remove rows 1865 to 1867

### LINKED ISSUE #4000: Fodderbeet observed file has duplicate dates for WaikatoNit0Splits2
The Fodderbeet simulation WaikatoNit0Splits2 always seems to fail the performance tests randomly. The performance site says it is a problem with the observed data being swapped around each run. When I look at the observed file is see this:

60775286-1a351a00-a164-11e9-9582-23e483525c0b

Notice how each date is duplicated i.e. there are two 2017-01-11. This is a problem.

hol353:
Resolves #4000


RobZys:
@hol353 look at 4002

PR #4002, opened 2019-07-08
Fixed splits N in waikato observed

### LINKED ISSUE #4000: Fodderbeet observed file has duplicate dates for WaikatoNit0Splits2

RobZys:
Resolves #4000

Adding 0 N in 3 splits is the same as adding 0 N in two splits


Tests/Validation/FodderBeet/FodderBeet.apsimx
```diff
@@ -1725,6 +1725,7 @@
                     {
                       "$type": "Models.Core.Simulation, Models",
                       "Descriptors": null,
+                      "Services": [],
```
hol430: @hol353 does this Services object really need to be serialised? Seems unnecessary to me

hol353:
No it shouldn't be serialised.

PR #4006, opened 2019-07-10
CLEM Updated pasture growth to treat suckling individuals separately

### LINKED ISSUE #4005: CLEM Growth rate of sucklings seems too low
While checking the ruminant model, the growth rate of sucklings seems too low compared with the NABSA model

lie112:
Resolves #4005


lie112:
How do I now locate the problem with a Pull performance test fail? It appears to provide a new screen without reporting the error.
Also when I added a new issue online, I got a previously filled in issue box.

hol430:
I don't think this was a problem with your pull request. @rcichota @RobZys this looks like a similar problem to the one you fixed yesterday in #4002? The performance tests seem to think the observed value has changed for WaikatoNit0Splits2.

hol353:
So I've had a look at this problem. The PR fixed yesterday (#4002) does fix the issue. There are now only 3 observed/predicted values for simulation WaikatoNit0Splits2 in the observed file. The performance tests site is still showing 6 values. So the problem has to be either:

1. The collector isn't deleting all existing performance tests data for the PR being run before reimporting the new data.   OR
2. The .db file produced on Muck still has 6 rows in the .db file for this simulation. When I run it on my laptop I see only 3 as expected.

I can't determine which of these 2 possibilities it is because the .db files are being deleted after a pull request runs. I don't know why. @hol430 if you can stop the .db files being deleted then we can determine what is causing the performance tests to fail.

hol430:
retest this please jenkins

PR #4008, opened 2019-07-11
Removed fodderbeet from under review dir.

hol430:
Working on #78


hol430:
@hol353 the performance tests collector is now picking up the correct values for the fodderbeet observed AboveGround.N for the WaikatoNit0Splits2 and WaikatoNit0Splits2 simulations. The previous accepted values were incorrect, probably due to the duplicate fodderbeet files in validation/under review directories. Accepting the changes and merging this pull request should get the other pull requests running green again.

Issue #4007, opened 2019-07-10
Refresh data store does not work


hol430:
Right clicking on a data store and selecting 'refresh' does not work - a message is displayed in the status panel saying 'Post processing models have successfully completed' but this appears to be a lie because the post simulation tools do not get run.

hol430:
Sometimes the datastore's filename gets set to :memory: which causes the excel input component to fail. This also causes any subsequent simulation runs to fail until the file is closed and reopened. This often seems to occur when running with the f6 multi-process job runner.

jbrider:
This is caused by not having any simulations to run.
The function PostAllRuns which will call the postProcessing tools only gets called when the last job completes.  No jobs, no event gets published.
Possible solutions:
1. a dummy job to ensure the event is published - only added when there are no jobs.
2. add the postprocessing functions as a job - this will also show that there is still 1 job running in the interface - would be useful.  It would need a flag of some sorts to wait until all the other jobs have finished - not sure how easy that is.
3. add a check for no jobs that then calls the postprocessing?

hol430:
Re-opening because error messages don't appear in the UI.

PR #4014, opened 2019-07-12
Fixed potential race condition in job runner code

hol430:
Working on #4007

It's not feasible for all simulations to have a child DataStore added at simulation runtime - what would its Parent be? Currently each simulation changes the parent property to point to itself but when simulations are run in parallel, the DataStore's parent could potentially point to a different simulation, which allows models/events to propagate across simulations.

The only reason the Simulation needs a child DataStore is for link resolution. The solution is to pass a reference to the DataStore into the Links object via its Services property. This will probably cause problems if we ever try to add any [Link] fields to the DataStore. There's also still the problem of one simulation being able to potentially access data from another simulation while it is being run; the simulation would have a reference to the DataStore which has a reference to the top-level Simulations node via its Parent property.

Not sure that this is a great solution but if it runs green on Jenkins then I think it would definitely be an improvement over the current system. It certainly seems to fix the problem on my PC.


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol353:
Sorry @hol430 This now has a conflict.

PR #4022, opened 2019-07-17
Added int/double support for event names on graph.

### LINKED ISSUE #4021: EventNamesOnGraph only works if x axis shows DateTime data
EventNamesOnGraph only works if x axis shows DateTime data. If the x axis shows, for example, day of year, EventNamesOnGraph *could* work fine, but it currently attempts to parse the x data to date time objects, which fails.

hol430:
Resolves #4021


hol430:
retest this please jenkins

PR #4024, opened 2019-07-17
Improved regular expression for graph filter.

### LINKED ISSUE #4023: Graph filters rarely work
Putting a filter on a graph tends to be an exercise in frustration. It will either have no effect, or it will have too much of an effect and cause nothing to be graphed. In my specific problem, this seems to be related to the regular expression used to identify field names in the filter.

hol430:
Resolves #4023 

@hol353...do we have any unit tests for this regex which I've changed?


hol353:
There are unit tests that will exercise this regular expression.

hol430:
Yeah they failed the first time, hence my second commit. Nice to know the system works for a change.

PR #4034, opened 2019-07-22
CLEM Added improved date filtering to pivot table

### LINKED ISSUE #4033: CLEM Pivot table data needs improved date filtering
It is not possible to filter by year/month, only by day. Extending the filter would be useful.

MikeStower:
Resolves #4033

### DIFF
--- a/ApsimNG/Presenters/CLEM/PivotTablePresenter.cs
+++ b/ApsimNG/Presenters/CLEM/PivotTablePresenter.cs
@@ -55,17 +55,17 @@ namespace ApsimNG.Presenters.CLEM
             // Attach events to handlers
             this.view.UpdateData += OnUpdateData;
             this.view.StoreData += OnStoreData;
-            this.view.ChangePivot += OnChangePivot;
             this.view.TrackChanges += OnTrackChanges;
 
             // Update the boxes based on the tracked changes
-            this.view.Ledger.ID = table.Ledger;
-            this.view.Expression.ID = table.Expression;
-            this.view.Value.ID = table.Value;
-            this.view.Row.ID = table.Row;
-            this.view.Column.ID = table.Column;
-            this.view.Pivot.ID = table.Pivot;
-            this.view.Time.ID = table.ID;            
+            this.view.FilterLabel.Text = table.Filter;
+            this.view.LedgerViewBox.ID = table.LedgerViewBox;
+            this.view.ExpressionViewBox.ID = table.ExpressionViewBox;
+            this.view.ValueViewBox.ID = table.ValueViewBox;
+            this.view.RowViewBox.ID = table.RowViewBox;
+            this.view.ColumnViewBox.ID = table.ColumnViewBox;
+            this.view.FilterViewBox.ID = table.FilterViewBox;
+            this.view.TimeViewBox.ID = table.TimeViewBox;                        
 
             // Update gridview data (initial loading of data)
             OnUpdateData(null, EventArgs.Empty);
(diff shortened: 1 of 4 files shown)


ApsimNG/Presenters/CLEM/PivotTablePresenter.cs
```diff
@@ -80,82 +80,206 @@ private void OnUpdateData(object sender, EventArgs e)
...
             // Don't try to update if data source isn't found            
             if (input == null) return;
 
-            // Find distinct values in the chosen pivot            
-            table.Pivots = new List<string>(
-                input
-                .AsEnumerable()
-                .Select(r => r.Field<object>(view.Pivot.Text).ToString())
+            // Update filters if the filter button was pressed
+            if (sender is Gtk.Button btn) { UpdateFilter(input, btn.Name); }
```
hol430: We're trying to avoid references to Gtk in our presenter code (as per [here](https://apsimnextgeneration.netlify.com/development/userinterface/1-overview/)). Perhaps you could create a different event inside the presenter which this particular button invokes when it is clicked? This new event could run the `UpdateFilter(...)` method. I also see another Gtk reference, inside the `OnTrackChanges()` event handler.

Also, for if statements, the code to be run should be on a new line. If the code to be run is only a single line, braces should be omitted. e.g.

```csharp
if (view.TimeViewBox.Text == "Daily")
    timescale = 10;
else if (view.TimeViewBox.Text == "Monthly")
    timescale = 7;
else if (view.TimeViewBox.Text == "Yearly")
    timescale = 4;
...
```

I was about to link to the coding style document on the website, but it looks like this is not covered. I'll have to update it when I get the chance.

lie112: I agree. I'll have to wait until I have some time to get in and fix this issue. This was a pull request based on Mike Stower's work before he left. We're very keen to get these working, so shouldn't be too long. I just need to sort out how to get to this repo to make the changes and apply them to this pull request.

lie112: I am fixing this SQL report creator that Mike Stower submitted to removal all reference to gtk from the presenter. Mike had a if (sender is Gtk.Button btn) which I have removed by changing the value of a local property when the view fires a timescalechanged event.
I assume we should also get rid of any view.TimeViewBox.Text that are reading the property of a gtk object in the presenter from the view as well or is that acceptable?

I assume you are trying to separate the view and the presenter completely so they talk via events updating local properties only. 

When I have completed this I will be submitting my own pull request based on the two issues created so we can delete Mike Stowers’ two outstanding pull requests.
How do we delete currently outstanding pull requests. Do I just add a comment and close them or do you have to do it?
Can I have a single pull request that resolves two outstanding issues?

lie112: Should all the Gtk components in a view be private as no presenter should be talking to the directly?

hol430: > I assume we should also get rid of any view.TimeViewBox.Text that are reading the property of a gtk object in the presenter from the view as well or is that acceptable?

Yeah, anything in the presenter which makes assumptions about the display technology used in the view should be removed. That way, we could change the views to use WinForms, or qt, or anything else and we shouldn't have to change the presenters at all. 

> Should all the Gtk components in a view be private as no presenter should be talking to the directly?

Gtk components should be private yes - the presenter should not be able to see them. However it's ok for the presenter to talk to the view directly. There is [a document](https://apsimnextgeneration.netlify.com/development/userinterface/1-overview/) on the website (written by Eric?) which describes this and is well worth a read. Basically, the presenter can tell the view what to do by setting properties, calling methods or whatever. The view can't talk directly to the presenter - all it can do is fire off events. I think from memory, some views (ViewBase?) do expose Gtk properties, and these should ultimately be removed as well. 

> How do we delete currently outstanding pull requests. Do I just add a comment and close them or do you have to do it? ... Can I have a single pull request that resolves two outstanding issues?

You can just close this pull request - there should be a close button at the bottom of the page. A pull request addressing two issues is fine.

lie112:
This issue will now be fixed by @lie112  and submitted on a new pull request

PR #4059, opened 2019-07-30
Simplified Ruminant.IsPregnant and BirthDue to fix timing of births in CLEM.

### LINKED ISSUE #4058: CLEM Ruminant births appear unexpectedly
There is a problem where births appear in the ruminant herd without pregnant females in months where births should not occur based on the timing of breeding

lie112:
resolves #4058


Models/CLEM/Activities/RuminantActivityBreed.cs
```diff
@@ -245,14 +245,16 @@ private void OnCLEMAnimalBreeding(object sender, EventArgs e)
             // for each location where parts of this herd are located
             foreach (var location in breeders)
             {
-                // determine all foetus and newborn mortality.
-                foreach (RuminantFemale female in location.Where(a => a.Gender == Sex.Female).Cast<RuminantFemale>().ToList())
+                // determine all foetus and newborn mortality of all pregnant females.
+                foreach (RuminantFemale female in location.Where(a => a.Gender == Sex.Female).Cast<RuminantFemale>().Where(a => a.IsPregnant | a.BirthDue).ToList())
```
hol430: Should this be logical or (double |)?
```suggestion
                foreach (RuminantFemale female in location.Where(a => a.Gender == Sex.Female).Cast<RuminantFemale>().Where(a => a.IsPregnant || a.BirthDue).ToList())
```

lie112: I do need the logical OR operator. One or the other must be true. My changing of some properties and when states change actually means I can delete the | BirthDue() part of the test and it will still work fine.

hol430: I don't understand the context here, but a single vertical bar is not a logical or, it's a bitwise or. They're mostly equivalent here since you're applying it to two bools, but in general it's a bit confusing/incosistent with the rest of the code to see the bitwise or here.

I see you've since changed this line of code anyway so it's a bit of a moot point now.

lie112: I always get a bit confused here with OR statements. I was going with the Microsoft documentation that states
**The | operator computes the logical OR of its operands. The result of x | y is true if either x or y evaluates to true. Otherwise, the result is false.**
The old way the model was working meant a female could be Pregnant or have birth due, buy not both. Therefore I wanted if either x is true or y is true.
Anyway, I have rewritten the code so that the multiple test is no longer needed and all we need to check is IsPregnant() now.

Issue #2245, opened 2018-01-18
Refactor soil class


hol353:
In Soil.cs, everytime a model calls any of the properties e.g. LL15, DUL, SAT, the property is checked to see if the layer structure is correct and maps the variable into the correct layer structure if necessary. This means that properties are duplicated. Currently there is a LL15 in Soil.cs (which provides mapped values) and there is an LL15 in SoilWater.cs. I don't like this duplication.

Instead, a better approach would be to have code somewhere (soil.cs?) to map all variables at the start of the simulation. That means that models can call LL15 in SoilWater.cs and expect to get the correct mapped versions.

hol353:
This refactor effort has now broadened to a general refactor the soil.cs and it's child models.
![image](https://user-images.githubusercontent.com/1943827/63064344-025f6b80-bf44-11e9-8651-47bdeb09855b.png)
The left hand side is the input side, the view that the user sees in the GUI. We are suggesting to rename child nodes of soil as per image and move properties from sample into other nodes eg NO3 and NH4 goes into a 'Chemistry' node. We want to migrate to having properties only specified in one place.  Variables on the input side are to be concentrations (ppm for nitrogen, volumetric for water variables)

The right hand side shows variables available during a model run. All child models except Initial will be removed from the simulation. All variables will be only available in one set of mass based units (kgha for nitrogen, mm for water). A unit specification functionality will be available in REPORT to convert units for output. All variables will be on a standard layer structure.

zur003:
At the moment, particle size distribution (which I happen to need for G-Range) falls under "Analysis", but it doesn't seem appropriate to move that to "Chemistry". Where will these values be placed?

I suppose there's a distinction between values that will routinely change during the course of a simulation (water, C, N, and other elements), and relatively "fixed" properties like DUL and LL15, but even these latter could change in the context of an erosion or pedogenesis model.

hol353:
Particle size numbers will probably stay under Analysis / Chemistry. Perhaps we need a better name.

We are still debating what goes in Initial. Perhaps all variables, even ones that normally don't change e.g. LL15.

rcichota:
[...]

PR #4084, opened 2019-08-12
Get compiling again. Remove NUnitsEnum - not needed

hol353:
Working on #2245


zur003:
So this means we will no longer be able to choose between using ppm and kg/ha in the GUI?

hol353:
No. You can still choose between kg/ha and ppm. There is a bug there somewhere though that I'm currently tracking down.

Issue #4077, opened 2019-08-12
Improve soil wat validation


HamishBrownPFR:
There is currently only one experiment in soil wat test set.  More needed

HamishBrownPFR:
This is superseded by efforts to improve the PO system

PR #4091, opened 2019-08-13
Soil water validation

HamishBrownPFR:
working on #4077   @hol353 the pull request #4090  will need to be merged before this one


HamishBrownPFR:
@hol353 I have added additional data to the test set for soil water which is why it failed performance tests.  We are still working on this test set so please merge in the mean time

PR #4100, opened 2019-08-16
Replace apex model into wheat

### LINKED ISSUE #4099: Use apex model for wheat
Test the wheat model to use apex model

byzheng:
resolves #4099


HamishBrownPFR:
@byzheng I don't think the apex model is working well enough over the validation set to be able to put it into release.  I will close this pull request

PR #4119, opened 2019-08-20
Add error handling to divide

### LINKED ISSUE #4117: Unprotected divide in Agpasture

HamishBrownPFR:
resolves #4117 
@sno036 @rcichota I encountered a fatal error in Agpasture while trying to get canopy rainfall interception code working properly.  I traced it back to this unprotected divide

### DIFF
--- a/Models/AgPasture/AgPasture.PastureSpecies.Organs.cs
+++ b/Models/AgPasture/AgPasture.PastureSpecies.Organs.cs
@@ -1638,7 +1638,7 @@ namespace Models.AgPasture
                 for (int layer = 0; layer < nLayers; layer++)
                 {
                     DMLayer[layer] += DMLayersTransferedIn[layer];
-                    NamountLayer[layer] += NLayersTransferedIn[layer] - (NRemobilised * (NLayersTransferedIn[layer] / NTransferedIn));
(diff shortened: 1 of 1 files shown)


rcichota:
Thanks @HamishBrownPFR 
There were quite a few of these in the original code, I thought I had fixed them, but clearly not all...

HamishBrownPFR:
@rcichota this bug fix has had an impact on AgPasture performance.  Are you happy for this to be merged?  It sort of has to be merged as it causes exceptions currently but is their a need to change anything else in Agpasture to account for this?

rcichota:
hmmm, I am getting a bit suspicious about this now. I have a better look at the code and realised that the fact that there wasn't a check for negative division is because such event should not happened at that stage (there is actually a check for very low values, defined by 'MyPrecision'). for that part of the code to be triggered it means that values for DM changed (growth) but without addition of nitrogen. This should not happen, any growth should have at least minimum N content. Are you able to tell me what was change and/or give me access to the branch that has the changes so I can have  closer look?

HamishBrownPFR:
@rcichota Pull the current tip revision of master and change the c_interception parameter on Microclimate to 0.3.  This will cause the RadcliffePoolburnD simulation to throw an a fatal

Issue #3532, opened 2019-02-19
Error to compile latest version of APSIMX


byzheng:
The new clone of APSIMX and APSIM.Shared cannot be complied after returning from two months.

It seems some packages are missing, but I can find them in the package folders.

Any ideas about these errors?

The errors list below.
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0246	The type or namespace name 'DocumentFormat' could not be found (are you missing a using directive or an assembly reference?)	APSIM.Shared	C:\Users\zhe00a\Documents\Working\04-Software\032-ApsimX\APSIM.Shared\Utilities\ExcelUtilities.cs	17	Active
Error	CS0246	The type or namespace name 'DocumentFormat' could not be found (are you missing a using directive or an assembly reference?)	APSIM.Shared	C:\Users\zhe00a\Documents\Working\04-Software\032-ApsimX\APSIM.Shared\Utilities\ExcelUtilities.cs	18	Active
Error	CS0246	The type or namespace name 'ExcelDataReader' could not be found (are you missing a using directive or an assembly reference?)	APSIM.Shared	C:\Users\zhe00a\Documents\Working\04-Software\032-ApsimX\APSIM.Shared\Utilities\ExcelUtilities.cs	20	Active
Error	CS0246	The type or namespace name 'FirebirdSql' could not be found (are you missing a using directive or an assembly reference?)	APSIM.Shared	C:\Users\zhe00a\Documents\Working\04-Software\032-ApsimX\APSIM.Shared\Utilities\Firebird.cs	13	Active
Error	CS0246	The type or namespace name 'FbConnection' could not be found (are you missing a using directive or an assembly reference?)	APSIM.Shared	C:\Users\zhe00a\Documents\Working\04-Software\032-ApsimX\APSIM.Shared\Utilities\Firebird.cs	25	Active
Error	CS0246	The type or namespace name 'FbServerType' could not be found (are you missing a using directive or an assembly reference?)	APSIM.Shared	C:\Users\zhe00a\Documents\Working\04-Software\032-ApsimX\APSIM.Shared\Utilities\Firebird.cs	30	Active
Error	CS0246	The type or namespace name 'FbTransaction' could not be found (are you missing a using directive or an assembly reference?)	APSIM.Shared	C:\Users\zhe00a\Documents\Working\04-Software\032-ApsimX\APSIM.Shared\Utilities\Firebird.cs	322	Active
[...]

PR #4121, opened 2019-08-21
Moved APSIM.Shared into ApsimX

hol430:
Working on #3532


hol430:
@zur003 @her123 @sno036 @rcichota @jbrider @BenAbabaei @byzheng @ver078 @lie112 @MikeStower 

FYI - we've moved APSIM.Shared into ApsimX. This means you no longer need to maintain a separate APSIM.Shared repository. You can safely delete APSIM.Shared from your PC if you don't work on one of our other repos.

zur003:
If we make changes to files on the APSIM.Shared subfolders, and have them committed, will those modifications automatically be merged into the APSIM.Shared repository? Should we do this manually? Or are the two versions of APSIM.Shared effectively going to be forked, and not necessarily be kept in common?

hol430:
The last one - the two repositories probably won't be synchronised and will diverge over time. I think the plan is to gradually phase out APSIM.Shared but keep it around for legacy purposes. The other codebases which reference it - performance tests, websites, etc - really only use a few utility methods and that's it (I think!). 

I don't know that there will be many changes which need to propagate to all of our other codebases.

hol353:
To stop the divergence over time, I'd probably lean towards closing the APSIM.Shared repo. This will force the issue for the other tools.

hol430:
If we want to check out an older version of ApsimX, we would still need to clone APSIM.Shared - deleting it would make this impossible. Although I agree that having two different APSIM.Shareds out there is not ideal.

PR #4122, opened 2019-08-21
remove interception losses from daily palm water demand

### LINKED ISSUE #4116: Interception losses not accounted for in Oil Palm water demand
The tree water demand should be discounted when interception losses are turned on in OP simulations otherwise the atmospheric demand can be met twice!

hut104:
Resolves #4116


hut104:
Conflict in code.
Note that the suggested change now also makes use of a property now deleted from MicroClimate and so will need to be implemented in a different manner.

PR #4129, opened 2019-08-21
Ensure released models are kept up-to-date

### LINKED ISSUE #4110: Ensure resources are kept up to date
- Write a unit test that will fail if any resource isn't current. 
- Remove simulations node at top of resource. Put version number into Model.cs instead?
- Remove oats1.json and potato1.json resources
- Some resources e.g. oats don't display correctly by default. If you click on 'source' tab at bottom of window it is Ok.
- In GUI, change the 'Convert XML file' to 'Convert files'

hol430:
Resolves #4110


zur003:
This patch, in file AddModelCommand.cs, introduced a local variable modelToAdd, which conflicts with a class level variable with the same name. I suspect this may have broken "Undo", since that makes use of the class-level variable.

PR #4125, opened 2019-08-21
On Windows, automatically copy deployment support files to Bin before…

hol430:
… building. This build step will not run on Unix machines.

Combined with #4121, this resolves #3532.

PR #4130, opened 2019-08-22
Fixed xcopy command in apsimng.csproj

hol430:
Resolves #4125 

Thanks Eric - that's a good catch. If I run the command manually I get the prompt you mentioned because the file already exists. Strangely though I'm able to compile ok even if the file already exists.


zur003:
One more thing to fix: the reference to pango-sharp.dll should have the HintPath point to the Bin64 folder, not the Bin folder.

PR #4148, opened 2019-08-28
Fix OnEvent Function so it evaluates the IFunction Value every time...

### LINKED ISSUE #4146: Bug in OnEventFunction
The OnEventFunction will only update its value when we change phases. However the child functions' values could potentially change at any time. This problem was identified because we have a sorghum cultivar which modifies the value of a constant function which is referenced by the OnEventFunction. The OnEventFunction continues to return the old value of the constant even after the cultivar changes have been applied.

@HamishBrownPFR is this the intended behaviour of this function?

jbrider:
resolves #4146 
 ...it is referenced, - not just on event change.

### DIFF
--- a/Models/Functions/OnEventFunction.cs
+++ b/Models/Functions/OnEventFunction.cs
@@ -15,7 +15,8 @@ namespace Models.Functions
     public class OnEventFunction : Model, IFunction, ICustomDocumentation
     {
         /// <summary>The _ value</summary>
-        private double _Value = 0;
+        private bool _preEvent = true;
+
 
         /// <summary>The set event</summary>
         [Description("The event that triggers change from pre to post event value")]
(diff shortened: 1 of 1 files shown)


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
@hol353 Error in jenkins, in the multi process job runner code:

```csharp
   at Models.Core.Run.JobRunnerMultiProcess.PipeServerTaskThread() in C:\Jenkins\workspace\PullRequest_v2\ApsimX\Models\Core\Run\JobRunnerMultiProcess.cs:line 130
```

I'm not sure how this code is supposed to work, but it seems odd that it's suddenly causing problems since we haven't changed anything in this space recently.

hol353:
reboot muck and try again?

hol430:
@jbrider - looks like this isn't going to work after all. Should this (and #4146) be closed?

jbrider:
Yes, they can be closed.
It looks like everywhere it's implemented is designed to use cached values.

PR #4158, opened 2019-09-05
Introduce a permutation model for factorials

### LINKED ISSUE #4157: Need a permutation model for factorials
We need more flexibility when creating factorials. My use case is to be able to run a series of weather files (forecast ensemble members) against a range of crops at different grid cells (lat/longs) e.g.
image

hol353:
Resolves #4157


Models/Core/ApsimFile/Converter.cs
```diff
@@ -1217,6 +1217,26 @@ private static void UpgradeToVersion65(JObject root, string fileName)
             }
         }
 
+        /// <summary>
+        /// When a factor is under a factors model, insert a permutation model.
+        /// </summary>
+        /// <param name="root"></param>
+        /// <param name="fileName"></param>
+        private static void UpgradeToVersion66(JToken root, string fileName)
```
hol430: @hol353 does this converter work? The first argument here is supposed to be a `JObject`, not a `JToken`.

Issue #1159, opened 2016-09-13
SQLite error running current revision


sno036:
I cannot run anything in APSIM X.  Have deleted everything and pulled down a completely fresh version and doing right-click run or F5 of simulations that should be Ok (e.g. examples) results either in no action at all or an error saying the SQLite is not open.  See attached screenshot from a debugger run.  Any suggestions?  Thanks, Val

![capture](https://cloud.githubusercontent.com/assets/12176269/18457361/43c9e9aa-79ac-11e6-9c55-142fce6247ce.JPG)

zur003:
Interesting problem. Are you running on 32-bit or 64-bit Windows? The "BadImageFormatException" suggests a possible mismatch between 32- and 64-bit. But that shouldn't happen, since UserInterface.exe should always run as 32-bit...

sno036:
I'm running Windows 64 bit.  Windows 7 Professional.  I last used APSIM X (and it didn't have any issues) last week on 5, maybe 6, September.

zur003:
I can confirm that I see exactly that behaviour, with that error, if I compile UserInterface.exe to run as a 64-bit application. However, the build you "pulled down" should have been built with the "Prefer 32-bit" option set, which instructs the application to run as 32-bit on all platforms. I'm also running on Windows 7, 64-bit, but I see no problems with the current version.

sno036:
![capture](https://cloud.githubusercontent.com/assets/12176269/18459906/79d549da-79c1-11e6-9d15-fcd5739216ed.JPG)

Just ferretted around in the properties of UserInterface and found that the "Prefer 32-bit" was not checked (screen shot attached).  I checked it, built it and it ran fine (as you suggest above!).  

I don't think that I changed this????? so ???

Thanks - as ever! - Eric.

zur003:
The way this stuff gets handled changed slightly with the move to version 4.5 of the .Net frameworks. Looks like I might not have had all the settings right in all configurations when making that change. It may be that the Prefer 32-bit flags wasn't set for debug configurations. I will check, and correct if necessary.

zur003:
Yeah. Mea culpa. The flags wasn't set for Debug builds. A fix is on its way.

PR #4171, opened 2019-09-10
adding an empty job object to force jobs finished event to fire

### LINKED ISSUE #4007: Refresh data store does not work

jbrider:
resolves #4007

### DIFF
--- a/APSIM.Shared/JobRunning/JobManager.cs
+++ b/APSIM.Shared/JobRunning/JobManager.cs
@@ -10,7 +10,8 @@
     public class JobManager : IJobManager
     {
         private Queue<IRunnable> jobs = new Queue<IRunnable>();
-        private int numJobsToRun;
+        /// <summary> number of jobs </summary>
+        protected int numJobsToRun;
         private bool initialised;
 
         /// <summary>Invoked when this job manager has finished everything.</summary>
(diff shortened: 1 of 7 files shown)


jbrider:
It's failing on a test: EnsureOnlyPostSimulationToolsAreRun which is testing for the wrong thing.
It's testing if no simulations are run, but not testing if postprocessing happens.

jbrider:
Having issues running tests. Getting an error opening the sqlite db - BadImageFormatException - which according to a previous issue (#1159) is something to do with 32bit + 64bit.
Other tests failing as well - is there anything you have to configure to get tests to run successfully?

hol430:
Yeah the setting you need is hidden away under the Tests ribbon in visual studio:

Tests -> Test Settings -> Default Processor Architecture -> X64

image

jbrider:
Mine didn't quite look like that - but found the same setting in the
testexplorer dialog.

jbrider:
While looking at the contents of a table in the datastore - how do you update the table display when you right click->Refresh on the datastore?  
explorerPresenter.Refresh(); didn't appear to do anything.

hol430:
Hmm...maybe something like this?

```csharp
(explorerPresenter.CurrentPresenter as DataStorePresenter).PopulateGrid();
```

jbrider:
That worked. 👍 
Should I create a separate issue for that, ,or just push it through with this one?

hol430:
I'd say just push it through with this one, it's a small enough change.

jbrider:
How do you work out why a test failed? It's running green on my machine.

PR #4182, opened 2019-09-12
When importing old .apsim files, add fert/irrigation nodes if needed.

### LINKED ISSUE #2762: import of old APSIM files bringing faulty nodes
SurfaceOM needs to be renamed SurfaceOrganicMatter when the import occurs

Also the crop specific characteristics within soil/water needs to be renamed CropSoil, not just 'crop' as it is at the moment.

hol430:
Resolves #2762

### DIFF
--- a/Models/Importer/APSIMImporter.cs
+++ b/Models/Importer/APSIMImporter.cs
@@ -374,6 +374,41 @@
                         XmlUtilities.SetValue(newNode, "InitialCNR", "80");
                     }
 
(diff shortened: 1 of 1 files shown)


jbrider:
Can you add a check if the irrigation is set to automatic - it needs to add
the automatic irrigation script with the relevant parameters.

hol430:
Good thinking - that would have been handy a few weeks ago

hol430:
retest this please jenkins

hol430:
Ahh...I don't think this one was good to merge

PR #4194, opened 2019-09-18
World modellers

### LINKED ISSUE #4193: Problem when two crops are present in a field
When plant A is sown, it publishes a PhaseChanged event. This event goes to plant B which reports the sowing as if it was it's own.

hol353:
Resolves #4193


hol430:
retest this please jenkins

PR #4207, opened 2019-09-24
Potato Example Cleanup

### LINKED ISSUE #4206: Potato Model Clean Up
A few issues have been indentified with the potato example, including issues with soil parameters and the manager script.  These should be cleaned up to provide a better starting point for people modelling potato.

hut104:
resolves #4206


hut104:
@HamishBrownPFR  may want to check the changes to this example

Issue #2575, opened 2018-05-04
Expand validation of Potato model


HamishBrownPFR:
We have access to the Substore validation set and some other data also which will be used to expand the potato validation

PR #4218, opened 2019-09-26
Potato

HamishBrownPFR:
working on #2575


HamishBrownPFR:
@hol353 I am wanting to share this progress with a @JJguri and @RobZys.   Could you please check and merge when you have a chance.

Thanks

hol353:
Why is this in prototypes? Potato is a released model.

HamishBrownPFR:
We have significantly expanded the validation from ~50 simulations to ~350  and are still have a few more to add and some more checking to do.  I want to get progress into the build but not have it as the base that the model is tested against yet.  
As part of this process we are establishing a procedure for rigorous checking of all aspects of  a validation set and I would like to discuss this with RP at some stage.
There it is also a possibility we will make some changes to the potato model that will need review.
All of this could be done through the released potato validation set but there are three of us working on this and it will be a bit quicker and cleaner to do pull requests via the dev folder while we are still setting things up

Issue #4202, opened 2019-09-24
Improve aggregated reporting


hol430:
Would be handy if report was more powerful in terms of aggregation, as this could eliminate a lot of manager scripts.

Currently we can aggregate to [Clock].Today, but not from variable dates such as report's last report date. As part of this it would be good to refactor some of the reporting code as this is quite complex to work around.

We also need to update the test suite and documentation to include tests/examples for all of these features.

hol430:
@hol353 the [Clock].EndOfSimulation event gets called before [Clock].DoReport. Is this intentional? It's causing some problems for me because the ReportColums need to perform some cleanup (manually unsubscribing from events) at the end of the simulation, but if we subscribe to the EndOfSimulation event, then we'll disconnect our event handlers before the final report.

hol353:
EndOfSimulation is in the wrong place!! It will be invoked every date. This needs to be moved.

What's the use-case for moving EndOfDay to after DoReport?

hol430:
EndOfSimulation will only be invoked on the last day of the simulation - there's a check:

```csharp
if (Today == EndDate && EndOfSimulation != null)
    EndOfSimulation.Invoke(this, args);
```

I don't have a specific use case for moving EndOfDay after DoReport; however the name EndOfDay implies that it happens at the end of the day, which is currently untrue.

If an end user without the source code is writing a manager script, they might make the (I think quite reasonable) assumption that end of day happens at the end of the day.

sno036:
It might not be the end of the day but functionally there are no more meaty actions left?

There are times in simulations when you do want to do a report on the last day of the simulation (in these cases it would often be the only report). I don't mind if you change the stuff above provided I can set a report as "end_sim" either in Report or from a Manager script., 

Hope that makes sense!

hol353:
Ok move the two events. This will still let you do a report on EndOfSimulation.

hol430:
@sno036 I've added a `[Report].DateOfLastOutput` date so both of the examples you posted in the pull request *should* work. The rules around whether the aggregation is inclusive of the to/from dates are complicated but let me know if it's not doing what you expect.

Also worth noting that ave does not work for computing the average - it's actually avg despite what the documentation might say. I think we need to move away from a static PDF for the report documentation - perhaps we should create a new page on the next gen website. At least that way it's easier to find and modify.

```
[...]

PR #4221, opened 2019-09-30
Implemented Report.DoReport event

hol430:
Working on #4202 

e.g. `sum of [Clock].Today.DayOfYear from [Report].DoReport to [Clock].Today as sum`

### DIFF
--- a/Models/Report/Report.cs
+++ b/Models/Report/Report.cs
@@ -79,6 +79,11 @@ namespace Models.Report
         [Description("Output frequency")]
         public string[] EventNames { get; set; }
 
+        /// <summary>
+        /// Date of the last report event.
+        /// </summary>
+        public event EventHandler DoReport;
+
         /// <summary>An event handler to allow us to initialize ourselves.</summary>
         /// <param name="sender">Event sender</param>
         /// <param name="e">Event arguments</param>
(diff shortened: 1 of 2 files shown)


hol353:
@sno036: DoReport sounds like a method name. I know it was the name we used in APSIM 7.10. 

Need to come up with an event name. How about 'Reported' or 'Outputed'. The example would be:

sum of [Clock].Today.DayOfYear from [Report].Outputed to [Clock].Today as sum

hol353:
or even better: Why not change report to have a DateOfLastOutput property so the example becomes:

sum of [Clock].Today.DayOfYear from [Report].DateOfLastOutput to [Clock].Today as sum

sno036:
I like that last suggestion. I assume that this will work with array operations as well - e.g.

Ave([Soil].SoilWater.SW) from [Report].DateOfLastOutput to [Clock].Today as AverageWaterContent
and (slightly ridiculously)
(Ave([Soil].SoilWater.Sat) + Ave([Soil].SoilWater.Dul)) from [Report].DateOfLastOutput to [Clock].Today as AverageSomethingSilly

hol430:
Closing this in favour of #4222

PR #4264, opened 2019-10-14
Edit file from terminal feature

### LINKED ISSUE #4262: Need a way to call Models.exe to change a property value
This would be useful when generating .apsimx files from a python or R script where we start with a template file and want to change a few parameter values before running the file again. Without this feature, the script would have to interact directly with the raw json which can potentially break if the user then upgrades to another version of apsim.

Current plan is to add an option to Models.exe which allows the user to specify a file and a path to a config file. The config file would look something like this:

```
[Clock].StartDate = 1/1/2019
[Weather].FileName = asdf.met
```

The internal parameter modification mechanism would work similarly to how the morris component currently works.

hol430:
Resolves #4262 

From commit message:

This allows the user to specify an .apsimx file and a config file when running Models.exe. The .apsimx file will not be run but instead, the changes listed in the config file will be applied to the .apsimx file, which will then be written to disk under the same filename.

The config file should contain lines of the form 'path = value'

e.g.

```
[Clock].StartDate = 2019-1-20
.Simulations.Sim1.Name = SimulationVariant35
.Simulations.Sim2.Enabled = false
.Simulations.Sim1.Paddock.Soil.Thickness[1] = 50
```

Notes:

- Command line arguments should look like:

Models.exe file.apsimx /Edit /path/to/config/file.conf

- Relative paths will be resolved to the first match. ie
[Clock].StartDate will match the first clock found in the file.

- Dates can be specified as yyyy-mm-dd or mm/dd/yyyy.

- Strings should not be quoted

- Array indices will be interpted as 1-indexed (😡). So the first
element in the array should have index 1 in the config file.

- The file will be upgraded to the latest file version as part of this
process.

### DIFF
--- a/Models/Core/VariableComposite.cs
+++ b/Models/Core/VariableComposite.cs
@@ -8,6 +8,7 @@ namespace Models.Core
     using Functions;
     using System;
     using System.Collections.Generic;
+    using System.Linq;
 
     /// <summary>
     /// This class encapsulates a list of IVariables that are evaluated when
@@ -169,7 +170,13 @@ namespace Models.Core
         {
             get
             {
-                return Object.GetType();
+                if (Object != null)
+                    return Object.GetType();
+
+                if (variables != null && variables.Count > 0)
+                    return variables.Last().DataType;
+
+                throw new Exception("Variable is null");
             }
         }
 
--- a/Models/Main.cs
+++ b/Models/Main.cs
@@ -226,6 +226,9 @@
             string[] lines = File.ReadAllLines(configFileName);
             for (int i = 0; i < lines.Length; i++)
             {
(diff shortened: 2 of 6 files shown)


jbrider:
What's the reason for the 1-indexing?

hol430:
The report/cultivar/factorial/morris/etc features all use 1-based indexing, and this feature uses the same code. I strongly disagree with the 1-based indexing but it's less work to use the code that's already there and at least this way it's consistent with other parts of apsim.

hol430:
@jbrider - just thought - this might be of interest to Mike who is procedurally constructing .apsimx files. I guess he'd still have to manually construct the json when adding components into the simulation, although if he started with a template file which contained all of the components he could possibly need, he could then just disable stuff he doesn't need.

Issue #4263, opened 2019-10-14
Property presenter is not handling DisplayType.Filename


zur003:
Properties marked with [Display(Type=DisplayType.Filename)], when viewed in the PropertyPresenter, should display both the filename text and a button which should invoke a file selection dialog when pressed. This isn't working. The button appears, but pressing it initiates editing of the text, and does not cause a dialog to appear.

hol430:
@zur003 do you remember how this used to work? It seems like in order for the CellRendererToggle's Toggled event to fire, we have to click on it while the cell is in edit mode. It seems like it would be more intuitive if we could click on the button even if the cell is not already selected.

zur003:
@hol430 - I have no idea why this isn't working as expected. I can't get the Toggled event to fire at all. I've submitted a pull request which provides a work-around, but we shouldn't have to do things that way. I thought this used to work (although it was never used very much), but I have no clue what may have broken it.

hol430:
This is not a problem with the new properties UI. The old gridview-based properties UI no longer exists so I'm closing this as fixed.

PR #4265, opened 2019-10-15
Added ugly hack to support file selection dialogs invoked from proper…

zur003:
…ty grids. Working on #4263 

This "fixes" the problem, but in a very ugly way. I have no idea why Gtk is not behaving as intended, and this is a clumsy workaround. Hence I do not regard this as a full resolution of the problem. @hol430 - do you have any idea why the "toggled" event is not being fired in the expected way?


zur003:
@hol430 - Is there something wrong with Jenkins builds? Looks to be some sort of problem with R scripting...

hol430:
Yeah this is the solution that I was considering but wasn't brave enough to implement. The event isn't reaching our OnChooseFile method because we need to set the return value of the button down event handler to true, which for some reason suppresses the toggled event. 

I think the error in Jenkins is a race condition and will disappear if we rerun the tests.

hol430:
retest this please jenkins

PR #4266, opened 2019-10-15
Ensure Property grid displays the description of enumerated values, n…

### LINKED ISSUE #4236: Enum properties in grid not working correctly
Enumerated properties displayed in a property grid seem to be broken in the cases where a Description attribute is used to provide the displayed text. Selecting a new value results in a System.ArgumentException being thrown (and not caught for proper handling).

Note that this only occurs when the Description text differs from the enumeration item name.

zur003:
…ot the value itself. Resolves #4236


zur003:
@hol430 - This seems to have stalled on Jenkins with the same R-related "error: package 'vctrs' not found" message as yesterday. Any clues?

zur003:
Retest this please Jenkins

PR #4280, opened 2019-10-22
Fix ln p cover values

### LINKED ISSUE #4279: Problems with observed data for L and P experiment
The observed cover green data was wrong (zeros when it should not be) for the first few months of this experiment.  This data is used in the slurp and soilwater validations where it is read in a manager script and pushed into slurp so it can do water extraction.  
Also the calculations of Profile Water Change were wrong for the LandP experiment in the observed file for the soil water validation

HamishBrownPFR:
resolves #4279


hol353:
retest this please jenkins

HamishBrownPFR:
@hol353  this can be merged.  Changes are expected because of changes to input values of CoverGreen for LanP simulations in Slurp and Soil water validations sets.  Given mostly improvements with some small reductions for deep soil profile soil water values

Issue #4275, opened 2019-10-21
Jenkins builds have been failing.


zur003:
Most of my recent pull requests have failed to run on Jenkins.  The log contains errors like this:

> StdErr:
>Warning message:
>package 'boot' was built under R version 3.5.3 
>Error: package or namespace load failed for 'sensitivity':
> .onLoad failed in loadNamespace() for 'pillar', details:
>  call: utils::packageVersion("vctrs")
>  error: package 'vctrs' not found
>In addition: Warning message:
>package 'sensitivity' was built under R version 3.5.3 
>Execution halted
>StdOut:
>   at Models.Sobol.Run() in C:\Jenkins\workspace\PullRequest_v2\ApsimX\Models\Sensitivity\Sobol.cs:line 390
>   at Models.Core.Run.SimulationGroup.RunPostSimulationTools() in C:\Jenkins\workspace\PullRequest_v2\ApsimX\Models\Core\Run\SimulationGroup.cs:line 269
>ERRORS FOUND!!

hol430:
I don't have time to look into this this week so I am going to opt for the simple solution and install the missing packages for now and look into this more thoroughly when I have time. Will leave this issue open so that we know there is an issue for now.

hol430:
I'm going to go ahead and close this as this particular problem appears to be fixed. Feel free to reopen if it shows up again.

PR #4281, opened 2019-10-23
Fixed bug in R component (I hope)

hol430:
So it seems that due to a bug in R(?) that when we load a package via library('asdf', lib.loc = '/packages/'), lib.loc is ignored in certain circumstances (maybe while loading dependencies or if dependencies are installed to a different location?). The simplest solution is to add this path to .libPaths for the current session.

Working on #4275


hol430:
retest this please jenkins

hol430:
@hol353 can you merge this please (after closely inspecting the changes of course) - it should get the other pull requests going again.

PR #4302, opened 2019-10-28
Fixed zone name in report

### LINKED ISSUE #4298: What is Zone column in a Report?
Why is the zone column value in reports simply set to the Report's Parent.Name? (Report.cs line 142)
I would have expected Zone to read up the tree until it found a parent of type Zone.
I was looking to run multiple CLEM instances (each effectively a Zone) within a Simulation and use the Zone column in reports to distinguish the output from each farm, but the report presents the name of the folder holding the report ("Reports") as the zone name and seeing each farm has the same structure, they all report "Reports" as the Zone name.
Is this the intentional use of the Zone column, or is the a hangover from a previous simple implementation that adding a more complex tree structure for CLEM simulations doesn't follow?

hol430:
Resolves #4298

### DIFF
--- a/Models/Report/Report.cs
+++ b/Models/Report/Report.cs
@@ -139,7 +139,8 @@ namespace Models.Report
         private string[] TidyUpVariableNames()
         {
             List<string> variableNames = new List<string>();
-            variableNames.Add("Parent.Name as Zone");
+            IModel zone = Apsim.Parent(this, typeof(Zone));
+            variableNames.Add($"[{zone.Name}].Name as Zone");
             for (int i = 0; i < this.VariableNames.Length; i++)
             {
                 bool isDuplicate = StringUtilities.IndexOfCaseInsensitive(variableNames, this.VariableNames[i].Trim()) != -1;


lie112:
Is there a reason why the table name drop down in datagrid is now disabled? This doesn't really matter when reporting a specific report, but when you click on the DataStore you are unable to access any of the tables present.

hol430:
Yep that was my bad - it's fixed in #4305

lie112:
Is it possible to filter data based on the Zone column in the DataStore view? If you select a report under a zone then the data displayed in the data tab should only be for the current zone as it is for the simulation. At present all zones are displayed. I've had a look, but don't want to break anything to address this. I guess one of the Presenters needs to determine the Zone parent of the report and add this to the GetData() filter in DataStorePresenter.GetData.

hol430:
The data store view should have a row filter text box. I always forget the exact syntax but I think it should be something like:

```
[Zone] = 'NameOfZone'
```

lie112:
Should this be automated as it is with Simulation name? I would suggest the inclusion of Zone is for the same reasons as Simulation and should only report the data matching both Simulation and Zone in the display. You can look in the tables in the DataStore to see all data across simulations and zones. I realise most people aren't using multiple zones, but the functionality would be good.
I guess we could automatically fill in the Filter for the report, but it gets a bit messy when the user has provided a custom filter.  So the same functionality as the SimulationName for Zone would be great.

hol430:
What do you mean by automated exactly - you mean that if we click on a report that's under a zone it should only show data in that zone? I think that's how it's supposed to work (could be wrong!) so if it's not working then maybe that's a bug or an oversight.

lie112:
Yes, that's what I expected to happen, but it does not.
It doesn't appear that Zone is in the filter and it's not blank as would be expected if it was incorrectly filtering on the old Parent.Name, so this has been missed. I wasn't quite sure how to add this.

lie112:
As far as I can see the DataStorePresenter only has the ability to use SimulationFilter.Name and not a Zone as filters. Is this where the new functionality would be needed? I don't think I'll go down the path of making these changes as it also means adding to the DataStore.Reader.GetData parameter list. I expect this functionality was never implemented and maybe noone else has zone based reporting.

hol430:
This should be implemented by #4338. I didn't want to change the DataStore.Reader.GetData parameter list so I just included the zone filter in the filter argument.

Issue #4285, opened 2019-10-23
A box-whiskers plot style would be useful


sno036:
On the wish-list

sno036:
As discussed @hol430  - this seems to have broken all graphs. Can this be your top priority now that Jenkins is up?

hol430:
This should be fixed now that #4305 is fixed...is it still broken for you?

sno036:
@hol430  - yes - sorry - missed that message. Just pulled and working again. Thanks.

hol430:
Closing this because the box plots have been implemented.

PR #4305, opened 2019-10-28
Fixed selection of items in x variable drop down in series view

hol430:
(Also fixes broken data store table name drop down).

Working on #4285


hol430:
retest this please jenkins

PR #4311, opened 2019-10-29
Cast enum values to strings for storage in database.

### LINKED ISSUE #2038: Report enum value as string in reports
Is it possible to have a default .ToString() when reporting enums so values are reported and not left blank? I can't see there is a way to add the .ToString() in the report creation, so it probably needs to be in the code.

hol430:
Resolves #2038


lie112:
This was working but my latest pull of master has reverted back to integer reporting of enums in the dataStore. Did some changes get lost?

PR #4314, opened 2019-10-29
Merge in latest sorghum changes

hol430:
Working on #572


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
@hol353 @jbrider I think this is good to merge. The error message about a missing filename.tablename is due to a bug which is fixed by APSIMInitiative/APSIM.PerformanceTests#27.

There are some legitimate changes in the sorghum stats which is to be expected. They seem to be mostly improvements though so it should be good to go.

There's also a minor change to agpasture stats somehow. Not really sure where this came from - maybe another bug in the performance tests? Either way it's an improvement so I guess it's not a problem...

hol353:
It is up to you guys. The model isn't released yet so you (as model owners) can decide if the stats are 'good enough'.

hol430:
Well they're closer to old apsim anyway (it never lies!) so I say go for it.

jbrider:
@hol353 Yep, it's right to be merged.
There are a few minor changes still to come before final release.
@sarahcleary The minor changes to come shouldn't affect a review so it can be done now.

Issue #3698, opened 2019-04-03
Is it time to start categorising the possibilities under "Add model"


sno036:
There are now at least 272 possibilities under some "Add model" right clicks. Is some categorisation past due? (I can't find the one I think I want!)

![image](https://user-images.githubusercontent.com/12176269/55455220-56298200-563f-11e9-8973-4a997fef7746.png)

hol430:
There is a search box up the top if that helps

sno036:
Maybe ...(I didn't see the search)

hol430:
![image](https://user-images.githubusercontent.com/36427516/55528054-dce16c00-56de-11e9-8bb9-66e03c1c8546.png)

hol353:
I think some categorisation would be useful. Need to think about how we want to do that though.

hol430:
These are all C# classes, so we could categorise by the namespace in which they're declared.

hol353:
Yes that would work. I guess we need a tree view in that window instead of a list.

lie112:
I have been considering ways of removing the CLEM models unless the "add model" is selected under CLEM component as I realise these are not needed by most users and are quite numerous. I could add some code to ignore CLEM components for all but CLEM "add model". It could be that some more sophisticated means of setting valid parents and valid NOT parent may clean up the situation.

We could also add a treeview component to the side of the "add model" presenter to list all available models APSIM.Models.XXXX and "core components" that would act as a useful filter.

her123:
A treeview for categories sounds good

lie112:
I had more of a thought. There are times when you don't need or want to select from a filter. If the Parent attribute has been set well you might only have a couple options to chose from and they should simply be displayed. It could be the filtering appears when a large number of items appear or you are working with an imodel or zone. Still needs a bit of thought for best usability,

hol430:
If I go into the add model view and expand or collapse a folder by double clicking on it, it will add a model of that name if it exists. e.g. If I double click on the folder for the graph namespace it will add a graph model because the namespace and the class have the same name:

![AddModel](https://user-images.githubusercontent.com/36427516/68553449-003a7980-046d-11ea-87e4-0d092fb5b18c.gif)

PR #4324, opened 2019-10-31
Changed AddModel view to use a tree control

hol353:
Working on #3698


ApsimNG/Presenters/AddModelPresenter.cs
```diff
@@ -31,45 +37,103 @@ public class AddModelPresenter : IPresenter
...
+                AddTreeNodeIfDoesntExist(modelThatCanBeAdded, rootNode);
+
+            tree.Populate(rootNode);
+            if (models.Count() < 10)
+                tree.ExpandChildren(".Models");
+        }
+
+        private static void AddTreeNodeIfDoesntExist(Apsim.ModelDescription modelThatCanBeAdded, TreeViewNode parent)
+        {
+            var namespaceWords = modelThatCanBeAdded.modelType.FullName.Split(".".ToCharArray()).ToList();
```
hol430: Isn't this the same as

```csharp
var namespaceWords = modelThatCanBeAdded.modelType.Namespace;
```

And same story a few lines down:

```csharp
var modelName = modelThatCanBeAdded.modelType.Name;
```

ApsimNG/Views/ViewBase.cs
```diff
@@ -103,5 +150,35 @@ public ViewBase(ViewBase owner)
...
+            control.Initialise(this, builder.GetObject(controlName));
+            return control;
+        }
+
+        /// <summary>
+        /// A method used when a view is wrapping a gtk control.
+        /// </summary>
+        /// <param name="ownerView">The owning view.</param>
+        /// <param name="gtkControl">The gtk control being wrapped.</param>
+        protected virtual void Initialise(ViewBase ownerView, GLib.Object gtkControl)
```
hol430: What's the idea with this Initialise method and the GetControl method above?

Models/Core/Apsim.cs
```diff
@@ -562,5 +611,42 @@ private static Locater Locator(IModel model)
...
+            /// <summary>Constructor.</summary>
+            public ModelDescription(Type t, string name, string resourceName) 
+            { 
+                modelType = t;
+                modelName = name;
+                resourceString = resourceName; 
+            }
+
+            /// <summary>Type of model.</summary>
+            public Type modelType { get; }
```
hol430: These properties should start with capital letters since they're public

ApsimNG/Presenters/AddModelPresenter.cs
```diff
@@ -76,14 +76,13 @@ private void PopulateTree(IEnumerable<Apsim.ModelDescription> models)
...
-            var namespaceWords = modelThatCanBeAdded.modelType.FullName.Split(".".ToCharArray()).ToList();
+            var namespaceWords = modelThatCanBeAdded.modelType.Namespace.Split(".".ToCharArray()).ToList();
 
             // Remove the first namespace word ('Models')
             namespaceWords.Remove(namespaceWords.First());
 
             // Remove the last word - model name.
-            var modelName = namespaceWords.Last();
-            namespaceWords.Remove(modelName);
+            namespaceWords.Remove(namespaceWords.Last());
```
hol430: Does this work? I don't think this line is needed because the namespace variable doesn't contain the model name. e.g. if modelType is GenericOrgan, namespaceWords would be ["Models", "PMF", "Organs"]

ApsimNG/Views/ViewBase.cs
hol353: The AddModel functionality doesn't have a view. Instead contextmenu.cs has this at line 540:
'''
        public void AddModel(object sender, EventArgs e)
        {
            object model = Apsim.Get(explorerPresenter.ApsimXFile, explorerPresenter.CurrentNodePath);
            explorerPresenter.HideRightHandPanel();
            explorerPresenter.ShowInRightHandPanel(model,
                                                   "ApsimNG.Resources.Glade.AddModelView.glade",
                                                   new AddModelPresenter());
        }
'''

This creates an instance of the ViewBase base and loads a glade resource. The presenter (AddModelPresenter in this case) can query the view base to get controls e.g. line 42 of AddModelPresenter.cs:
'''
            tree = (view as ViewBase).GetControl<TreeView>("treeview");
            addModelButton = (view as ViewBase).GetControl<ButtonView>("button");
            filterEdit = (view as ViewBase).GetControl<EditView>("filterEdit");
'''

I suspect quite a lot of our view classes can be removed if we do it this way.

ApsimNG/Presenters/AddModelPresenter.cs
hol353: Yes you are correct - removed.

ApsimNG/Views/ViewBase.cs
hol430: Ah ok - makes sense. Does this assume that the presenter has to know the name of the gtk widget in the glade file? By the name I mean the string that you pass into the GetControl method.

hol353: Yes.

PR #4336, opened 2019-11-06
Added excel file type and custom column names to crop file reader in CLEM

### LINKED ISSUE #4335: Allow custom column names and excel files for crop data reader
Need to

- Allow the crop data reader to read excel files as well as prn txt files

- Let users define the column names to use rather than hard coded column names

lie112:
resolves #4335


lie112:
Is there a reason that Jenkins is stalling? Submitted yesterday afternoon.

sno036:
@lie112  - apparently "Scoop is offline" - the tests part of my pull req. is still running (more than a day) because of this. @hol353 and @hol430 are both on holiday and I don't know if anyone else knows what to do.

PR #4341, opened 2019-11-11
Link refactor

### LINKED ISSUE #2683: too picky about capitalisation (Ok in some instances but not others)
lowercase 'd' is allowed in the word 'depth' when used in the irrigation module manager script.
but a lowercase 'd' in 'depth' in the fertiliser module script will cause an error.

### LINKED ISSUE #2863: Change the different types of links to be a single [Link] attribute
Currently we have [Link], [LinkByPath], [ScopedLinkByName], [ScopedLink], [ChildLink], [ChildLinkByName], [ParentLink]. I think we need a single [Link] class with perhaps an enum argument e.g.

[Link(type=LinkType.Path, path="xyz")]
[Link(type=LinkType.Scoped, name="xyz")]
[Link(type=LinkType.Child, name="xyz")]
[Link(type=LinkType.Parent, name="xyz")]

hol430:
resolves #2863

~Do not merge - this is a work in progress commit. Just want to ensure I haven't broken anything yet.~


hol430:
This pull request removes all of the different types of links in favour of a single Link attribute. e.g.

```csharp
[Link(Type=LinkType.Scoped, ByName = false, IsOptional = false)] // default
[Link] // Equivalent to above

[Link(Type=LinkType.Path, Path = "[Phenology].ThermalTime")]
[Link(Type=LinkType.Child, ByName = true)]
[Link(Type=LinkType.Parent, IsOptional = true)]
```

The Type argument controls how the link will be resolved and is basically equivalent to the old `ScopedLink`, `LinkByPath`, `ParentLink`, etc. The `ByName` argument means that the name of the target model must match the name of the field or property to which the link attribute is applied.

If no arguments are provided, the default behaviour is a scoped link with `ByName = false`. This is exactly the same as what we had previously, meaning that an ordinary [Link] attribute from before this change will usually work exactly the same as an ordinary [Link] attribute after the change. The exception is when we are linking to an `IFunction` or `Biomass` object. Previously, an ordinary link acted as a `ChildLinkByName` when applied to a field or property of one of these types but I've removed this behaviour because (to me) it's unclear and unexpected. I guess I can revert to the old behaviour but personally I'd rather the link system be clear and explicit even if it is a bit more wordy. The upshot of this is that many `IFunction`s which previously had something like this:

```csharp
[Link]
private XYPairs XYPairs = null;
```

will now have to have something like this:

```csharp
[Link(Type = LinkType.Child, ByName = true)]
private XYPairs XYPairs = null;
```

I've applied this change to the existing code and written a converter to fix this in existing manager scripts but model developers will need to take this into account when writing new code, particularly if the new code is linking to `IFunction`s.

Issue #2731, opened 2018-07-01
New Soybean Model


hut104:
A new soybean model is to be developed within APSIM Next Generation in collaboration with Iowa State University.

sarahcleary:
@yashvirchauhan - please raise your question in relation to difference between this new model and APSIM Classic Soybean model within this issue.

yashvirchauhan:
I would like if the Next gen developers of soybean and chickpea models could enlighten me in what respect the science in the apsim nextgen soybean and chickpea  from the APSIM classic models? Thanks.

hut104:
Hi,
Firstly, note that the NG chickpea and soybean models are not yet released and so the science is not as yet settled.
Having said that, much of the basic approach is similar in that we use RUE for photosynthesis, similar approaches for thermal time, etc  However the implementations are slightly different and so many parameters have different meanings and so need to be altered. A more significant difference is in the calculation of transpiration - the new models only support penman monteith via the microclimate model (this was optional in apsim classic).  The new legume models have a much more mechanistic approach to N fixation and partitioning within the plant.  The arbitration rules are therefore also different to APSIM classic (see recent paper by Brown et al on the APSIM NG plant arbitrator).  Uptake of soil water is still via KL, and uptake of N follows the approach from APSIM classic.  The phenology models are basically the same, but extra stages have been incorporated into the model to match the VR system used in soybean etc.

yashvirchauhan:
Thanks Neil
There are a few additional issues in both chickpea and soybean models that I think that may need to be fixed for the APSIM NextGen. In chickpea I have found solution to poor prediction of various phenological events especially flowering time.  Had a brief presentation at the todays reference panel meeting on this aspect. As a new team in DAF has been formed on crop improvement  modelling, will be happy to chat with you guys to keep ourselves abreast with the latest developments and contribute to your efforts where you think we could e.g. on chickpea phenology.

sarahcleary:
@sarchontoulis - FYI

hut104:
@sarahcleary This model should now be ready for review.  Soterios and I suggest Enli Wang as a potential reviewer given his recent experience in Soybean modelling with a student in China.

sarahcleary:
@APSIMInitiative/reference-panel - please note that this model is ready for review.  @EnliWang - are you willing to review this model? @APSIMInitiative/reference-panel - anyone else that should be nominated as a reviewer?

sarahcleary:
[...]

PR #4344, opened 2019-11-11
Soybean Development

hut104:
working on #2731


hut104:
@HamishBrownPFR  and @hol353 
I fixed a bug in the senescence methods for generic organ.  Only a few models actually try to senesce generic organs.  They are now a bit different.
What do you think?  The new code looks to be doing the right thing after comparison with detailed DM an N data for senescence for each organ in Soybean.  Can we accept these minor diffs?

HamishBrownPFR:
@hut104, looks ok to me.  It has caused red clover to light up.  @rcichota does this bug fix cause any issues for the red clover model

rcichota:
I think the changes are very sensible (maybe the name of the variable could explicitly be ```senescedFraction```, to be pedantic). The changes in the results of clover only affect variables that were already not that good and are fairly minor. I see no problem in accepting them.

PR #4379, opened 2019-11-19
Adding checks for division by zero in AgPasture

### LINKED ISSUE #4117: Unprotected divide in Agpasture

rcichota:
Resolves #4117 
Not sure whether the original problem was resolved, but I found few functions with potential division by zero problem. I will fix them in a few commits.
Not to be merged yet


rcichota:
retest this please Jenkins

rcichota:
This resolves #4117 
I believe this can now be merged. The stats have deteriorated a bit, but are only because of one site. The current results for that site (low production) are a result of a bug (glfN was not properly handled in cases of extreme deficiency, causing one of the species to crash - interestingly the simulation went on). Now we have more production because both species keep growing...
It would be good for @sno036 and @mpandreucci to check, and be aware, of this fix.

Issue #4374, opened 2019-11-18
Pest/Disease plant interface


HamishBrownPFR:
To implement pest/disease impacts on crop/pasture production we need an interface for plant to provide status info to the pest/disease model and it to send status changes back to the plant.  
-  This will be the same interface that stock use for interacting with plants
- Some incomplete thinking about the components of this interface below

**Reduce population**
Pests/Diseases that kill entire plants (e.g greasy cut worm).
•	Parameter “PlantMortality” specified for each life cycle stage
•	“Plantmortality” needs to access daily Plant population from plant model and assemble functions to return how many plants it kills each day.
•	Life cycle model aggregated plant mortality from each stage, determines the extent of biomass, LAI and RLD removal.  This would be done in proportion to plant mortality
•	Cycle stages that return a positive “PlantMortality” will not need to return “BiomassRemoval” or “LAIRemoval” or “RLDReduction”
•	These stages may still return positive values for these damages if the pest/disease causes damage additional to the death of specific plants (i.e sub fatal plant damage)
**Remove biomass**
Pests/Diseases that remove biomass from one or more organ (e.g caterpillars)
•	Parameter “OrganFeedingDemand” specified for each life cycle stage 
•	“OrganFeedingDemand” determines how much tissue each cycle stage will consume from the specified organ if feeding is unconstrained.
•	Multiple “OrganFeedingDemand” functions may be specified for each lifecycle if the pest/disease stage feeds on more than one organ (struggling to think of an example here)
•	Life cycle model gets feeding demands from all stages, reduces them if feed availability from the plant is constrained and then tells the plant how much biomass is removed from each organ.
•	Removal of biomass must cause a reduction in LAI (if from leaf) or RLD (if from root).  However, it is up to the plant to translate loss of biomass into  and pest model is to work out how much area/length is reduced
**Reduce effective LAI or RLD**
Pests/Diseases that remove the effective area of organs (e.g. diseases that cause neurotic spots/ chlotitic spots or obscure radiation absorption or gas exchange or leaves or pests that block conductivity of vesicles limiting hydraulic conductivity of roots)
•	Parameter “ReduceEffectiveDimension” specified for each life stage
•	“ReduceEffectiveDimension” determines the reduction in effective LAI or RLD.  Disease works out how much it can potentially grow.  
•	LAI is not constrained to Leaf organs.  Other organs can implement Icanopy and could be influenced by a reduction in effective dimensions (eg spots on pea pods)
[...]

PR #4396, opened 2019-11-21
Damage interface

hol353:
Working on #4374


hol353:
@HamishBrownPFR: These diffs are caused by our new damage KLModifier turning on due to a call to root remove biomass with a removetype of 'cut' or 'graze'. Do we turn this function off temporarily?

HamishBrownPFR:
Hmm, yes disable for the meantime

PR #4425, opened 2019-11-27
Handle AddUrine in both SoilNitrogen and Nutrient

### LINKED ISSUE #4406: Stock sends urine to SoilNitrogen but not Nutrient
Should use the INutrient interface when talking to SoilNitrogen or nutrient. Will probably need to add a AddUrine method to the interface.

zur003:
Resolves #4406


hut104:
Is it worth thinking about whether we need the AddUrine method?  The method simply performs a calculation using the data from the caller to calculate a delta which is then passed to an Add method for a soil nutrient pool.
Perhaps a cleaner way would be to simply do the calculation in the caller, and then pass the delta to the existing add method.  ie we don't need to add a new method to Nutrient.
If one of our aims is to keep the interface to all classes as small as we can (within limits), perhaps we do something like this.  This will allow Nutrient to remain ignorant of urine, and just know about its pools.

hut104:
The FOMCNR value has also been changed to include urea.  I'm not sure that this is required.  The CNRF for FOM decomposition in soil N only included NO3 and NH4.

PR #4442, opened 2019-11-29
Soil water error increment

### LINKED ISSUE #1418: Warnings for incorrect soil properties
I had an experiment where SAT > total porosity.  It created millions of warnings and a DB file that was about 1GB.  That caused APSIM to run out of memory.
Really, the error doesn't need to be echoed endlessly, in fact, it could simply be changed to a fatal error.

hut104:
Resolves #1418


hut104:
@hol353 I think this can be merged.  The only failure is for R2 for profiles soil water and the R2 changed from 0.025739 to 0.025281.  This was caused when I had to change SAT to be below total porosity in a NZ dataset.


Predicted Observed Tests: | Filter
-- | --
VariableTestAcceptedCurrentDifferenceDifferencePercentPassedTest ▲IsImprovementProfileWaterR20.0257390.025281-0.0004581.779401%FalseFalse | ProfileWater | R2 | 0.025739 | 0.025281 | -0.000458 | 1.779401% | False | False

ProfileWater | R2 | 0.025739 | 0.025281 | -0.000458 | 1.779401% | False | False

Issue #4445, opened 2019-11-29
Sorghum crop failing in 2nd year


jbrider:
Check variable resets
Add multi year example.

PR #4446, opened 2019-11-29
fix reset of variables in SorghumLeaf and the Arbitrator

jbrider:
working on #4445


Models/Plant/Organs/SorghumLeaf.cs
```diff
@@ -558,18 +558,24 @@ public void Clear()
         
             LeafInitialised = false;
             laiEqlbLightTodayQ = new Queue<double>(10);
```
hol430: @jbrider why are we using a queue here?

jbrider:
It's calculating a moving avg, and pretty sure that's what it was in
classic.

Is there an issue with queue's?

hol430:
Ah ok makes sense. No issue - was just curious

PR #4450, opened 2019-11-29
Support for database views added

### LINKED ISSUE #4449: Support for views as well as tables in data store
The style of reports written to the DataStore by CLEM (e.g. ledgers of transactions) means that post run data synthesis (pivot style analysis) is often required for reporting and graphing. For this we use SQL statements that, to date, have been created new tables, thus duplicating the data stored and bloating the datastore size.

A much cleaner approach is to allow for the creation of database views in APSIM and expose these for display and use in graphing.

The CLEM/CustomQuery model allows for the simple creation of views.

I have written the updated code to add views to datastores and readers and it's ready to be accepted if considered a useful upgrade.

PS I couldn't find links to previous discussion about this in the issues.

lie112:
Resolves #4449


lie112:
I didn't realise how many other objects derive from IDataStore and IStorageReader
These did not show up in any of my builds or validation checks and it looked like it was all running fine.
Maybe I can just add the throw new NotImplemented exceptions for these as they wont support views.
- StorageViaSockets
-  TextStorageReader
- Storage\MockStorage

hol430:
StorageViaSockets is in ApsimX/APSIMRunner/StorageViaSockets.cs. This can throw a not implemented exception as it's only used for inter-process communication by child processes when using the experimental multi-process job runner.

MockStorage is in ApsimX/Tests/UnitTests/Storage/MockStorage.cs. This can also safely throw a not implemented exception here as it's only used for the unit tests.

TextStorageReader is in ApsimX/Tests/UnitTests/TextStorageReader.cs. I'm pretty sure this is only used in the unit tests as well - given all the other functionality which is not implemented in this class it's probably safe to throw a not implemented exception here too.

PR #4451, opened 2019-11-29
Use an enumeration rather than string for reproductive type in Stock Buy method.

### LINKED ISSUE #2902: Make the Stock.Buy event use type rather than "sex"
At present Stock.Buy requires a "sex" field that contains descriptors like "ewe", "steer" but these are converted into **types** "Empty" for any female, "Male" for entire males including cryptos and "Castrated" for castrated males.  It would make the maintenance and usability of Manager scripts much more robust if they could directly specify **type** rather than sex.

@her123

zur003:
Resolves #2902


hol353:
Can you bring this PR up to date? There are a bunch of files in it that shouldn't be showing in GitHub.

zur003:
Hi Dean,

I’ve issued a new pull request after tidying things up a bit. If it’s still a problem, let me know which files are of concern.

Cheers,
Eric

PR #4452, opened 2019-11-29
Make TapRoot below ground

### LINKED ISSUE #4421: Taproots are being defined as above ground and therefore are all being grazed
@HamishBrownPFR  - would you like to look at this?

hol353:
Resolves #4421

### DIFF
--- a/Models/Resources/Chicory.json
+++ b/Models/Resources/Chicory.json
@@ -4018,6 +4018,7 @@
               "FixedValue": 0.0,
               "Units": "g/m^2",
               "Name": "InitialWtFunction",
+              "IsAboveGround": false,
               "Children": [],
               "IncludeInDocumentation": true,
               "Enabled": true,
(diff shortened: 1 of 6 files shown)


Models/Resources/Chicory.json
```diff
@@ -4018,6 +4018,7 @@
               "FixedValue": 0.0,
               "Units": "g/m^2",
               "Name": "InitialWtFunction",
+              "IsAboveGround": false,
```
hol430: Why does this contsant have an IsAboveGround property?

hol353: It isn't a constant any more. We need it to be settable. In several crops, the taproot organ is flagged as being above ground when it is below ground in the real world.

hol353:
@HamishBrownPFR Making taproot below ground produced a small diff in the stats for fodderbeet. Are you ok if I accept the changed stats?

HamishBrownPFR:
Look like quite a big effect on only two simulations.  Better to fix this.  I think there are a few that have grazing and your change has effected the grazing rule.  I will have a closer look
image

HamishBrownPFR:
There is a folder in the fodder beet tests called grazing test that has two simulations.  The garzing rule there could be altered.  However, I think a more appropriate fix might be to not call the taproot of fodder beet below ground as it is just as much above ground

image

Models/Resources/Chicory.json
hol430: But this isn't an organ, this is a constant function...right?

PR #4456, opened 2019-12-01
Fixed faulty soils in fodderbeet validation set

hol430:
Working on #1418


hol430:
@RobZys - I've fixed some problems with the soil setups in the fodderbeet validation set, mainly air dry values being higher than crop lower limits. I've fixed these which has resulted in some minor changes to the stats. The only real difference was in Fodderbeet.SCS which had N = 2 anyway. Are you happy with these changes?

RobZys:
@hol430 - Accept those, however I will re look at the ashley dean values (which the SCS are a symptom of) as these represent a case of a very shallow soil, so lowering the bucket size has a larger effect

PR #4462, opened 2019-12-02
Fixing Euc Soils

hut104:
Working on #1418


hut104:
@hol353 I have fixed the bad LL15 values in the Australian datasets.  The Brazillian ones will need more work.
For now I think these results can be accepted. The corrected soil water data improve nearly all predictions apart from some branch weights.  I have discussed this with Phil Smethurst.

PR #4466, opened 2019-12-02
Fixed soil in soilwat unit tests.

hol430:
Working on #1418


hol430:
@hol353 these pull requests are all failing due to that old bug in the performance tests....I think you need to accept the changes to fodderbeet again.

hol353:
@hol430  Need to fix this problem. Can you rerun all PRs?

PR #4467, opened 2019-12-02
Don't show LL15 region series in soil profile graph legend.

### LINKED ISSUE #4404: LL151 in Soil Graphs
A line appears in soil water graphs in the Soil/Physical presenter.  It is Labelled "LL151" .
Should this be there?  I find nothing in the files with this name.

It appears in the fodderbeet example file but it may be in all files.

hol430:
Resolves #4404


hol430:
retest this please jenkins

PR #4469, opened 2019-12-02
Soil errors - the mission

mpandreucci:
Working on #1418 . Hopefully not forever


hut104:
@mpandreucci that which doesn't kill us can only make us stronger.

mpandreucci:
Not sure if strong is the word Drew is thinking about every time he sees these errors.

hol430:
retest this please jenkins

PR #4471, opened 2019-12-02
Pressing enter while editing a grid cell should move the cursor down …

### LINKED ISSUE #4470: Grid cell navigation should be more excel-like
In Excel, pressing enter or tab while editing a cell will take the user out of edit mode and move the cursor to the next cell down or right, respectively. We should emulate this behaviour in our grids.

hol430:
…a row. Resolves #4470


hol430:
retest this please jenkins

PR #4472, opened 2019-12-03
Allow scrolling via up/down keys in add model view.

### LINKED ISSUE #4402: Add Models Presenter does not scroll
If you go to add a model to a node, and then choose the function folder, a scroll bar appears.  However, if you use down arrow to move down the list, the selection will disappear off the bottom of the screen and the scroll bar does not move.

hol430:
Resolves #4402


hol430:
retest this please jenkins

PR #4473, opened 2019-12-03
Fix Eucalyptus

hut104:
Working on #1418


hol430:
retest this please jenkins

hut104:
@hol353 This is probably as good as we can do for now after fixing the various soils issues.  The results are mostly within tolerance, with only minor changes for a few things.

hol353:
retest this please jenkins

PR #4474, opened 2019-12-03
Allow for disabling composite factors and their model overrides.

### LINKED ISSUE #4373: Disabling a factor has no effect
If I have a factor which overrides a model, and I disable one of the children of this factor, the number of simulations generated is still the same. I'm guessing apsim will override the model with the now-disabled model, although I'm not sure if this occurs before or after we remove all disabled models from the simulations tree. For example:

image
image

hol430:
Resolves #4373


hol430:
retest this please jenkins

PR #4475, opened 2019-12-03
LessThanFunction should not cache its child functions.

### LINKED ISSUE #4358: LessThanFunction caches children between simulation runs
If I run a simulation, then reorder the children of a less than function and then re-run the simulation without closing the GUI, the less than function will still use the previous ordering of the children.

hol430:
Resolves #4358


hol430:
retest this please jenkins

PR #4476, opened 2019-12-03
Rename create documentation context menu option.

### LINKED ISSUE #4349: Create documentation menu item needs revision
The create documentation menu item in the context menu requires renaming as it is not wholly context specific to the component clicked on. It could read 'Create simulation documentation'
May be related to #4333

hol430:
Resolves #4349 

I also changed this so that the option only shows up after right clicking on the Simulations node. Don't know that it really makes sense otherwise since it's not context-sensitive.


hol430:
retest this please jenkins

PR #4486, opened 2019-12-04
Intelliparent

### LINKED ISSUE #4463: Intellisense finds wrong component
The attached sample shows intellisense being asked about component M2, but returning information about M1
[x1.zip](https://github.com/APSIMInitiative/ApsimX/files/3909298/x1.zip)

image

peter-devoil:
Resolves #4463


Tests/UnitTests/Report/ReportTests.cs
```diff
@@ -59,25 +59,64 @@ public void InitSimulations()
                                         Name = "Report",
                                         VariableNames = new string[]
                                         {
-                                            "[Clock].Today.DayOfYear as n",
-                                            "sum of [Clock].Today.DayOfYear from [Clock].StartDate to [Clock].Today as TriangularNumbers",
-                                            "sum of [Clock].Today.DayOfYear from [Clock].StartOfWeek to [Clock].EndOfWeek as test",
-                                            "[Clock].Today.Year as Year",
-                                            "sum of [Clock].Today.DayOfYear from 1-Jan to 31-Dec as SigmaDay",
-                                            "sum of [Clock].Today.DayOfYear from 1-Jan to 9-Jan as HardCoded"
+                                            "//[Clock].Today.DayOfYear as n",
```
hol430: Doesn't this break some of the other unit tests if you comment out these reporting variables?

peter-devoil:
@hol353 / @hol430 - fixing this is beyond me today. There's a unit test that demonstrates the problem - the wrong numbers are being reported. Drew suspects the DB retrieval of the manager component is wrong; I tracked it as far as the locator to find a manager component with a script different from its children.

Tests/UnitTests/Report/ReportTests.cs
peter-devoil: oops. forgot debug code..

hol430:
The problem occurs when we clone the simulation before running it. The assembly resolve event is not firing for each different manager script - it only fires once, for M1 (the first script). If I run a simulation which has multiple, very different manager scripts, this event gets fired once for each manager.

This seems to imply that the runtime is able to resolve the M2/M3 manager script assemblies without needing our assistance. My theory is that it's mistakenly using the same assembly that we pointed it to for the M1 script. No idea why it's doing this - but these assemblies are extremely similar.

hol353:
Like the title of the PR. Is there any such thing as an intelli parent?

zur003:
@hol353 - Perhaps you should ask @hol430 about the existence of intelli parents.

hol430:
Some questions are better left unanswered 👍

hol430:
retest this please jenkins

hol430:
@peter-devoil I think I've fixed bug 2 in #4463 (the one where the wrong manager properties are stored by report). However this branch is conflicted and I'm a little scared to merge upstream. git seems to think the entire ReportTests.cs file is different, even the parts that haven't changed. Maybe a line endings problem...

peter-devoil:
Yes, it must be about time to do another merge.

hol430:
Duplicate of #

hol430:
retest this please jenkins

hol430:
retest this please jenkins

PR #4489, opened 2019-12-04
Improve the GenotypeNamesDefined() function

### LINKED ISSUE #4426: Stock to expose the genotype names as a public property
Expose a list of standard genotype names with the custom ones appended to it so that Manager scripts can access them.

her123:
resolves #4426

Fixed function. Improved function by using already created StockList parameter set.


hol353:
retest this please jenkins

Issue #4495, opened 2019-12-09
Add all other AgPasture species.


hol353:
Species:
AGPBrownTop
AGPCocksfoot
AGPKikuyu
AGPLucerne
AGPPaspalum
AGPPhalaris
AGPRedClover
AGPRhodes
AGPTallFescue

PR #4496, opened 2019-12-09
Added extra AgPasture species

hol353:
Working on #4495


ApsimNG/Presenters/AddModelPresenter.cs
```diff
@@ -119,15 +119,17 @@ public void Detach()
         /// <param name="e">Event arguments</param>
         private void OnAddButtonClicked(object sender, EventArgs e)
         {
-            Type modelType = typeof(IModel).Assembly.GetType(tree.SelectedNode);
```
hol430: @hol353 what is the background on this change? This appears to just revert my change in 0bce914 and re-introduces a bug described [here](https://github.com/APSIMInitiative/ApsimX/issues/3698#issuecomment-552251467). I'm happy to re-apply my change (and create a test so this doesn't get changed back again) but thought I'd better check first and make sure this isn't a deliberate change.

hol353: I had to revert it as it wouldn't allow the user to add an instantiation of PMF or PastureSpecies e.g. wheat. At least that is what I think it was.

hol430: Ok, thanks. I've fixed and added tests for both of these use cases in #4758.

PR #4499, opened 2019-12-09
Remember previous location of split-screen divider.

### LINKED ISSUE #4316: split screen size when re-clicking
When clicking the split-screen option the default is half original simulation and half the new (split) simulation. I usually make the right (new) split screen to about 75% of the screen for readability, copy if I need anything and then close the split-screen to make the original full size. When re-clicking the split-screen the screen goes back to 50:50. Is it possible to make the size of the split screen when re-clicking the size of when it was last closed - otherwise have to keep re-sizing. I hope that makes sense!

hol430:
Resolves #4316

### DIFF
--- a/ApsimNG/Interfaces/IMainView.cs
+++ b/ApsimNG/Interfaces/IMainView.cs
@@ -60,6 +60,10 @@ namespace UserInterface.Interfaces
         /// </summary>
         bool SplitWindowOn { get; set; }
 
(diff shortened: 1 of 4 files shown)


jbrider:
Will this work when re-opening Apsim, or just when you toggle the option.

hol430:
It will be persistent across restarts of apsim

Issue #4167, opened 2019-09-06
Can models be saved backward-compatible?


jal-frezie:
I'm working on ApsimX models and using the latest sources on Git. If I try to open a model with the latest pre-built version, it says I cannot because they were saved with a newer version. So, is there any way I can make usable versions to send to colleagues who are using pre-built ApsimX?

lie112:
This has started to catch me out when my CLEM users are not up to date with APSIM and I send them example apsimx files. Is this to avoid known changes that cause errors, or just version control of the simulation files?

jbrider:
There is an iterative conversion process that runs when a file is opened.
If the file is version 60 - and the version you're running is 66, it will
convert it to 61, then to 62, then to 63 and so on.
It will then record the version in the file as 66.

Unfortunately it does't work in reverse.
This process is contained within converter.cs.

People using an earlier version will not be able to open your files.
You would have to make changes to the raw json to be able to open them in
an older version - and it won't be a straight forward process in most
instances - most likely it will cause errors - and they may not be
immediately visible ones.
It would be easier to have them use the most recent version.

rcichota:
Just want to add that we have experienced this issue a lot too. The lack of 'proper' versions and the nature of continuous improvements of APSIM can be quite a nightmare to navigate. Here we decided to created branches to freeze 'versions' of the model for use in specific projects and if all people involved are aware, it works fine. This isn't good when working with people that use a release version...

hol430:
I think the real problem here is that there have been no new releases/pre-built version of apsim for almost 3 weeks now due to a bug in the continuous integration system, which means the compiled code will be several version numbers ahead of the released/pre-built version. I'm hoping that #4168 will fix this. Once fixed, if you do a `git pull` whenever your colleagues upgrade apsim, then your versions should generally be compatible in terms of sharing .apsimx files.

hol430:
Closing this as it's been answered and fixed. The latest pre-built version should be able to open files created with the latest compiled version.

Feel free to reopen or open a new issue if this isn't working.

lie112:
Would it be possible to try and deserialise the aspimx file before reporting the version error? If the file opens fine, there is no reason to report the error and the file can be assumed to be compatible. I believe this is the only reason to have the version checking. We could report a non-critical warning about the version issue and suggest they upgrade.

[...]

PR #4506, opened 2019-12-09
Move custom build check to upgrade button in popup rather than menu

lie112:
working on #4167 

There are a couple problems I can't solve

1. I don't understand the GTK# render order. I can't change the settings before the widgets render. Thus there is a pause with the rego setup while the upgrades list is generated before it updates with the the custom build warning and disabled button. This is the same issue with the HTML view #3934

2. All sorts of strange things happen if a simulation is open when you then try to upgrade.

- I used to get an error on close of the upgrade window, but this has gone away
- The open simulation disappears from the tab
- When you try open another or same simulation get a null ref error line 797 of MainPresenter.cs as the APSIMXFile is null

### DIFF
--- a/ApsimNG/Presenters/MainPresenter.cs
+++ b/ApsimNG/Presenters/MainPresenter.cs
@@ -1213,13 +1213,10 @@
             {
                 ShowError("You are on a custom build. You cannot upgrade.");
             }
-            else
+            if (this.AllowClose())
             {
-                if (this.AllowClose())
-                {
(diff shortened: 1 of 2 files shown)


zur003:
While you're working on this, maybe you can also fix the problem with doing upgrades on Linux, where (at least the last time I tried), the original instance of ApsimX terminates with a crash and stack dump, rather than closing gracefully.

PR #4514, opened 2019-12-11
Sorghum/dlt lai

jbrider:
Working on #572


jbrider:
retest this please jenkins

jbrider:
these are ok to be accepted @hol353 
Not sure if we're supposed to click the accept stats as well, or if that's yours to do as well.

Issue #4521, opened 2019-12-11
Port the maize model from Apsim 7.10 into Next Gen


hol430:
Unsure if this is meant to replace or coexist with the current maize model. Am working with Jason Brider and Greg McLean to make this happen.

HamishBrownPFR:
Would be good to clarify the intent here please.

jbrider:
@Hamish Brown <Hamish.Brown@plantandfood.co.nz> We're bringing the old
maize model across as a new entity in the first instance as it is based
very closely on the sorghum code.
Following this will be the new DCaPS photosynthesis model, a limited
transpiration model, and a grain abortion model that were all developed on
this code base.

The goal is to merge all of this into 1 maize model eventually, but I don't
really understand the differences in the science quite well enough to know
how easy or hard this may be at this point.
However that is managed will require the old model to be brought into the
new framework in order to utilise the testing system to know what effect
changes will have - the system used in ApsimClassic is not effective.
The easiest way will to do that will be to put this work in the UnderReview
folder as all changes are tested on the models there.
The new model is now giving the same results as the old one - Drew has got
that working really quickly.

HamishBrownPFR:
@jbrider thanks for the clarification here.  This is a good step toward harmonizing the different approaches.

PR #4522, opened 2019-12-11
Port the maize model from Apsim 7.10 into Next Gen

hol430:
Working on #4521

@jbrider - at this point the model is in basically the same state as sorghum. The code (and model setup in the PMF) is ugly as hell but gives basically the same results as old apsim.


hol430:
@hol353 these differences in the sorghum stats all look pretty minor to me and are to be expected due to some small changes to the sorghum code. This could probably be merged if @jbrider is also ok with the changes.

jbrider:
These can be accepted.
Will accept these and re-submit the other pull request I pushed earlier.
It will have more significant changes that should also be right to be accepted.

Issue #4423, opened 2019-11-27
Possible errors with scoping


hut104:
In resolving #3554, several issues were identified.
One comes from the use of a variable reference function in the inflorescence organ in White Clover that seemed to resolve the link by grabbing a function of the same name in another organ. (see #4412)
There are a few ways we could think about addressing this.
1) We could stop allowing links to resolve without some sort of parent info (e.g. don't allow "[NNN]" but only something with "[PPP].NNN").  This would make things less likely, but is not foolproof.
2) Try adding [ScopedModel] to organs to stop linking across organs unless explicitly told to do so.
3) VariableReference function uses locator.get to find the function.  We could get the locator to check that it only finds one matching function in scope and throw if it finds >1 - this would require developers to be explicit where scoping is inadequate to make it clear.
4) A similar issue may arise with other functions that use links. (e.g. linearinterp).  Perhaps linearinterp could be changed to have XProperty as a child (which could be a VariableReference) and then the fix above is captured into that function as well.
5) We need to consider how to resolve this for expression functions as well.

hut104:
@hol353 @hol430 @HamishBrownPFR @rcichota 
We need to be thinking about how we resolve this.

HamishBrownPFR:
Explicit is always good to be sure we know what is going on.  The only thing that can be safely assumed is the scope should be within the existing crop model.  Hmm, actually this is not even always the case as we use soil and climate variables also.  
Option 3 seems like a good bet??  If we implement something to solve this for variable reference function, we could then use variable reference functions where ever a variable is addressed.  For example, instead of Linint having XProperty as a property, it would need to have a variable reference child function called XProperty.  Equations in Expression function would only be able to reference children and would need to have a child variable reference function for each external reference the expression uses.

hol430:
Another problem with the scoping rules is that any model which is not a sibling of a plant will not receive any events from the plant or any model inside the plant. E.g. if I have a management folder and I place a fertilise on sowing rule inside the management folder then the fertilise script will not receive the sowing event from the plant if the plant is not also inside the folder.

I suspect the reverse will also be true - if I have several plants in a single paddock I might decide to place them in a plant folder. Bad idea - the scoping rules will hide their events to other models which are not also in the plant folder.

hut104:
Can we just get rid of management folders? They seem superfluous.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

HamishBrownPFR:
[...]

PR #4538, opened 2019-12-15
Tidy up variable referencing

HamishBrownPFR:
Working on #4423.  Make the x property for Linint an I function and replace current XProperties with VariableReference funciton.  Aim is to reduce the amount of code where locator is used to make it easier to tidy up scoping

### DIFF
--- a/Models/Core/ApsimFile/Converter.cs
+++ b/Models/Core/ApsimFile/Converter.cs
@@ -16,7 +16,7 @@
     public class Converter
     {
         /// <summary>Gets the latest .apsimx file format version.</summary>
-        public static int LatestVersion { get { return 70; } }
+        public static int LatestVersion { get { return 71; } }
 
         /// <summary>Converts a .apsimx string to the latest version.</summary>
         /// <param name="st">XML or JSON string to convert.</param>
@@ -1382,6 +1382,21 @@
                     genotypes[i]["Conceptions"] = genotypes[i]["Conception"];
                     genotypes[i]["GenotypeName"] = genotypes[i]["Name"];
                 }
+            }
+        }
+
+        /// <summary>
+        /// Alters all existing linint functions to have a child variable reference IFunction called XValue instead of a
+        /// string property called XProperty that IFunction then had to locate
+        /// </summary>
+        /// <param name="root">The root JSON token.</param>
+        /// <param name="fileName">The name of the apsimx file.</param>
+        private static void UpgradeToVersion71(JObject root, string fileName)
+        {
+            foreach (JObject linint in JsonUtilities.ChildrenRecursively(root, "LinearInterpolationFunction"))
+            {
+                JsonUtilities.AddVariableReference(linint,"XValue", linint["XProperty"].ToString());
+                linint.Remove("XProperty");
             }
         }
 
--- a/Models/Core/ApsimFile/JsonUtilities.cs
+++ b/Models/Core/ApsimFile/JsonUtilities.cs
@@ -348,6 +348,31 @@ namespace Models.Core.ApsimFile
         }
 
         /// <summary>
+        /// Add a variable reference function to the specified JSON model token.
+        /// </summary>
+        /// <param name="modelToken">The APSIM model token.</param>
+        /// <param name="name">The name of the variable reference function</param>
+        /// <param name="Path">The variable reference path</param>
+        public static void AddVariableReference(JObject modelToken, string name, string Path)
+        {
+            if (ChildWithName(modelToken, name) == null)
+            {
+                JArray children = modelToken["Children"] as JArray;
+                if (children == null)
+                {
+                    children = new JArray();
+                    modelToken["Children"] = children;
+                }
+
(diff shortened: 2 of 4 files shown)


Models/Core/ApsimFile/Converter.cs
```diff
@@ -1385,6 +1385,21 @@ private static void UpgradeToVersion70(JObject root, string fileName)
...
+        /// Alters all existing linint functions to have a child variable reference IFunction called XValue instead of a
+        /// string property called XProperty that IFunction then had to locate
+        /// </summary>
+        /// <param name="root">The root JSON token.</param>
+        /// <param name="fileName">The name of the apsimx file.</param>
+        private static void UpgradeToVersion71(JObject root, string fileName)
+        {
+            foreach (JObject linint in JsonUtilities.ChildrenRecursively(root, "LinearInterpolationFunction"))
+            {
+                JsonUtilities.AddVariableReference(linint,"XValue", linint["XProperty"].ToString());
```
hol430: Could probably remove this AddVariableReference function altogether and just do something like this:

```suggestion
                VariableReference variable = new VariableReference();
                variable.Name = "XValue";
                variable.VariableName = linint["XProperty"].ToString();
                JsonUtilities.AddModel(linint, variable);
```

HamishBrownPFR:
@hol353 @hol430 .  Hi guys.  I am testing some stuff here but have a unit test failure that I don't know how to fix.  Any pointers appreciated

hol430:
The key line is about halfway down:

> Resource 'Models.Resources.AGPRyegrass.json' is not up to date - version is 70 but latest version is 71.

There's a unit test to ensure that all of the model resources (wheat.json, maize.json...) are kept up to date in terms of file version. You've added a new converter version so you'll need to update all of the resources to the latest version. You should be able to use the file converter tool in the GUI to do this I think.

HamishBrownPFR:
OK, thanks Drew.

Models/Core/ApsimFile/Converter.cs
HamishBrownPFR: I tried doing something like that for a start but could not get it working.  Don't really understand the converter well enough.  I will  try what you have suggested.

HamishBrownPFR:
@hol353 @hol430, is there a problem with Jenkins?

hol430:
There was, but I think it's fixed. It's running this pull request now.

HamishBrownPFR:
@hol430 @hol353, I cant work out what this error message means, can you help me interpret it??

Models/Core/ApsimFile/Converter.cs
```diff
@@ -1383,6 +1384,48 @@ private static void UpgradeToVersion70(JObject root, string fileName)
...
+        {
+            foreach (var varRef in JsonUtilities.ChildrenRecursively(root, "VariableReference"))
+                varRef["VariableName"] = varRef["VariableName"].ToString().Replace(".Value()", "");
+
+            foreach (var report in JsonUtilities.ChildrenOfType(root, "Report"))
+                JsonUtilities.SearchReplaceReportVariableNames(report, ".Value()", "");
+
+            foreach (var graph in JsonUtilities.ChildrenOfType(root, "Series"))
+            {
+                graph["XFieldName"] = graph["XFieldName"].ToString().Replace(".Value()", "");
```
hol430: I think this might be the problem. There are probably some graphs out there (e.g. in standard toolbox) whose XFieldName is set to null. You could try something like this:

```suggestion
                if (graph["XFieldName"] != null)
                    graph["XFieldName"] = graph["XFieldName"].ToString().Replace(".Value()", "");
```

You'd probably want to add the same check for the y field name as well.

HamishBrownPFR:
@hol353 this could be merged now.  The failure in the performance tests is because some variables have been renamed (.Value() removed from them).  

I am a bit worried about the consequence of a change the converter is making.  It is replacing null for x2 and y2 properties with blank text (see diff snip below).  Not sure if this is a problem or not and not sure how to fix it
image

Models/Resources/AGPRyegrass.json
```diff
@@ -12,6 +12,9 @@
       "R50": 200.0,
       "SpeciesFamily": 0,
       "PhotosyntheticPathway": 0,
+      "InitialShootDM": 1500.0,
+      "InitialRootDM": 450.0,
+      "InitialRootDepth": 750.0,
       "ReferencePhotosyntheticRate": 1.0,
```
hol353: Can't have InitialShootDM, InitialRootDM and InitialRootDepth in the AgPasture resources. I'm guessing the conversion tool is putting them in. Can you manually remove the 3 lines from the resources file. Perhaps @hol430 can look into why they are being added.

hol353: Actually I'll solve this another way. Leave it as is and I'll look into it.

HamishBrownPFR: Why does this not throw an error?

HamishBrownPFR:
@hol353, Good to go now?

PR #4547, opened 2019-12-17
Fix several reporting defects

### LINKED ISSUE #4520: Problems with report
In the attached simulation there are multiple problems with the report model.

1. In 'report1', the two WaterUptake variables don't produce the correct results. There should be a value of uptake for each layer for two zones e.g. columns should be:
[Eucalyptus].FineRoot.Zones[1].WaterUptake[1]
[Eucalyptus].FineRoot.Zones[1].WaterUptake[2]
...
[Eucalyptus].FineRoot.Zones[2].WaterUptake[1]
[Eucalyptus].FineRoot.Zones[2].WaterUptake[2]
...

2. In 'report2', the columns output should be same as above.
3. In 'report3', an exception should occur. Array indexing in report should be 1 but when I access index 0 it produces output instead of throwing an exception.
4. In 'reporttop', the commented variable throws an exception when it shouldn't.
5. The 'CumulativeGraph' is blank. I have ticked 'Cumulative' beside the Y variables in both series. When I do that the graph disappears
[Test.zip](https://github.com/APSIMInitiative/ApsimX/files/3948524/Test.zip)
.

hol353:
Resolves #4520

### DIFF
--- a/Tests/UnitTests/Report/ReportDataTests.cs
+++ b/Tests/UnitTests/Report/ReportDataTests.cs
@@ -40,26 +40,96 @@
             "2017-01-02,2.000,   2, def\r\n");
         }
 
-        /// <summary>Convert an array to a table.</summary>
+        /// <summary>Convert whole arrays to a table with correct headings.</summary>
         [Test]
-        public void ArraysToTable()
+        public void WholeArraysToTable()
         {
             var data = new ReportData()
             {
                 CheckpointName = "Current",
                 SimulationName = "Sim1",
                 TableName = "Report",
-                ColumnNames = new string[] { "Col1", "Col2" },
-                ColumnUnits = new string[] {   "mm",    "g" }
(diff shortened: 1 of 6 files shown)


hol430:
retest this please jenkins

Tests/UnitTests/Report/ReportDataTests.cs
```diff
@@ -62,29 +62,6 @@ public void WholeArraysToTable()
             "     5,     6,                  8.000,                  9.000,     7\r\n");
         }
 
-        /// <summary>Convert partial arrays to a table with correct headings.</summary>
-        [Test]
```
hol430: Why delete this tests?

hol353: I didn't. I renamed the test and added extra ones.

PR #4549, opened 2019-12-17
Fixed crash in manager when reformatting code which doesn't compile.

### LINKED ISSUE #4548: Manager hard crash
If you change a manager script so that it doesn't compile and then right click and do a 'reformat' it will crash the GUI.

hol430:
Resolves #4548


hol430:
retest this please jenkins

PR #4564, opened 2019-12-20
Fixed many bugs in sorghum model which were screwing up multi-year runs

hol430:
Working on #572


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
@jbrider there are some legit changes to the sorghum stats here which seem to be caused by adding surface OM to all of our sims. If I check out the master branch's version of the .out files and delete surface OM from the paddocks, the results basically match those from old apsim - implying that the other changes here don't have a big effect (on single-season sims).

jbrider:
@Holzworth, Drew (A&F, Toowoomba) <drew.holzworth@csiro.au>
Does that mean that the Surface OM was being ignored before and moving it
to the correct place has a small effect?
Do the New vs Old comparisons should still match with these changes?

hol430:
Yes to both - new and old still give the same results. I had to add surface OM into the old apsim simulations as well which has changed the results in old apsim. Will need to create a pull request for that change.

jbrider:
I guess the question is if it wasn't in the old ones, do we need to have it
present in the new ones?
Probably adding it to both is the right way to go. I assume it doesn't
change it by very much?

hol430:
> do we need to have it present in the new ones?

I don't think it was doing anything when at the god-level(?). Neil indicated that we wouldn't be able to release the model without it anyway. Certainly no other model has surface OM at the god-level. Moving it from the simulation-level into the paddock had a noticeable impact when comparing it to both the observed results and old apsim results. Marrying up the differences vs old apsim was as simple as adding surface OM nodes into the old apsim simulations.

hol430:
Ok I think this should be ok to merge then. The changes in stats are smaller after reverting my change to the maturing TT target.

PR #4575, opened 2020-01-08
Miscellaneous small fixes to get the performance tests working again

### LINKED ISSUE #4574: "Annual" table missing from AgPasture validation set
If you look in the agpasture datastore, you'll notice that the Annual table is missing, despite there being a PredictedObserved model called Annual.

### LINKED ISSUE #4576: Problems with performance tests
There are currently a lot of missing records in the PredictedObservedValues table in the performance tests database. These missing records cause graphs to fail to appear on certain pages on the portal. The missing data seems to be due to minor discrepancies in the tables in the validation datasets and/or related to bugs in the performance tests code.

The errors are all written out to a log file which no one checks, but I've changed this so that they will now cause jenkins to fail. Jenkins is now running red for all pull requests.

hol430:
Resolves #4574
Resolves #4576

@sno036 @mpandreucci fyi - I've changed the simulation names in the ObservedAnnual sheet of the agpasture observed file (ObservedAll.xlsx).

The problem is that the simulation names in apsim are called "RadcliffeLocationxxxx" but in the observed data they're called "RadcliffLocationxxxx". I've changed the simulation names in the .xlsx file and the Annual predicted/observed table is now showing up.


mpandreucci:
Hi Drew.
A while ago I went through some of the observed spreadsheets to fix some typos but I missed those ones. Sorry about that.
Mariana

Get Outlook for iOS<https://aka.ms/o0ukef>

hol430:
@HamishBrownPFR I'm not sure if you're in charge of the potato model so feel free to pass this along to someone else.

Just wanted to let you know that I've fixed a minor typo in the potato observed file - the variable name (PlantandHarvest.Script.FinalTag) used to match the FinalYieldData predicted/observed tables was different (in case) between the observed file and apsim. This was causing problems for the performance tests.

See my commit message (for 2dd4859) for more details.

hol430:
@sno036 I've had to make a small change to SoilNitrogenPatchValidationOfPatchiness.apsimx (what a name!). One of the factors (SoilOM) has two Organic children which were originally called "On" and "Off" respectively, but were recently renamed to both be called "Organic" due to a bug in the converter (see 96b6b4accbfaa55a17271510d2a9dba5e0f72275) which I think has since been fixed. The end result was that the PredictedObserved node was failing to merge the predicted and observed tables because the simulation names were different between apsim and the observed file.

@hol353 this highlights a flaw in the tests system - why did the tests not pick up on this breakage? The old Tests node (the one with with the insect icon) is no longer checked by Jenkins (I think) and the performance tests only check the Tests/Validation and Tests/UnderReview directories, so this file is not really being tested at all except to ensure that it runs without errors.

hol430:
@hol353 I think the stats are showing changes because there are two new tables which were previously not being tracked (the tables previously didn't exist). I think this should be safe to merge - no other pull request will run green until this is merged anyway.

Issue #4565, opened 2019-12-20
SOBOL Example Improvement


hut104:
The current SOBOL example is incomplete and not consistant with the Morris example. It has no graphs to show results.
The SOBOL example should be changed to allow direct comparison between methods.

hol430:
@hut104 it looks like the sobol example now generates ~11k simulations and takes about an hour to run on jenkins. We need to find some way to trim down the number of simulations.

hut104:
I’ll reduce the execution time.  However, I might get you to help fix a bug in the  UI that makes it hard to alter the experimental design without doubling the size of the table!

hol430:
@hut104 can this be closed?

hol430:
I'll take that as a yes.

PR #4589, opened 2020-01-13
Fixed a bug where morris/sobol parameter array never gets emptied.

hol430:
Working on #4565


hol430:
@hol353 if you're happy with this change can you merge it please? The sobol/morris UI is unusable until this is merged and we need to make changes to the sobol example so we can get the jenkins runtime back down.

PR #4618, opened 2020-01-21
Agro forestry

hut104:
working on #56


hut104:
Trying to move Agroforestry tests and Examples into appropriate folders in preparation for release - which needs to happen prior to the APSIM Symposium.

hut104:
@hol353 I suggest accepting the stats from these tests.  This commit does not affect any stats on any test outside of the agroforestry model.  The stats for soil water have 3 stars (ie very good).  Yield may need improving, but n=7 only.  Phil Smethurst might try to add some improvements at some time.
For now, initial results appear to be good enough.

PR #4630, opened 2020-01-22
Throw exception if soil cannot find child models.

### LINKED ISSUE #4602: Soils downloaded in NextGen don't come with solute nodes
Can we please change this so that when the soil is downloaded, the solute nodes are already included in it?

hol430:
Resolves #4602


hol430:
@hol353 in the SoilStandardiserTests, we create a soil without a SoilWater or WEIRDO child. Is this a valid thing to do?

hol353:
No.

PR #4653, opened 2020-01-29
Don't ignore .csv files.

### LINKED ISSUE #4613: should csv file types be in the gitignore file?
Not sure why we are ignoring csv files from git? 
I asked Drew and he does not know either. 

Should this line be removed from the gitignore file? (Perhaps others should go also?)

Makes it hard to commit test datasets using graphical tools such as gitkraken. 
Have to use the command line instead.

hol430:
Resolves #4613


hol430:
retest this please jenkins

Issue #4634, opened 2020-01-23
Add another option to Simple Grazing model for grazing on a expression criteria


hol353:

sno036:
The no grazing window is not working. Working on it.

PR #4660, opened 2020-01-30
Fixes to SimpleGrazing

mpandreucci:
Working on #4634
Several fixes to the model.
The operations component doesn't need to come after the SimpleGrazing in the tree anymore (the order doesn't matter).
Fixed the bug for when SimpleGrazing generates a "manager". This was confusing one of the graphs name with the name of a parameter.
Deleted the option of TargetMassLength. Now lengths thresholds for rotations can be set for both TargetMass and FlexibleGrazing.


hol353:
retest this please jenkins

PR #4663, opened 2020-01-30
Improves the performance of CLEM runs with large herds. 

### LINKED ISSUE #4662: Time to optimise performance for large herds in CLEM
Find and improve CPU bottlenecks in CLEM code. This is especially needed for large herds.

### LINKED ISSUE #4665: Need to update to .Net 4.7.2
This need relates to issue #4662 and the errors occurring in pull request #4663 
Can we have the reference assemblies for Framework 4.7.2 provided for Jenkins?
After discussions with @zur003 and @hol430 there should be no reason that APSIM next gen can't be using the latest of the 4.7 .Net release and coding principles.
I want to use private protected in one of my models base classes.

lie112:
Resolves #4662 Resolves #4665


hol430:
retest this please jenkins

lie112:
Arrghh. C#7 is .Net Core. 
I'll get a work around for now as .Net core still needs work for UI right?

zur003:
The GUI needs a LOT of work before it will run under .Net Core; on the other hand, the non-GUI components build readily.

lie112:
I don't think we're ready for 7.0 yet. I assume the current GUI has a fit if the rest of the model is in .Net Core, and so we're not ready to move Models to c# 7.0 as I would need.

lie112:
retest this please jenkins

hol353:
- What about the APSIM installer? Doesn't look like it has been changed to check and install the newer version of .NET.
- PR's should only resolve 1 issue (unless they are related). Changing .NET versions should really be in it's own PR.

zur003:
Are there any implications for installations using Mono (Linux and OS X)?

lie112:
Sorry, this all started because Visual Studio 2019 assumes you can use the latest c# language of the installed .Net framework and I started using private protected in this merge. It was suggested that all the projects could be updated to .Net 4.7.2 but then I realised that I can't use c# 7.2 in the visual studio 2015 Jenkins runs under, so I have reverted to using the protected statement. I did create a second issue regarding 4.7.2 and checked with @zur003 and @hol430   I can go back through all the

lie112:
What should I do to resolve this issue. Revert all .Net changes back to where they were? I can't test if any implication on iOS for Linux. It would be nice if APSIMx was always at the best version of .Net and c# language possible.

hol353:
You don't have a choice. You'll have to revert all the .NET changes back. If I merge this then it will likely not run on some users computers because the APSIM installer won't install the newer .NET. This is a good example of why you shouldn't resolve two different issues in the same PR. Keep the PRs simple!

zur003:
I just had a quick look into the "will it work on Linux and OS X" issue. I haven't actually tested it, but the Mono docs claim support for C#7 starting with Mono 5.18.0. This means we would need to do a bit of tinkering with the installers for Linux and Mac, as we currently specify a Mono version of 4.0 for Linux and 4.4 for Mac. Ubuntu could be a bit tricky, and the "standard" mono packages for the older long-term Ubuntu distributions are behind the times (Mono 4.2.1 for Ubuntu 16.04 and 4.6.2 for Ubuntu 18.04). I mention this now as a reminder that this will need updating as part of an upgrade of the .Net Framework version.

lie112:
that's fine. I'll get this fixed up after my workshop on Thursday.

hol430:
@zur003 which version of msbuild are you using on your Ubuntu installation? I'm getting some unused variable errors with msbuild 14.1.0.0 (which I guess are worth fixing anyway) but I'm guessing you're not seeing these errors?

hol430:
Sorry Eric - ignore my previous comment. I somehow managed to replace my msbuild with an older version. Reinstalling the latest version of the package has fixed the problem.

zur003:
@hol430 - On the rare occasions when I build on Ubuntu, I use MonoDevelop and don't really pay attention to what, if anything, that's using in the background. However, I have had problems when building APSIMClassic from the command line.

hol430:
> I have had problems when building APSIMClassic

Haven't we all!

lie112:
Anyone know how to fix the current GIT error not being able to checkout the branch?

hol430:
retest this please jenkins

lie112:
Anything else I need to get this pull request merged?

Tests/UnitTests/Properties/Resources.Designer.cs
```diff
@@ -19,7 +19,7 @@ namespace UnitTests.Properties {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
     [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
```
hol353: Why has this changed?

Models/Importer/Importer.csproj
```diff
@@ -11,7 +11,7 @@
     <RootNamespace>Importer</RootNamespace>
     <AssemblyName>Importer</AssemblyName>
     <FileAlignment>512</FileAlignment>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
     <TargetFrameworkProfile />
```
hol353: This doesn't look right.

lie112:
Can someone explain how I resolve rejected hunks and remove the rej files?
The last changes requested by @hol353 can't be reverted in SourceTree
It might be because Importer is a project nested inside the Models project
Any ideas?

hol430:
To remove the .rej files - delete them from your pc and they should show up as changed in source tree. You should be able to just add them and commit.

As for reverting the .net changes - my guess would be the reason you can't revert is that you've  partially reverted the changes in 893b15f, and trying to do an actual git revert will cause conflicts. You might be able to reset to before 893b15f and then try a git revert.

lie112:
I have stuffed up as my .Net change commits accidentally when to my CLEM branch and not the new .Net change branch I created so they went into the huge number of changes that were building up waiting for the merge that crashed because of the private protected issue with C# 7.2. I can't revert back and lose these changes.

lie112:
Maybe rather than revert the changes I should have manually gone through and changed all the .Net frameworks back to what they were but couldn't remember which one was 4.5 and which were 4.6

lie112:
Can I simply revert my workingCLEM branch back one commit to before I tried to revert to 4.6 and make the changes manually. Then pull CLEM again. Or am I in trouble. I can't lose all the changes in the commits.

lie112:
I still don't fully understand git

hol430:
Yes - use git reset to step back one commit, then revert the two commits (0e7e0ea and 4087990) which change .net framework versions. If you're worried about accidentally overwriting some of your actual changes, you can checkout a new branch to test the changes

```bash
git checkout -b testbranch
git reset --hard a752c11e2d8b0ab1786eb1cd17f7ff1c229ecf6e
git revert 0e7e0ea
git revert 4087990
```

lie112:
Can I do this in SourceTree? this is the only interface I have learned

hol430:
I'm sure it's possible, but I don't use source tree so I'm not sure of which exact buttons to click. I think you can right click on individual commits to do a reset or a revert(?). Maybe someone else can advise here.

If you checkout a new branch then you can go nuts without worrying about damaging this branch. Then once you're happy, you can merge back into this branch or even delete this branch in favour of the clean branch.

zur003:
Yeah, definitely check out a new branch. And maybe also make a backup copy of everything (the old-fashioned way) if you're as paranoid as I am.
I haven't worked out the complete mapping between SourceTree and command-line git, so I'm not going to attempt to give full instructions. Right-clicking on a commit in SourceTree gives you lots of options, but I'm still a bit unclear about rebase vs. reset vs. reverse, and have fouled up more than once.
Good luck.

lie112:
Am I going about this correctly? Can you delete commits in the history.
I'd like to delete my two commits that take me to 4.7.2 and one that takes me back to 4.6 so that the changes for c# 7.2 were never made. Possible?

hol430:
As far as I know, you can't delete a particular commit. What you can do is "reset" your branch. This effectively takes your branch back in time. So if you reset to a752c11, it will be as if you never made your most recent commit (the partial manual reversion of the .net framework changes).

You can also revert a commit - which doesn't delete the commit, but instead creates a new commit which is basically the inverse of the commit you're reverting - it will undo all of the changes in a particular commit. So after resetting to a752c11, you should be able to revert the two commits where you modified .net framework versions (0e7e0ea and 4087990).

lie112:
Thanks,
I tried going back a step, but then there were all the uncommitted changes which I discarded. Now there's an outstanding pull and I can't get my working copy on desktop back to the branch. Maybe that's the pull I need to do
I think I have myself tied up in a huge Git knot. Not sure what to do anymore.

lie112:
Nope. Pulling the pull takes me back to the latest incorrect to 4.6 commit.

hol430:
Sounds like you did a soft or mixed reset. As you've discovered, these will keep the changes you've made in the intervening time. What you want is a hard commit (I should have mentioned that sorry!) which resets your branch to a particular commit and discards any changes you've made since that commit.

jbrider:
Git should be installed on your path, so you should be able to got to the
command prompt and run the commands as Drew wrote them out.

SourceTree is a bit limited in some of those extra's.
I have had more success with https://git-fork.com/
I'm not sure it has the hard commit option - but it has a right click ->
Reset to here sounds like it would work for you.

Git can be daunting, but it's problems like this one that will give you a
better understanding of it as you try to sort through it.

lie112:
Thanks Drew.
I was able to repeat the steps you provided from command prompt and have the reverted testbranch and the model runs. After working out how to name the commit with wq:
How do I check out the current CLEMWorking branch to do the same steps? i.e. Whats the command to set the current working branch rather than create a new one. When I do it in SourceTree I get a pull that prevents other things and when I pull it I am back to where I started.

hol430:
To checkout a different branch on the command line:

```
git checkout branchname
```

Issue #4639, opened 2020-01-23
unable to create new cultivars in the maize model


JJguri:
I cannot create new cultivars in the maize model because it is not editable in the GUI, i.e. you can drop a cultivar folder into it and then a new cultivar in the cultivar folder. However, if you save the file and then re-open it you cannot see the new editions.  I tried this with my own simulation and the maize example and does not work. @hol430 @HamishBrownPFR 

Here is my simulation: https://www.dropbox.com/s/d3kaddb2kxcscpk/Trials.apsimx?dl=0

hol430:
@hol353 @hut104 how do we want this to work? Should we be able to drop a cultivar onto a plant? This may be tricky to implement due to how the serialization works with the crop models.

hol353:
Yes, we do want to be able to drop a cultivar onto a plant. Someone here at AgResearch tried to do that yesterday with wheat. It will need changes to ModelCollectionFromResource.cs to make it smarter wrt child models dropped by the user.

PR #4678, opened 2020-02-04
Improved null handling in ModelCollectionFromResource

hol430:
Working on #4639

### DIFF
--- a/Models/Core/ModelCollectionFromResource.cs
+++ b/Models/Core/ModelCollectionFromResource.cs
@@ -47,7 +47,13 @@
         {
             get
             {
-                List<Model> officialChildren = GetResourceModel().Children;
+                if (string.IsNullOrEmpty(ResourceName))
+                    return Children;
+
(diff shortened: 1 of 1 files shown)


hol430:
@hol353 my previous pull request (#4671) introduced a bug where any .apsimx file containing a non-released model (e.g. a prototype like soybean or chickpea) would throw an exception when you try and save it from the GUI. This is a pretty serious bug which is obviously not being tested by Jenkins.

It turns out that when you run a file from the command line, the file is not saved after being read. This is different to running a file from the GUI, where the file is saved immediately before being run. Is this the intended behaviour? I guess it makes sense that we don't modify the input file when running from the command line but if that's the intended behaviour then we should think about some other sort of test for the 'save file' mechanic.

hol353:
Don't really want to save a file from the command line and don't really want to save the file in the GUI unless the user has changed it. I'll merge this for now to hopefully fix the problem. Need to think about this.

PR #4686, opened 2020-02-06
Changed sorghum root senescence

hol430:
Working on #572


hol430:
@jbrider the changes to the stats here are pretty marginal, and I think they can be accepted. What do you think?

jbrider:
Checking now - stats seem fine, looking more closely at the effects on
individual simulations - particularly low resourced ones.

jbrider:
The change in the stats is acceptable.

PR #4707, opened 2020-02-12
Soybean

hut104:
working on #2731


hut104:
@hol353 Please merge these changes - new results arising from movement of files to underreview folder.  I will then create an example file etc.

PR #4739, opened 2020-02-19
Changed operator

### LINKED ISSUE #4738: Error in TT Accounting in GenericPhase
All phases are passed a fraction of the day remaining for them to use in their calculations of plant development.  There is an error in Generic Phase in the calculation of the fraction of the day remaining after it has met its target.  

The error will only occur when the phase is passed a fraction of the day < 1 and processes through the day in a single timestep (e.g. a zero or very small length stage).  In most cases, the fraction passed in is one and the error will not occur.
The coding change is 
propOfDayToUse = (ProgressThroughPhase - Target) / ProgressionForTimeStep;
should be
propOfDayToUse *= (ProgressThroughPhase - Target) / ProgressionForTimeStep;

hut104:
resolves #4738

### DIFF
--- a/Models/Plant/Phenology/Phases/GenericPhase.cs
+++ b/Models/Plant/Phenology/Phases/GenericPhase.cs
@@ -74,7 +74,7 @@ namespace Models.PMF.Phen
                 if (ProgressionForTimeStep > 0.0)
                 {
                     proceedToNextPhase = true;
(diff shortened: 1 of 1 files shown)


hut104:
@hol353 @hol430 @HamishBrownPFR 
Wow.  We dodged a bullet here. I was worried for a while that changing such an important line of code might have lit the tests up like a Christmas tree.
Had no effect because we always come in with the argument = 1 and we don't have a stage < 1 day in length.

PR #4726, opened 2020-02-17
Np iwork

HamishBrownPFR:
Resolves #4725

PR #4740, opened 2020-02-19
Changes to excel multi input component

### LINKED ISSUE #4713: ExcelMultiInput not merging files
I have two excel files with observed data (worksheets named 'Observed') that I am trying to read in using the ExcelMultiInput object.  It will read in the data from the first file into and Observed table.  It sort of reads the 2nd file but does not merge in the simulation name so the data is not visiable in the UI and can't be matched with predicted values.

Interestingly, if I read one of the files in with the ExcelInput object, then change the file name on the ExcelInput object to the other file it will read it in and merge it properly with the data already in the Observed table.  That works but is painful.  

I can provide the two files I am working with to test this, just fire me an email

hol430:
Resolves #4713

### DIFF
--- a/Models/PostSimulationTools/ExcelMultiInput.cs
+++ b/Models/PostSimulationTools/ExcelMultiInput.cs
@@ -73,7 +73,7 @@ namespace Models.PostSimulationTools
                         // Write all sheets that are specified in 'SheetNames' to the data store
                         foreach (DataTable table in dataSet.Tables)
                         {
-                            if (StringUtilities.IndexOfCaseInsensitive(this.SheetNames, table.TableName) != -1)
+                            if (SheetNames.Any(str => string.Equals(str.Trim(), table.TableName, StringComparison.InvariantCultureIgnoreCase)))
                             {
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@hol353.  This bug fix causes the soybean tests to fail because the multi-input excel object had been duplicating data and now it isn't so the stats are now correct.

when you merge this can you please to a retest on #4726 which needs this bug fix to run its tests correctly.  Many thanks

hol430:
Actually, can you hold off on merging this. I want to double check that this is doing what I think it's doing. All of those diffs in the stats are a bit alarming.

HamishBrownPFR:
@hol430.  Sorry to winge again.  I need this bug to be resolved before I can complete #4726.  Is there anything I can do to help progress this one?

hol430:
Ok, I just had a quick look and everything appears to be working so @hol353 I guess this is ok to merge.

HamishBrownPFR:
Thanks @hol430 .  I think we can close this issue now

Issue #4773, opened 2020-03-02
Wheat phenology model improvements


HamishBrownPFR:
Model changes, new cultivars and updated cultivar parameters for NPI project.  To be put through release to ensure improvements are generically applicable and broadly available and to ensure everyone in the project is working with the same model

PR #4814, opened 2020-03-04
Npi

HamishBrownPFR:
working on #4773 Adding in Gatton simulations


HamishBrownPFR:
@hol353 The performance tests changed on this because I added more simulations and haun stage data to wheat.  All good to merge from my perspective

Issue #4826, opened 2020-03-04
Root Class Refactoring


hut104:
The Root Class is in desperate need of refactoring.  
1) It has a large number of optional functions - to be removed
2) It uses switches to allow multiple approaches (should be using functions to avoid this)
3) It has constants in the class (these should use existing constants)

PR #4827, opened 2020-03-04
Root Refactoring

hut104:
working on #4826


HamishBrownPFR:
@hol353 I have reviewed the diffs on performance tests here.  They are all small and there are more greens then reds.  I cannot provide a full explanation why the diffs occur at all but this pull request refactors the code make it easier to diagnose problems with biomass partitioning and allow new functionality to be added to root.  Recommend merging

PR #4833, opened 2020-03-04
Lowered min acceptable co2 level to 300 ppm.

### LINKED ISSUE #3381: Minimum acceptable CO2 is too high for simulations starting before 1990

hol430:
Resolves #3381

### DIFF
--- a/Models/Functions/SupplyFunctions/RUECO2Function.cs
+++ b/Models/Functions/SupplyFunctions/RUECO2Function.cs
@@ -47,7 +47,7 @@ namespace Models.Functions.SupplyFunctions
                 if (temp >= 50.0)
                     throw new Exception("Average daily temperature too high for RUE CO2 Function");
 
-                if (MetData.CO2 < 350)
+                if (MetData.CO2 < 300)
                     throw new Exception("CO2 concentration too low for RUE CO2 Function");
                 else if (MetData.CO2 == 350)
                     return 1.0;


Chloe0317:
Dear Dean @hol353 and Neil @hut104 , it appears that when introducing the StomatalConductanceCO2Modifier function based on Elli et al., (2020), the decision to lower the RUE function requirement for CO2 to 300 ppm was not adopted for the modifier function. https://github.com/APSIMInitiative/ApsimX/issues/7983 https://github.com/APSIMInitiative/ApsimX/issues/8286
Considering the previous discussions on historical CO2 data, I think it's sensible to lower the requirement for the StomatalConductanceCO2Modifier function to 300 as well.

Issue #1005, opened 2016-08-02
SoilWater needs a major refactor and speed up.


hol353:
We also want to make separate small classes for each soil water process so that an individual process can be swapped for a different implementation.

We also need to remove the excessive creating of arrays. This is causing the garbage collector to continually pause a running simulation to clean up the allocated arrays.

hol353:
New water balance model:

This is very close to being complete. Jenkins runs all sims without error. There are major diffs in the stats in the AgPasture validation and minor diffs in all other files. I have already fixed one bug to do with the calculation of coverTotSum (Line 222 of EvaporationModel) that helped with the AgPasture sims, probably because there are multiple plant canopies. I suspect there are more cover bugs in the new water balance model that need fixing.

To test, uncomment the converter999 at the bottom of Converter.cs

hol353:
Simulations for some of the bigger diffs in stats are:
wheat - Lincoln2015Nit250IrrFull
sugarcane - 13 Kunia 1991_93 H73_6110 P
soybean - all soilwater - Sutherland2017SowEarly
maize - HustWt - Lincoln2012NitLowIrrFull
Barley - spikewt - MCPD09_10CultBoomaNit0IrrNil

hol430:
@hol353 I'm getting null errors whenever I open a file. The errors are handled internally, so it's not a critical problem or anything, but it slows things down a bit. Looks like it occurs when we copy properties from the official WaterBalance json inside the OnCreated method of ModelCollectionFromResource. Is this supposed to happen?

![image](https://user-images.githubusercontent.com/36427516/76817812-9cc3bc80-684f-11ea-90be-a119f91368ab.png)

hol353:
That's right I'd forgotten about that. Last week I couldn't find a way around the throwing of exceptions in this code. For WaterBalance we need to copy the parameters from the resource. Problem is that the model isn't initialised properly at that point. This is a common problem with many of our models. Perhaps we need to put a whole bunch of if statements protecting the calls to WaterTableModel.Depth property, evaporationModel etc. A bit ugly. Any suggestions?

hol430:
I guess adding null checks is probably the way to go.

hol430:
Actually - can't we ignore those properties with an xmlignore or jsonignore attribute?

PR #4872, opened 2020-03-10
New Soilwater model

hol353:
Working on #1005


Models/Soils/SoilTemp/SoilTemperature.cs
```diff
@@ -394,8 +394,8 @@ private void OnProcess(object sender, EventArgs e)
         /// <summary>
         ///     ''' Initialise soiltemp module
         ///     ''' </summary>
-        [EventSubscribe("Commencing")]
-        private void OnInit2(object sender, EventArgs e)            // JNGH - changed this from Init1.
+        [EventSubscribe("StartOfSimulation")]
```
hol430, line 397: @hol353 do you recall why we changed this from Commencing to StartOfSimulation? It's since been moved to an even later event called FinalInitialise. The result is that it's hard to setup an initial report which runs once at start of simulation.

hol353: Only that 'commencing' is usually too early for models to do their initialisation. The preferred way is for them to use StartOfSimulation. Unsure what you mean be ' setup an initial report'

hol430: If the user wants to report initial conditions on the first day of the simulation, what is the recommended way to achieve that? Setting the event to `StartOfSimulation` is intuitive but it doesn't work. Setting the event to a date as yyyy-MM-dd, but this is not always possible - e.g. if clock start date is a factor in an experiment, or if the report is up under replacements.

Models/Soils/MultiPoreWater/WEIRDO.cs
```diff
@@ -1115,6 +1115,10 @@ private void doDiffusion()
         private double ProfileSaturation { get; set; }
         private double SODPondDepth { get; set; }
         private double EODPondDepth { get; set; }
+        /// <summary>The efficiency (0-1) that solutes move down with water.</summary>
+        public double[] SoluteFluxEfficiency { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
hol430, line 1119: @hol353 do you remember why this was necessary? This causes an exception to be thrown when you try and run weirdo from the gui (probably swim too, since we throw a not implemented there as well).

It seems that these two properties are not actually referenced anywhere outside of WaterBalance. Perhaps they shouldn't be part of ISoilWater?

Models/Soils/SoilTemp/SoilTemperature.cs
hol353: I still don't get what you're trying to do. Don't know what you mean by 'initial conditions'. Give me a concrete use case that currently isn't possible.

Models/Soils/MultiPoreWater/WEIRDO.cs
hol353: Can't remember why these aren't implemented. You are right. Perhaps they shouldn't be part of the interface.

Models/Soils/SoilTemp/SoilTemperature.cs
hol430: Use case: I want to output total soil OC on day one of the simulation.

Models/Soils/MultiPoreWater/WEIRDO.cs
hol353: What's wrong with seting reporting frequency to 'StartOfSimulation'?

hol353: Sorry comment above is related to the other comment above.

Models/Soils/SoilTemp/SoilTemperature.cs
hol430: > What's wrong with seting reporting frequency to 'StartOfSimulation'?

Report has some initialisation code which, among other things, is responsible for subscribing to the events specified in reporting frequency. This initialisation code is run during `FinalInitialise`, which doesn't occur until after `StartOfSimulation`. So everything works perfectly, except it doesn't subscribe itself to the events until after certain events have already fired.

If the initialisation code ran during the `Commencing` event, as it did until this commit, then this would still work. That's why I'm wondering why it was changed - I'm guessing there was a reason.

hol353: Ahh ok. For a while now I've thought our use of events from clock is wrong. Anyway...
You can try putting changing this back to Commencing and see what happens on Jenkins. That might fix the immediate problem for WaterBalance properties but I suspect that there will be similar problems with properties from other models. Would be nice to have a more robust solution. I don't know what that is though atm.

PR #4877, opened 2020-03-10
MicroClimate

### LINKED ISSUE #4876: Strip Cropping Does Not Work
Recent work on micromet has caused some functionality to no longer work.  The test set shows the issues.
Predicted-Observed Tables should be added to the test set so that the results are evaluated by Jenkins.

hut104:
Working on #4876


hut104:
@hol353 These results are different because I've started adding stats for a change!!!  Will we merge?

PR #4886, opened 2020-03-10
Soil water warnings should be fatal errors.

### LINKED ISSUE #4882: Soil wat warnings should be fatel
There are some warnings being thrown from soil that should throw exceptions 

eg see the summary file for the Emerald experiment in the Cotton prototype

hol430:
Resolves #4882


hol430:
@hol353 we looked at these stats yesterday and agreed that they're minor enough to be accepted. If you're happy, this pull request can be merged.

PR #4902, opened 2020-03-11
Fixed jenkins installs

### LINKED ISSUE #4901: New installs are not being built
^

hol430:
Resolves #4901


lie112:
I don't understand what problems CLEM was causing. The "Reason" property of the ResourcesLedger report is a required value. It is a public string property of the ResourceTransaction type being reported. The pricing variables are also needed if pricing has been provided. This has always worked and I'm not sure why these were throwing errors. They are both working in my current version of APSIM. This will now affect every CLEM apsimx simulation not just the examples. The problem must have arisen from other APSIM changes. 

I expect that the problem may arise from the fact that the ReportResourceLedger derives from Report and so changes in Report have not been handled in the few overwirtten methods. What changes were made to report?

hol430:
We changed report so that it throws an exception if it can't find any of its variables. This caused a lot of breakages because a lot of .apsimx files have variables in the report which aren't valid. One such variable is the reason property you mentioned. I have no idea why it's invalid - I didn't look into every single breakage or I'd still be going next year haha. Feel free to look into it and add it back in. If you pull down latest changes and reinstate this variable into your reports, you should see the error.

lie112:
The problem was some of my properties return a null (a string Reason and an object ConvertTo). This null returned by the value() is inferred to mean the property was not found even though the property has been added to the property list of the IVariable and simply has a real value of null. Basically, APSIM no longer allows null values to be returned and reported to a report even if the property exists. 
I could ensure that these are all trapped and return "" instead of null, but hard to trap every possible outcome.
A better approach would be to add a bool property to IVariable that states Valid. This would be set to false at all locations where "return null" is fired during Locator.GetInternal suggesting the variable path was not successfully transversed. This is a more accurate means of trapping the problem.
Then in locations where the errors are thrown it would be 
if(!variable.Valid) throw exception
This would allow anyone to report "" from a property with null value. 
Shall I implement this solution?

lie112:
Upon more investigation and better still,  an _out bool Success_ parameter passed to the GetInternal() method of locator would do the job better than mucking with all the derived classes of IVariable.
We set _out bool Success_ to false at the start of GetInternal and set it to true if either a cached variable is returned or a variable is successfully added to the cache. This adds limited code to achieve the new functionality and allows the code to handle null values from known properties and methods.

hol430:
If you want a quick fix, then returning an empty string sounds like the quickest and easiest solution. In the longer term I'd prefer that the locator throw if it can't find a variable, as this would be consistent with some of our recent changes to soil checking and reporting. I'm not sure how much damage this will cause, but `locator.Get()` is only called from report, variable reference, and string comparison function, two of which already throw if it returns null, so in theory this change might not be as breaking as it sounds. I will put up a pull request and see what happens.

PR #4914, opened 2020-03-12
Report handle position should persist between restarts of apsim.

### LINKED ISSUE #4903: Less screen real estate for reporting frequency in report presenter
Half the screen in the report presenter is taken up with reporting frequency even though 99% of all simulations only use one line within this list.  The frequency list could be smaller in area, allowing the user to see more of the reporting variables.

hol430:
Resolves #4903


hol430:
retest this please jenkins

PR #4930, opened 2020-03-17
Consider XmlIgnore and JsonIgnore when copying properties

hol430:
Working on #1005

### DIFF
--- a/Models/AgPasture/PastureSpecies.cs
+++ b/Models/AgPasture/PastureSpecies.cs
@@ -65,7 +65,6 @@
 
         /// <summary>Gets or sets the canopy albedo for this plant (0-1).</summary>
         [Units("0-1")]
-        [XmlIgnore]
         public double Albedo
         {
             get { return myAlbedo; }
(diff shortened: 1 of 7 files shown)


hol430:
@hol353 this is showing big diffs in the agpasture stats. I had a quick peek in PastureSpecies.cs and it appears that a lot of properties with XmlIgnore attributes (e.g. albedo) are serialized in the .json resource files. In fact, as far as I can tell, it seems like most if not all of the properties in the json have an XmlIgnore in the code. Is this deliberate?

hol353:
They have XmlIgnore on them so that they aren't written to the .apsimx files. I now remember trying the change you have in this PR and got very different stats as well. We need to change the approach we're using for deserialising from resource files. This is a mess.

hol430:
I think the problem is that some of these properties with an XmlIgnore are state variables (LAI, CoverGreen, ...), and some are actually model inputs. The problem is I'm not sure which are which in this case. If you look at PastureSpecies.cs, the default value for SpecificLeafArea is 25, but in AGPRyegrass.json, SpecificLeafArea is set to 20. How did you manage to do this? By editing the json? If this property (and others like it) are really a model input, shouldn't they have a display attribute so that they appear in the UI?

hol430:
retest this please jenkins

PR #4932, opened 2020-03-18
Grapevine

junqi108:
working on  #1822
updated the yield data
it seems the changes in the upstream causes errors in opening the prototype by indicating the
waterTableModel was null.
however, using the released version is OK


junqi108:
retest this please Jenkins

PR #4959, opened 2020-03-26
Add 'on' keyword to report. Refactor ReportColumn.

### LINKED ISSUE #4958: Add an 'on' keyword to REPORT
This optional keyword will control when REPORT collects values for variables. If not specified then the existing behaviour of '[Clock].DoReportCalculations' will be retained. 

This keyword is necessary when using means of variables that are zero most days but have a value occasionally. For example, [AGPRyegras].HarvestedWt is zero most days and has a non-zero value when a harvest occurs. When a mean is calculated in REPORT, the user doesn't want the zeros as part of the calculation.

hol353:
Resolves #4958


hol353:
retest this please jenkins

PR #4963, opened 2020-03-27
Refactor/locator

### LINKED ISSUE #4911: `locator.Get()` could throw if it returns null
This would be consistent with recent changes to reporting/soil checking, and would allow for error-free fetching of variables with a value of null.

hol430:
Resolves #4911

### DIFF
--- a/Examples/Agroforestry/Tree Belt Example.apsimx
+++ b/Examples/Agroforestry/Tree Belt Example.apsimx
@@ -349,7 +349,6 @@
                     "[Clock].Today",
                     "[Zone].Area",
                     "Parent.Distance as Distance",
-                    "Radn",
                     "[LocalMicroClimate].Radn",
                     "[TreeProxy].Shade"
                   ],
--- a/Models/Core/Locator.cs
+++ b/Models/Core/Locator.cs
@@ -293,6 +293,8 @@
                         VariableProperty property = new VariableProperty(relativeToObject, propertyInfo, arraySpecifier);
                         properties.Add(property);
                         relativeToObject = property.Value;
+                        if (relativeToObject == null)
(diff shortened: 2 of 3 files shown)


hol430:
retest this please jenkins

lie112:
The "allow locator to return null if property/method" change does not work.
The example is in CLEMExamples\CLEM_Example_Cropping\single and mixed cropping where Labour.lastTransaction.Reason is null, but the property exists.
If the property.Value is null the Locator.GetInternal returns null not the link to the property that was successfully found causing all the errors to be fired.
The _return null;_ should be a _break;_ to get out of the for loop of namePath items. This means if a property mid way through the namePath is null it will ignore the remainder of the namePath which is acceptable. An error will fire if later the value is not null and a further entry is an invalid property or method.

lie112:
Break; isn't correct either. I'll keep exploring.

lie112:
The problem is in Locator.Get(string namePath). It's not clear if this is meant to return the value or the object. It is currently returning the value of the path. Then in VariableGroups.StoreValue if value is null it throws an error. Value will be null here as its the result not the object, whereas you are expecting the object. What is returned from Locator.Get needs to be cleaned up.

hol430:
Locator.Get should return the value of the object that the path is pointing to. This could be a number, a model, or even an IVariable instance (e.g. if the path points to a property of type IVariable). The locator will throw if the variable is not found, so it seems to me like you could just remove the null check from VariableGroup.cs - the only situation that I can think of where that null check will trigger will be if the variable exists but has a value of null.

hol353:
And there will be a lot of times where that happens (i.e. variable exists but has a value of null).

PR #4968, opened 2020-03-27
Variable ref error

### LINKED ISSUE #4967: Variable reference needs better fail message
Currently only tells the name of the Variable reference model but not where it is located in the simulation.  More useful error message would include a full simulation tree address to the variable reference model that is throwing.

HamishBrownPFR:
resolves #4967 

@hol353 This may not be the best way to achieve what I want to here but shows the intent.  Please advise if there is a better way.

### DIFF
--- a/Models/Core/Model.cs
+++ b/Models/Core/Model.cs
@@ -30,6 +30,25 @@
         /// </summary>
         public string Name { get; set; }
 
+
+        /// <summary>
+        /// Gets the simulation tree adress of the model
+        /// </summary>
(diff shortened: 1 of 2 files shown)


hol353:
@HamishBrownPFR there is already a function to return a pull path try this:

 throw new Exception("Unable to locate " + VariableName.Trim() + " called from the variable reference function " + Apsim.FullPath(this));

HamishBrownPFR:
Yep, that does it.  Thanks @hol353

PR #4978, opened 2020-03-30
Life cycle

### LINKED ISSUE #4977: Lifecycle enhancements
Lifecycle needs to be able to:

1.  reproduce into parallel lifecycle (As in a population reproducing resistant and non-resistant ofspring
2. migrate into parallel lifecycle (As in individules from on population moving to another)
3. Vector transmission (As in individules moving from one population (in vector) to anohter population (in host) as a result of vector activity)

HamishBrownPFR:
resolves #4977


hol353:
@HamishBrownPFR : This is now conflicted.

hol430:
@HamishBrownPFR just so you know, the lifecycle docs are over 1000 pages long. Might want to double check which nodes are being documented and which nodes need to be documented.

hol353:
I'm sure every word on those 1000 pages is important.

PR #4979, opened 2020-03-30
Allow for the user's locale when downloading soil descriptions.

### LINKED ISSUE #4969: Soil downloader doesn't work in German locale
Attempts to use the soil download in a German locale fail because the latitude and longitude values in the request uri get written in a culture-sensitive manner (using a comma as decimal place separator):

```
http://www.asris.csiro.au/ASRISApi/api/APSIM/getClosestApsoil?maxCnt=5&longitude=151,26&latitude=-27,18
```

zur003:
Resolves #4969


hol353:
@zur003 This is now conflicted. Is this the same as the @hol430 fix I merged this morning?

zur003:
Same net effect, using a different approach.

PR #4990, opened 2020-04-01
Fixed Cumulative Gliricidia Harvest Data

hut104:
working on #1227


hut104:
@hol353 I think can be merged in.  The results are only different because I fixed a bug in a script (that improved some results) and added new data.

hut104:
You can merge this now please @hol353

PR #4996, opened 2020-04-01
Added reduce plant population to SimpleGrazing.

### LINKED ISSUE #4992: SimpleGrazing not working with PMF plant models
- SimpleGrazing.GrazedDM is always zero for PMF and AgPasture models
- PostGrazeDM calculation is using the AboveGroundWt variable and so it is different to PreGrazeDM. Make sure units are kg/ha for both variables regardless of which plant model is being run.
- Add population reduction parameter into SimpleGrazing that works the same as the Defoliate manager. It will work with PMF crops only, not AgPasture ones.

hol353:
Also fixed SimpleGrazing.GrazedDM variable.

Resolves #4992

### DIFF
--- a/Models/AgPasture/SimpleGrazing.cs
+++ b/Models/AgPasture/SimpleGrazing.cs
@@ -138,7 +138,6 @@
         public string NoGrazingEndString { get; set; }
 
         /// <summary></summary>
-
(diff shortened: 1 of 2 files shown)


hol353:
retest this please jenkins

PR #5003, opened 2020-04-02
Allow string based land id in txt/xls crop data reader

### LINKED ISSUE #5002: Problem using string based soil/land id in xls crop data reader
The crop reader is not working with xls files when land id is the newly allowed string

lie112:
Resolves #5002


lie112:
Again, not sure why i'd get validation errors from this pull request.

hol430:
The exception is thrown while running CLEM_Sensibility_IndividualAnimal.apsimx. Do you get this error when you run the file locally?

lie112:
Thanks.. I’ll fix that one..
I wasn’t able to see what was causing the error.

lie112:
I fixed the CLEM validation error, but it fails to pass and I can't work out why except that a few validations are too low.

lie112:
What/Who is involved in accepting the validation stats?
I get a 93% for What HarvestObsPred

hol430:
retest this please jenkins

hol430:
I'm pretty sure the failed stats are a false positive and are due to a bug in the performance tests, possibly due to @hol353 accepting changes from another pull request while this one was already running. Rerunning the tests should fix the problem (I hope).

Models/Report/VariableGroup.cs
```diff
@@ -49,8 +49,6 @@ public class VariableGroup
         public void StoreValue()
         {
             object value = locator.Get(variableName);
-            if (value == null)
-                throw new Exception($"Unable to locate report variable: {variableName}");
```
hol353, line 53: Why do you want to remove this check. @hol430 spent quite a lot of time getting it in there.

lie112: This was based on discussions #4963 with @hol430 where this line was not considered necessary because it was checking the value of the object and not a null object. There is no reason to stop parameters that were found that have a null value, where other checks now report errors when the parameter supplied is not found.

hol430: I'm pretty sure this is unnecessary now because the locator will throw if the variable is not found.

Issue #4803, opened 2020-03-03
New FloralInitiation Stage in Wheat


hut104:
The wheat model currently has a terminal spikelet stage but no floral initiation stage.  the later should be added to enable users to more closely compare model output to observations.

hol353:
@APSIMInitiative/cereals : fyi

hut104:
The team has changed its mind.  The stage should really be called VernalSaturation.

PR #5009, opened 2020-04-02
Refactor Phenology/Structure dependencies

hut104:
working on #4803


hut104:
@hol353 Could you merge this please?  The PvO table is different because a variable name changed

PR #5011, opened 2020-04-03
Grapevine

junqi108:
working on #1822
added a beta growth function of the form y = Ymax * (1 + (te - t)/(te-tm))* (t/te)^(te/(te-tm))
 See Yin et al 2003 A Flexible Sigmoid Function of Determinate Growth; 
it has similar pattern as the sigmoid pattern but has better performance. it is widely used in the wageningen crop models.


junqi108:
@hol353 @HamishBrownPFR  I am not sure why there is a shift of columns in the wheat test. I just set up a new function that no other model is using.  any suggestions? 
image

junqi108:
retest this please Jenkins

PR #5020, opened 2020-04-06
Change vernalisation to use local value for delta HS

hut104:
working on #4803


hut104:
@hol353 Could you merge this please?  The results are different as a result of some refactoring on the wheat model.  Some data are better, some are not.  However, the current implementation avoids some bad design practices.

PR #5043, opened 2020-04-16
Leaf 'GrowthDuration' should not be reduced if the final cohort has not fully appeared

### LINKED ISSUE #5042: Leaf 'GrowthDuration' should not be reduced if the final cohort has not fully appeared

BenAbabaei:
Resolves #5042 

I am not sure about this. But it is worth inspecting. 
**Admins**, please feel free to close the pull req if the proposed change is not needed.


hol430:
retest this please jenkins

PR #5055, opened 2020-04-17
Grape vine development  

### LINKED ISSUE #5054: Add improved priority arbitration method
Current priority methods specify order of priorities by order of organs in the json file and prioritize structural and metabolic allocation ahead of storage.  This creates cases where some organs get no allocation and storage gets no allocation.

We need a method where priorities can be specified by coefficients for each demand type on each organ so all demands can be arbitrated at the same time, all pools will get some allocation but the higher priority organs will get more.
In situations where supply is not limiting we need to ensure that all organs get their full demand 
Propose including a PriorityCoefficient on each organ which specifies priorities for each pool type.  Two pass allocation in arbitrator where each demand is multiplied by its priority coefficient to give a scalled demand, and all scalled demands are summed to give a plant total.  If the first pass allocation is done based on scalled demand relative to total  scalled demand.  For second pass, allocate based on unmet demand for each pool relative to total unmet demand

HamishBrownPFR:
resolves #5054 

@junqi108 , this changes the results for the grapevine model and I have not made any effort to fit the priority parameters, just set values to check they are being passed around correctly.

@hol353, @jbrider, @hol430 , @hut104   I have added DMDemandPriorityFactor into the IArbitration interface to facilitate the passing of these parameters from each organ to the arbitrator.  This means I have had to implement redundant DMDemandPriorityFactor in all organs.  Can't think of a better way of doing this but open to suggestions.


hol353:
Almost seems like you need each type of allocation to have its own interface e.g. IQPriorityThenRelativeAllocation, IPriorityAllocation, IPriorityThenRelativeAllocation etc. Each organ then implements which ever of the allocation methods it wants to. This gets around the problem of putting dummy implementations for say QPriorityThenRelativeAllocation in a lot of different organs.

Downside: this might take a bit of redesigning of arbitrator to make this work. Might also end up with a lot of allocation interfaces.

hut104:
I even wonder if we need to look at each of these allocations to see if we can remove some, or redesign a more flexible allocation rule to replace the others.  Hamish, could the new approach replace the existing approach?

I think we need to be careful about optional (or unused) public members.

jbrider:
@HamishBrownPFR I was looking at something similar when doing the sorghum arbitration. I think what @hut104 suggested could work. 
Do you need to have a concrete implementation of DMDemandPriorityFactor - couldn't you just add an IFunction to each of the Organs in the tree and then look for it by name at the beginning of Arbitration - then throw an exception if it's missing?

HamishBrownPFR:
Hi Guys, thanks for your comments.  I suspect this arbitrator method could superseed the others as it give alot more flexibility.  However, it would be a big job replacing this and refitting it for all the crop models.
We need to make some progress on fitting the grapevine model.  @hol353 can we get this merged now and then open a separate issue to  rationalize the number of arbitration methods.

PR #5057, opened 2020-04-19
SetActualUptakes must account for the proportion of root in each layer

### LINKED ISSUE #5056: SetActualUptakes must account for the proportion of root in each layer

BenAbabaei:
Resolves #5056


BenAbabaei:
@hol430 Could you please have a look at this?

Does Zone.Water consider the proportion of each layer that is occupied by roots? It seems it is justa copy of Soil.Water, which only depends on the thinknesses of the layers. If this is the case, I think Zone.Water must be scaled with these proportions, which would result in reduction in water supply as compared with the current setting.

Issue #4617, opened 2020-01-21
Refactor the parameters for Stock


sno036:
So that they are APSIM-like, settable from Managers, and in the JSON file

zur003:
This would require some fairly significant refactoring. It may also mean that GrassGro and AusFarm users with customized parameter sets would not be able to use those parameter sets directly in Apsim simulations. I think we need to consider what we're trying to achieve with this, and decide whether a less radical change would allow us to accommodate most use cases.

her123:
Testing by @sno036 is suggesting that dairy breeds may also be putting too much of their intake into body growth. This may need examination of the parameter.

her123:
First step is to enable the use of external parameter files. By using the grazplan parameter editor it will be simple to edit these data sets and use them in Experiments.
Discussion of improved handling of animal/pasture/crop parameters needs to take place and then Stock could use a new technique for parameter storage.

hol353:
AgResearch are funding my time to refactor STOCK, in particular the parameter handling code and what the user sees in the user interface. I agree this needs to be done carefully and will require some significant refactoring. We think it can be achieved though in the next few months.

1. I have been working on this for a couple of weeks and have come up with a way to read the ruminant.prm file into an AnimalParamSet. This will allow continued use of GrazPlan .prm files into the future solving the compatibility issue. It would be nice to be able to go the other way around as well i.e. take a user genotype parameterisation from APSIM and have it run in GrazPlan. This is a much lower priority for us at the moment.

2. My plan is to allow instances of AnimalParamSet classes to be put onto the STOCK model as a child in the user interface. The user can then change the parameters of these either directly by clicking on the parameter set and modifying parameters or by manager scripts or by factors in an experiment. I currently have this working. There are a lot of parameters so some grouping will probably be desirable at some point.

3. I also want to allow new cross breeds to be specified in the GUI by letting the user add child instances of SingleGenotypeInits to the STOCK model. This will be consistent with the idea of adding AnimalParamSet to the STOCK model. This means we can remove the 'Genotypes' tab on the STOCK view.

4. I then want to do the same for animal groups i.e. let the user add an AnimalInit instance to the STOCK model. This means we can then remove the 'animals' tab on the STOCK view so STOCK in the GUI won't have any parameters at all. Instead it will rely on child models for its parameterisation.

To achieve the above I will raise a number of pull requests in the next few weeks / months. This will provide an opportunity for peer-review as I progress and allow testing of the above by others. 

[...]

PR #5061, opened 2020-04-20
Refactored genotype handling code in STOCK

hol353:
Working on #4617


hol430:
retest this please jenkins

Tests/UnitTests/Stock/StockTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        [Test]
+        public void EnsureNoDotsInGenotypeNames()
+        {
+            var genotypes = new Genotypes();
+            var allGenotypes = genotypes.GetGenotypes();
+            foreach (var genotypeName in allGenotypes.Select(genotype => genotype.Name))
+                Assert.IsFalse(genotypeName.Contains("."));
+        }
+
+        ///// <summary>Make sure nested, hierarchical parameter sets are read.</summary>
```
hol430, line 86: Don't we normally try and avoid commenting out code?

Tests/UnitTests/Stock/StockTests.cs
```diff
@@ -0,0 +1,100 @@
...
+                      "     <set name=\"sheep\">" +
+                      "        <par name=\"c-w-0\">0.999</par>" +
+                      "     </set>" +
+                      "  </set>" +
+                      "</parameters>";
+            var genotypes = new Genotypes();
+            genotypes.LoadPRMXml(xml);
+            var animalParamSet = genotypes.GetGenotype("sheep");
+
+            Assert.AreEqual(animalParamSet.sEditor, "Andrew Moore");
```
hol430, line 35: I'm pretty sure the expected value should be the first parameter passed into this function. It makes a difference when the test fails because the error message will have the expected/actual values around the wrong way if we leave it like this.

Models/Stock/grazparam.cs
```diff
@@ -371,14 +387,19 @@ public class ParameterSet
...
-        private string FName;
         private string FEnglishName;
         private string[] FLocales = new string[0];
         private Translation[] FTranslations = new Translation[0];
         private ParameterSet FParent;
-        private ParameterSet[] FChildren;
+        /// <summary>
+        /// 
+        /// </summary>
+        public ParameterSet[] FChildren;
```
hol430, line 397: This should be a property (and same for `FDefinitions`).

APSIM.Shared/Utilities/ReflectionUtilities.cs
```diff
@@ -46,6 +48,20 @@ public static List<FieldInfo> GetAllFields(Type type, BindingFlags flags)
...
 
+        /// <summary>
+        /// Return all properties. The normal .NET reflection doesn't return private fields in base classes.
+        /// This function does.
+        /// </summary>
+        public static List<PropertyInfo> GetAllProperties(Type type, BindingFlags flags)
+        {
+            if (type == typeof(Object)) return new List<PropertyInfo>();
+
+            var list = GetAllProperties(type.BaseType, flags);
```
hol430, line 59: Do we really need this function? I think (could be wrong) `BindingFlags.FlattenHierarchy` gives access to inherited members.

Tests/UnitTests/Stock/StockTests.cs
hol353: Yep. Fixed.

Tests/UnitTests/Stock/StockTests.cs
hol353: I didn't know that. I've been doing it wrong for years. Fixed.

Models/Stock/grazparam.cs
hol353: Fixed.

APSIM.Shared/Utilities/ReflectionUtilities.cs
hol353: BindingFlags.FlattenHierarchy doesn't give access to privates from base classes. Don't know why. That's why I wrote this method.

hol430: Oh ok, well I guess it does make some sense since if they're private, they don't really belong to the derived class. That being the case though:

1. Does it really make sense to access private members from the base class, and if it does, should they be changed to protected?
2. If they're private should they converted be fields?

hol353: I'm using this to serialse all public and private fields and properties to JSON. This allows me to capture the compete state of an object at a point in time. Really useful for debugging as I can then use a diff tool to compare states of objects.

hol430: Makes sense

hol430:
retest this please jenkins

lie112:
The commit [a593230](https://github.com/APSIMInitiative/ApsimX/pull/5061/commits/a5932301ad8bdd0c1b90638f6c2aa4488f74cdfb) on 20 April 2020 added the BindingFlags.DeclaredOnly to selecting the list of available properties to display to the user. Prior to this the PropertyPresenter included the public properties of a base class that included a description attribute. This was used by CLEM to provide some base properties (e.g. CLEMActivityBase.OnPartialResourcesAvailableAction from CLEMActivityBase for all Activity models) that are no longer available for the user to set.
Clearly, the stock model needed to ignore some base properties for this change. 
Is there another way to ignore these properties (e.g. an [Ignore] Attribute or simply having no description attribute will suffice as the next line filters by those with description attribute) from the stock model and return the original functionality?
It's messy to add CLEM specific code to read a different list of properties to all other models and I can't think of an easy way to allow these to be included without reverting the bindingflags change.
Removing the BindingFlags.DeclaredOnly fixes the CLEM issue and does not affect any of the saving of property state in the PropertyPresenter.

hol430:
I expect this is an oversight -it will affect more than just clem components. E.g. I would expect to see variables like altitude/slope for rectangular zone, which are inherited from zone. @hol353 do you recall why this change was made?

hol353:
Nope can't remember why. There is a lot of inheritance used in the stock cultivars. Try changing it back. I'll then pull down the PR and see what the impact is on the STOCK cultivar parameters shown to the user.

lie112:
What is the best way to resolve this in GIT? Do we revert the specific change (it is actually part of multiple changes in the commit), create a new issue, reopen this as working on, or what? Shall I change the line 251 of PropertyPresenter.cs to remove the BindingFlags.Declared and pull as a new issue?

hol430:
Yeah easiest is probably to just manually change the binding flags on that one line of code, then create a new issue and submit a pull request against that issue.

PR #5069, opened 2020-04-21
Implementing a skip-row configuration

### LINKED ISSUE #5068: Implementing a skip-row configuration
If there is not, can we implement something like this?

image

BenAbabaei:
Resolves #5068


hut104:
These changes seem to be implementing switches (not usually a good way of implementing things). Could we simply get people to enter the two rowspacings on left and right?  That would avoid the use of flags, would be explicit, and would be easier to understand and would allow infinitely more options, especially for agroforestry or more complex simulations.

BenAbabaei:
Hello @hut104 . I will make the change. What would you do with rowConfig? Currently SkipPlant and SkipRow are not accessible via Sowing routine. Should I move everything in that routine and make them 'input' parameters?

hol430:
retest this please jenkins

BenAbabaei:
@hol430 can this be merged?

PR #5073, opened 2020-04-21
Add a root shape function

### LINKED ISSUE #5072: Add a root shape function
This function calculates the area occupied by root in each soil layer. It requires root angle.

BenAbabaei:
Resolves #5072 

**Ready to be merged.**


BenAbabaei:
Sorr for that @jbrider. I fixed it now.

BenAbabaei:
Are you sure Sorghum did not use to use the 'simple' semi-circular pattern? yet, the differences are quite small.

jbrider:
The code for sorghum/C4Maize differentiates the calculation of area using the RootAreaCalcType switch.  You have implemented the C4Maize routine for both of them. Is that what you meant by sorghum using the 'simple' calculation? In the short term, it would be good to reproduce the difference between the 2 models as that is how it was implemented in Classic.

BenAbabaei:
Actually no. By 'simple' I meant a simple circle, not that code written by Greg. So, which one must be used for Sorghum?

jbrider:
Looks good @BenAbabaei.

hol430:
retest this please jenkins

BenAbabaei:
@hol430 could you please let me know what I can do with this failed unit test?

jbrider:
@BenAbabaei This is what happens when another pull request before yours that changes the version in converter and has been merged.  You just need to change yours to 95 instead of 94.  Sometimes this can be really messy and everyone hates when it happens... but no one has come up with a better way.

BenAbabaei:
Thanks @jbrider . It never happened to me before. I hope this fixes it. I am a bit suprised that I cannot see any newly merged pull.

jbrider:
Sorry @BenAbabaei I saw a version mismatch and assumed it was the converter. Your version (94) was right, but the test is failing on the resource file check. So you also need to update the resource files to the new version as well.  The resource files are the released crop models.
On the menu bar in Apsim there is a convert files option - you need to click that button and update the resource files in the models\resources folder to the latest version, then commit them all.

BenAbabaei:
I really cannot say why we have red lights.

Models/Interfaces/IRootShape.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Models.PMF.Interfaces;
+using Models.PMF.Organs;
+
+namespace Models.Interfaces
+{
+    /// <summary>
+    /// Interface for arbitration methods
```
hol430, line 7: Forgot to update description?

BenAbabaei: Sorry for that. I'll fix it.

BenAbabaei:
@hol430 The red lights is because of fixing a problem in the calculation of semi-circular pattern that previsouly existed. The old code, did not account for the coordinate of the centre of the circle. We now have a semi-elliptic pattern instead, which with a root angle of 45 will do the same job as the old code was supposed to do.

https://github.com/APSIMInitiative/ApsimX/blob/6d800b3495975ab4d7b8c3363a6bb038d08d89ae/Models/Plant/Organs/Root.cs#L967-L991

hol353:
@jbrider Are you happy with these changed stats to Sorghum? Look only minor to me.

hol430:
The actual changes in stats (rmse, nse, ...) are minor but if you go in to the drilldown graphs you'll see that there are a lot of predicted/observed pairs showing up which don't have a matching accepted datapoint:

image

BenAbabaei:
@hol430 What does it mean then?

jbrider:
@hol430 that's not the same graph as the one I'm seeing.
Mine is showing no diffs in the TPLA graph.

@BenAbabaei if you make the structural changes first and get it to run green then it is much easier to review the calculation fixes after that.

@hol353 it's ok to be accepted.

BenAbabaei:
@jbrider here changes and fixes are the same.

hol430:
@jbrider can you link me to the page where you're seeing the graph with no diffs?

jbrider:
[link](https://apsim.csiro.au/APSIM.PerformanceTests/ValuesCharts.aspx?PO_Id=122315)

image

hol353:
So my guess is that @jbrider is looking at the graph via the graphical results button. @hol430 is looking at the graph via the tabulated button. Which one is correct?

jbrider:
It looks like the tabulated method is broken - all of the graphs are wrong if I go in that way.

BenAbabaei:
retest this please Jenkins

BenAbabaei:
@hol430 can this be merged?

hol353:
I can't merge this until we sort out which set of graphs on the performance site are correct. @hol430 is working on it.

hol430:
I'm hoping that #5097 will fix the problems with the graphs. Might have to retest this one after it gets merged though.

BenAbabaei:
@hol430 Thanks. I am making another push. I will wait till your push gets merged.

BenAbabaei:
@hol430 and @jbrider My work on this is done.

BenAbabaei:
@jbrider These are the stats when I fixed that problem with the formula of the circle in Sorghum root pattern. I leave it to you to decide what to do with Sorghum. I reverted to the old one as it took too much time to get some green lights for Sorghum.

Sorghum Stats with New Function

hol430:
Ok, I think the graphs for this pull request are fixed, but if you notice any weirdness I'd suggest just closing this pull request and opening a new one.

BenAbabaei:
The stats abd graphs seem OK to me.

PR #5077, opened 2020-04-23
Bug Fixed: CalculateRootActivityValues must not include ProportionThroughLayer

### LINKED ISSUE #5076: Root activity must not use RootProportionInLayer
Proportion of the last soil layer through which root has penetrated is already accounted for when calculating water uptake.

BenAbabaei:
Resolves #5076 
Proportion of each soil layer through which root has penetrated is already accounted for when calculating water uptake.

Working on #5079


BenAbabaei:
@hol353 I am almost sure about this fix. But it makes some major changes in the stats. Would you mind commenting?

hol353:
@HamishBrownPFR, @jbrider, @hut104 Can you please check this small change to root. It creates a lot of changes to the stats.

PR #5085, opened 2020-04-23
Fixed a bug where croptype was null when crop was added via 'Add Mode…

### LINKED ISSUE #5084: Problem with adding new crops to simulation
When you right click on a paddock, choose 'Add Model' and then choose a crop e.g. plantain or oats, the CropType parameter for that crop is set to null even though CropType has a value in the resource. When running, you get an error saying unknown residue type. This is due to a null CropType.

hol430:
…l' button.

Resolves #5084

### DIFF
--- a/ApsimNG/Presenters/AddModelPresenter.cs
+++ b/ApsimNG/Presenters/AddModelPresenter.cs
@@ -142,11 +142,7 @@
                     IModel child = (IModel)Activator.CreateInstance(selectedModelType.ModelType, true);
                     child.Name = selectedModelType.ModelName;
                     if (child is ModelCollectionFromResource resource)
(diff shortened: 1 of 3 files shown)


hol353:
Why is this necessary? The plantain.json resource has this:

> {
>       "$type": "Models.PMF.Plant, Models",
>       "CropType": "Plantain",
>       "IsEnding": false,
>       "DaysAfterEnding": 0,
>       "ResourceName": null,
>       "Name": "Plantain",

I thought that the properties (including CropType) would be copied from the resource.

hol430:
Good point. That being the case, I'm not sure why this isn't being copied from the resource model.

PR #5089, opened 2020-04-27
Working clem

### LINKED ISSUE #5088: Improve marketplace functionality and transaction reporting
The market needs to be updated to include model cloning and transaction bubbling up to correct resource holder

lie112:
Resolves #5088 
Fixes a couple minor bugs identified while producing update


hol353:
Any reason why you aren't using Apsim.Clone(model) for objects that implement IModel 
or
ReflectionUtilities.Clone(object) for all other objects? This will save you writing a bunch of code

lie112:
I didn't know about Apsim.Clone() or the ReflectionUtilities.Clone. Quite elegant approach. It is hard to know what functionality is provided and I didn't stumble across it even in my search for all "Clone" the other day. I assume that approach will also copy all the children of the model that I require?

Is the binarySerialise like serialising the apsimx object or does it also include the non property private parameters? I guess because its binary it doesn't worry about [XmlIgnore] attributes. 

I will look to revert back to using solely Apsim functionality. 

My slight problem is that I was able to include functionality to work out just what properties I set so it wasn't a complete clone. I am thinking about how I could add this to set the values I don't want after the clone call.

I need to use ReflectionUtilities.Clone as I'm not just working with IModel but classes derived from IModel. I am unable to use ReflectionUtilities.Clone as an error says Models.Core.Events is not marked as serializable. I'm not sure where my class uses Core.Events.

hol353:
Binary serialising will serialise all members (private and public). As a result it will serialise all 'child' objects as well. It is a complete clone. It is not a json or xml serialise so it ignores XmlIgnore attributes.

If you don't want full clones then you can put an attribute of [NonSerialized] on fields or properties you don't want cloned. You might need to do this if you have a field of type 'Events'. I don't think we've made it serialisable.

If your object is derived from IModel or Model then you should use Apsim.Clone. Otherwise use ReflectionUtilities.Clone. I have used ReflectionUtilities.Clone in other, non APSIM, projects. It should work.

lie112:
Thanks,
I don't know where my model class is using Models.Core.Events (CLEM/Resources/HumanFoodStoreType). 
I do have a System.EventHandler. Is it coming form the EventSubscribe attribute? But these are on methods not fields or properties. I can't find any use of Core.events in CLEM (or any model outside Links.cs, Simulations and SimulationGroup)
I notice you use a private field behind all properties that need to be NonSerialized and so can't use {get; set;} on properties, this will require a bit of recoding. I can never remember what version of C# we are up to, but from 7.3 onwards you can use [field: NonSerialised] for properties.
Any idea how I avoid the Core.Events issue?

lie112:
Thanks.. 
I just had to deserialise all [Link] attributes.
I guess this means the cloned objects won't have access to the [link] properties. Is there a way of hooking them up is needed at the time of cloning?

hol353:
Unsure what you mean by "I just had to deserialise all [Link] attributes.". Links shouldn't be serialised. They can be hooked up at runtime if need be. e.g.

`var links = new Links();
links.Resolve(obj, true);`

where obj is the object that needs links resolved.

I'm also unsure what you're trying to do. I'm not sure there are any models in APSIM that need to resolve links during a simulation run. Resolving links is a really low level thing to do.

lie112:
Hmm, From the start of developing CLEM Shaun had used [Link] Clock to make the clock available in all my models. I have since also hooked up the Summary so any Model can write to summary and my activities have a link to the Resources holder. Am I better off just having a public property that is resolved at startup to share this information around models from base class? I may have been led down the wrong path for years here.
They weren't serialised, but were inside serialised Models.
I was able to get Apsim.Clone to work by

`        /// <summary>
        /// Link to summary
        /// </summary>
        [Link]
        [NonSerialized]
        public ISummary Summary = null;`

hol353:
Using links is a good thing. It is what I recommend to avoid having to write 'plumbing' code to get models like clock and summary. I guess my question is why are you cloning models? Are you adding/removing them during a simulation? If so, why?

lie112:
The example is when a farm tries to sell a resource to a marketplace and the market doesn't currently contain that store, the model clones a resourcetype with associated converters and pricing and places it in the market resources, Then based on the children list approach of APSIM the resource is available and transactions between the market will commence. Most time I suggest the user has the market all set up, but it makes sense to dynamically create the market for the DARPA project.

One other related question: is the apsim project file disconnected to the simulation? That is, do changes to properties in the simulation tree get stored in the apsimx file. Originally I though this might me the case so I duplicated some pricing structures so they could be modified during the simulation.

hol353:
I think the answer to your question in yes. Before running any simulation the simulation is cloned and the clone is run. That way any changes to the simulation during a run are not saved back to the .apsimx file.

lie112:
Good to know. I still need Cloning for marketplace and I have changed it all to Apsim.Clone(model). This should be good to pull now.

PR #5090, opened 2020-04-27
When generating PDF files, always be explicit about page orientation.

### LINKED ISSUE #4796: Autodoc page orientation
Autodoc is currently creating PDF's with pages in landscape mode.  Formatting is being affected with many near-empty pages.

### LINKED ISSUE #5087: The Soybean model documentation has been generated in landscape mode
The Soybean model is under review.  I just checked the pdf file and it has been generated in landscape model and so figures do not fit on pages.  It should be in portrate.

This has happened before.  I cannot remember what caused this.  Can we add a test to our system to check if pagination is portrate or landscape?

hol430:
Resolves #5087 (I think)
Resolves #4796


hol430:
@hol353 as you can see I've changed the DataStoreWriterTests. For some reason we were getting UnauthorizedAccessExceptions when changing the current directory. I'm not really sure why we suddenly started getting these errors but then I'm also not really sure why we're changing the directory in the first place. ¯\\\_(ツ)\_/¯

Issue #5093, opened 2020-04-28
Tidy up initialization of biomass in organs


HamishBrownPFR:
Organs have biomass in 3  pools which demands for are specified separately in .json.  However initial biomass is specified by a single value in .json and all partitioned to structural.
Each organ should have an InitialWt and InitialN object in Json of type BiomassPoolType with Structural, Metabolic and Storage specified seperately.

HamishBrownPFR:
To general to be useful

PR #5094, opened 2020-04-28
Initial dm

HamishBrownPFR:
working on #5093.  
If this works the same approach can be implemented for N and for all organs to give more flexibility in model configuration in JSON.


HamishBrownPFR:
@hol353 could you please review  this and get back to me if any changes needed.  @junqi108  is keen to get on with parameterising the initial DM for the Grapevine model

PR #5102, opened 2020-05-03
working on root storage calc

### LINKED ISSUE #5101: Root storage biomass is not updating
Root storage biomass is not calculating correctly. this affects a lot of the performance of perennial plants.

junqi108:
resolves #5101


junqi108:
@HamishBrownPFR @hol353 
It seems changing the root storage calculation function would affect the performance of Gliricidia model. not sure how to solve the performance issue, although the change was not very big. 
image

HamishBrownPFR:
@hut104  This pull request is failing because @junqi108 has fixed the root code so it is allocating DM to structural and storage DM.  There was a bug so it was only allocating to structural even if there was a storage demad.  Glericida is the only model that parameterises a storage demand for root which is why it is failing.
Solutions are either to accept this as a bug fix, or change the parmeterisation so glericidia does not put up a storage demand.  Please advise how you would like to proceed

HamishBrownPFR:
@junqi108 try setting structural fraction to 1.0 in gliricidia root which should reproduce the current behaviour
image

junqi108:
retest this please jenkins

junqi108:
@HamishBrownPFR @hut104 the stats looks still the same, although not so much different from the originals.

HamishBrownPFR:
@hut104  Changing the structural fraction to 1.0 did not reproduce the existing behaviour.  I am guessing because when we make all the demand structural it is treated with a higher priority that storage.
We are keen to get this bug fixed so we can progress fitting of the grapevine model.  How do you suggest we proceed?

hut104:
@hol353 This bug fix has slightly altered the results, but the overall stats are not significantly affected.  We can revisit retranslocation from roots at some time in the future if required.  I suggest we merge this in.

Issue #5103, opened 2020-05-04
Allow different responses for AirTemperatureFunction


HamishBrownPFR:
The AirTemperatureFunction is hard coded to use XY pairs to specify temperature response shape that it applies at each 3h period.  However, it would be useful to use other responses functions such as the WangEagelTemperatureFunction in the same sub-daily routine as suggested by @junqi108

PR #5104, opened 2020-05-04
Generalise temp interp

HamishBrownPFR:
working on #5103


HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
@hol353.  This is OK to merge.  Two very small diffs in the wheat set caused by some unexplained changes in controlled environment simulations where Min and Max temperatures are the same.

HamishBrownPFR:
@hol353 this is OK to merge.  There are two small diffs in wheat for stats which are already poor.  These are caused by differens in controlled environment where min and max temp are both set to the same value.  Something subtle but unimportant going on there.

HamishBrownPFR:
@junqi108 , try changing the structuralFraction for root in gliricidia to 1.0 which should reproduce the existing model behaviour
image

PR #5114, opened 2020-05-05
Generalise temp interp

### LINKED ISSUE #5103: Allow different responses for AirTemperatureFunction

HamishBrownPFR:
resolves #5103


HamishBrownPFR:
@hol353 .  This is OK to merge.  There are some small diffs that I have traced back to the PopulationBasedDemandFunction which Maize, Oat and Barley have been using for their spike or rachis DM demands.  This was getting yesterdays value for thermal time and the changes made mean it is now getting the correct value.  Differences in stats are all small so this is OK to merge.

HamishBrownPFR:
@hol353  Please merge.  Ignore my previous comment, I found a way to remove those diffs and tidied up the code and made it more generic in the process

PR #5119, opened 2020-05-06
Renamed Genotype to GenotypeWrapper and AnimalParameterSet to Genotype.

hol353:
Working on #4617


ApsimNG/Presenters/GenericPresenter.cs
```diff
@@ -57,16 +62,46 @@ public void Attach(object model, object view, ExplorerPresenter explorerPresente
...
             this.genericView.SetContents(html, false, false);
         }
 
+        /// <summary>
+        /// For each image markdown tag in a paragraph, locate image from resource and save to temp folder.
+        /// </summary>
+        /// <param name="paragraph">The paragraph to scan.</param>
+        private void FindImagesInParagraph(AutoDocumentation.Paragraph paragraph)
+        {
+            var regEx = new Regex(@"!\[(.+)\]\((.+)\)");
```
hol430, line 84: An image name could theoretically have `](` in it which might cause problems with this regex. Might be worth replacing the first `.` in the regex with something like `[^\]]`. And maybe the second `.` should be `[^\)]`

hol353: Not sure there is any difference between you suggestion of `[^\]]` and my `[(.+)]`. Both will match on any character except `]`. Yes?

hol430: As I understand it, `[(.+)]` will match one or more of any character.

I thought we could already include images in markdown anyway?

Issue #5113, opened 2020-05-05
hourly temperature interpolation 


junqi108:
using hourly temperature for some calculations like the grapevine budburst make the algorithm perform better. The mean daily temperature is a summary of the whole-day temperature. Its application for calculating chilling is ambiguous as chilling is accumulated more instantly.

HamishBrownPFR:
@junqi108 I have a pull request open at the moment #5114 that will allow implementation of this along side the existing approach.

Issue #5121, opened 2020-05-06
Tomorrows met data


HamishBrownPFR:
@hol353 @hol430 @hut104 @junqi108 we are working on the grapevine model and bud burst is quite sensitive to short periods of chilling and are needing a more detailed daily air temperature interpolation.  The problem is the approach we want to use needs to know tomorrows minimum temperature.  APSIM currently has no notion of tomorrows weather.

I can program another a MinTtomorrow variable into IWeather and have it go get this from the met file.  However, this feels quite messy and I am wondering if anyone can think of a better way.

hut104:
I assume that you are using the parton model for diurnal temperature.  We would need to work through a use-cases perhaps.  We use diurnal temperatures currently to:

Calculating TT in plant model - need to iterate through the day
Calculating VPD in microclimate - need to iterate through the day
Calculating soil temperatures - not sure, but may need temperatures at different specific times

It may be that we could provide an array of hourly temperatures, or a method that returns the temperature for a time of day (0-24hours).  We could do a similar thing for solar radiation for the photosynthesis models.

This would also allow us to provide tests of the model because we could have Pvs O diurnal temperature data in there.

________________________________

HamishBrownPFR:
@hut104 Take a look at the AirTemperatureFunction.  I have already modified it to be able to use different interpolation methods.   This can be used by any model, not just plant.  Perhaps this needs to be generalized further so it can interpolate other met variables such as radiation.  That would not be to hard to do.  Would require a different interpolation method to what is used for temperature but I have structured it so you can change in different interpolation methods.  I have already set it up so it can return the average or sum of sub daily values.  Perhaps we should generalise AirTemperaturefunction further to be a SubDailyIntepolationFunction.  That would not be hard.  @jbrider how are you doing the sub daily interpolations for the canopy photosynthesis modelling?

Anyway, we are implementing a 24 hour interpolation that uses a sign during the daylight period and exponential during the dark and needs to know tomorrows Tmin to do this.  My question here is what is the best way to go about getting TminTomorrow.

Keith-Pembleton:
There are a number of use cases where it would be handy to have APSIM look forward in the met data.  For the simulations behind CropARM we needed to do this and had to come up with a work around that was very messy and slowed down the simulations.  What ever happens here it should be expanded beyond the next day and also beyond temperature.

HamishBrownPFR:
@Keith-Pembleton  How much further ahead do you need to look?

[...]

PR #5123, opened 2020-05-06
Hourly temp

junqi108:
working on #5113 

 calculating the hourly temperature based on Tmax, Tmin and daylength
  At sunrise (th = 12 − d/2), the air temperature equals Tmin. The maximum temperature 
   is reached when th equals 12 + p h solar time.The default value for p is 1.5 h.
    The sinusoidal curve is followed until sunset.Then a transition takes place to an
    exponential decrease, proceeding to the minimum temperature of the next day.To
    plot this curve correctly, we first need the starting point, the temperature at sunset
    (Tsset).


BenAbabaei:
I have implemented something here, if you wanna have a look:
https://github.com/APSIMInitiative/ApsimX/issues/5121#issuecomment-624965539

junqi108:
> I have implemented something here, if you wanna have a look:
> [#5121 (comment)](https://github.com/APSIMInitiative/ApsimX/issues/5121#issuecomment-624965539)

Thanks Ben.

PR #5126, opened 2020-05-07
Map should automatically zoom out to fit all markers

### LINKED ISSUE #3184: Map doesn't work properly in auto-documentation
The google map doesn't seem to zoom to an appropriate region of the globe when maps are inserted into auto-documentation.  The eucalyptus validation file inserts a map of Africa.  There are no experiments from Africa.

hol430:
Resolves #3184

Note that if all markers are already visible onscreen, the map *shouldn't* zoom out.

### DIFF
--- a/ApsimNG/ApsimNG.csproj
+++ b/ApsimNG/ApsimNG.csproj
@@ -394,6 +394,7 @@
     <Compile Include="Views\TitleView.cs" />
     <Compile Include="Views\TreeProxyView.cs" />
     <Compile Include="Views\ViewBase.cs" />
+    <Compile Include="Views\WindowView.cs" />
     <Compile Include="Views\XYPairsView.cs" />
     <Compile Include="Utility\MacUtilities.cs" />
   </ItemGroup>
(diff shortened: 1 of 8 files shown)


zur003:
Is this desirable? I can imagine a situation where there are markers around the world, but for whatever reason, the user wants to focus on just those in Australia.

hol430:
It's not really ideal but the problem with the way that this currently works is that the size of the map in the apsim gui is different to the size of the map rendered by the autodocs (the autodocs uses a fixed size whereas the map in the gui seems to be pased on the user's screen resolution). So even if I get the map looking just the way I want it in the UI, when the autodocs are generated the map is a different resolution which leads to unexpected results.

Actually I just did some extra tinkering on this branch and my 'fix' doesn't work well if the markers are clustered together. I added a map to the wheat example which has only 1 met file and the map zoomed in on the dalby eagle boys pizza (which I can relate to but it isn't really what we want it to do...).

I wonder if a better solution would be to use a fixed-sized map in the UI? This would get around the problem of incosistencies between autodocs/UI although it might make it a bit difficult to use the UI on low-res devices like phones.

zur003:
I shudder at the thought of using the APSIM GUI on a phone. Perhaps we can add an option in the MapView to allow the user to preview the map at the size used in the auto-docs?  That would probably require some additional fields for storing/using the auto-doc size and position separately from the interactive.

hol430:
Yeah, not a bad idea. Will see what I can do.

Issue #5137, opened 2020-05-11
Addition of N2O datasets to Test Set


hut104:
Improvements can be made to the APSIM Nutrient model for N2O emission modelling.  Test data will need to be added to the APSIM test set before this can be undertaken.
Data will incorporated from
Mielenz Henrike, Thorburn Peter J., Harris Robert H., Officer Sally J., Li Guangdi, Schwenke Graeme D., Grace Peter R. (2016) Nitrous oxide emissions from grain production systems across a wide range of environmental conditions in eastern Australia. Soil Research 54, 659-674.

https://doi.org/10.1071/SR15376

PR #5141, opened 2020-05-11
Commit initial version of Horsham dataset

hut104:
working on #5137 
@hol353 please do not merge this as yet


hut104:
this can be committed now

PR #5151, opened 2020-05-12
Add small time-based delay when exporting map to PDF

### LINKED ISSUE #3184: Map doesn't work properly in auto-documentation

hol430:
Resolves #3184 

@zur003 - would you be able to test the autodocumentation on macos when there's a map in the simulation?

### DIFF
--- a/ApsimNG/Classes/PDFWriter.cs
+++ b/ApsimNG/Classes/PDFWriter.cs
@@ -392,13 +392,20 @@
                                 map.HideZoomControls();
 
                             popupWin.ShowAll();
+
                             while (Gtk.Application.EventsPending())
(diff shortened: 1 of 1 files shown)


zur003:
@hold430 - Before or after pulling from hol430:bug/Map?

hol430:
After - I'm hoping it should be fixed.

zur003:
@hol430 When I generate documentation, I see the map appear for a couple of seconds on the display, but end up with just a grey box in the PDF.

PR #5161, opened 2020-05-14
Changed folder view so it doesn't hide legends by default

### LINKED ISSUE #5153: Graph legends don't appear as expected
When using multiple graphs on one view the legends don't show up until you click on them with the mouse. Noticed on Mac and Windows.
image

hol430:
Resolves #5153 

To keep the legends hidden, disable them in the series settings.


hol353:
This change may annoy a few people (eg @HamishBrownPFR, @hut104). It is more consistent though. If you tick 'show legend' in the graph series the legend will appear on the graph regardless of whether it is 'full window', part of a page of graphs or part of a page of graphs in the auto-doc. Previously, legends were hidden on graphs that are on a page of graphs, regardless of whether 'show legend' is ticked.

PR #5165, opened 2020-05-15
PredictedObserved should check for duplicate observed datapoints.

### LINKED ISSUE #4708: Duplicate Observed Data
The Wheat model test set appears to have duplicate data for flag leaf and anthesis date for plantings in the CPT dataset.  this may be due to differences in observations in the original plot data.  It would be worth investigating to get a single value for flag/flowering dates for each treatment.

### LINKED ISSUE #4710: Duplicate observed data
I raised an issue with duplicate observed data in the wheat validation set (#4708 ).
I only noticed this because I had changed an input file and the Performance Tests were documenting changes to my observed data (!).  The observed data was not in fact changing, but the order in which it was reporting these when P&O data was merged did change and so it looked like individual values were changing.
My deeper question is, should we be throwing an error if our merging of predicted and observed data finds multiple values of either. e.g.
1) there are two outputs from the model on a given day for a given simulation
2) there are two values for observed data on a given day (ie ambiguous value for expected result)
Our merging criteria, in some way, excludes the idea of duplicates given that we only merge on simulation name and date.

I know that people have tried to merge on rep/plot level data in the past to show variation in the observed data.  However, that might be OK for timeseries data, but not so much for P vs O data tables.

Our lives may be easier if we stop people doing this. Our tests are a little bit confusing with this sort of thing.

hol430:
And by check, I mean throw a fatal error if any are found.

Resolves #4710 and resolves #4708

@hol353 - this is for damage assessment purposes and will not run green. If it does run green, please don't merge it.

### DIFF
--- a/Models/PostSimulationTools/PredictedObserved.cs
+++ b/Models/PostSimulationTools/PredictedObserved.cs
@@ -278,6 +278,7 @@ HAVING N > 1";
             if (duplicates.Rows.Count > 0)
             {
                 StringBuilder msg = new StringBuilder();
+                msg.AppendLine($"Error in file {System.IO.Path.ChangeExtension(dataStore.FileName, ".apsimx")}:");
                 msg.AppendLine($"Error in PredictedObserved {Name}: duplicate records detected for {duplicates.Rows.Count} simulations in observed table {ObservedTableName}:");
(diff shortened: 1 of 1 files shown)


hol430:
I'm going to close this until there's more action on #4710.

PR #5173, opened 2020-05-18
Stops Stock failing when NaN value for N from AgPasture

### LINKED ISSUE #5171: AgPasture biomass N causing Stock to fail
When the biomass Dead.N is queried by stock and the Dead.Wt is 0 the N value is found to be NaN. This has caused Stock to fail with a cascade of NaN values.

her123:
resolves #5171


hol430:
retest this please jenkins

hol430:
retest this please jenkins

PR #5178, opened 2020-05-18
Don't use %root% when creating relative paths

### LINKED ISSUE #2371: errors in the address of file when using the released APSIMX vesion
the address of file going wrong when using the released APSIMX version for those models in the prototype folder. see the photo below.
image
selecting the data again and run it can make it works.

hol430:
Resolves #2371


hol430:
retest this please jenkins

PR #5179, opened 2020-05-19
Force emergence

### LINKED ISSUE #5168: Need to specify emergence date from a manager script
In model validations we often want to force emergence to occur on observed dates so all the time coarse data lines up properly there after.
Currently this is done by twidling sowing depth which is cumbersome and in time will be forgotten and the sowing depths in simulations may be taken as an actual record of sowing depth.
We need to create a ForceEmergence event that can be called from a manager script and have the EmergingPhase subscribe to in and push it progress to the target so emergence occurs that day.

HamishBrownPFR:
resolves #5168


HamishBrownPFR:
@hol353  could I get this one merged please

PR #5188, opened 2020-05-21
Added code into Model/IModel to locate other models

### LINKED ISSUE #3061: Make Apsim API use generics
Calls to most of the static Apsim methods which go looking for a model (such as `Apsim.Parent()`) are almost always followed by a cast. If we make these methods generic, we can avoid the need for such a cast.

For example, instead of 
```
public static IModel Parent(IModel model, Type typeFilter) { }
...
Simulations parentSimulations = Apsim.Parent(this, typeof(Simulations)) as Simulations;
```

We could have
```
public static T Parent<T>(IModel model) { }
...
Simulations parentSimulations = Apsim.Parent<Simulations>(this);
```

hol430:
This will eventually resolves #3061 

@hol353 - please don't merge this yet. Just putting it up here so you can see what I've implemented so far. As you can see, most of the search by type stuff is very simple - one line only. I haven't added any search by name functions yet (they will be very easy to implement but writing tests takes time). Class model is maybe bit cluttered, but I think we should at least have these functions somewhere. Many users who write manager scripts will struggle with linq, and even for C# programmers, one of these is easier to read than the other.

```csharp
Experiment expt = zone.Ancestor<Experiment>();
Experiment expt = zone.Ancestors().OfType<Experiment>().FirstOrDefault();
```

What are your thoughts?


hol430:
retest this please jenkins

hol353:
Looks good. Going to be a big API, although there is a pattern with the methods (returning single instance matching on type, returning multiple instances matching by type, returning all instances).

What about 'Get'?

What about changing the name Find and FindAll to better match the other methods and to better say what they return? InScope and InScopeAll (don't like the 'All' suffix but can't have InScopes)? Other suggestions?

hol430:
How about `AllInScope` instead of  `FindAll` and `InScope` instead of `Find`?

hol353:
Trying to avoid the verb 'Find' because none of the other methods have one. Lets go with 'InScope' and 'InScopeAll'.

hol430:
@hol353 how would you feel about renaming the `Get()` function to something more meaningful? The name `Get` doesn't really convey much meaning by itself...what are we getting? The key difference between `Get` and the other functions is that you pass a path rather than a name into `Get`. None of the other function names in this API contain a verb, so perhaps we could call it something like `VariableOrModel(...)`, or `FromPath(...)`. I don't really like either of these all that much.

`InScope()` and `InScopeAll()` are also a little odd - `InScope()` in particular sounds like a boolean function designed to check if another model is in scope. I could get used to these though, and they at least give you some idea of what the function does, unlike `Get`.

Any thoughts?

hol353:
According the the Microsoft naming guidelines we should use verbs for method names

https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/4df752aw%28v%3dvs.71%29

This will make the names longer but clearer. So perhaps Descendants should be FindDescendants() etc. This doesn't match the 'Children' property though. The other requirement is that properties have a very simple implementation and are not computationally complex.

So perhaps FindInScope works. The problem with Get is that it can return a variable value or a model. It is very general - it gets something. Perhaps FindByPath?

hol430:
FindInPath sounds good. I wonder if we need both Get and GetVariableObject. Currently Get is essentially defined as

```csharp
return GetVariableObject(args)?.Value;
```

I'm not sure that this is actually useful.

hol353:
I don't think we need the other version of FindInPath that simply returns the .Value of FindInPath.

Perhaps FindInPath  should be FindUsingPath or FindByPath. 

The rest of the methods look good. There is a lot of them.

hol430:
@hol353 another problem is that Apsim.Parent() can return the model you pass in. E.g.

```csharp
Simulation sim = ...
IModel parentSim = Apsim.Parent(sim, typeof(Simulation));
```

This will return the sim which we pass in. The new FindAncestor methods I've added don't do this (is a model really its own ancestor?) but this "feature" seems to be quite widely used in our codebase. How should we handle this? Do we want to add another method called something like FindModelOrAncestor? Is it the caller's responsibility to check the type of the model being passed in? Or do we modify the FindAncestors method to return the model you pass in as well (I would rather not do this but it would certainly be the easiest solution).

hol353:
FindAncestor shouldn't return the model that is passed in. I don't think we want another method that does though. We'll have to make sure the callers are aware of this new functionality.

hol430:
Agreed, it will be painful but doable. What about manager scripts?

hol353:
Hopefully there aren't many (any?) manager scripts that call the existing Apsim.Parent. Don't know if we can write a converter to move them to the new API. Probably very difficult.

hol430:
@hol353 this is failing for me in the stock tests. I have added a property to IModel.cs called FullPath, which is not supposed to be serialized because it doesn't have a setter. However, the unit tests are failing because there seems to be some special stock serialization code which is serializing the FullPath property. Why are we not using the standard serialization code (which would ignore non-writable properties)?

hol353:
Because the stock tests want to test the value of every field / property regardless of whether the property is settable or not. The standard serialisation code can't do this.

hol430:
retest this please jenkins

hol430:
@hol353 this is now at the point where the code in Apsim.cs is just calling the new model locator API. I am ready to start removing the functions from Apsim.cs (in fact I've already started with Apsim.Get()). However, writing converters for manager scripts is going to be difficult (impossible?) for the generic methods since you can't pass a runtime type as a generic parameter. Well, technically it is possible via reflection, but I really don't want to do that from the converter.

I'm thinking the best solution might be to slap an `[Obsolete]` attribute on these methods, telling users that their script will break in a future update and then remove these methods in say 6 months time. Or we just remove the methods now and break people's scripts with no warning - which we've done before and will probably do again. This will only affect the methods which filter by type - I should be able to fix the other method calls via regex.

hol430:
@hol353 see above comment, what do you think is the best approach here?

hol353:
Which generic methods are you talking about? Show some code for old syntax vs new syntax. I don't think users are using any of the generic methods in Apsim.cs.

hol430:
I mean the methods which accept a runtime type as an argument and which have been converted to generic methods in the new API. E.g.:

```csharp
// this usage should be possible to automatically fix with converter:
// old:
IModel sim = Apsim.Find(this, typeof(Simulation);
// new:
Simulation sim = this.FindInScope<Simulation>();

// this will be very difficult to automatically fix with converter:
// old:
IModel model = Apsim.Parent(this, x.GetType());
// new:
IModel model = Apsim.FindAncestor<???>();
```

I suspect that there are quite a few manager scripts out there which use these methods where you search for a model by type - e.g. scripts which expose variables for a report.

hol353:
I suspect the first usage is by far the most common so why not do the conversion for it.

For the second usage, just have the converter do what you suggest and convert it with ??? e.g. ```IModel model = this.FindAncestor<???>();```

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

PR #5194, opened 2020-05-22
One manager can call another manager

### LINKED ISSUE #4859: Can't call a method in a manager script from another manager script
Currently, if you link to a manager model in another manager, it has not object called script so it is not possible to call methods that are declared in that manager.

hol353:
Resolves #4859


Models/Core/Simulation.cs
```diff
@@ -107,6 +107,9 @@ public Locater Locater
             }
         }
 
+        /// <summary>Is the simulation running?</summary>
+        public bool IsRunning { get; private set; } = false;
```
hol430, line 111: This doesn't need to be serialized, right?

```suggestion
        [JsonIgnore]
        public bool IsRunning { get; private set; } = false;
```

Models/Manager.cs
```diff
@@ -97,274 +107,104 @@ private void OnCommencing(object sender, EventArgs e)
...
 
         /// <summary>Rebuild the script model and return error message if script cannot be compiled.</summary>
         public void RebuildScriptModel()
         {
-            // This is called from manager presenter on detach so go refresh our parameter values.
-            if (Children.Count == 1)
-                GetParametersFromScriptModel(Children[0]);
-
-            if (Enabled && isCreated && Code != null && (Code != CompiledCode || Children.Count == 0))
+            if (Enabled && afterCreation && !string.IsNullOrEmpty(Code))
```
hol430, line 117: Does this mean the script will be recompiled even if it hasn't changed?

Models/Manager.cs
```diff
@@ -87,8 +96,9 @@ public string Code
         private void OnCommencing(object sender, EventArgs e)
         {
             // Need to ensure script parameters are up to date.
-            if (Children.Count > 0)
-                SetParametersInObject(Children[0]);
+            afterCreation = true;
+            RebuildScriptModel();
```
hol430, line 100: I think this will cause a test failure. If everything runs green, ignore this comment. Otherwise, you might want to look at 6d25e5d4312c54defb45e156573e7ef692166356. I also noticed you commented out a line in SimulationDescription.cs:122 which is related to this.

Models/Core/ScriptCompiler.cs
```diff
@@ -0,0 +1,239 @@
...
+                        compilation.CompiledAssembly = result.CompiledAssembly;
+                    }
+                }
+                else
+                    newlyCompiled = false;
+
+                if (compilation != null)
+                {
+                    // We have a compiled assembly so get the class name.
+                    var regEx = new Regex(@"class\s(\w+)\s");
```
hol430, line 87: Do we need `\s+` here in case there are multiple spaces between the keyword class and the class name?

Models/Core/Simulation.cs
hol353: It won't be serialised anyway because it has a private setter.

Models/Manager.cs
hol353: Nope. The new ScriptCompiler instance looks after that.

Models/Core/ScriptCompiler.cs
hol353: Fixed.

Models/Manager.cs
hol353: Reverted back to how it was. The sequencing of OnCreated and the fact it is called multiple times is a mess.

hol353:
retest this please jenkins

Issue #5199, opened 2020-05-25
Root class can not handle retranslocation and reallocation


junqi108:
Root class can not handle retranslocation and reallocation. 
in short term we can use generic organ to represent the storage part. in long term we may want to add this capacity, especially for perennial plant. 
 public void SetDryMatterAllocation(BiomassAllocationType dryMatter)
        {
            double TotalRAw = 0;
            foreach (ZoneState Z in Zones)
                TotalRAw += MathUtilities.Sum(Z.CalculateRootActivityValues());

            Allocated.StructuralWt = dryMatter.Structural * dmConversionEfficiency.Value();
            Allocated.StorageWt = dryMatter.Storage * dmConversionEfficiency.Value();
            Allocated.MetabolicWt = dryMatter.Metabolic * dmConversionEfficiency.Value();
..............
}

junqi108:
The increase in storage biomass is calculated. but when there is re-translocation, the amount of biomass retranslocated does not minus from the total storage biomass.

PR #5200, opened 2020-05-25
retranslocation error

junqi108:
working on #5199

### DIFF
--- a/Models/Plant/Organs/Root.cs
+++ b/Models/Plant/Organs/Root.cs
@@ -677,6 +677,9 @@
             GrowthRespiration = (Allocated.StructuralWt + Allocated.StorageWt + Allocated.MetabolicWt) * growthRespFactor;
             if (TotalRAw == 0 && Allocated.Wt > 0)
                 throw new Exception("Error trying to partition root biomass");
+            
+            if (dryMatter.Retranslocation > 0 | dryMatter.Reallocation > 0)
+                throw new Exception("Error Root class can not handle Retranslocation or Reallocation > 0");
 
             foreach (ZoneState Z in Zones)
                 Z.PartitionRootMass(TotalRAw, Allocated);
--- a/Models/Plant/Organs/Root.cs
+++ b/Models/Plant/Organs/Root.cs
@@ -65,9 +65,6 @@
     [ValidParent(ParentType = typeof(Plant))]
     public class Root : Model, IWaterNitrogenUptake, IArbitration, IOrgan, IOrganDamage
     {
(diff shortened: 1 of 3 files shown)


junqi108:
@HamishBrownPFR only grapevine model through errors in the root storage retranslocation

HamishBrownPFR:
Thats good.  Delete Lines 127 - 146 as well.  These functions are redundant so we can tidy up the root class.  That will also fix the error grapevine is flowing.  For completementss it would be good to put the same error checks in the N Allocation code for root.

image

junqi108:
@HamishBrownPFR @hut104 OK, I will delete those lines first. How difficult it would be to include the re-translocation function in Root? it seems to add structural and storage biomass is very easy as we did last time in the PartitionRootMass? will make the TotalDMAllocated negative works if the retranslocation is very big? 
public void PartitionRootMass(double TotalRAw, Biomass TotalDMAllocated)
        {
            DMAllocated = new double[soil.Thickness.Length];
 
            if (Depth > 0)
            {
                double[] RAw = CalculateRootActivityValues();

                for (int layer = 0; layer < soil.Thickness.Length; layer++)
                    if (TotalRAw > 0)
                    {
                        LayerLive[layer].StructuralWt += TotalDMAllocated.StructuralWt * RAw[layer] / TotalRAw;
                        LayerLive[layer].StorageWt += TotalDMAllocated.StorageWt * RAw[layer] / TotalRAw;
                        LayerLive[layer].MetabolicWt += TotalDMAllocated.MetabolicWt * RAw[layer] / TotalRAw;

                        DMAllocated[layer] += (TotalDMAllocated.StructuralWt + 
                                               TotalDMAllocated.StorageWt +
                                               TotalDMAllocated.MetabolicWt)* RAw[layer] / TotalRAw;
                    }
            }
        }

junqi108:
@rcichota @HamishBrownPFR Hi Rogerio, i removed some unwell performed function in the root class. it seems it affects the clover model performance marginally. Just would like to discuss with you whether we can set root retranslocation factor to zero for clover?

rcichota:
Hi @junqi108, happy to discuss this further. Are your simply deleting retranslocation functions or replacing them? Isn't this an important factor for re-growth over spring?
I had played a bit with adding retranslocation from roots in the models I was building, these are quite important for perennial forages (though specific data is not that easy to come by). I had removed from Chicory and Plant when I submitted them, but I think I left them in WhiteClover and maybe RedClover, the effect is small as most retranslocation should come from stolons and taproot. We can either replace the logic or even maybe delete and accept the changes if small...

Issue #4080, opened 2019-08-12
Port the old apsim SoilTemp2 model


hol430:
SoilTemp2 is a numerical solution of soil temperature changes. It is a port of the original model in Classic and is based on the solution in Campbell's "Soil Physics with Basic". There have been various reports that this model performed better than the older CERES soil temperature model. The proposal is that once SoilTemp2 is ported, the CERES model would be removed.

sarahcleary:
Comment from @yashvirchauhan:  I have been using Soiltemp in APSIM classic for quite a while. Tthe major difference in soil temperature simulated by Soiltemp and that simulated by soilN module is when soil is covered by the crop canopy. Soil N based simulated temperature is much higher (>2-5oC) indicating it is not able correctly simulate soil temp when radiation reaching soil is reduced - if I correctly remember. Soil temperature is used in predicting aflatoxin risk in peanuts and the values predicted by soil temp were more realistic.

sarahcleary:
@sarchontoulis to provide Validation data

HamishBrownPFR:
@hol353 I have some data that I would be keen to have used in the soil validation and I think will be very usefull.

1. Soil temperature at 2, 15, 30, 90, 120, 150 and 180 cm, heat flux at 15cm, and crop radiation interception over entire crop duration's under barley, wheat, oats, fodder beet and potato with 6 treatments (and 4 reps for each crop) giving differences in canopy cover patterns
2. Soil temperature at 7.5, 22.5, 37.5 and 52.5 cm in lysimeters with 0, 30 and 50% stones in the top soil layer in factorial combination with gravels at 30 or 60 cm.

These two sets should provide a wide range of soil temperature profiles and fluxes

3. Soil temperature at 1.5 cm depth under wheat and barley crops for a range of sowing dates.  It is going to be important to have the surface temperature well validated for soil temp so we can look to use it for driving early crop development.

I am happy to pull datasets and simulations together if you want to include them

hol353:
Yep that would be great. The more validation data we have the better.

sno036:
Sounds good to me. The WWEPP validation is looking pretty good at the moment.

zur003:
@sno036 - Are you actively working on this? I may have a bit of time in the next few weeks to have a go with it, if it's a priority.

hol353:
No one at AgResearch is working on SoilTemp at the moment. Yes, it is a priority. It seems to be working - just needs some tests and validation data.

[...]

PR #5201, opened 2020-05-25
Allow the "@" character to be used in Report aliases (along with the …

zur003:
…current alphanumeric, "_" and "." characters). Working on #4080


hut104:
@hol353 and @zur003 Is there a reason why we would allow "." in variable names?  That should surely cause confusion if I could alias [Wheat].Leaf.LAI to Wheat.LAI

zur003:
@hut104 - Why would aliasing with the "." character cause any more confusion than is possible without it? I can easily alias [Wheat].Leaf.LAI to RootingDepth, which would be even more confusing than your example.

hut104:
People may start to think that there is an LAI member on Wheat.  I guess it comes back to the fact that "." means something already.  It may become really confusing when debugging things if it is not clear what is real and what is alias.  At least the "_" character is not currently used in APSIM for anything (It took us years to get rid of it to clean things up).

I think I also suggested this becaues I know that we have moved away from this several times in the past for these reasons.

Issue #5204, opened 2020-05-25
Change Wheat Model to differentiate between heading and flowering


hut104:
The current wheat model has a heading stage and a flowering stage, but they occur together.  They need to be pulled apart.  Datasets will need to be added to test this.

PR #5205, opened 2020-05-25
Wheat Heading Date

hut104:
working on #5204


hut104:
@hol353 and @hol430 OK, so this failed the test, but there is nothing different in the tables.

I think this is because I added a new field "HeadingDAS" but it doesn't appear in the table. It is correct that the stats are different.  It just doesn't seem to be showing all the results.

hol430:
Looks like the observed data has changed somehow:

image

hut104:
retest this please jenkins

hut104:
@hol353 I think we can merge this.  I've added new data and these are the only things that have changed.

hol353:
retest this please jenkins

PR #5210, opened 2020-05-26
Added Chinese Data to Soybean Tests

hut104:
working on #2731


hut104:
@hol353 I've added the new data provided by Enli as part of his Soybean model review.  It changed the stats.  I think we can commit this. No change to the model itself.

PR #5214, opened 2020-05-26
Grapevine

junqi108:
working on #1822


junqi108:
retest this please Jenkins

junqi108:
@hol353 is there something changed in the performance stats? i just changed the grape model, so should not affect other model's performance.

hol353:
There is currently a problem with the performance site. @hol430  is working on it.

hol353:
There is an 'is optional' function in this PR. Can we find a different way to do this?

HamishBrownPFR:
The alternative is to write a converter that puts a constant function with a value of zero in every Leaf object.   Which is the worse evil, optional functions or the proliferation of redundant parameters?  I can help @junqi108 write a converter if that is what you prefer.  Let us know

HamishBrownPFR:
@hol353 sorry, forgot to tag you to previous comment

junqi108:
@hol353 hi dean could you merge this one first as it has lots of commits. if needed we can change the isoptional in a different pull request, see my latest pull request

hol353:
Instead of an optional function, I think it is better to write a converter that adds a constant to all leaf models in all plants that use Leaf. Agree @hut104 ?

PR #5225, opened 2020-05-28
Trim model names upon renaming in gui.

### LINKED ISSUE #5108: Trim model name on rename in UI tree
Is anyone opposed to trimming spaces from the name of a model when the user changes it in the tree with "Rename"? A space on the end of a model name is currently accepted but now causes problems with the new invalid path error reporting in Locator as the model names do not match the namebits that have had empty spaces trimmed.

hol430:
Resolves #5108


hol430:
retest this please jenkins

PR #5227, opened 2020-05-29
add optional canopy width and depth to the leaf class

junqi108:
working on #1822 

for using the DoTreeRowCropShortWaveRadiation method, we need a different canopy width and depth function than the one in current leaf class. 
This is quite specific to strip crops with more than two zones. other models will not require it. so probably make it optional is tidier and easier. 

 foreach (MicroClimateCanopy c in tallest.Canopies)
            {
                if ((c.Canopy.Height - c.Canopy.Depth) > 0)
                    TallestIsTree = true;
            }
            if (TallestIsTree)
            else
                DoStripCropShortWaveRadiation(ref tallest, ref shortest);

### DIFF
--- a/Models/Plant/Organs/Leaf.cs
+++ b/Models/Plant/Organs/Leaf.cs
@@ -189,10 +189,10 @@ namespace Models.PMF.Organs
 
         /// <summary>Gets the depth.</summary>
         [Units("mm")]
-        public double Depth { get { return Structure.Height; } }
+        public double Depth { get; set; }
 
         /// <summary>Gets the width of the canopy (mm).</summary>
-        public double Width { get { return 0; } }
+        public double Width { get; set; }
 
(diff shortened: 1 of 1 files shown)


hut104:
Hi, 
The team has been trying to remove all the optional functions.  If it is optional, and is not used in the code, perhaps these could just be added to the crops that need it as functions in the Json?

HamishBrownPFR:
All crops need it because it is in ICanopy.  However, Leaf has it set to zero in the background (Simple Leaf is the only one that sets it at the moment).  The idea of making it optional on Leaf was so it could be specified in JSON for the crops that need and the default remains for those where it is not specified.
Next best thing is to write a converter to put a default value into the json for all crops

junqi108:
got a conflict in the file. i will make a new pull request with the latest updates

PR #5243, opened 2020-06-02
Leaf canopy width depth

junqi108:
working on #1822

for using the DoTreeRowCropShortWaveRadiation method, we add canopy width and depth function than the one in the leaf class.
The canopy width and depth are only calculated after plant is emerged in the leaf class. this is not appliable for perennial crop at the start of simulation. so width and depth are reset at sowing. 

fixed couple of errors for the grapevine model after using the canopy depth and widht as well.


junqi108:
@HamishBrownPFR i have submitted the pull request.

junqi108:
thanks, Dean. this was quick.

PR #5244, opened 2020-06-03
Apsim wheat

HamishBrownPFR:
working on #4773


HamishBrownPFR:
@hol353   This is OK to merge, some new results added and some management changed which is why stats have changed.  Some poor fits are from new cultivars that have been added but not fitted yet

zur003:
Merging this pull request seems to have broken builds for me. The treeview image AGPBrowntop.png is included in the project, but seems to have been deleted.

hol430:
I think it's because there used to be AGPBrowntop.png and AGPBrownTop.png. Hamish deleted AGPBrowntop.png, but because ntfs is not case-sensitive, things get a bit mucked up (the problem doesn't show up for me on Linux). Discarding the changed should fix the problem.

zur003:
@hol430 - I can see how that might have caused the problem, but I'm surprised it didn't break the Jenkins build.

zur003:
@hol430 -  And for me, it's not working on Linux either. There I have a AGPBrownTop.png file, but it's given as AGPBrowntop.png in the .csproj. This needs to be fixed.

hol430:
Hmm, good point about jenkins, I think the reason it ran green is because at the start of the build, we  discard any local changes and ignored/untracked files, to prevent any leakage between builds. This would probably have un-deleted AGPBrowntop.png, allowing the build to pass.

I'm now getting the same result as you on Linux on latest master - no missing file, but compilation fails. I must have been on the wrong branch when I checked earlier.

hol430:
Regarding the name - is browntop one word or two? I would say if it's one word, then lower case t makes sense.

zur003:
Google searches seem to show is as "browntop" or "brown-top", not as "brown top", but you'd have to ask the Kiwis.  I'd go with AGPBrowntop.png.

PR #5248, opened 2020-06-04
Grapevine

junqi108:
working on #1822


junqi108:
@HamishBrownPFR I added a new vine strip crop light interception method as the tree row crop method does not apply for the grapevine condition. the canopy width of a grapevine is smaller than the strip width (herbicide strip), about half, and its leaves are all concentrated in the above part. so it is different to both the tree row and crop strip light interception method.

junqi108:
changed the sequence of depth updating in leaf as it is not updating for perennial trees before budburst

junqi108:
retest this please Jenkins

PR #5261, opened 2020-06-04
Improved error handling in main presenter

### LINKED ISSUE #5260: apsimx closes when opening the standard toolbox
The title says it pretty much, I have tried to open the standard toolbox with a couple of versions of APSIM, including the most recent one (APSIM2020.06.04.5250), and the window would systematically close without any error or warning message. I am using APSIM with Windows 10 Pro, 64 bits, with most of default parameters set for French (apart from the dot that I use as the decimal deparator).

hol430:
Resolves #5260

@her123 @zur003 this crash was caused by an error in GrazSupp.cs:1611, which happens when opening the standard toolbox on windows with French regionalisation settings. The error message is `Error reading supplement library - line for Green maize forage is invalid`. I've fixed the crash, and I've added a quick and dirty fix to GrazSupp to allow the standard toolbox to be opened, but I suspect there may still be an issue with GrazSupp which may be worth looking into.

### DIFF
--- a/ApsimNG/Presenters/MainPresenter.cs
+++ b/ApsimNG/Presenters/MainPresenter.cs
@@ -479,15 +479,22 @@
         /// <param name="e">Event arguments.</param>
         public void OnStandardToolboxClick(object sender, EventArgs e)
         {
-            Stream s = Assembly.GetExecutingAssembly().GetManifestResourceStream("ApsimNG.Resources.Toolboxes.StandardToolbox.apsimx");
-            StreamReader streamReader = new StreamReader(s);
-            bool onLeftTabControl = true;
-            if (sender != null)
+            try
             {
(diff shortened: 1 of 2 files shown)


zur003:
@hol430 there is definitely still an issue stemming from parsing the supplement resource values with the current culture rather than the invariant culture. It's easily fixed. Should it be done as a separate issue? If not, how should I merge the needed changes with yours (which should also be applied)?

hol430:
I guess the easiest way might be to just push up to this branch?

zur003:
@hol430 can I do that?

hol353:
Have you guys had enough fun with this PR. Want it merged?

zur003:
I say go for it.

hol430:
Yes, we have had enough fun

PR #5262, opened 2020-06-04
Develop

### LINKED ISSUE #5260: apsimx closes when opening the standard toolbox

zur003:
Resolves #5260
Includes mods from PR #5261


hol430:
That's probably the easiest solution, yeah

Issue #480, opened 2015-10-19
PMF class doco for developers


HamishBrownPFR:
The view for each PMF class should display some infromation that is useful to developers and is in sync with the source code.
1. Some instructions that will guide developers but would not be included in science documentation
2. A list of the classes dependencies
3. A list of child functions that provide parameters to the class
4. These need to fit in with the current grid view for setting class properties.

1 could be achieved by picking up some kind of tag out of the source code and displaying it on the class view.  Problem is the Science doco already uses the summary tag.  Not sure of the solution here
2 could be done by making a list of Link not of type Function
3 could be done by making a list of Links of type Function and also displaying if they are optional or not.

HamishBrownPFR:
This is now relevant to FunctionL

sno036:
Can this be bumped up the priority list (along with all the other top-priority issues :-))?  

From Hamish following him fixing one of my simulations:
"Someone needs to write a class view for the UI that displays a list of the classes properties and parameters (including units) so you don’t have to go to the source code to work them out.  There is an issue (# 480).  I suggest posting something on this saying lack of doco is causing problems and this might spur some action in getting it working.

I think you may have inadvertently changed something in you SLURP model under replacements.  I deleted this and ran the simulations again (which means it will use the XML in the resources file) and the N taken up numbers became sensible.  This is the down side of using the replacements node to view the model, it makes accidental changes possible.

A solution to this would be if the SLURP node (or any crop for that matter) showed the tree view of its structure so users can look at the model and learn how it works but doesn’t allow setting of parameters.  Not sure how feasible this is.  In the meantime copy a fresh lot of The SLURP xml from the resources file and then comitt.  Then if you are getting weird behaviour diff your .apsimx file and check there are not any unintended changes to the Slurp xml."

HamishBrownPFR:
@hol353 
This issue has always bugged me but the need for some better developer info on LifeCycle models has moved it up my priority list.  I can have a go at implementing something here but not really sure how the views and presenters work.  
Would I be able to have a chat with you at some stage to get some guidance on how best to approach this.

hol353:
Happy to have a chat any time. If we can spec it up in sufficient detail then @hol430 could implement. The question is what do we want?

HamishBrownPFR:
Some updated thoughts following discussion with @hol353 .
Need to distinguish between two types of documentation.
[...]

PR #5272, opened 2020-06-08
Display some documentation in the user interface

hol430:
Working on #480

This will only affect models which don't have a view name or presenter name specified. Plant is one example - so if you open the wheat example and click on the wheat model, you should see some details about its inner workings.


hol353:
Why not use GenericPresenter with a HTMLView for models that don't have a specified presenter / view. STOCK uses the generic presenter to show model documentation.

hol430:
If I recall correctly, the generic presenter just shows the contents of the model's summary tag (or calls the `Document()` method if the model is an `ICustomDocumentation`). This presenter is a bit different - it shows more technical info about the model's inputs/outputs/events/etc.

PR #5285, opened 2020-06-11
Grapevine2

junqi108:
working on #1822

previous pull requests slightly affect the stock model. do not know why. so made a new clean version


junqi108:
retest this please jenkins

junqi108:
@hol353 hi dean, i tested in debug mode by adding breaks to the changes i made in microclimate.cs. Stock model does not go into that method. so my changes should not affect the stock performance. maybe there is some slight changes in the stock model stats.

Issue #3989, opened 2019-07-02
temporary db files left behind


rcichota:
This is as much a question as an issue. I noticed that there are now some temporary db files that are not cleaned up after a simulation run (see figure). These can be deleted manually without problem, but shouldn't this be done by Apsim/Sql? This can be cumbersome when running hundreds of simulations and can mess up some R scripts we have. Was it intended that these files were left behind, or is it an indication of something else wrong, or is simply a side-effect of latest sql version?
![image](https://user-images.githubusercontent.com/12175986/60479061-9b119300-9cd8-11e9-9688-c2a68f2e04ee.png)

zur003:
Those should be cleaned up automatically when the database is closed (SQLIte does the cleanup, not Apsim). If they're not being removed, it's a sign that the database isn't being closed properly. If I recall, it all gets a bit tricky because of the .net garbage removal mechanism, so database objects can be kept open longer than you might expect. Or maybe we have a bug somewhere that's keeping garbage collection from working as expected.

rcichota:
This seem to be only an issue when the simulations doesn't fully finished. Frank here noticed that when running in our HPC environment, the only files of that type left behind are from the simulations that for some reason did no run properly (maybe a linux thing because re-running them will get things cleaned up). 
Running on my PC I noticed that those files are created as soon as the simulation is opened and are delete when APSIM is closed (not when the simulation is closed, as I'd expect). However, if the simulation is stopped before it finishes (using the stop button), the files will not get cleaned up, even after closing APSIM...  Not sure if this is an issue that needs attention or simply a 'feature' of how Apsim uses sql...

hol430:
When #3999 is merged, these temporary files will be cleaned up properly when you close a tab, even if you cancel a run (or fail a run due to errors). I'm unable to reproduce the second part of the problem however - if a simulation throws an error when run from the command line, the .db-wal files are still cleaned up properly. Haven't tested on Linux though - perhaps it is a peculiarity related to mono's garbage collector. Will look into it.

zur003:
This may need to be re-opened, as it seems to be breaking Jenkins. I should note that I encountered something similar when running tests of Models.exe compiled in .Net Core, The DataStore wasn't closing at the end of a command-line run. I added some Dispose code to let me run my examples, but didn't merge this into the main branch, as I wasn't sure it was the "right" solution to the problem.

hol430:
[...]

PR #5293, opened 2020-06-12
Ensure the datastore gets closed at the end of a command-line run.

zur003:
Working on #3989


hol430:
Won't the datastore automatically be disposed of when we exit Main() anyway?

zur003:
I had expected disposal to be automatic, but it certainly was causing problems with my .net core builds. I don't recall the details, but I think it might even have been keeping the process alive at the end of the run.

zur003:
PR #5294 appears to have corrected the problem in a better way, so I'm dropping this pull request.

Issue #2192, opened 2017-11-22
Check/test implementation of pseudo-patches in SoilNitrogen


sno036:
Testing the implementation

sno036:
Test (sorry Dean!)

hol353:
Closing as I've created a new issue (#5327) specifically getting urine patch functionality working with Nutrient.

PR #5314, opened 2020-06-17
Converted an APSIM 7.10 urine patch simulation to NextGen

hol353:
Working on #2192


Models/Clock.cs
```diff
@@ -256,6 +256,14 @@ public double FractionComplete
             }
         }
 
+        /// <summary>Is today the end of the month?</summary>
+        [XmlIgnore]
+        public bool IsEndOfMonth { get; private set; }
```
hol430, line 261: Why do we need these properties? They seem like they should be a utility function

hol353: Yep. Good point. I've moved them to DateUtilities.

PR #5319, opened 2020-06-18
Apsim wheat

HamishBrownPFR:
working on #4773
Also has fixes to remove duplicate obs data for #4710


HamishBrownPFR:
@hol353 Please merge.  Changes in statistics are expected because I have removed some duplicated data from the wheat observation file.  @hol430 @hut104 this should deal to the duplicate data issues for wheat.  I have also added some missing data and deleted some simulations that were effectively duplicates (different names but no factors not set up to create different configurations)

hol353:
There are some big diffs in stats here. Plus the changes in stats are probably due not only to the removed duplicated data but the changes to wheat phenology and vernalisation. Are you sure you want this merged.?

HamishBrownPFR:
Can you point out the specific ones you are looking, i had a pretty good check and was comfortable but may have missed some

hol353:
Pretty much all of these:
image

HamishBrownPFR:
Thanks for the extra checking here @hol353 I have had a closer look and there are some zero values for predicted where there should be something, I will investigate further.  May I retract my previous request, please do not merge

hol430:
Maybe it would be easier to remove the duplicate data in a separate pull request, to help isolate which changes in stats are caused by your changes to the wheat model and which are changed by removal of observed data.

HamishBrownPFR:
It was easy enough to find them once I managed to get to the Obs Pred graphs on the performance site.  I still get time out errors most of the time when I try to get there

HamishBrownPFR:
@hol353 all good to merge now.  There where a number of the new simulations added that were regestering zeros for observed values.  Some additional work was also required to complete the management and cultivar description for these experiments.

PR #5323, opened 2020-06-21
Generalise temp interp

### LINKED ISSUE #5121: Tomorrows met data

HamishBrownPFR:
resolves #5121.

I have changed the met object to read a day ahead to get tomorrows met data and retain the past two reads to use for today and yesterdays met


HamishBrownPFR:
@hol353 @hol430.  I think there is something odd happening with the performance testing.  This pull request has changes on n for the wheat test but no changes to wheat simulations or observations.  I think these changes are being retained from the previous pull request?

hol353:
retest this please jenkins.

I think I merged the other PR just as you submitted this one. Might have been a timing issue. Anyway see if re-running it works.

HamishBrownPFR:
@hol353 it has failed for a different reason this time, didn't seem to run the tests

HamishBrownPFR:
@hol353 it has worked, just takes some time for the graphs to appear on the performance website

HamishBrownPFR:
Is it surprising that it failed the performance tests in less that 14 min?

hol353:
Jenkins says it took 18 minutes which is about right:
image

HamishBrownPFR:
Thats weird, when I looked at it your retest comment was made 14 min ago and it already had a red X.  I kept looking at it and the number of graphs kept increasing.  Does it show it has failed as soon as one goes red but then keep working its way through the tests and adding the graphs to the performance site

HamishBrownPFR:
I will reopen this pull request.  I cannot reproduce the diffs it is failing on so want to check it is not an issue with the performance tests

PR #5324, opened 2020-06-22
Generalise temp interp

### LINKED ISSUE #5121: Tomorrows met data

HamishBrownPFR:
resolves #5121


HamishBrownPFR:
@hol353 @hol430 there is something going wrong with the build system.  I opened this pull request 1 min ago and it is already telling me it has failed the performance tests

HamishBrownPFR:
@hol353 I don't know why this has failed the performance tests, there are no differences that I can see

hol430:
@hol353 when was the last time you accepted changes in stats?

hol353:
Well no change in stats but there are changes if you drill down on the 3rd sheet from the front page:
image
The performance site should probably look only at stats. It must be looking at the 'passed tests' as well.

hol353:
I last accepted stats about 5 hours ago.

HamishBrownPFR:
So do you merge this and accept stats or do I need to do something else?  I am not sure how to resolve those changes.

hol353:
@hol430 What is the performance site doing?

hol430:
Looks like a bug/inconsistency between the performance tests service (aka backend) and portal (aka frontend). The only tests which have failed are SEslope, SEintercept, and r2 for some of the wheat.leaf.cohortMaxSize values, but the portal's tabulated results page (and graphical results too) only show the rmse, nse, rsr, and n columns. So it is failing, but the portal isn't showing the failures.

I thought that only the rmse, nse, rsr and n are supposed to be taken into consideration when the pass/fail status is calculated, so this may be a bug more than an inconsistency.

hol430:
retest this please jenkins

Issue #5325, opened 2020-06-22
Generalise interpolation method


HamishBrownPFR:
Following on from #5121 which looked to make yesterdays and tomorrows temperature available.  The AirtemperatureFunction needs to be generalised so it can do sub daily interpolations of any variable.
Much of this has already bee done under #5121 but some renaming needed

hol430:
@HamishBrownPFR  if a simulation ends on the same date as the last weather reading in the met file, we will get an exception thrown on the final day of the simulation. The exception is handled internally so it's not really a huge problem, but it might be more convenient to just not attempt to read tomorrow's met data in this situation. If I implement this, will it muck up your interpolation code? I think what is happening now is that the `TomorrowsMetData` variable holds today's data on the final day, so that behaviour wouldn't be changed.

HamishBrownPFR:
So long as the behavior is not changed you can changed I am happy for you to change what ever you need to

PR #5326, opened 2020-06-22
Generalise interp

HamishBrownPFR:
Working on #5325


hol430:
@HamishBrownPFR looks like you've accidentally changed the line endings in a few files. Have you changed computers or any of your git config settings recently?

HamishBrownPFR:
Yes, I changed some git setting to surpress an error I got about line endings every time I committed.

hol430:
This is probably because `core.autocrlf` is set to false, but I have no idea how to change that from git fork. @hol353 I assume that we want lf (not crlf) in all files? If so, I can create a separate pull request and add a .gitattributes file so that this is not dependent on each user's git config settings.

HamishBrownPFR:
I set it in this file

image

hol353:
@hol430 I don't care what the setting is. I do care that the line endings on windows computers are crlf.

hol430:
@HamishBrownPFR I've standardised the line endings of all files to LF. You will need to pull down latest changes and merge conflicts. Then, to fix the files you've added CRLF to, you'll need to open a terminal in the ApsimX directory (there's probably a "console" button in your git client - do you use sourcetree?), and run the following command:

```
git add --renormalize .
```

HamishBrownPFR:
@hol430.  Im a bit confused by this.  I made the change to my gitconfig file a few weeks ago and have had a number of pull requests merged into APSIM master since so not sure why it has come up as an issue now.  Will I need to change all the files I have committed since making the change?  Do I need to change the setting in my gitconfig file to stop it changing them again?

hol430:
I think your gitconfig is correct now, as core.autocrlf is set to true. Previously, this setting determined git's line endings behaviour, but I've changed the .gitattributes file, so this behaviour should be tied to the repository now, rather than to each user's git settings.

I'm not sure exactly how these files ended up with the wrong line endings. The only files whose line endings you'll need to fix is the files you've changed as part of this pull request - and maybe files on your other branches which aren't merged, but I wouldn't worry about them for now. You can fix up these files by running the command I posted above, and then committing any changes.

HamishBrownPFR:
@hol430 I have followed your instructions above but it does not create changes to any files.  I will push to this branch again and see how it looks?

HamishBrownPFR:
Looks OK?  Was merging in your changes along with the change to the gitattributes file enough to fix it?

HamishBrownPFR:
No, sorry, it did make changes put staged them all automatically.  I wasn't looking for them there

Issue #5327, opened 2020-06-22
Create a nutrient patching model


hol353:
The use case for this model is to simulate urine patches. Currently the old SoilNitrogen model maintains a cohort of 'patches'. The plan is to replicate this behaviour in a new model that works with Nutrient but is separate to Nutrient. At the end of this process, we should be able to delete the old SoilNitrogen model.

hol353:
@APSIMInitiative/soil

sno036:
Done! Paper submitted even.

PR #5328, opened 2020-06-22
New nutrient patching model

hol353:
Working on #5327


HamishBrownPFR:
@hol353 @hol430 I think there is something wrong with Jenkins, he has been chewing on this for over 40 min

hol430:
I noticed this in the jenkins log:

```
[Validation] Grevillea has finished. Elapsed time was 558.2 seconds.
```

I can't see anything obviously wrong with Muck, so I've aborted the build. @hol353 could this be caused by your changes?

Models/Core/ApsimFile/Structure.cs
```diff
@@ -45,6 +45,9 @@ public static IModel Add(IModel modelToAdd, IModel parent)
                 links.Resolve(modelToAdd, true);
                 var events = new Events(modelToAdd);
                 events.ConnectEvents();
+
+                // Call StartOfSimulation events
+                events.PublishToModelAndChildren("StartOfSimulation", new object[] { parent, new EventArgs() });
```
hol430, line 50: Why are we publishing a StartOfSimulation event if it's not the start of the simulation? Seems a bit counterintuitive since this code has nothing to do with running a simulation.

hol353: Yes this is the bit of code I am testing. I want to add a model (Nutrient in this case) at runtime after a simulation has started. If I don't send this event then the nutrient model I'm adding hasn't been initialised and doesn't run correctly.

PR #5338, opened 2020-06-24
Add radiation interpolation method

### LINKED ISSUE #5325: Generalise interpolation method

HamishBrownPFR:
resolves #5325

### DIFF
--- a/Models/Functions/HourlyInterpolation.cs
+++ b/Models/Functions/HourlyInterpolation.cs
@@ -13,7 +13,7 @@ namespace Models.Functions
     /// Uses the specified InterpolationMethod to determine sub daily values then calcualtes a value for the Response at each of these time steps
     /// and returns either the sum or average depending on the AgrevationMethod selected
     /// </summary>
-    
+
     [Serializable]
     [Description("Uses the specified InterpolationMethod to determine sub daily values then calcualtes a value for the Response at each of these time steps and returns either the sum or average depending on the AgrevationMethod selected")]
     [ViewName("UserInterface.Views.GridView")]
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@BrianCollinss @jbrider @hut104 @zur003 

I have sort of completed the job of making a generic approach for interpolating things on sub daily basis.

I have renamed the IFunction that was called AirTemperature function to HourlyInterpolation.  It has two children, the function that we want to calculate values at each timestep (named Response) and the method we use to interpolate sub daily inputs for calculating the response (named Interpolation method).   The HourlyInterpolation class calls the InterpolationMethod which returns the SubDailyInput values for what ever variable and at what ever frequency specified.  It then passes each SubDailyInput into the Response function and agregates the SubDailyResponses using the AgregationMethod specified (sum or average).
Currently there are three options for InterpolationMethods, ThreeHourSin (the method used in the original AirTemperature function), HourlySinPpAdjusted (This probably needs a better name but calculates hourly temp using the method we discussed in #5121) and HourlyRadiation (Which sort of uses @BrianCollinss code but I have not tested this and am pretty sure there are errors in it).
The Response function needs to be IIndexedFunction, XY pairs is one of these and I have written a vernalisation function that also takes inputs.
image
For photosynthesis, you would have a HourlyInterpolation method called Photosynthesis on the Leaf class with the HourlyRadiation class as the InterpolationMethod and need to write a HourlyPhotosynthesis class that implements IIndexedFuction to take radiation values from HourlyRadiation and return values that the HourlyInterpolation method would then sum to return a Value for daily photosynthesis.

Issue #4075, opened 2019-08-08
Create a new peanut model


hol353:
The APSIM Peanut model is being developed from the original APSIM Classic Peanut model described by Robertson et al (2002) using the Plant Modelling Framework of Brown et al (2014).

Phenology is represented as a sequence of field-observable states, the progression of each responding to temperature and (in some cases) soil water. DM partitioning is based on vegetative/reproductive state. Grain filling follows the 2002 HI model that partitions DM to grain via a target HI.

Funded by DNRME. To be built by Neil Huth. Support from Brendan Power.

sarahcleary:
@yashvirchauhan - please discuss peanut developments with @hut104

sarahcleary:
Please note action from RP 8/19 meeting: @yashvirchauhan - to discuss peanut developments with @hut104

sarahcleary:
@hol353 - just following up the concept note.  For background, please refer to https://github.com/APSIMInitiative/ReferencePanel/issues/57#issue-595681585

EnliWang:
@hut104 
In what stage is the model now?
 
At least 5 years of data from CAU (one cultivar only, but with different N treatments) can be made available to support the development and validation. These include dynamics of LAI, leaf/stem/pod/grain biomass, soil water usage.

CAU colleagues would be happy to contribute. Happy to chat in the new year?

hut104:
Sounds good.
I hope to have an initial version of the model ready for the RP soon.
Neil.

yashvirchauhan:
Hi Neil
It appears that the APSIM 7.10 is not able to adequately simulate the effect of high temperature. This results in a substantial gap in simulated and observed yield in hotter environments. I am attaching a paper on this aspect and if something can be done to incorporate the high (max) temperature effect on pod yield and we will have project to validate this in the northern tropics. 
[Hightemperatureeffectonpeanuts_Challinoretal2005.pdf](https://github.com/APSIMInitiative/ApsimX/files/5675981/Hightemperatureeffectonpeanuts_Challinoretal2005.pdf)

hut104:
This certainly looks useful. We may be able to add something similar to this, however it would be useful to have test data for it.

________________________________

yashvirchauhan:
[...]

PR #5339, opened 2020-06-24
Peanut

hut104:
working on #4075


hut104:
@hol353 you can merge this. This is as far as I need to go right now.

PR #5343, opened 2020-06-25
Netcore3

zur003:
I'm leaving this here in case you want to have an up-to-date .Net Core 3 version (without a GUI) available from the main repository. Note that this is a request to pull into the netcore3 branch, not the master branch.

Working on #5344


hol430:
retest this please jenkins

hol430:
@hol353, @zur003 this is getting close to going green on Jenkins. It's just the unit tests and UI tests which are still failing. The performance tests are green which is a good sign. The UI tests are failing because there is no UI on this branch. The unit tests are failing because the unit tests project hasn't been converted to netcore yet and added back into the solution, so it isn't being built. The unit tests project actually depends on ApsimNG, so we can't add them back in until we add ApsimNG back in. 

I can go and add the .net framework ApsimNG/UnitTests back into the solution if you want. I've already done half of this (ApsimNG) on another branch. The other option would be to just merge this for now - note that the pull request is against the netcore3 branch, not master, so it won't affect the master branch. Then I raise a pull request on my netcore3 branch, which contains both framework and netcore3 builds of ApsimNG, and we proceed from there.

On a side note, I wonder if jenkins should be running the netcore version of Models.exe on this branch...

hol353:
I don't mind which way you do it. 

Jenkins should run the Framework version of Models.exe as that is what users are using atm.

hol430:
Ok, well in that case, could you merge this one please? Sounds like it will be less work for me (merging, maintaining 2 branches) if I just start a new pull request against netcore3.

hol353:
Can't merge this because it is red.

hol430:
How about now?

hol353:
Still red.

PR #5351, opened 2020-06-25
Life cycle

### LINKED ISSUE #5350: Lifecycle improvements
Migration events create a new cohort from each cohort that it migrating.  Need to consolidate into a single new cohort otherwise we end up with to many.
Cohorts need to hold information about where they came from.

HamishBrownPFR:
resolves #5350


Models/Lifecycle/Cohort.cs
```diff
@@ -26,7 +27,10 @@ public class Cohort
         public double Progeny { get; set; }
 
         /// <summary>Number of migrants leaving this cohort</summary>
-        public double Migrants { get; set; }
+        public double Emigrants { get; set; }
+
+        /// <summary>Information about where this cohort came from</summary>
+        public SourceInfo sorceInfo { get; set; }
```
hol353, line 33: Can this be spelt correctly? Will potentially save having to write a converter later.

hol430: Maybe it's supposed to be emigrants (electronic migrants) as in ebook?

HamishBrownPFR: It is spelt correctly.  Migration is the net term for moving, immigration is moving in and emigration is moving out.

zur003: Yes, but "sorceInfo" does seem like an odd spelling.

HamishBrownPFR: Information about the source of a cohort.  Suggestions?

hol353: Can we spell it sourceInfo rather than sorceInfo?

HamishBrownPFR: Ah, I see what you are on about now.  Spellings not my strong point.

HamishBrownPFR:
@hol353 spelling fixed

Issue #5356, opened 2020-07-02
AgPasture science review.


hol353:
It is time for a review of the science in AgPasture. Two things to kick off the list:

- There are 4 options for doing water and solute uptake. Need to standardise on one. Currently AgPasture uses the 'basic' option. Should look at one of the other options.
- From @MarkLieffering: the CO2 response in AgPasture needs a to have no effect for when CO2 is below the reference level of 380ppm. It should be like this from EcoMod http://imj.com.au/wp-content/uploads/2014/08/GrazeMod.pdf page 4.12:
![image](https://user-images.githubusercontent.com/1943827/86410768-c02b4e00-bcfe-11ea-89ee-277b7a69edd4.png)

peter-devoil:
Can we also have establishment / killing routines so it works in a rotation?

hol353:
@peter-devoil Can you elaborate a bit on this? What do you mean by establishment routines. 

The model does respond to a Kill management call. It also implements the plant and organ damage routines so that crop consumers (pest and diseases) can damage a pasture species.

peter-devoil:
Sowing routines. Perhaps I am missing something, or thinking of classic. One day it would be good to run a crop/pasture rotation.

sarahcleary:
@jwhish - you may wish to comment?

hol353:
I've found a problem in AgPasture that needs fixing. 

Soil Arbitrator sends 4 soil N states to a pasture species and expects an estimate back for each on how much N would be taken up. AgPasture does this but always returns the same estimate back to soil arbitrator regardless of the soil state passed to it. The result is that each species will be given full access to all the available N in the soil which means that there is no competition between the species. 

When I fix the bug I get a dramatic over prediction in the validation harvested wt:
![image](https://user-images.githubusercontent.com/1943827/86556100-a9525900-bf95-11ea-9b29-e1d5c6bfdaa9.png)
The blue dots are with the bug fixed. The orange dots are the original validation values.
RMSE changes from 2045 kg/ha to 3529 kg/ha
NSE changes from 0.57 to -0.27
R2 changes from 0.57 to 0.56

This needs further investigation.

rcichota:
> Sowing routines. Perhaps I am missing something, or thinking of classic. One day it would be good to run a crop/pasture rotation.

@peter-devoil ,
[...]

PR #5372, opened 2020-07-09
Changed AgPasture N uptake method to DefaultAPSIM rather than Basic 

hol353:
Also Fixed bug with n uptake - arbitrator.
Working on #5356


hol353:
retest this please jenkins

PR #5375, opened 2020-07-09
Changes to sorghum/c4maize to mirror recent changes in classic

hol430:
Working on #572

@jbrider, I noticed that you changed maize's frost kill threshold to 1, but left sorghum's at -3.5. Is this deliberate?

### DIFF
--- a/Models/Plant/Organs/SorghumLeaf.cs
+++ b/Models/Plant/Organs/SorghumLeaf.cs
@@ -12,6 +12,7 @@ using Models.PMF.Struct;
 using System.Linq;
 using Models.Functions.DemandFunctions;
 using Models.Functions.SupplyFunctions;
+using System.Text;
 
 namespace Models.PMF.Organs
 {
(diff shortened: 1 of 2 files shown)


hol430:
@jbrider the changes to the stats are fairly minor, are you happy for this to be merged?

jbrider:
@hol430  Sorry, this slipped through yesterday. Maize and Sorghum should be the same - were they different in Classic, or in NextGen.  Can you remove the sdRatio change out for now - want to make sure it's not affecting the frost change outputs.

hol430:
Ah, I was talking about classic when I said the frost_kill values were not the same, but I just double-checked and they are indeed the same. I guess I was looking at the changes in APSIMInitiative/APSIMClassic#1847 where maize.xml was changed but sorghum.xml wasn't.

jbrider:
Yeah, there was one update that was out of sync. I thought I'd managed to do it again!

jbrider:
This can be merged.

Issue #5377, opened 2020-07-09
Supplement library isn't thread safe


hol353:
I just got a bunch of errors from Jenkins in code I didn't change - see errors at bottom.

If you look at lines around GrazSupp.cs:line 773:

```c#
for (int idx = 0; idx < SupplementLibrary.DefaultSuppConsts.Count; idx++)
{
    string suppName = SupplementLibrary.DefaultSuppConsts[idx].Name.ToUpper();
    ...
}
```
The static property SupplementLibrary.DefaultSuppConsts is used on the for loop line as well as inside the for loop. The exception was thrown inside the for loop. The only way this can happen is if some other thread changed DefaultSuppConsts sometime between the 'for' line and the inner 'suppName = ' line. This SupplementLibrary.DefaultSuppConst property is implemented as:

```c#
        public static SupplementLibrary DefaultSuppConsts
        {
            get
            {
                if (GDefSupp == null)
                {
                    GDefSupp = new SupplementLibrary();
                    SetupDefaultSupplements();
                }
                return GDefSupp;
            }
        }
```
Where there are more than one thread accessing this property there can potentially be problems due to the lack of locks. I think this is what happened on Jenkins (although I thought Jenkins ran things multiprocess rather then multithreaded).

```
----------------------------------------------

System.Exception: Error in file:C:\Jenkins\workspace\PullRequest_v2\ApsimX\Tests\Simulation\DairyFarmManager\PlantainGrazing.apsimx ---> System.NullReferenceException: Object reference not set to an instance of an object.

   at Models.GrazPlan.FoodSupplement.ParseText(String suppSt, Boolean nameOnly) in C:\Jenkins\workspace\PullRequest_v2\ApsimX\Models\Grazplan\GrazSupp.cs:line 773

   at Models.GrazPlan.FoodSupplement..ctor(String suppSt) in C:\Jenkins\workspace\PullRequest_v2\ApsimX\Models\Grazplan\GrazSupp.cs:line 279

   at Models.GrazPlan.SupplementModel.AddToStore(Double suppKg, String suppName, Int32 roughage, Double dmp, Double dmd, Double medm, Double cp, Double dg, Double ee, Double adip2cp, Double phos, Double sulf, Double ashAlk, Double maxPass) in C:\Jenkins\workspace\PullRequest_v2\ApsimX\Models\Grazplan\Supplement.cs:line 241

[...]

PR #5386, opened 2020-07-12
Add lock when loading default supplement constants.

zur003:
Working on #5377.


hol353:
Why not make it non static instead?

zur003:
@hol353 - It's intended as an optimisation. The default supplement constants don't change, but do need to be parsed from a resource, and there's no point incurring the computational expense of parsing the values more than once. These values get accessed in the FoodSupplement constructor, and that constructor gets called frequently when feeding supplement to stock. 

Probably the whole thing would benefit from some refactoring, but for now can you recommend a better way of providing a singleton instance of the default supplement values?

PR #5390, opened 2020-07-13
Moved more functionality for leaf number out of leafappearancephase.

hut104:
working on #4773

### DIFF
--- a/Models/Core/ApsimFile/Converter.cs
+++ b/Models/Core/ApsimFile/Converter.cs
@@ -19,7 +19,7 @@
     public class Converter
     {
         /// <summary>Gets the latest .apsimx file format version.</summary>
-        public static int LatestVersion { get { return 105; } }
+        public static int LatestVersion { get { return 106; } }
 
         /// <summary>Converts a .apsimx string to the latest version.</summary>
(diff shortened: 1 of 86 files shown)


hut104:
@HamishBrownPFR and @hol353 I have removed a dependancy between LeafAppearancePhase and Leaf. It changes the results (somewhat expected) when I changed the following code
            if (leaf.ExpandedCohortNo >= (leaf.InitialisedCohortNo))
to
            if (LeafNumber.Value()>= FinalLeafNumber.Value())
to determine when the stage is over.

The earlier can only work if you use the complex leaf model and is a little bit of a kludge.  The new code actually improves some phenology predictions, but improving the phenology makes some biomass output worse.

In one way, there may be no need for this class at all, given that models like Teff has managed to be written within the use of this phase type - it uses a generic phase with the length calculated from current and final leaf number.

As we continue to improve the phenology model we will continue to affect other output.  We will need to decide on an approach for these situations.  Do we committ phenology improvements, and start running optimisers to fix up some of the damage on an ongoing basis?

hol430:
retest this please jenkins

PR #5394, opened 2020-07-14
Added some helper functions to Nutrient and INutrient

### LINKED ISSUE #5376: Nutrient model needs some helper properties.
We are writing manager script and need to access nutrient properties:
Denitrification
Nitrification
dlt_n_min_res
MicrobialC
MicrobialN
HumicN
dlt_urea_hydrol

There are probably others. These need to be added to Nutrient and INutrient.

Accessing some of these are easier from REPORT because the intellisense can prompt you down to the e.g. child Microbial pool but this doesn't happen in manager scripts. My suggestion is to 

1. provide access to the child pools from nutrient e.g. ```public NutrientPool Microbial { get; set; }```. This will solve problems of access to Microbial and Humic etc. Perhaps we should have an interface for NutrientPool so that we don't expose it's inner workings to manager scripts.

2. provide helper properties for the others e.g. dlt_n_min_res and Denitrification.

3. I don't think the existing helper property MineralisedN is correct. Currently it is returning the MineralisedN from just the SurfaceResidue pool. This needs checking.

hol353:
Resolves #5376


hol353:
retest this please jenkins

PR #5396, opened 2020-07-14
Added Root property to PastureSpecies 

hol353:
So that manager scripts can change root parameters. Working on #5356


hol430:
retest this please jenkins

Issue #5344, opened 2020-06-25
Could we start actively maintaining a .Net Core 3 build?


zur003:
It's currently possible to compile all components other than the ApsimNG.exe user interface with .Net Core 3. @her123 has been working on using such builds in Docker containers for the World Modelers work, as they are more compact and much speedier than using Mono on Linux-based systems. Would it be possible to add this into the main branch in some way? I believe it is possible to have both Frameworks and Core projects in the same solution, though there may be some complications in doing so. It would be nice to ensure that a .Net Core version is always current, readily available, and able to run most or all of the test suite.

hol430:
Good idea - maintaining a separate branch is going to be painful in the longer term. I have a netcore3 branch on my repo in which I have a somewhat-working gui on linux. The branch started out as your netcore3 branch but probably isn't quite up-to-date at the moment.

What would be the best way to merge this with the master branch? Do we just maintain a separate .csproj file for each solution?

her123:
It would be nice to have the same code shared by two different solutions/projects. A dotnetcore solution may be able to live alongside the main solution in the main branch.

hol430:
Thinking about it some more, I'm not sure if it will be possible to use exactly the same code for both solutions. The Gtk#3 api is a bit different to the gtk#2 one in some places, and I've had to make some changes which won't compile under gtk#2. No idea if you guys have had to make similarly breaking changes in the other projects to get them working in .net core.

her123:
Is there too much to have in conditional sections?

zur003:
Yes, what @her123 said is what I had in mind. The differences between Framework and .Net Core are very minor, if GUI code is excluded. It's just the code in Models\Core\ScriptCompiler that has any real differences.

zur003:
More specifically, I was thinking we could have .Net core 3 builds of APSIM.Shared, APSIMRunner and Models, but not include any of the GUI stuff just yet. Not sure about UnitTests and Updater (Updater really only makes sense when driven by the GUI).

hol430:
Ah, ok. Maybe we create two csproj files for each project except ApsimNG, then create a copy of ApsimNG which is based on .net core. That way we can still use the same code for framework/core for everything except ApsimNG?

hol430:
Or yeah, leave the ApsimNG changes on a branch. Maybe that's the way to go.

her123:
Sounds like a good way to go

zur003:
[...]

PR #5399, opened 2020-07-15
.net core/gtk3 ApsimNG

hol430:
Working on #5344

Converted ApsimNG to the sdk-style csproj format and have both netcore and framework builds using the same files. Also added the unit tests project back into the solution, ~although I haven't converted it to the sdk-style format yet.~


hol430:
Thanks Eric- I guess this means you got it working ok on your machine?

zur003:
@hol430 - Yes, I got it working, although there are still a few rough edges. For example, sqlite3.dll is not being automatically copied into NetCoreBin. Also, I'm having trouble starting a debug session from the IDE. But mostly it's working as expected. The next thing I'd like to attempt is getting basic text/code editing working with gtkSourceView. I have this mostly working, but need to tidy it up a lot before it's ready to be merged. That currently uses a GtkSharp 3 wrapper for SourceView that I compiled myself - I need to check to see whether anyone has put a similar version on nuget.

hol430:
Good point about sqlite3.dll, although in the long run it might be worth switching to the nuget package. For now I'll just modify the csproj file to fix this.

For the sourceview, I've submitted a pull request against the GtkSharp repo to add some bindings. In the long run this will mean we should end up with a nuget package for this. For now, I've manually generated the .nupkg files and uploaded them to myget ,and added a reference to this to ApsimNG.csproj. The only caveat is that - for now - you will have to install libgtksourceview4 via msys and then place it somewhere where the clr will find it (e.g. in the NetCoreBin directory). I've been in touch with the maintainer of GtkSharp about packaging this with the other native binaries which are automatically downloaded when you build a project using GtkSharp, so this shouldn't be an issue for developers in the longer term. We will need to think about how we ship these to our end users however, as there's no official way of doing this yet - although there may be something in the works there as well.

There are also some issues with debugging in visual studio which are caused by a bug in the particular version of the native GLib library provided by the GtkSharp package. There's an issue in the GtkSharp repo about this and it should hopefully be fixed sometime in the next few weeks when we update the native libraries provided by GtkSharp.

image

zur003:
@hol430 - So the target will be GtkSourceView 4 rather than GtkSourceView 3? That will be good, although I noticed that the Windows Gtk 3 distribution that I'm using (which I think may have been auto-downloaded by the GtkSharp stuff) packages SourceView 3 rather than 4. I need to go back and see what the differences were.

hol430:
Yep, I have targeted sourceview 4(.0.0 from memory), and yeah it's not included in the files which are auto downloaded by GtkSharp (ie the files downloaded to %LOCALAPPDATA%\Gtk\3.24), but we're working on adding it to those files.

hol430:
@hol353 this could probably be merged, I don't know if I'll be getting back onto this this week (note that the PR is against the netcore3 branch, not master).

Current status is that all projects now build under netcore and netframework. We have a working code editor with syntax highlighting, code completion, etc. Intellisense for other components (e.g. report) seems to be working. Probably lots of things which don't work but the main things which still need to be implemented:

- Some sort of HTML/markup renderer. Currently I've implemented IHTMLView by sticking the raw HTML into a textview widget
- Syntax highlighting in report
- Ctrl+shift+space to trigger method signature helper dialog in code editor
- Find/replace in code editor (this will actually be more powerful than the current version, since gtksourceview api supports regex)
- Few issues with grids such as colours in the profile view, column header vertical alignment

I wonder if at some point in the future we would want to merge this into master. This would mean that anyone compiling apsim would need to install the .net core sdk (>=3.0 I think), which can be done via the visual studio installer, or via microsoft's [dotnet-install](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) scripts. For the average user, this wouldn't have much of an impact (once the sdk is installed), since the .net framework (gtk2) build is still the default when you hit run. Switching to .net core builds can be done via a dropdown in visual studio. Any thoughts?

PR #5421, opened 2020-07-23
Refactor of CLEM Ruminant filter groups

### LINKED ISSUE #5420: CLEM refactoring to clean up code and functionality
A number of naming approaches and duplicate components need cleaning up in CLEM

- Use RuminantGroup Filter rather than RuminantFilterGroup and replace RuminantDestockGroup
- Add RuminantActivityGroup for simulation level filtering
- Replace month (int) properties with enum
- Continual improvement of validation, summary descriptions and default values

lie112:
Resolves #5420


Models/CLEM/Activities/RuminantActivityPredictiveStocking.cs
```diff
@@ -219,7 +219,7 @@ private void HandleDestocking(double animalEquivalentsforSale, string paddockNam
             ruminantHerd.PurchaseIndividuals.RemoveAll(a => a.Location == paddockName);
 
             // remove individuals to sale as specified by destock groups
-            foreach (RuminantDestockGroup item in this.Children.Where(a => a.GetType() == typeof(RuminantDestockGroup)))
+            foreach (RuminantGroup item in Apsim.Children(this, typeof(RuminantDestockGroup)))
```
hol430, line 222: @lie112 is this actually a valid cast from RuminantDestockGroup to RuminantGroup? I'm just trying to work out how this works, as I need to change this call to Apsim.Children as part of my work on removing the Apsim.xxx api.

lie112: No, that's a mistake. There is no longer such a thing as RuminantDestockGroup so the typeof should be RuminantGroup.
I did not delete RuminantDestockGroup as that would break all users simulations and make upgrading difficult for them
RuminantDestockGroup will soon be deleted. That's why id didn't pick this up.
RuminantDestockGroup can be changed to the new general RuminantGroup in the line above.

PR #5424, opened 2020-07-23
Does using integer for SQLite boolean columns cause issues?

### LINKED ISSUE #5393: Write boolean as integer (0-1) to SQLite reports
The current Sqlite.GetDBDataTypeName(Type type) method reports boolean properties as TEXT for database column creation.
Is there a reason for this or could we create boolean based columns as integer (0-1) in our SQLite reports?

lie112:
Resolves #5393

### DIFF
--- a/APSIM.Shared/Utilities/SQLite.cs
+++ b/APSIM.Shared/Utilities/SQLite.cs
@@ -1031,6 +1031,8 @@ namespace APSIM.Shared.Utilities
                 return "real";
             else if (type.ToString() == "System.Double")
                 return "real";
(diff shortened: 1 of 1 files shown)


lie112:
I'm going to need a hand working out what went wrong here. 
I have gone down a rabbit hole into another world.
I assume the two tests use a text based boolean (True/False) somewhere in the validation tests.

hol430:
This is most likely a bug in the performance tests, I'll do some investigating.

hol430:
retest this please jenkins

PR #5426, opened 2020-07-23
Move removal of views code occur after check for all empty databases …

### LINKED ISSUE #5425: Database VIEWS are not deleted on empty datastore when _Checkpoints exist
Need to fix droppping of views for simulations with _Checkpoints

lie112:
Resolves #5425 

Now works regardless of _Checkpoints
If all tables are empty and deleted there is no reason to keep any views.

### DIFF
--- a/Models/Storage/EmptyCommand.cs
+++ b/Models/Storage/EmptyCommand.cs
@@ -59,11 +59,6 @@
                     if (!tableName.StartsWith("_"))
                         database.ExecuteNonQuery(string.Format("DELETE FROM [{0}]", tableName));
                 }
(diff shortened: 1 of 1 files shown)


lie112:
I think all my different branches pulled are somehow tied together.
This change should not have cause the same error as previous pull.

lie112:
retest this please jenkins

lie112:
Seems to have the same performance issues as #5424 but retest does not solve the problem

hol430:
Yeah, the job is still running. The performance tests' status doesn't get updated until it finishes the validation stage. Would be good to fix that at some point, I think there's an issue for it but I haven't done anything yet...

PR #5427, opened 2020-07-23
Updated Gatton Data

hut104:
working on #4773


hut104:
@hol353We have simply updated an observed data spreadsheet.  If you look at the results, the only changes are for outputs where we have increased the number of observed data.

hol430:
If you look on page 3 on the 'home screen' of the performance tests, you'll see changes to wheat HarvestObsPred. This was due to an issue with the performance tests earlier this morning. If we rerun the tests it should go away (that fixed it for one of the other pull requests anyway).

hol430:
retest this please jenkins

hut104:
Is this right to merge now? We added new OBS data.

PR #5432, opened 2020-07-29
Store font as a string rather than Pango.FontDescription

hol430:
This also improves the robustness of the font resolution algorithm, which should be helpful if switching between the master and netcore3 branches.

Working on #5344

### DIFF
--- a/ApsimNG/Utility/Configuration.cs
+++ b/ApsimNG/Utility/Configuration.cs
@@ -125,7 +125,7 @@ namespace Utility
         /// Stores the user's preferred font.
         /// </summary>
         /// <value></value>
-        public Pango.FontDescription Font { get; set; }
+        public string Font { get; set; }
 
         public ApsimFileMetadata GetMruFile(string fileName)
         {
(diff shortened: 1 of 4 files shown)


hol430:
@hol353 this will cause problems for anyone who runs a version of apsim which includes these commits, then switches to an older version of apsim from before these commits. The problem is that the older versions don't have a proper fallback font setup, and you can end up with this on the home screen:

image

If this happens you have 3 options:

1. Upgrade to a newer version
2. Change the font (this is the easiest solution if you know how to change the font in apsim, since the problem doesn't seem to affect the font chooser dialog:

image
image

3. Delete %appdata%\ApsimInitiative\ApsimX\ApsimX.xml (which will also other settings such as recent files, theme, ...)

I'm not sure if there's a way to avoid this issue since the bug is in the older version of apsim.

hol353:
I don't like this because there will be people that have multiple versions on their computer that they swap between. Telling them to upgrade isn't an option because they may have chosen that version for the life of the their project and don't want to change. Your option 2 and 3 would probably work.

The other option is that in configuration.cs don't change Font from a Pango.FontDescription to a string. Create a new property instead. Is this possible?

hol430:
Good thinking - if we remove the old Font property altogether (by changing the name of the property to FontName) then there's no problem. The older versions are able to handle a null value for Font, it's just when there's some info but it's the wrong type that it starts to get confused.

Issue #5478, opened 2020-08-02
Need to allow TreeProxy properties for Height, ShadeModifier and NDemand to be settable from scripts


hut104:
These changes would allow APSIM agroforestry models to be more easily operated within larger whole-of-farm models in Simile etc.
It would also allow for more flexible manager rules within APSIM.

hol430:
While you're at it, if you could refactor out the `public List<List<string>> Table { get; set; }`, that would be great.

PR #5480, opened 2020-08-03
Tree Proxy

hut104:
working on #5478


hut104:
might be worth merging this lest our collaborators run into the same problem as I'm trying to help them

PR #5483, opened 2020-08-03
Move ptq

HamishBrownPFR:
working on  #4773


hol353:
Why are there lots of diffs in nutrient.json resource?

HamishBrownPFR:
Converter has changed a high level indent.  Its all white space, the only actual diff is the version number

PR #5484, opened 2020-08-03
Speed Up

hut104:
working on #3509
@hol353 check that I've implemented your patching stuff correctly (assuming that this passes your tests)


hol430:
@hut104 - this is caused by the resource name in Nutrient.json being "Nutrient" (line 216 in Nutrient.json) - it should be set to null, as it is in the other resource files. Changing this fixes the stack overflow exception for me.

hut104:
@hol353 apart from a failed unit test, the tests are green if you want to merge with your work to test overall speed improvement

hol430:
@hut104 this build has stalled on jenkins. I've killed it, but there are some exceptions from failed simulations which should help with debugging.

hol430:
@hut104 this stalled again on jenkins. It may be possible that you have an infinite loop somewhere. It stalled while running prototypes and also when running the validation sets.

PR #5487, opened 2020-08-04
Refactor of Plant

### LINKED ISSUE #5443: Namespace for Plant code and folder in studio are different
Plant.cs has 
namespace Models.PMF

but the folder in studio is called "Plant"

### LINKED ISSUE #5444: Why is there a using for System.XML when we don't do XML anymore?
See issue title for issue

All XML ignores should be Json ingnores

### LINKED ISSUE #5445: Plant.cs uses regions
These are outside our coding standards
see         #region Class links

### LINKED ISSUE #5446: Plant class has identical members
???
        /// <summary>The Canopy</summary>
        [Link(IsOptional = true)]
        public ICanopy Canopy = null;
        /// <summary>The leaf</summary>
        [Link(IsOptional = true)]
        public ICanopy Leaf = null;

### LINKED ISSUE #5447: Use interfaces for links in Plant.cs
Plant.cs has links that are resolved on type rather than interfaces

        /// <summary> The plant's zone</summary>
        [Link]
        public Zone Zone = null;

        /// <summary>The phenology</summary>
        [Link]
        public Phenology Phenology = null;
        /// <summary>The arbitrator</summary>
        [Link(IsOptional = true)]
        public IArbitrator Arbitrator = null;
        /// <summary>The structure</summary>
        [Link(IsOptional = true)]
        public Structure Structure = null;
        /// <summary>The Canopy</summary>
        [Link(IsOptional = true)]
        public ICanopy Canopy = null;
        /// <summary>The leaf</summary>
        [Link(IsOptional = true)]
        public ICanopy Leaf = null;
        /// <summary>The root</summary>
        [Link(IsOptional = true)]
        public Root Root = null;

### LINKED ISSUE #5454: cultivarDefinition can be removed from class
use a local variable instead and no need for the unapply

        /// <summary>The current cultivar definition.</summary>
        private Cultivar cultivarDefinition;

hol430:
Resolves #5443
Resolves #5444
Resolves #5445
Resolves #5446
Resolves #5447
Resolves #5454


hol430:
retest this please jenkins

hol430:
@hol353 could you have a look at this and let me know if you want changes made, or merge it otherwise? I want to continue the work on plant, but want to make the changes in smaller chunks in the future. Maintaining a PR of this size is just not practical.

hol353:
Why are there 4 .apsimx files in here?

hol430:
They have complicated manager scripts which weren't fixed correctly by the converter (it's pretty much impossible to be 100% accurate using regexes to change C# code). Mostly it was related to changing plant's links to be links to interfaces - IPhenology, IRoot, etc. These interfaces don't contain some members which were being accessed by the manager scripts. This is related to the other issue where we want to change how these children are accessible in manager scripts. The interim solution is a link by path.

I've also added zadok to IPhenology. I'm assuming you're ok with this since you didn't reply to my comment on that issue.

hut104:
I note that the Peanut Test set now has issues because DaysAfterSowing is used on the Xaxis of some graphs

PR #5497, opened 2020-08-06
Speed up increment

hut104:
working on #3509


hut104:
@hol353 This is the last green commit - ie the one you pushed up

hol353:
Can you remove potato.apsimx?

hol353:
Hmm - potato is still there.

PR #5506, opened 2020-08-07
Fixed, refactored and tested Azure functionality

### LINKED ISSUE #4340: Azure functionality is broken
From @sno036:

> I just tied this with Azure and got an error. It might be that the easiest if I supplied a Azure creds to someone who knows what they are doing so they can test. Just let me know who.
>
> System.IO.IOException
HResult=0x80070020
Message=The process cannot access the file 'C:\Users\snowv\OneDrive - AgResearch\Desktop\Temp\TestAzure.db' because it is being used by another process.
Source=mscorlib
StackTrace:
at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
at System.IO.Compression.ZipFileExtensions.DoCreateEntryFromFile(ZipArchive destination, String sourceFileName, String entryName, Nullable1 compressionLevel) at System.IO.Compression.ZipFile.DoCreateFromDirectory(String sourceDirectoryName, String destinationArchiveFileName, Nullable1 compressionLevel, Boolean includeBaseDirectory, Encoding entryNameEncoding)
at System.IO.Compression.ZipFile.CreateFromDirectory(String sourceDirectoryName, String destinationArchiveFileName, CompressionLevel compressionLevel, Boolean includeBaseDirectory)
at UserInterface.Presenters.NewAzureJobPresenter.SubmitJob_DoWork(Object o, DoWorkEventArgs e) in C:\aaUnderVC\apsimNextGen\ApsimNG\Presenters\NewAzureJobPresenter.cs:line 338
at System.ComponentModel.BackgroundWorker.OnDoWork(DoWorkEventArgs e)
at System.ComponentModel.BackgroundWorker.WorkerThreadStart(Object argument)

@hol353 on a side note, do you think we should add some unit tests for this Azure stuff? This would require us to purchase a small Azure VM which we can use for simple jobs on Jenkins but I doubt it would cost much. Would probably also want to do the same for other platforms as we add them (ie aws, google compute, etc).

hol353:
Resolves #4340


ApsimNG/Cloud/Azure/AzureInterface.cs
```diff
@@ -227,34 +227,37 @@ public async Task SubmitJobAsync(JobParameters job, CancellationToken ct, Action
         /// </summary>
         /// <param name="ct">Cancellation token.</param>
         /// <param name="ShowProgress">Function to report progress as percentage in range [0, 100].</param>
-        public async Task<List<JobDetails>> ListJobsAsync(CancellationToken ct, Action<double> ShowProgress)
+        public async void ListJobsAsync(CancellationToken ct, Action<double> ShowProgress, Action<JobDetails> AddJobHandler)
```
hol430, line 230: Why are we returning void here? Doesn't this mean that this method can't be awaited? Pretty sure the recommended pattern is for async methods with no return value to return a Task.

hol353: Rather than return a list of jobs (which is very slow) it calls the AddJobHandler for each job which in turn updates the GUI every time a job is loaded.

ApsimNG/Cloud/Azure/AzureInterface.cs
```diff
@@ -301,57 +304,102 @@ public async Task DeleteJobAsync(string jobID, CancellationToken ct)
...
+            var resultsZip = Path.Combine(options.Path, options.Name + ".zip");
+            resultsZip = DirectoryUtilities.EnsureFileNameIsUnique(resultsZip);
+
+            // Create a temporary workarea which will be zipped later.
+            var resultTempDirectory = Path.Combine(options.Path, Path.GetFileNameWithoutExtension(resultsZip));
+            if (!Directory.Exists(resultTempDirectory))
+                Directory.CreateDirectory(resultTempDirectory);
+
+            // Get a list of outputs generated by cloud run.
+            var outputs = await GetJobOutputs(options.JobID, ct);
```
hol430, line 317: Not necessarily a problem here, just wondering what is our policy on using var. Personally I would prefer to see an actual type in cases like this one, where the return type of GetJobOutputs is not immediately obvious.

ApsimNG/Cloud/Azure/AzureInterface.cs
```diff
@@ -301,57 +304,102 @@ public async Task DeleteJobAsync(string jobID, CancellationToken ct)
...
+        /// <param name="ct"></param>
+        /// <returns></returns>
+        private static async Task DownloadAndProcessFiles(List<CloudBlockBlob> outputs, Action<double> ShowProgress, string resultTempDirectory, CancellationToken ct)
+        {
+            // Download all run outputs.
+            for (int i = 0; i < outputs.Count; i++)
             {
-                ShowProgress(i / toDownload.Count);
-                CloudBlockBlob blob = toDownload[i];
+                ShowProgress(i / outputs.Count);
```
hol430, line 358: Does this progress reporting work? I would have thought the result will be cast to an integer

ApsimNG/Cloud/Azure/AzureInterface.cs
hol430: If you use `yield return` you can lazily return results as they're loaded. I just looked into this a bit and unfortunately it seems that it's hard to use this with the async pattern until C#8, so I guess this is kind of a moot point for now. Once we move to .net 4.7.2 I think it would be worth adopting this pattern however.

I also think that we still want a return type of Task on this method, so that the caller can actually await it. I see that when this is called from the view, it's run in a separate thread and so you don't await it, but it would be good if this method was more usable in other contexts.

ApsimNG/Views/ListView.cs
```diff
@@ -17,10 +17,28 @@ public interface IListView
...
         /// <summary>Gets or sets the selected row in the data source.</summary>
-        int[] SelectedRows { get; set; }
+        int[] SelectedIndicies { get; set; }
 
         /// <summary>Sets the render details for particular cells.</summary>
         List<CellRendererDescription> CellRenderDetails { get; set; }
+
+        /// <summary>Add a column to the list view.</summary>
+        /// <param name="columnName">The column heading.</param>
+        void AddColumn(string columnName);
```
hol430, line 27: Why are we exposing a public AddColumn function (and AddRow, RemoveRow, etc) when we already have a public DataTable with getter and setter? I just feel like this ListView is trying to do too much. Would be much simpler if there was one way to do things, and if we need another way of doing things, we have a different view/interface for that.

I notice that the AddColumn method also doesn't interact at all with the tree model, whereas AddRow does modify the tree model, which implies that these functions need to be called in a specific order.

ApsimNG/Views/ListView.cs
```diff
@@ -246,5 +386,62 @@ private void OnTreeClicked(object sender, ButtonReleaseEventArgs e)
...
+            tree.Selection.Mode = SelectionMode.Multiple;
+            tree.RubberBanding = true;
+            tree.CanFocus = true;
+        }
+
+        /// <summary>the sort column.</summary>
+        private void SetTreeSortModel()
+        {
+            if (sortColumn == null)
+                tree.Model = null;
```
hol430, line 411: Why do we set the tree model to null here? Setting this to null means no data will be shown in the tree view. You'll notice this if you click on an experiment - nothing is shown in the UI.

ApsimNG/Views/ListView.cs
```diff
@@ -246,5 +386,62 @@ private void OnTreeClicked(object sender, ButtonReleaseEventArgs e)
...
+            {
+                var sortColumnIndex = columns.FindIndex(c => c.Title == sortColumn);
+                if (sortColumnIndex != -1)
+                {
+                    var sortMultpilier = 1;
+                    if (!SortAscending)
+                        sortMultpilier = -1;
+                    sort = new TreeModelSort(new TreeModelFilter(store, null))
+                    {
+                        // By default, sort by start time descending.
```
hol430, line 422: This is a debug comment? View doesn't (and shouldn't) know what a start time is

ApsimNG/Views/ProgressBarView.cs
```diff
@@ -0,0 +1,65 @@
...
+            owner = null;
+        }
+
+
+        /// <summary>
+        /// A method used when a view is wrapping a gtk control.
+        /// </summary>
+        /// <param name="ownerView">The owning view.</param>
+        /// <param name="gtkControl">The gtk control being wrapped.</param>
+        protected override void Initialise(ViewBase ownerView, GLib.Object gtkControl)
```
hol430, line 56: Not really related to this PR, but do you recall why we made this parameter a `GLib.Object` and not something more specific, such as `Widget`?

ApsimNG/Views/ProgressBarView.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>Sets the visibility of the progress bar.</summary>
+        public bool Visible { get => progressBar.Visible; set => progressBar.Visible = value; }
+
+        /// <summary>
+        /// Cleanup objects
+        /// </summary>
+        /// <param name="sender">The sending object</param>
+        /// <param name="e">The argument parameters</param>
+        private void OnMainWidgetDestroyed(object sender, EventArgs e)
+        {
```
hol430, line 45: Need exception handling here - if mainWidget is null apsim will crash!

ApsimNG/Cloud/Azure/AzureInterface.cs
hol353: Hmmm I didn't notice all these comments before I merged this. Oh well...

Yes probably should to a return type of Task so that I don't have to do this:
 Task.Run(() => cloudInterface.ListJobsAsync(cancelToken.Token,
                                                        p => UpdateProgressBar(p),
                                                        job => AddJobToView(job)));

ApsimNG/Cloud/Azure/AzureInterface.cs
hol353: I actually like var. I guess we don't have a policy.

ApsimNG/Cloud/Azure/AzureInterface.cs
hol353: There is no longer any progress reporting when downloading files. Probably an oversight that needs fixing one day.

ApsimNG/Views/ListView.cs
hol353: I wanted to be able to add columns and rows manually without having to go through a DataTable. In this PR rows will be added one at a time when they come in from Azure. So I guess I'm just supporting manual loading of a ListView in much the same way as the WinForms ListView does.

ApsimNG/Views/ListView.cs
hol353: This has been fixed in another PR.

ApsimNG/Views/ListView.cs
hol353: Stray comment. That must have been there originally. I don't remember adding it.

ApsimNG/Views/ProgressBarView.cs
hol353: I was wondering that as well. I wasn't game to change it.

ApsimNG/Views/ProgressBarView.cs
hol353: Hmmm maybe. Why would mainWidget be null? The other controls (listview, button) don't check for null mainwidgets.

hol430: The other controls catch exceptions so even if it's null it won't cause a crash. Right now it's hard to imagine any scenario where mainWidget would be null, but who knows if that will still be a valid assumption in 6 months' time. I don't see any reason *not* to catch any exceptions in here, when the alternative risks a crash...

PR #5512, opened 2020-08-07
Refactor of the locator API

### LINKED ISSUE #3061: Make Apsim API use generics

hol430:
Resolves #3061


hol430:
retest this please jenkins

hol430:
retest this please jenkins

PR #5524, opened 2020-08-08
Move Links

### LINKED ISSUE #5522: MaxNUptakeRate, MaxDiffusion and NSupplyFraction Properties in Root Class could be moved to C4NitrogenUptakeMethod
The MaxNUptakeRate, MaxDiffusion and NSupplyFraction properties of root are not referenced within this class at all.  They are only called referenced in C4NitrogenUptakeMethod and so could be moved into this class.  They could then be made mandatory because they would only then need to be linked when this uptake method is used.

hut104:
resolves #5522


hut104:
@jbrider @hol430 @hol353 Are you guys happy with this approach?

PR #5531, opened 2020-08-11
Change Soybean Sample to be similar to Wheat example

hut104:
working on #3509


hut104:
retest this please jenkins

PR #5535, opened 2020-08-12
Slightly improved handling of weather files in the UI

### LINKED ISSUE #5440: using excel for met file instructions
it is really handy being able to use excel files as the met file in NG: https://apsimnextgeneration.netlify.app/usage/usingexcelforweatherdata/

one of the requirements of the met file has been to have values for Tav and Tamp. It has been brought to my attention that these are now internally calculated by APSIM and are not required for the met file (I have checked and this appears to be the case). It would be good to update the text in the URL so that people do not need the angst of calculating these values.

Thanks Drew - I somehow submitted issue somehow without finishing before

hol430:
Resolves #5440

image


hol430:
retest this please jenkins

PR #5538, opened 2020-08-13
Update Multispecies simulations (prototype)

rcichota:
Working on #5356 
Need to update the simulation testing multiple species (AgPasture species no longer working)

Start by changing/updating the name of root length density output to match PMF

Not to be merged just yet, will update simulation after checking this change


rcichota:
Sorry, just noticed that this was not the right tread to get into. I'm closing int and will re-submitt in a new pull request

Issue #4970, opened 2020-03-29
Refactor AgPasture


hol353:
Perform tidy up of code by renaming and grouping variables. e.g.

StandingHerbageWt becomes Harvestable.Wt
StandingHerbageN becomes Harvestable.N
StandingLiveHerbageWt becomes HarvestableLive.Wt
...

Do this for other variables as well. e.g. AboveGround, BelowGround, Harvested.

PR #5542, opened 2020-08-13
Update multi-species simulations

rcichota:
Working on #4970 
Need to update the simulation testing multiple species (AgPasture species no longer working)
I started by changing/updating the name of root length density output to match PMF...
Changed/updated the nodes and output references to ryegrass and lucerne that are based on AgPasture
Can be merged as I this is not (not yet at least) a validation simulation...


hol353:
What's the reason for renaming RootLengthDensity to LengthDensity?

rcichota:
Hi @hol353 ,
I changed it because having ```Root.RootLengthDensity``` is rather silly and also to align with how it is done in PMF...

PR #5550, opened 2020-08-18
Removed duplicate AboveGround member from plant.

### LINKED ISSUE #5448: Plant.cs has superfluous member for abovegroundharvestable mass
The harvestable one should be removed

        /// <summary>Above ground weight</summary>
        [Link(Type = LinkType.Child, ByName = true, IsOptional = true)]
        public Biomass AboveGround { get; set; }

        /// <summary>Above ground weight</summary>
        public Biomass AboveGroundHarvestable { get { return AboveGround; } }

hol430:
Resolves #5448

@hut104 (and maybe @HamishBrownPFR fyi) the reason we had a separate AboveGround and AboveGroundHarvestable is because IPlantDamage contains these two as separate biomasses (biomi?). I have removed AboveGroundHarvestable from IPlantDamage (and Plant). Let me know if we do actually want both here.

### DIFF
--- a/Models/AgPasture/SimpleGrazing.cs
+++ b/Models/AgPasture/SimpleGrazing.cs
@@ -405,7 +405,7 @@
             foreach (var forage in forages)
             {
                 PreGrazeDM += forage.AboveGround.Wt;
-                PreGrazeHarvestableDM += forage.AboveGroundHarvestable.Wt;
+                PreGrazeHarvestableDM += forage.AboveGround.Wt;
             }
 
             // Convert to kg/ha
--- a/Models/PMF/Interfaces/IPlantDamage.cs
+++ b/Models/PMF/Interfaces/IPlantDamage.cs
@@ -19,9 +19,6 @@
         /// <summary>Total amount of above ground biomass.</summary>
         Biomass AboveGround { get;  }
 
-        /// <summary>Total amount of harvestable above ground biomass.</summary>
-        Biomass AboveGroundHarvestable { get; }
-
         /// <summary>Plant population.</summary>
         double Population { get; }
 
--- a/Models/PMF/Plant.cs
+++ b/Models/PMF/Plant.cs
@@ -60,9 +60,6 @@
         [Link(Type = LinkType.Child, ByName = true, IsOptional = true)]
         public Biomass AboveGround { get; set; }
 
-        /// <summary>Above ground weight</summary>
-        public Biomass AboveGroundHarvestable { get { return AboveGround; } }
-
         /// <summary>Used by several organs to determine the type of crop.</summary>
         public string CropType { get; set; }
 


hut104:
I'm not familiar with the need for AboveGroundHarvestable.  If I developed it, I've since forgotten and cannot think of a reason for it.

hol430:
Hmm looks like it was added back in April by @hol353 in 747cf26bcc13e9ee1280e854620a450151599bc2. Unfortunately the commit message is not very informative so I have no idea why it's here or why we need it. All I can say is that PastureSpecies does actually implement AboveGroundHarvestable in a different way to AboveGround. So maybe we do need it...@hol353 can you provide some background here?

It's only referenced [in SimpleGrazing](https://github.com/APSIMInitiative/ApsimX/blob/master/Models/AgPasture/SimpleGrazing.cs#L408).

zur003:
I'm not sure, but it may be to allow for the fact that grazers can't remove _all_ the aboveground biomass, but only down to within a centimeter or two of the surface. For pasture models, this is important. (Although if that is the case here, it might be better to call it "AboveGroundGrazeable" rather than "AboveGroundHarvestable")

mpandreucci:
The white clover in Agpasture, for an example, has AboveGroundWt, which includes stolons. However, the AboveGroundHarvestable only considers 30% of the stolons considered in the AboveGroundWt. So, I guess for the pasture species this is needed.

hol353:
As @mpandreucci  says above, in AgPasture WhiteClover not all of the stolons organ is 'harvestable' or 'grazable'. What should probably happen is that AgPasture be changed to have an above ground stolon organ and a below ground organ stolon. This would probably be cleaner from a code perspective. Until this happens though you can't remove AboveGroundHarvestable. Doing so would mean that SimpleGrazing (used widely in AgResearch) just won't work properly.

mpandreucci:
I guess that would still have the problem that even though you have stolons above ground, not all of them are harvested.

hol353:
No I'm suggesting that we put 30% of the current stolon organ into an above ground stolon and 70% of the current stolon organ into a below ground organ. We then flag the above ground stolon organ as above ground (harvestable) and the below ground one is not above ground (and hence not harvestable). It would mean that we can do away with the concept of harvestable in AgPasture. An organ would either be harvestable or not.

mpandreucci:
Yeah, I get that. But I think we need to check how stolons are accounted for as green material that does photosynthesis. Do you remember there was something around considering stems and stolons as green LAI when the cut/graze was too drastic? I need to check how that happens because I think in some situations the above ground stolons are considered in this way (so that wouldn't just be the 30% of the stolons). Does this make sense?

hut104:
Tropical pasture has Leaf, stem, crown.  We currently assume that the crown is not harvested (ie "Cut" when simulating trials that were harvested with a mower).  Not sure if we use the harvestable thing or not.  I should check.

Anyway, we have similar issues with Tropical Pasture and models like oats, wheat, etc.

hol430:
There should probably be some tests to ensure this works...this pull request really shouldn't have passed. Anyway, I'll close this for now and assign @hol353 to #5448 to implement the change suggested above.

Keith-Pembleton:
Not sure if I fully follow what this thread, but it should be possible to have below ground organs that are grazeable, take fodder beets  for example (or are those organs classified differently).    Also in extreme overgrazing situations sheep will  dig out clover stolons and lucerne crowns but not sure if that extreme situation needs to be represented.

PR #5573, opened 2020-08-23
AgPasture root organ refactor

hol353:
Working on #4970

### DIFF
--- a/Models/AgPasture/PastureBelowGroundOrgan.cs
+++ b/Models/AgPasture/PastureBelowGroundOrgan.cs
@@ -13,16 +13,13 @@
     [Serializable]
     public class PastureBelowGroundOrgan : Model
     {
-        /// <summary>Minimum significant difference between two values.</summary>
-        private const double Epsilon = 0.000000001;
-
         /// <summary>Nutrient model.</summary>
         [Link(Type=LinkType.Ancestor)]
         private PastureSpecies species = null;
 
         /// <summary>The collection of tissues for this organ.</summary>
         [Link(Type=LinkType.Child)]
-        internal RootTissue[] Tissue = null;
+        private RootTissue[] tissue = null;
 
         /// <summary>Soil object where these roots are growing.</summary>
         private Soil soil = null;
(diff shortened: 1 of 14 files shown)


Models/AgPasture/PastureBelowGroundOrgan.cs
```diff
@@ -2,91 +2,74 @@
...
     using System;
     using System.Linq;
-    using Newtonsoft.Json;
 
     /// <summary>Describes a generic below ground organ of a pasture species.</summary>
     [Serializable]
     public class PastureBelowGroundOrgan : Model
     {
+        /// <summary>Minimum significant difference between two values.</summary>
+        private const double Epsilon = 0.000000001;
```
hol430, line 17: Lower case e here...also why is this required? Is this used by code which should be using `MathUtilities`?

Models/AgPasture/PastureBelowGroundOrgan.cs
```diff
@@ -2,91 +2,74 @@
...
-        private INutrient SoilNitrogen;
+        /// <summary>Soil nutrient model where these roots are growing.</summary>
+        private INutrient nutrient;
 
         private double[] dulMM;
         private double[] ll15MM;
 
         /// <summary>The NO3 solute.</summary>
-        public ISolute NO3 = null;
+        private ISolute no3 = null;
```
hol430, line 37: Why can't we use a link for no3 and nh4?

Models/AgPasture/PastureBelowGroundOrgan.cs
hol353: This is a work in progress. I am part way through refactoring this organ to remove things like this. There will be more commits to this PR.

Models/AgPasture/PastureBelowGroundOrgan.cs
hol353: Can't use links because for multi-zone simulations, a root organ will be told which zone its roots are in.

Issue #5472, opened 2020-07-31
ReduceCanopy method should be in Leaf models


hut104:
In fact it is probably not required as LAI is settable in the ICanopy interface

        /// <summary>
        /// Set the plant leaf area index.
        /// </summary>
        /// <param name="deltaLAI">Delta LAI.</param>
        public void ReduceCanopy(double deltaLAI)
        {
            var leaf = Organs.FirstOrDefault(o => o is ICanopy) as ICanopy;
            var lai = leaf.LAI;
            if (lai > 0)
                leaf.LAI = lai - deltaLAI;
        }

hol430:
This is required by the `IPlantDamage` interface, I think it will have to stay. It could be made more efficient though.

hut104:
Can we change IPlantDamage? It seems silly to have two ways of doing something so basic.

hol430:
Well `PastureSpecies` also implements `IPlantDamage`, but its `ReduceCanopy()` implementation is a bit different - it manually calls `leaf.RemoveBiomass()`. Theoretically there could be other `IPlantDamage` implementations as well which also do it differently. So it's not as simple as just exposing LAI in the interface with a setter. I'm not sure what the best solution is.

hut104:
We should investigate why changing LAI is handled differently.  Is one of these doing something incorrectly?

jbrider:
Isn't the damage interface allowing for removal of actual leaf not just reduction in potential growth.  How is the biomass and N accounted for if you just change LAI in isolation?

HamishBrownPFR:
I never finished implementing the damage stuff from Lifecycle so it is quite possible there is some weird stuff in there.  Even it I did finish it there is still a high posibility of there being weird stuff in there

hut104:
I'm not sure they should ever be done in isolation. If you remove leaf area you should remove the biomass related to it.

________________________________

hol430:
The ReduceCanopy function is called from 2 places, once from PlantOrganFunctionalDimensionReduction and once from Plant.RemoveBiomass(). However it gets more confusing because Plant has 3 methods called RemoveBiomass and only one of them manually reduces the Canopy. To make things even more confusing, I don't think the PlantOrganFunctionalDimensionReduction actually works. This is the code:

```csharp
[EventSubscribe("DoPestDiseaseDamage")]
private void DoPestDiseaseDamage(object sender, EventArgs e)
{
[...]

PR #5580, opened 2020-08-25
Streamline canopy biomass removal

### LINKED ISSUE #5471: This method should be part of the ICanopy Interface
/// <summary>
        /// Remove biomass from an organ.
        /// </summary>
        /// <param name="organName">Name of organ.</param>
        /// <param name="biomassRemoveType">Name of event that triggered this biomass remove call.</param>
        /// <param name="biomassToRemove">Biomass to remove.</param>
        public void RemoveBiomass(string organName, string biomassRemoveType, OrganBiomassRemovalType biomassToRemove)
        {
            var organ = Organs.FirstOrDefault(o => o.Name.Equals(organName, StringComparison.InvariantCultureIgnoreCase));
            if (organ == null)
                throw new Exception("Cannot find organ to remove biomass from. Organ: " + organName);
            organ.RemoveBiomass(biomassRemoveType, biomassToRemove);

            // Also need to reduce LAI if canopy.
            if (organ is ICanopy)
            {
                var totalFractionToRemove = biomassToRemove.FractionLiveToRemove + biomassToRemove.FractionLiveToResidue;
                var leaf = Organs.FirstOrDefault(o => o is ICanopy) as ICanopy;
                var lai = leaf.LAI;
                ReduceCanopy(lai * totalFractionToRemove);
            }
        }

hol430:
Resolves #5471

### DIFF
--- a/Models/PMF/Organs/Leaf.cs
+++ b/Models/PMF/Organs/Leaf.cs
@@ -1439,6 +1439,10 @@ namespace Models.PMF.Organs
                 needToRecalculateLiveDead = true;
             }
 
+            // Reduce LAI
+            double fractionToRemove = amountToRemove.FractionLiveToRemove + amountToRemove.FractionLiveToResidue;
+            LAI *= (1 - fractionToRemove);
+
             if (amountToRemove != null && totalBiomass != 0.0)
             {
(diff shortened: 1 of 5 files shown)


hol430:
I'm going to close this, as this approach is not going to work. We need to standardise the RemoveBiomass functions in plant and decide whether this is actually doing the right thing first as per #5472.

PR #5581, opened 2020-08-25
Added subsurface drain and water table capability to SWIM

hol353:
Working on #1254


hol353:
@hut104: This should allow SWIM tile drain and water table sims to work. Needs testing.

PR #5597, opened 2020-08-28
Fix SetEmergenceDate feature when running on mono

### LINKED ISSUE #5596: Plant.SetEmergenceDate does not work in other cultures
Any simulation which uses plant.SeEmergenceDate will fail when run on Linux (and most likely Mac too). This seems to be due to `DateTime.ToString("d-MMM")` giving a different result on mono vs .net framework ("1-Jan" vs "1-Jan."). The solution is to not convert this date to and from strings, we can just stick with a DateTime object.

Edit: it is a culture problem, not a mono problem.

hol430:
Resolves #5596


Models/PMF/Phenology/Phases/EmergingPhase.cs
```diff
@@ -76,7 +76,7 @@ public double FractionComplete
         /// Date for emergence to occur.  null by default so model is used
         /// </summary>
         [JsonIgnore]
-        public string EmergenceDate { get; set; }
+        public DateTime? EmergenceDate { get; set; }
```
hol353, line 79: Changing this from string to DateTime will mean the user must enter a year doesn't it? The user needs to have the flexibility to either put in '1-jan' or '1-jan-2020' or '1/1/2020'

hol353: Actually in this case the user doesn't need to put in a year so perhaps this is ok.

hol353: Actually (again) the user will be forced to put a  year in which isn't what we want. For emergence date the user will want to enter '1-jan'

hol430: Yeah there is an automatic call to `DateTime.Parse()` in the operations code. The phase itself will ignore the year component anyway. I expect that using a DateTime will be more robust than using a string anyway.

hol353: But the intuitive way to use this property is without a year.

hol430: In operations you can still do `2020-01-01 [Wheat].SetEmergenceDate("1-jan")`. There is code which parses each argument 1 by 1. I've updated this to account for DateTimes and call DateTime.Parse(). In other contexts (e.g. scripts) you can do `wheat.SetEmergenceDate(DateTime.Parse("1-jan"));`

hol430: Actually I've just found a simpler (one-line) fix which doesn't involve changing this from string to DateTime. I'd prefer to see DateTimes here but if you prefer strings I can implement the simpler fix.

hol430:
Closing this, there is a simpler fix available, although it still means passing around of strings and multiple redundant conversions between strings and DateTimes.

PR #5601, opened 2020-08-30
Fixed a bug in SetValueOfFieldOrProperty function

### LINKED ISSUE #5599: Cannot assign particular enum in PropertyPresenter
I have encountered a problem I cannot solve.

I have two models with a MonthsOfYear (CLEM/common.cs as didn't want to add this to Clock) enum (IntervalTimer and MonthRangeTimer) (see Simulation/CLEM/Activities folder in example aspimx attached)

I can change the value of the enum in the interval timer model in the property presenter
However, when I try to change the value in the month range model, it always reverts to January (for start month, i.e. the value it was, or second value in 0 based enum list with int 1)
Upon investigation
In PropertyPresenter.OnCellChanged
-> SetPropertyValue
-> Refresh
-> Grid.EndEdit --- this again fires the OnCellChanged event and goes through code again
The second time through by the time GetCellValue is called on property (no longer in edit mode) object newvalue is correct 
By the time the ChangeProperty.Do is fired the this.oldvalue is set to the new value and WasModified is set to false and so the the old value is used and nothing changes.

For some reason in this model I cannot get SetPropertyValue to change the value before being displayed. The code knows the correct old and new value, but must be in edit mode for the enum drop down when changed is fired.

I can't find any difference between the two models.
I thought it might be because somewhere enum is saved a string rather than the int representation of enum, but all values in aspimx file are int

Is this because OnCellChanged is fired while an enum dropbox is still in edit mode?

[EnumChangeExample.zip](https://github.com/APSIMInitiative/ApsimX/files/5140497/EnumChangeExample.zip)

hol430:
Resolves #5599

### DIFF
--- a/APSIM.Shared/Utilities/ReflectionUtilities.cs
+++ b/APSIM.Shared/Utilities/ReflectionUtilities.cs
@@ -114,7 +114,7 @@
             }
             else
             {
-                BindingFlags Flags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.IgnoreCase;
+                BindingFlags Flags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic;
                 FieldInfo F = obj.GetType().GetField(name, Flags);
                 if (F != null)
                 {


lie112:
Well found. I had considered the two private variables with the same name and different capitalisation in my model may have been involved as I couldn't see any other difference to the Models that worked. Thnx

PR #5607, opened 2020-08-31
Fix autodocs

### LINKED ISSUE #5258: Documentation page is out of date
The documentation page at
https://apsimnextgeneration.netlify.app/modeldocumentation/
is out of date.  
Several models there are said to be under review, but they are now released.  When released, the paths changed and so the documentation is not being created for them here.

hol430:
This should fix #5258


hol353:
Why are you changing 2 .apsimx files instead of using a converter?

hol430:
Because I don't know why they're broken. There's nothing to convert. My guess is some reports were deleted at some point, but the graphs which referenced those reports weren't removed. These broken graphs are causing the autodocs to fail.

PR #5609, opened 2020-08-31
Apsim wheat

HamishBrownPFR:
working on #4773  Fixed calculations of DAS in Observed filed for LaTrobe


HamishBrownPFR:
@hol353  Please merge.  Small diffs in Wheat but Observed data was incorrect so needed to be changed

PR #5616, opened 2020-09-01
Fixed a bug in linear interp code

### LINKED ISSUE #5612: Using photoperiod function in xypairs results in unexpected values
Hi there: 
it's me on photoperiod function again. 
I find using photoperiod function in xypairs results in unexpected values. the result is not what you put in the xypairs. I tested in two different models, and hope i did not make it wrong. 
Input
image
Output
image
Junqi
@junqi108 @HamishBrownPFR

hol430:
Resolves #5612 

Note: this will certainly slow down the linear interp. Perhaps it would be better to throw an exception, telling the user to reorder their x/y pairs, and maintain faster execution speed?

### DIFF
--- a/APSIM.Shared/Utilities/MathUtilities.cs
+++ b/APSIM.Shared/Utilities/MathUtilities.cs
@@ -417,6 +417,13 @@ namespace APSIM.Shared.Utilities
             if (dXCoordinate.Length == 0 || dYCoordinate.Length == 0 || dXCoordinate.Length != dYCoordinate.Length)
                 throw new Exception("MathUtilities.LinearInterpReal: Lengths of passed in arrays are incorrect");
 
+            List<Pair> xyPairs = new List<Pair>();
+            for (int i = 0; i < dXCoordinate.Length; i++)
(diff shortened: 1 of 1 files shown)


hol353:
LinearInterp gets called a lot! Let's not do a sort every time it gets called. Throw an exception instead. You could also change LinearInterpolationFunction to do the sort once at start of simulation.

hol430:
Yeah that sounds like a good idea. The only problem I see is when you consider a file like the wheat validation, that linear interp is still going to re-sort its x/y pairs 2000 times (once per simulation). Which seems a bit excessive. Maybe an exception and no fix-up  would be better?

hol353:
Yep you're probably right. Just throw an exception.

hol430:
Please don't merge @hol353, I don't think this is working yet

hol430:
Ok, I've tested this and it will give a nice error message when the array is in descending order. If the array is not sorted at all, you just get weird behaviour. We could fix this but it would impose a bigger performance hit. What would you prefer?

hol353:
Let's leave it as it is. I don't want a performance hit checking for something that rarely (never?) happens.

Issue #5624, opened 2020-09-02
Eucalyptus: include temperate genotypes


sme016:
Add new met and updated obs files

hut104:
@APSIMInitiative/reference-panel We are adding new genotypes to the eucalyptus model.

This model is a little different to other crop models, which are species models.  This model includes genotypes for a range of species, and hybrids between species, and clones.

It is worth us discussing this at the next RP meeting so that everyone understands how we handle this when merging new parameterisations. (ie when we add Eucalyptus nitens and globulus, they are not new models per se).

FYI @APSIMInitiative/forestry-and-agroforestry

sme016:
The Eucalyptus and Pinus models are for use in the forest plantation industry. They can be viewed as genus level models rather than the usual species level models elsewhere in APSIM.

PR #5625, opened 2020-09-02
Add met and updated obs file to Eucalyptus validation

sme016:
Working on #5624


hol430:
That's strange...@hol353 have you changed the jenkins settings or is our bot starting to become self-aware?

hol353:
I haven't changed anything.

hol430:
@sme016 - Hi Phil, looks like the path to the observed file in the Eucalyptus.apsimx file is an absolute path. From the logs:

```
System.Exception: Error in Observed: file 'C:\APSIMxSourceTree\Tests\Validation\Eucalyptus\ObservedEucalyptus 200825a.xlsx' does not exist
```

If you open up the .apsimx file in the gui, and edit the path to the excel file, and just delete the path (so the textbox just contains `ObservedEucalyptus 200825a.xlsx`), then I think that should fix the problem.

sme016:
Ah – yes – that problem again. Thanks

PR #5627, opened 2020-09-02
Updated Gatton2011 FinalLeafNumber Data

hut104:
working on #4773


hut104:
@hol353 Please merge this.  We have removed some incorrect data and this has changed the results.
FYI @EnliWang

Issue #5628, opened 2020-09-02
Refresh documentation


hol353:
In particular, the Reference Panel want clearer guidelines for users submitting new software and science. Restructure the docs to meet this need.

hol430:
Is there more to do here, or can this be closed?

PR #5629, opened 2020-09-02
Refreshed the documentation website

hol353:
Working on #5628

also

https://github.com/APSIMInitiative/ReferencePanel/issues/61


Docs/content/Development/Software/UnitTests.md
```diff
@@ -0,0 +1,63 @@
...
+                            }
+                        }
+                    }
+                }
+            };
+
+            simulation.Run();
+
+            var soil = simulation.Children[2] as MockSoil;
+            Assert.AreEqual(soil.NO3, new double[] { 1, 2, 103 });
```
hol430, line 57: These arguments are the wrong way around, the expected value should be the first argument. Otherwise, the error message when the assertion fails is a bit misleading.

Docs/content/Development/Software/Modeldesign.md
```diff
@@ -136,4 +138,14 @@ foreach (ICrop crop in Apsim.FindAll(this, typeof(ICrop)))
 Simulation simulation = Apsim.Parent(this, typeof(Simulation)) as Simulation;
 ```
 
- 
+## Attributes 
+
+1. **[Link(IsOptional=true)]**: Applies to class fields. When applied to a field, APSIM will locate an object  of the specified type and store a reference to it in the field. Will throw an exception if not found. When IsOptional = true, APSIM will not throw an exception when an object cannot be found.
```
hol430, line 143: Is it worth describing the different types of links? Child links in particular are widely used. Linking to a collection can also be useful on occasion.

Docs/content/Development/Software/Modeldesign.md
```diff
@@ -136,4 +138,14 @@ foreach (ICrop crop in Apsim.FindAll(this, typeof(ICrop)))
...
+5. **[ViewName(string ClassName)]**: When applied to the model class, this attribute specifies the name of the User Interface view class that should be used when showing the model in the ‘Right hand panel’. Class names need to include the namespace.
+6. **[Description(string Text)]**: Provides a description of the associated class, field or property. The user interface PropertyPresenter and ProfilePresenter classes uses these to display a description in their grid controls.
+7. **[EventSubscribe(string eventName)]**: Indicates the method is an event handler and that the APSIM framework should call the method whenever an event of the specified name is published by a model in scope.
+8. **[ValidParent(Type model)]**: When added to a model, it specifies that the model can only be a child of the specified parent model. The ExplorerView looks for these attributes to determine whether a model can be dragged and dropped (or copied and pasted) onto another model.
+9. **[Display(Format="F2", ShowTotal=true, DisplayType="TableName")**: Used by the ProfileView to determine the display format (e.g. "N3") or whether a total should be shown at the top of the column.
```
hol430, line 151: This looks like a very old example, I'm pretty sure DisplayType is an enum these days, not a string. The comment also doesn't describe what the DisplayType does.

Docs/content/Development/Software/UnitTests.md
hol353: Fixed.

Docs/content/Development/Software/Modeldesign.md
hol353: Yes, I've added examples of the different types of links.

Docs/content/Development/Software/Modeldesign.md
hol353: Yep. I've rewritten the attributes section. I've decided to turn it into examples.

hol430:
@hol353 what do you think of adding this info to apsimdev, alongside the API documentation? It seems like most of the docs on netlify are intended as user docs, whereas this info is intended more for developers. I don't really mind where it goes, just maybe something to think about.

hol353:
How do I get to the API doc? There doesn't seem to be a link to it anywhere.

hol430:
http://apsimdev.apsim.info/apsimx-docs/

Link at top of page to API documentation. There's a link next to it for articles as well, which is where this stuff would probably end up if we move it over to there.

PR #5635, opened 2020-09-03
Attempt to compile disabled managers when opening a file

### LINKED ISSUE #5634: Managers don't work properly after being re-enabled
If I open a file containing a disabled manager and then re-enable the manager, its script's properties are not visible in the properties tab until I click onto the script tab and back onto the properties tab.

This is because, when we open the file, managers are only compiled if they are enabled. While the file is open, they are only compiled after editing the script tab (ie when going back to the properties tab in this case). I can think of 2 solutions:

1. Attempt to compile disabled managers when opening the file. If the script compiles with errors, don't display the errors
2. Only call `OnCreated()` for enabled models. Modify the Enabled property so that when its value is changed to true it calls `OnCreated()`

hol430:
Resolves #5634 

@hol353 - I'm not sure I like this approach. There is another (maybe better) solution to this issue - see #5634 for details. I didn't opt for this solution because it involves the OnCreated method.

### DIFF
--- a/ApsimNG/Menus/ContextMenu.cs
+++ b/ApsimNG/Menus/ContextMenu.cs
@@ -755,14 +755,23 @@
                 IModel model = explorerPresenter.ApsimXFile.FindByPath(explorerPresenter.CurrentNodePath)?.Value as IModel;
                 if (model != null)
                 {
+                    // Toggle the enabled property on the model, and change the enabled property
+                    // on all descendants to the new value of the model's enabled property.
                     List<ChangeProperty.Property> changes = new List<ChangeProperty.Property>();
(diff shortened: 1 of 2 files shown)


hol353:
It seems odd to me that we need to try and compile disabled manager scripts. I probably prefer your option 2.

hol430:
Thinking about this some more, I'm not so sure that option 2 will work. The enabled property will be modified lots of times when we don't necessarily want to call OnCreated, such as during deserialisation.

The solution I've opted for is to call OnCreated from the GUI after toggling enabled to true. I think this should be safe enough and it fixes the original problem.

PR #5641, opened 2020-09-06
Tweaked code style in hugo theme

### LINKED ISSUE #5640: Code blocks in documentation are dificult to read
Looks good.
In the code examples the comments are easier to read than the code itself - the code is a gray text.

_Originally posted by @jbrider in https://github.com/APSIMInitiative/ReferencePanel/issues/61#issuecomment-686910150_

hol430:
Resolves #5640

Before:
image

After:
image

### DIFF
--- a/Docs/static/theme-original/style.css
+++ b/Docs/static/theme-original/style.css
@@ -668,8 +668,8 @@ code {
   border: 1px solid #ffeeb8;
   padding: 0px 2px; }
 
-.hljs {
-  background: #ffffff; }
+/*.hljs {
+  background: #ffffff; }*/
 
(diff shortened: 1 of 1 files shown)


hol353:
But we don't want a black background. Most users don't like it. Perhaps we just need to choose a different theme.

hol430:
Thanks - I didn't end up getting back to this yesterday. What does the `color: #000000;` affect exactly? Is that all text inside the code blocks?

hol353:
Yes, it seems to just make the font brighter inside the code blocks. It doesn't actually seem to alter any colours.

Issue #5568, opened 2020-08-21
Add P to Nutrient Model


hut104:
Nutrient is now the released model for nutrient flows in APSIM.  It currently accounts for stocks and flows of C and N within the soil.  The model should be enhanced to allow for the stocks and flows of P.  The model should follow the existing design for N via classes for C pools and flows.  P pools should follow the design for N pools.

Keith-Pembleton:
Also need to be able to relate the P pools to a soil P test value.  @Chloe0317 et al. has just finished up work on this in classic that would be applicable.

sarahcleary:
I note action from last meeting: @Keith-Pembleton to follow up internally with @Chloe0317.  @Keith-Pembleton - complete?

Keith-Pembleton:
I think I did, @Chloe0317 did you and I chat about this?

hut104:
I think you did, and then Chloe forwarded some information.

pmasikati:
Greetings, am doing simulations for soybean production in southern Africa (Malawi, Zambia and Mozambique) and want to assess P response in soybean. P is a limiting factor in soybean and other legume production in the region. Can someone assist with information on how I can do this in APSIMx

PR #5648, opened 2020-09-10
Add Labile P

hut104:
working on #5568


hut104:
@hol353 This seems to be working properly.  I will do further testing using lab data in the near future. This can probably be merged.

PR #5664, opened 2020-09-14
Refactor soil class

### LINKED ISSUE #2245: Refactor soil class

hol353:
Resolves #2245


hol430:
retest this please jenkins

PR #5686, opened 2020-09-23
Remove empty samples from soils

### LINKED ISSUE #5647: Samples in Examples
The wheat example as an empty sample node in the soil.  This is probably due to refactoring of soil and data into the chemical node.
The empty sample node is in other examples as well.  These should be removed.

hol430:
Resolves #5647


hol430:
retest this please jenkins

PR #5687, opened 2020-09-23
Apsim wheat

HamishBrownPFR:
working on #4773


HamishBrownPFR:
@hol353 this is OK to merge.  Diffs in wheat because I have added another experiment to test set. 

I also note there are a number of duplicate observations which are causing diffs because their order is changing relative to the accepted observation they are being diffed against.  I will remove the duplicates in a separate pull request

hol353:
retest this please jenkins.

Need to do this because I've accepted stats in another pull request.

Issue #5642, opened 2020-09-07
lists in filter


MarkLieffering:
Screenshot is of a DataStore filter. Note how the values for the line "List columns to include (csv). Leave empty for all columns" runs off the screen. Is it possible to make this a vertical list (i.e. going down by rows) to improve readability?

Cheers

![image](https://user-images.githubusercontent.com/32279744/92342254-66873b80-f114-11ea-8a9f-85cd23ee41a9.png)

MarkLieffering:
@hol430 - just a follow on from this filter issue - is there a reason that outputs using square brackets as in [AGPRyegrass].AboveGroundLiveWt are not in the filtered datastore? The are only filtered if they are renamed in the report. Here are some screenshots to hopefully illustrate this issue:

the report properties:

![image](https://user-images.githubusercontent.com/32279744/92528711-52ab1900-f27d-11ea-8755-6298bf32752f.png)

the report data:

![image](https://user-images.githubusercontent.com/32279744/92529275-53907a80-f27e-11ea-8081-bc84cf1352ae.png)

the filter:

![image](https://user-images.githubusercontent.com/32279744/92529366-7de23800-f27e-11ea-9719-c738ef08d9e9.png)

the datastore  - only the outputs that have been renamed are filtered (note numbers are different than the report because of the filtering):

![image](https://user-images.githubusercontent.com/32279744/92528961-d402ab80-f27d-11ea-92b6-309928d8421b.png)

hol353:
@MarkLieffering  I've fixed the bug you identify in the 2nd comment above.

@hol430 : Re first comment above: it would be good to have a multi-line option in the filter (and PropertyPresenter in general). After looking at the grid control we use, I think this will be difficult. Perhaps we need to rethink the PropertyPresenter (i.e. don't use a grid but instead some sort of Flow Layout with normal controls) or find another grid control (or write one ourselves).

hol430:
Yeah the standard input mechanism for arrays is comma-separated values. It works well for short arrays but not for long ones. I agree that if we want this feature, we might have to think about moving away from a grid (GtkTreeView) layout for the property presenter. We could instead just generate a form on the fly (from code) using a combination of labels and Entry/CheckButton/whatever widgets for input. This would be more consistent with other software packages but I'm still not sure what would be the best mechanism for an array input though - maybe something like this would work?

![image](https://i.stack.imgur.com/SDDAs.png)

hol353:
[...]

PR #5691, opened 2020-09-25
Change property presenter to not use a grid UI

hol430:
Well, technically I haven't actually changed the property presenter yet, I've just added a new SimplePropertyPresenter and a view to go with it, which should eventually supersede the current PropertyPresenter. I've also added a demo .apsimx file demonstrating most of the UI types (ApsimX/Examples/Tutorials/PropertyUI.apsimx). **This probably shouldn't be merged yet** - would be good if someone other than me can test this thing.

There should probably be a part 2 of this at some point, where we clean this code up. For now I've just tried to mimic the existing property presenter.

image

Working on #5642


hol353:
I've pulled the branch. Seems to work really well.

I notice one problem. When I resize the window and make it narrower, a scroll bar appears and the user has to scroll to get at the drop down buttons e.g.

image

I also notice (like the old PropertyPresenter) that there is custom code in there to get lists of cultivar names, residue types, life cycle classes, table names, field names etc). Can we remove that somehow? I was thinking instead of the manager code having:

```[Display(Type = DisplayType.CultivarName)]```

they could call directly into your PropertyPresenterHelpers likelike:

```[Display(Type = DisplayType.DropDown, Values=PropertyPresenterHelpers.GetCultivarNames)]```

Simpler for the user as there will only ever be one way of specifying drop downs. The user must always specify Values which is the name of a method (local to the manager script or somewhere else).

It would make your Property.cs simpler as well - less custom code for particular models. Adding new drop down helpers over time becomes easier as well.

Doable?

**Comment by hol430** [comment]
Hmm, that scrollbar is setup so that it appears when the user resizes apsim to the point that the space allocated to the right hand panel is no longer large enough to fit the contents of the right hand panel. In this case, it looks like there's plenty of extra space still, but if you look in the dropdown for cultivar names, there's a really long one about halfway down. If you select that cultivar then mess around with resizing you'll see that the scrollbar appears as soon as you resize it to be narrower than that cultivar name.

We can change the scrollbar policy so that the window can't be resized to be smaller than the minimum width requirement. The other option would be to leave as-is and fix that really long cultivar name.

> I also notice (like the old PropertyPresenter) that there is custom code in there to get lists of cultivar names, residue types, life cycle classes, table names, field names etc). Can we remove that somehow?

Yeah this is a real mess. My work so far has mainly been on reproducing what the existing property presenter does, but we absolutely should clean this up. Unfortunately it's not that simple though - in the case of the cultivar name dropdown, there needs to be some code which figures out which plant to show cultivar names for (e.g. in a rotation). Same goes for life cycle names, table names, etc. I'm not sure what the best solution is.

**Comment by lie112** [comment]
The new style looks good. Two points to make with respect to CLEM

1.  I can easily move all the CLEM code from property presenter to a public static class stored in models\CLEM to generate the needed entries for drop down items (dynamic resource list, crop and pasture data file reader lists) and remove this code from the property presenter. I placed it there because all the other drop down generation code was there. 

2. This new approach will BREAK some of the functionality of CLEM (CLEM.PropertyTablePresenter) whereby a parent holding a range of child models of the same type produced a table displaying all entries where each child could be edited on the same page (i.e. Master property editor of all children). I am not sure if there was ever any need for this approach in other APSIM models (Eg soil with soil layers) . This also relates to the CLEM.PropertyTreePresenter that allows property categories to clean up huge lists of properties for the user and provide general and advanced properties breakdown.  This also had a parent table presenter. Much of this was designed by Shaun and I think from looking at the code it is a copy and modify of the presenter rather than an inherited model. It could probably do with the same cleaning up, but I am concerned that the multi child display and edit of properties will not be as easy to implement with this new approach. If only single model or the category tree is used, we can utilise the new style, but the multi-model property pages wont work.

As for long names. What if the length of a name in the drop down was limited and displayed with a "..." if too long.

**Comment by hol430** [comment]
1. Yep sounds good, and with this approach we could eventually remove some of the `DisplayType` enum values, by using the `Values` property of the attribute instead.
2. There is precedent for this - the BiomassRemoval model works similarly in that the property presenter UI will show properties for child models. On master branch this is implemented in the `BiomassRemovalPresenter`, which is already quite generic and could be made usable for other models with a minimum of effort. I expect that porting it to the 'new' property presenter also shouldn't be too hard

**Comment by lie112** [comment]
Once you have an example of moving to display type.dropdown with values, I'll convert all the CLEM code. 
I do have need for some additional detail passed by attribute details. In the example below the values would call the PropertPresenterHelpers or CLEM.GetResourceNames(), but I also need to identify the resource types acceptable for this property and any extra entries (e.g. Not specified) to provide parameters for the method before called. 

`        /// <summary>F
        /// Managed pasture to muster to
        /// </summary>
        [Description("Managed pasture to muster to")]
        [Models.Core.Display(Type = DisplayType.CLEMResource, CLEMResourceGroups = new Type[] { typeof(GrazeFoodStore) }, CLEMExtraEntries = new string[] { "Not specified - general yards" })]
        public string ManagedPastureName { get; set; }`

The CLEM category presenter will be fine as it is a wrapper and will just use the new simple property presenter

This is an example of the current multi property presenter and category tree property presenter
MultiPresenter
CategoryPresenter

How do I test/see the current working on the property presenter? Is it a branch somewhere?

I haven't been happy with the legacy approach used in CLEM for some of the reports and property presenters as they were copies of models and therefore poor coding as these models get updated or replaced. I did want to rework the report so that it could be inherited and a few methods overwritten to provide custom functionality (e.g. the initialisation of the variables and events as used in CLEM reports that allow different values to be provided or completely build the variables list from scratch). I didn't want to make the required changes without discussion and before I got to really understand APSIM

I could imagine
class CustomReport: Report
    public void new OnCommencing()

or it could be done by overriding the TidyUpVariableNames

This also applies to the GRIDView when it upgrades from the old treeview style as I have bespoke displays based on the old gridview.

**Comment by lie112** [comment]
PS how do you insert formatted code in these GIT comments? The insert code was a bit messy.

**Comment by hol430** [comment]
Yeah the need to specify the resource groups sounds like it's a similar problem to `DisplayType.FieldName`, where we need some additional logic to determine which table the fields can come from. I'm not sure what the best solution is here - we want to keep things simple for people writing manager scripts.

To view/test this branch, you will need to add my remote (url is https://github.com/hol430/ApsimX). I'm not sure the exact steps in source tree, but I imagine there's a "New remote" button somewhere? Then just checkout the branch on my repo called `feature/SimplePropertyPresenter`. Although if you only want to see the code changes, github [does show that](https://github.com/APSIMInitiative/ApsimX/pull/5691/files), in the "changed files" tab of this pull request.

Agreed about cleaning up some of the duplicated classes, I imagine there are a number of bugfixes which haven't made it across to the duplicated sections.

For formatting code, just wrap the code in triple backticks (and optionally specify a language to enable syntax highlighting). E.g.: 

    ```csharp
    public static void Main()
    ```

[This doc](https://guides.github.com/pdfs/markdown-cheatsheet-online.pdf) is a brief but useful cheatsheet for the github markdown flavour.

**Comment by lie112** [comment]
I haven't needed to move to the world of users and manager scripts, or even developing them myself yet.

I've been thinking and probably the best approach is a style of overloading where there are three optional properties of DisplayType attribute

Values : string[] used to fill the list if ValuesGenerator null (i.e. missing)
Extras : object[] used to store additional information needed for generator
ValuesGenerator: string Name of a method of type string[] MethodName (string[] values, object[] extras) to calculate and replace values string[] . These could all be stored in the PropertyPresenterHelper class so easily located as most using manager scripts wouldn't need to create custom versions of this would they?  

So 
1. The user can just provides a string[] in Values (e.g. new string[] { "a", "b", "c"})
2. The user provides the name of the method in ValuesGenerator to return type string[] (e.g. GetCultivarNames)
3 The PropertyPresenter will locate and fire this method (2) to get the values[] if this string is provided and pass values and extras automatically. This would be sufficient for CLEM as values could provide the extra entries to include and extras provides the list of resource types (as strings or types) from which to generate the list for the enum. If the method doesn't need extras this would be null as not provided by the user.

An advanced option would allow some form of ordering to be included (e.g. alphabetical, or put the string values provided at top or bottom of generated list. I think I currently handle this by the order the properties are provided in the attribute. 

This way the user can know the name of the generator (easily obtainable from Helpers with a particular flag or name in description), add extra information, or simply provide an array of strings to fill the enum. Regardless of the method, only the values and extras arrays need to be provided.

My 2c worth

**Comment by hol430** [comment]
Yeah something along those lines is probably where we want to end up. I think it's probably useful to be able to specify a custom ValuesGenerator function inside the declaring class (e.g. in a manager script). So maybe we could look for the method inside PropertyPresenterHelpers first, and check in the declaring class if that fails. Or maybe we add another enum representing the different built-in dropdown types which exist inside PropertyPresenterHelpers. So instead of `[Display(Type = DisplayType.DropDown, ValuesGenerator = "GetCultivarNames")]` we could have `[Display(Type = DisplayType.DropDown, DropDownType = DropDown.CultivarNames)]` or something like that. (I'd prefer to avoid hardcoding names of methods as strings if possible.)

Another potential problem with this approach is that any arguments we want to pass to the ValuesGenerator would need to be constants. E.g. we can't do this:

```csharp
[Display(Type = DisplayType.DropDown, ValuesGenerator = "GetCultivarNames", Values = new object[] { this.Crop })]
```

(Since you would need an object reference in order to access this.Crop.) Even if we could do this, it's starting to get awfully complicated.

I'll have a bit more of a think about this, but for now I think I'll extract a separate issue from this so we can move forward with this pull request. The pull request doesn't really make the DisplayAttribute situation any worse, it just doesn't make it any better.

I think as things stand, this could probably be merged. The new property presenter is disabled by default, but can be enabled by editing the user settings file and changing `UseNewPropertyPresenter` to true. The file is located at `%appdata%\ApsimInitiative\ApsimX\ApsimX.xml` on windows or `~/.config/ApsimInitiative/ApsimX/ApsimX.xml` on Linux. The `UseNewPropertyPresenter` property won't exist inside the file until you open and close apsim after pulling down this code. Let me know if you guys notice any issues - I haven't noticed any glaring problems but I'm sure there will be some.

Issue #4381, opened 2019-11-20
Where to we go from here with the ApsimX user interface


zur003:
Three years ago, I ported the Apsim user interface from Windows.Forms to Gtk# to allow for cross-platform usage. That has worked reasonably well, despite it not being a particularly mainstream technology (more precisely, the Gtk side is mainstream on Linux, and .Net is mainstream on Windows, but their intersection isn't that widely encountered). 

Now the question arises of how (and if) we can maintain this platform as the associated technologies evolve. On the Gtk side, Linux is gradually moving away from Gtk 2 (which we are currently using as the native base layer), and towards Gtk 3, with Gtk 4 on the horizon. On the Windows side, Microsoft has introduced .NET standard (and the associated .NET Core 3 implementation), and indicated their intent to move things further in that direction. Apparently .NET Framework 5 will be base on .NET Standard.

I've spend some (too much!) time recently looking into these technologies. ApsimX's Models,exe builds and runs readily under the .NET Core 3 implementation, and gives noticeably better performance than Mono on Linux (and presumably OS X) platforms. The problems arise when we turn to the user interface. This seems like a good time to consider transitioning from Gtk 2 to Gtk 3, which is not too radically different, although Gtk 3 uses a very different CSS-based mechanism for controlling the style of user interface elements, and has a bit more support baggage associated with it. Most of the existing ApsimX user interface works under Gtk 3 with only minor modifications, but there are three areas with some difficulties:

1.  The Text Editor: we have been using Mono.TextEditor, which is based on Gtk 2 and was taken from version 5.9 of MonoDevelop. MonoDevelop is now at version 8.4 (and still using Gtk 2), but the editor code can no longer be readily abstracted from the rest of that project. I have tried porting the "old" Mono.TextEditor to Gtk 3, but haven't been fully successful, and it's not clear that we want to rely on an old, complex codebase that no-one else is working on.  If we migrate to Gtk 3, I suggest that we switch to GtkSourceView as the editor. It continues to be under active development, and provides useful stuff like a code completion interface. The newer versions target Gtk 3, not Gtk 2. I have been able to successfully build a .Net wrapper for it, but have tested only its most basic features.

[...]

Issue #4333, opened 2019-11-06
Documentation development needed


sno036:
- Need need need a way to break out of generating the documentation (escape, control+C)
- Would it be possible to right click on a component and have the option to exclude all components of the same name from documentation?
(I am trying to get the Lifecycle documentation down from 500 pages long)

hol430:
> Need need need a way to break out of generating the documentation (escape, control+C)

Maybe a progress bar with a cancel button, similar to when we run simulations? Would be nice if creating autodocs didn't freeze up the UI as well.

> Would it be possible to right click on a component and have the option to exclude all components of the same name from documentation?

What is your use case here? Normally, removing a component from autodocs will also remove all descendant nodes as well. Would this new feature behave similarly and find all components of the same name and then remove them and their descendants from autodocs?

sno036:
Yes to the first point.

My use case was over an hour trimming Jeremy Whish's Lifecycle documentation from 500 to 50 pages by excluding all the models that didn't need to be included. The kicker is that all it takes is 1 second of inattention or mouse-clumsiness near the top of the simulation tree and it is all undone with all 500 pages back!

hol353:
I spent a few hours trying to put the auto-doc code on a thread so that it can be cancelled. The problem is that much of the code can only be run on the user interface thread because it requires the UI e.g. Map, Nutrient bubble diagram, graphs etc. In the end I gave up. Need to redesign chunks of the code to not need presenters and views.

hol430:
Can't you just wrap the UI code in `Application.Invoke`?

hol353:
I tried putting in code to run parts of it on the UI thread. This kinda worked but made the code messy. It was also a problem when run from the command line (i.e. what Jenkins does).

hol430:
Well we could make the views responsible for ensuring that appropriate code runs on the UI thread. It is still messy of course (you may need to wait for the Gtk event loop to process its queued events?) but I'm not sure how you're going to render graphs, maps, images, etc without the frontend. That being said, maps may be possible to render without gtk - I've done some experimenting with a library called [SharpMap](https://github.com/SharpMap/SharpMap), and I was able to rewrite most of the map view to use SharpMap instead of the HTMLView. It's far more primitive than the current javascript implementation (zoom/position are only editable via property presenter currently) but if you only care about autodocs, I expect it will be a lot more reliable than the embedded browser widget.

[...]

PR #5705, opened 2020-10-01
Refactor the map view to not use a javascript approach

### LINKED ISSUE #5241: Graph not working in Gliricidia AutoDoc
image

### LINKED ISSUE #5662: Maps not working in Autodocumentation
I observed that the maps were not working in AutoDoc today during a live client demonstration of the APSIM website. I think this may be related to an existing APSIM issue.

hol430:
Resolves #5662 
Resolves #5241
Relevant to #4381, #4333, maybe others

The new widget uses [SharpMap](https://github.com/SharpMap/SharpMap). Generally it's a little more clunky than the old map view but it does work and should be more reliable for autodocs purposes. Panning/zooming with the mouse works ok. I've modified the installers, so that they install the necessary shapefiles. I've also added a converter which resets the center coordinates and zoom level of all map components to default values which make sense for the new map view.


zur003:
Any reason not to use BruTile to let us pull in better base maps (e.g., from OpenStreetMap)?

hol430:
No particular reason - it's a good idea. The api is pretty powerful and I haven't really explored much of it. I managed to get it working using a simple shapefile for testing and never ended up changing it.

zur003:
@hol430 - I had a go at trying to use OpenStreetMap as the source for base maps. As is all too often the case, things didn't work quite as smoothly as hoped. For reasons I don't yet understand, the base maps appear to be retrieved, but aren't rendering. This is complicated somewhat by the version of SharpMap available via NuGet dating back to early 2014, even though the SharpMap code itself has had substantial (and on-going) revision since then. I'm not sure whether the old code never worked quite as intended, or OSM has changed in the intervening years. Here are some options:

1. Continue with the current crude mapping, which doesn't work very well at a local scale.
2. Try using an updated version of SharpMap (that is, build our own using the current source). This could probably be made to work, but at the cost of not being able to use NuGet to automatically pull in SharpMap with its dependencies. SharpMap itself pulls in stuff from a number of other toolkits, so it may be a bit fiddly to get everything set up nicely, and working across all platforms and frameworks.
3. Use the NuGet version of SharpMap, but try to come up with a work-around to get it working properly. We can easily get the source code for that version, so debugging is possible, but there's no guarantee a work-around will be possible.
4. Use a different tool for mapping. I've had a look at osm-gps-map, which provides a Gtk widget for the display of OSM maps, but it provides only a native platform .dll (or .so or .dylib), so we would need to generate a Gtk# wrapper to be able to use it. There's also the question of whether we could get it working with both Gtk#2 and Gtk#3 across all platforms.

Any thoughts about the best direction to take? Why are these things never simple or easy?

hol430:
Hi Eric - thanks for looking into this. Agreed that it would be good to improve the current mapping implementation, as it is rather crude. I think I had some similar problems to you when I was getting this working in the netcore3 branch - the version in nuget was really outdated. I ended up compiling the latest source myself, and uploading it to [my myget feed](https://www.myget.org/feed/hol430/package/nuget/SharpMap). I added this feed to the [NuGet.config file](https://github.com/APSIMInitiative/ApsimX/blob/netcore3/NuGet.config) on the netcore3 branch. I've used this approach for a couple of other packages which I've had to modify (OxyPlot.GtkSharp3, and, at one point, the gtksourceview package, although this is now available in nuget proper).

Depending on which exact revision of sharpmap you need, it may be possible to use a similar approach. I think the version on my myget feed targets .net standard 2.0, so in theory it may work on the main branch (although somehow I doubt it will be that simple :).

PR #5741, opened 2020-10-14
Autodocs for maps should not depend on gtk/ApsimNG

hol430:
Working on #5662


Models/Models.csproj
```diff
@@ -120,6 +120,28 @@
...
+    <Reference Include="GeoAPI">
+      <HintPath>..\packages\GeoAPI.1.7.2\lib\net40-client\GeoAPI.dll</HintPath>
+    </Reference>
+    <Reference Include="Common.Logging">
+      <HintPath>..\packages\Common.Logging.2.0.0\lib\2.0\Common.Logging.dll</HintPath>
+    </Reference>
+    <Reference Include="ProjNET4GeoAPI">
+      <HintPath>..\packages\ProjNET4GeoAPI.1.3.0.3\lib\net40-client\ProjNet.dll</HintPath>
+    </Reference>
+    <Reference Include="BruTile">
```
hol353, line 139: I wonder about adding all these extra references to the Models assembly. These are really display (export) assemblies that I'd rather weren't in Models. Can you move Map.Export back to ApsimNG so that we can keep Models just about running simulations and not displaying / visualising things. Having said that I notice PdfSharp and MigraDoc are referenced. I wonder why.

hol353: I also notice System.Windows.Forms is included in Models.exe!!!!! That needs to go.

hol430: Ok I'll close this one then if we want to keep this in ApsimNG. PdfSharp and MigraDoc in Models because they are used by summary to export the summary file to rtf. After some investigative work it appears that the reference to System.Windows.Forms was added back in 2013 but is not actually used and can probably be safely removed.

Issue #5334, opened 2020-06-24
Some tidying up is needed


hol430:
As part of my work on refactoring the Apsim.Get API, I've come across some strange code which is either inefficient, hard to read, doesn't meet our style guidelines, or some combination thereof. Specifically:

- APSIM.Shared: feels like it was written in another language (to be fair it probably was originally)
- IFunctions
- SorghumArbitrator (and possibly other parts of sorghum too, will investigate)

I may update this as I find more problems. Mainly posting this so that I don't forget about it.

jbrider:
SorghumArbitrator should now be redundant and can be deleted.

hol430:
@hol353 I notice that we have a UserInterface directory in the repo - as far as I can tell, this appears to be the legacy winforms UI which was deleted several years ago but was accidentally reintroduced in a dodgy merge (9e6a1c46e422bab9d20734d79ba83a4bc473f18c). Is this correct? If so I will go and delete it again.

hol353:
Yep. That's a legacy directory. You can delete it.

hol430:
As part of our work on .net core builds, we've had to explicitly remove several files from compilation in Models.csproj. This is because none of these files are included in the Models project in the master branch, and they do not compile. You might be surprised to hear that DataStore.cs doesn't compile, but we seem to have 2 copies of DataStore.cs, one in Models/DataStore.cs, and one in Models/Storage/DataStore.cs. @hol353 can all of these files all be removed?

```xml
  <ItemGroup>
    <Compile Remove="DataStore.cs" />
    <Compile Remove="Functions\BoundFunction.cs" />
    <Compile Remove="GRange.cs" />
    <Compile Remove="Lifecycle\LifeStage.cs" />
    <Compile Remove="Lifecycle\LifeStageGrowth.cs" />
    <Compile Remove="Lifecycle\LifeStageImmigration.cs" />
    <Compile Remove="Lifecycle\LifeStageMortality.cs" />
    <Compile Remove="Lifecycle\LifeStageReproduction.cs" />
    <Compile Remove="Lifecycle\LifeStageTransfer.cs" />
    <Compile Remove="Plant\BiomassSummariser.cs" />
    <Compile Remove="Plant\Phenology\PhotoperiodEffect.cs" />
    <Compile Remove="Plant\Phenology\VernalisationEffect.cs" />
    <Compile Remove="Report\RowEnumerator.cs" />
  </ItemGroup>
```

[...]

PR #5745, opened 2020-10-15
Remove reference to winforms from Models.csproj

hol430:
Working on #5334


hol430:
retest this please jenkins

PR #5749, opened 2020-10-15
Implemented collision detection between labels on map

hol430:
Working on #5662


hol430:
retest this please jenkins

Issue #5734, opened 2020-10-09
What does "Undo" do?


jal-frezie:
Used to text editors, I expect it to reverse changes I recently made to the script of a manager module. But instead it just seems to change focus to the top level without affecting the model at all.

hol430:
Yeah the undo button doesn't really integrate with the manager script text editor, which has its own undo-redo buffer which is usable via ctrl+z/ctrl+y. Once you're finished editing your manager script, and you click onto another node in the tree, the changes are saved to the manager model and should be undoable via the undo/redo buttons. The undo/redo buttons at the top of the screen should undo all other changes you can make. Clicking on a node counts as a change for some reason though, so clicking undo can sometimes do unexpected things. Possibly related is #5692.

PR #5766, opened 2020-10-20
Changes to ui command system

### LINKED ISSUE #5692: UI Command system could be improved
Some UI actions are currently implemented as commands but aren't actually undoable. E.g.

- Generating docs
- Running a simulation
- Clicking on a node (this one can actually be undone, but I'm not sure it makes sense to do so since it doesn't change anything)

These actions could be changed to not be commands, which would make the undo/redo buttons a bit more reliable.

hol430:
- Clicking on a node no longer counts as an undoable action
- Every ICommand keeps track of a model which it affects. This model will then be selected when clicking undo/redo

Resolves #5692
Also relevant perhaps to #5734


hol430:
retest this please jenkins

PR #5772, opened 2020-10-21
Remove unused properties from OrganArbitrator

### LINKED ISSUE #5770: WDemand and WSupply are not used in the Arbitrator
It seems `WDemand` and `WSupply` are not used in the Arbitrator anymore. I don't notice any codes to update the values and they are zero in the report.

hol430:
Resolves #5770


hol430:
retest this please jenkins

PR #5779, opened 2020-10-22
Add reference to BruTile nuget package

### LINKED ISSUE #5778: Unable to use "Add Model" context menu option twice in one session
Attempting to add a model via the 'add model' context menu option more than once fails with a type initialisation exception for a type defined in BruTile (a dependency of SharpMap).

hol430:
Resolves #5778

### DIFF
--- a/ApsimNG/ApsimNG.csproj
+++ b/ApsimNG/ApsimNG.csproj
@@ -1075,6 +1075,14 @@
       <HintPath>..\packages\SharpMap.1.1.0\lib\net40-client\SharpMap.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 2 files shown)


hol353:
Have you referenced the right issue for this PR? What does BruTile (something to do with maps) have to do with a bug in Add Model?

hol430:
As per commit message - when we use the 'add model' context menu option, we search for all known IModel implementations. Sometimes, some of these type checks fail with a type initialization exception due to the missing references to brutile which I've added in this PR. This seems to happen for me if I  click the 'add model' context menu option multiple times without restarting apsim. The first time I try it, it works fine, but subsequent attempts fail.

PR #5791, opened 2020-10-25
Improve image support in markdownview (e.g. in memos)

### LINKED ISSUE #5789: Mac_Markdown
is there a way to include pictures without having to provide the full path from the root directory

the abbreviated methods used on windows are not working in the Mac version 

would be good if images in the same folder as the simulations could be picked up and included within the markdown document

hol430:
Resolves #5789


hol430:
retest this please jenkins

PR #5823, opened 2020-11-02
Fixed gui for LinearAfterThresholdFunction

### LINKED ISSUE #5821: APSIMX-functions. LinearAfterThresholdFunction
the LinearAfterThresholdFunction is not calculating values correctly if you use the lifecycle tutorial 2.2 to check it the x trigger value is set to 10 and the slope s set to 0.01 however the image created shows a slope that starts from 0.1 not 0 so the function appears to be ignoring the trigger value and initiating from 7 this means the rate of development is faster than expected.

hol430:
Resolves #5821


hol430:
retest this please jenkins

Issue #5844, opened 2020-11-05
WEIRDO Solute Transport - Accessing the N-solute


Chloe0317:
I'd like to enable solute transport in WEIRDO and I attempted to port the simple solute cascading from the SoilWater model, basically updating the solute status in each soil layer by calculating the solute flux contained in the upward and downward water movements. 

A detailed rationale with changes to codes can be accessed via: 
https://github.com/Chloe0317/APSIMsims/blob/main/WEIRDOSoluteTransport.ipynb

Changed WEIRDO.cs and HourlyData.cs can be found via:
https://github.com/Chloe0317/APSIMsims

I ran into the error "Object reference not set to an instance of an object." when I tried to map the solute from the physicalThickness to WEIRDO layers (line 894 of [WEIRDO.cs](https://github.com/Chloe0317/APSIMsims/blob/main/WEIRDO.cs)). 

           MappedNH4 = Layers.MapMass(nh4.kgha, Thickness, targetThickness);

I have three questions:
1. Is it sensible to port the solute transport of SoilWater to WEIRDO this way? Could someone give me some pointers on how to best enable solute transport in WEIRDO please?
2. I think the error is because that I'm not accessing the ISolute properly. If so, how should I do it please?
3. Is it sensible to use the MapMass function for variables such as solute.kgha please?

@hol353 @hol430 @HamishBrownPFR @hut104 
Thank you very much!

hol430:
Looks like you're accessing the solutes the right way. The problem is that the soil standardiser runs before links are resolved. The other soil components 'work' because they don't use links to locate their dependencies. I'd say this is a bug - the soil should not be standardised until after links are resolved, possibly in the StartOfSimulation event. I'll put up a PR to try this - hopefully it doesn't break anything.

Chloe0317:
Dear Drew, 

I tried the code in the PR, unfortunately it cannot be compiled.

What I then did is to move the map solute to layers out of the MapVariables function, and put it in the OnSimulationCommencing function (line 715 of [WEIRDO.cs](https://github.com/Chloe0317/APSIMsims/blob/main/WEIRDO.cs))

            // Initialise the efficiency factor
            SoluteFlowEfficiency = MathUtilities.CreateArrayOfValues(1.0, ProfileLayers);
            SoluteFluxEfficiency = MathUtilities.CreateArrayOfValues(1.0, ProfileLayers);

            // set the solute parameters
[...]

PR #5845, opened 2020-11-05
Soils should not be standardised until after links have been resolved.

hol430:
Working on #5844


hol430:
Closing - this approach isn't going to work.

PR #5853, opened 2020-11-11
Add jenkins check to ensure that unit tests delete all temp files the…

### LINKED ISSUE #5486: Unit tests don't cleanup after themselves
Running the unit tests causes a large number of temp files to be generated which are not being cleaned up. The UI tests also explicitly change the user theme to light mode, but don't change it back to its original value when finished.

hol430:
…y create. Resolves #5486


hol353:
My PR gives the same Jenkins UI error.

hol430:
retest this please jenkins

PR #5858, opened 2020-11-13
Created documentation for build models tutorial

### LINKED ISSUE #5857: Create some documentation tutorials on how to build a model.

hol353:
Resolves #5857


Docs/content/Development/Tutorial/BuildModelTutorial.md
```diff
@@ -0,0 +1,219 @@
...
+    [Serializable]
+    [ValidParent(ParentType = typeof(Simulation))]
+    public class RainfallModifier : Model
+    {
+    }
+}
+```
+
+1. All models in APSIM must derive from *Model* - ```public class SimpleModel : Model```
+2. To allow users, via the user interface, to add the model to another model (as a child), the model developer needs to specify the valid parent models - ```[ValidParent(ParentType = typeof(Simulation))]```. This will allow users to add *RainfallModifier* to a *Simulation* model. Other common options could have been *Zone* or *Folder*. Multiple valid parents can be specified by duplicating the *ValidParent* attribute.
```
hol430, line 32: It may or may not be worth updating this segment, but I feel I should mention that any model can be added to a folder, regardless of whether there's a valid parent attribute for folder.

Docs/content/Development/Tutorial/BuildModelTutorial.md
```diff
@@ -0,0 +1,219 @@
...
+        public double OriginalRain { get; private set; }
+    }
+}
+```
+
+The new property is call *OriginalRain*. APSIM makes all public properties visible and accessible to other models in a simulation. For example, the user can use the *Report* model to output the value of this property by specifying the output: ```[RainfallModifier].OriginalRain```
+
+The *OriginalRain* property in this example also defines a private setter so that other models cannot modify the value of this property. Only the *RainfallModifier* model is allowed to modify *OriginalRain*. To allow other models to modify the property. The *private* designator can be removed leaving:
+
+ ```public double OriginalRain { get; set; };```
```
hol430, line 112: Syntax error? Probably want to remove the trailing semicolon

Docs/content/Development/Tutorial/BuildModelTutorial.md
hol353: Thanks. Fixed.

Docs/content/Development/Tutorial/BuildModelTutorial.md
hol353: Thanks. Fixed.

PR #5862, opened 2020-11-16
Fixed an unprotected division in NitrogenUptakeMethod

### LINKED ISSUE #5507: NANs in arbitrator
public double TotalUptakeSupply { get { return MathUtilities.Sum(UptakeSupply); } }

This code above from BiomassArbitrationType.cs should use Uptake.Sum() however, UptakeSupply often has NAN values provided by organs.  The source of these NAN values should be investigated.

hol430:
Resolves #5507

### DIFF
--- a/Models/PMF/Arbitrator/Methods/NitrogenUptakeMethod.cs
+++ b/Models/PMF/Arbitrator/Methods/NitrogenUptakeMethod.cs
@@ -89,7 +89,7 @@ namespace Models.PMF.Arbitrator
             //Reset actual uptakes to each organ based on uptake allocated by soil arbitrator and the organs proportion of potential uptake
             //NUptakeSupply units should be g/m^2
             for (int i = 0; i < Organs.Count(); i++)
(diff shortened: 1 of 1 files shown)


hol430:
@hol353 this has caused some marginal changes to the stats for barley, euc, maize, and wheat. The changes are all very minor so they could probably be accepted if you're happy with them.

hol353:
Hmm they aren't so small:

image

Would be good to know what is happening in Eucalyptus LAI sims. Can you try debugging one of the points that have changed a lot e.g. Wagga. Was there a divide by zero before and now there isn't? If that was the case though then I would expect the accepted LAI value to be zero. Something else is going on here.

hol430:
Ok, the differences are occurring when all organs except one have 0 uptake supply, and the other organ has a very small uptake supply (<1e-5). In this scenario, on the master branch, that organ's uptake supply is set to the total N supply, because, for example, 1e-9 / 1e-9 = 1. After my change, that organ's uptake supply is 0 because the MathUtilities.Divide() function will return 0 if the denominator is close to zero (ie <1e-5).

I'm not sure which way is better. It should be easy enough to reproduce the old behaviour if we really wanted to though.

PR #5872, opened 2020-11-17
Make sobol more flexible in terms of table/variable

### LINKED ISSUE #5870: sobol method doesnt work on [Clock].Today
Sobol method like morris crashed when i wanna run simulation on single year "clock.today
On a [Clock].Today.Year with mullti year its Ok.the last i  updated last night morris problem fixed but sobol problem still remains

hol430:
Resolves #5870


masooandy (commented):
hi 
i don't understand how this solved #5870 ? should i Chang sobol.cs file in apsimx/model/sensetivity with this one?

PR #5889, opened 2020-11-25
Refactoring Models' CLI

### LINKED ISSUE #5878: Models.exe should give an error if .apsimx file doesn't exist
When Models.exe is passed an invalid file name, it produces no output and returns exit code of 0, indicating success. This should be changed to produce an explicit error message and return non-zero exit code.

hol430:
Resolves #5878

@hol353 - I ended up using a nuget package for this but I've maintained backwards compatibility via a series of text find/replaces on the provided arguments. I think this will give us the best of both worlds. Please don't merge just yet, as there's more work to do - I just wanted to see what jenkins thinks.


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
@hol353 looks like it's failing in APSIM.POStats. Is this a bug?

hol353:
Yes it is a bug in the current PerformanceTests. The new POStats is correct I think.

hol430:
Ok, so looking at the failures in the POStats, it claims that it's missing data for soybean - specifically, Soybean.Phenology.DaysAfterSowing. However, if I run this file locally (on this branch), the data exists in the PredictedObserved table. That's if I'm reading this correctly - I'm assuming the bracketed numbers are the accepted values?

hol353:
Hmm I'm wondering if POStats is only picking up double types and not integers in the .db files. I'll have a look.

hol430:
You might be on to something - all of the missing variables are integer types as far as I can see. I think there was code in the old performance tests to only include the variable if it's of certain types, including int, double, char, and maybe DateTime as well, you'd have to double check.

On a side note, could we change the ticks on the portal webpage to be green? They're a bit hard to see against a dark background.

hol353:
retest this please jenkins

The ticks are green:
image

hol353:
retest this please jenkins

hol430:
Strange, they're not green for me:

image

jbrider:
image
looks very green on light background... and clear :)

hol430:
It's a problem for me on a light background too, maybe it's a problem with the font I'm using. If I add some regular text into the same span as the check mark it shows up as green but the checkmark is still black.

image

hol353:
retest this please jenkins

hol353:
retest this please jenkins

PR #5894, opened 2020-11-27
Should be able to subscribe to an event with a public method.

### LINKED ISSUE #5893: Cannot subscribe to an event via a public method
The events code apparently only looks for non-public event handlers, so any attempts to subscribe a public method to an event will silently fail.

hol430:
Resolves #5893


hol353:
What is the use case for wanting public event handler methods?

hol430:
Well, what is the use case for *not* wanting public event handler methods? I try it from time to time when I forget that it can't be done, and waste a few minutes each time since it just silently fails and it takes me a while to work out why it's failing. It seems a little unintuitive to me but I guess I could be persuaded otherwise. I think my specific use case is that I had a method which was subscribing to an event but could also be called from other classes. Don't recall why.

As an alternative (or supplementary) change solution, perhaps we should add a page to the documentation explaining the event subscription model.

hol353:
I don't have any real reason other then wanting as much functionality private as possible - a good programming practice in general.

Issue #5195, opened 2020-05-22
Implement parameter optimisation via SticsRPacks/CroptimizR


hol430:
I will be working with @hut104 to implement a component which performs parameter optimisation using the R code [here](https://github.com/SticsRPacks) and the apsimx wrapper functionality [here](https://github.com/hol430/ApsimOnR).

hol430:
To be implemented:

- Allow different parameter values to be applied to different simulations within a single run
- Apsim shouldn't read observed data every time the file is run
- Add support for alternative optimization methods (currently only simplex is supported)
- Improve results display. Currently, results can only be viewed/saved to a god-level summary file

---

@hut104 this should now be more or less working, so you can give it a go if you want. The example file is in Prototypes/CroptimizR/template.apsimx. Just right click simulations node and hit run. I expect it will fail while attempting to install R packages, but you never know. If it works, it should take about 2-3 minutes to run.

frank0434:
@hol430 
Both @junqi108 and I had issues with the installation of `CroptimizR`. 
According to the report message below, I tried to regenerate a new token for the laptop I use and the installation succeeded. 
But for junqi, it still reported the same message.
Wonder how apsimx handle the R package? Does it source the R packages directly from the existing R libraries or it has to install those dependencies into apsimx environment? 
The `CroptimizR` has quite a few dependencies which might be the reason cause the configuration process failed for Junqi. 

```
StdErr:
Downloading GitHub repo SticsRPacks/CroptimizR@master
Downloading GitHub repo SticsRPacks/SticsRPacks@master
Error: Failed to install 'unknown package' from GitHub:
  Failed to install 'unknown package' from GitHub:
  HTTP error 403.
  API rate limit exceeded for 203.211.93.47. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
  Rate limit remaining: 0/60
  Rate limit reset at: 2020-06-25 09:17:32 UTC
  To increase your GitHub API rate limit
  - Use `usethis::browse_github_pat()` to create a Personal Access Token.
  - Use `usethis::edit_r_environ()` and add the token as `GITHUB_PAT`.
Execution halted
```

hol430:
@frank0434 - this is probably too late to help you but apsim will now [use docker](https://apsimnextgeneration.netlify.app/usage/croptimizr/) behind the scenes to run croptimizr sins, which solves the rate limiting issue you mention above.

frank0434:
[...]

PR #5898, opened 2020-11-30
Editing a file via command-line should also run the file

hol430:
Working on #5195 

This will improve performance in optimisation scenarios, as it saves us having to open/close the file more than once per iteration.


hol430:
retest this please jenkins

hol430:
@hol353 can this be merged if you're happy with the change? It seems to provide a decent improvement to the optimiser on the machines we've tested it with, and the latest version of the R package is broken until this is merged.

PR #5903, opened 2020-12-01
Add autosave option and settings dialog, accessible from homescreen

### LINKED ISSUE #5553: How to stop ApsimNG overwriting saved version of model?
ApsimNG frequently saves the model you are working on, overwriting the previous saved version, without any indication it is doing so. I just deleted a simulation from a model for a test, and found it had overwritten the complete version and I could not get the deleted component back. The only way to avoid this is to always save any model you are working on under a different name before editing it.

hol430:
Resolves #5553

Please don't merge yet - there are a few things I want to try and add on top of this. Let me know if this is basically what you guys had in mind.

image

### DIFF
--- a/APSIM.Shared/Utilities/ReflectionUtilities.cs
+++ b/APSIM.Shared/Utilities/ReflectionUtilities.cs
@@ -384,6 +384,25 @@
             return StringToObject(dataType, newValue, CultureInfo.InvariantCulture);
         }
 
+        private static readonly HashSet<Type> numericTypes = new HashSet<Type>
+        {
+            typeof(decimal),
+            typeof(short), typeof(ushort),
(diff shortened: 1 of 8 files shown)


jbrider:
Perhaps we could add 'on run' so the user knows it only autosaves on run and not after a set time.

her123:
It looks great Drew. 
I'm wondering if it would be worth considering that when a few more options get added, having them grouped on tabbed pages could be useful?

hol430:
Good suggestions. I've updated the description for the autosave feature, and it has a tooltip as well on mouseover.

Organising the settings into tabs/sections would be consistent with other software packages. Unsure of the best way to divide the (few) available options though - maybe an appearance section and general section?

PR #5906, opened 2020-12-01
Apsim wheat

HamishBrownPFR:
Working on #4773
- Added FileName into IWeather so it can be reported
- Added more validation simulations and met files
- Altered CAMP to use external Pp response function


hol430:
Not sure I agree with adding file name to IWeather. The interface should really describe functionality common to all weather implementations, which can't really be said of the filename, which is more of an implementation detail.

HamishBrownPFR:
I am not sure putting FileName into IWeather is the best way but I could not think of something better.  We need to add FileName to a report as MetaData.  The report is in replacements (needs to be because we need to standardize lots of simulations) some of the simulations have climate control and some have Weather so need to use IWeather or else the ones with climate control throw.  So the only way I could see to get it was to put it in IWeather.  Sure the use of FileName is an implementation detail but it is also useful metadata where a simulation returns "This Simulation did not read weather from a file" as metadata

HamishBrownPFR:
retest this please Jenkins

HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
@hol353 stats have changed because additional data has been added and some duplicate data removed

hol353:
I don't like filename in weather file. Would a manager script in each sim (and in replacements) that captures metadata (like weather filename if it exists, and other metadata?) would work. Would this be better?

HamishBrownPFR:
I not sure that would work.  If it is in replacements it would still need to access Weather or Controlled Environment and would throw and error for the one that is missing in the specific simulation.

HamishBrownPFR:
Perhaps using find to find the weather class inside a try would work.  I don't really like the manager solution because there are literally hundreds of experiments and simulations that would need to have this added in and it will be a slow job to get them all added.

HamishBrownPFR:
retest this please Jenkins

HamishBrownPFR:
@hol353 I think there is a problem with the performance tests

hol353:
retest this please Jenkins

hol353:
retest this please jenkins

PR #5935, opened 2020-12-08
Remove use of HTMLView from CLEM

### LINKED ISSUE #5934: Remove reliance of CLEM on HTMLView for migration to .Net Core
As Webkit no longer supports Windows we cannot use HTMLView moving forward with .Net Core
CLEM is extensively using HTML views for the display of html/css formatted information (versions, messages and descriptive summaries and the labour allocation table). These need to be converted to markdown or provided as a link to view in a web browser outside the APSIM Next Generation UI.

Relates to #4381

lie112:
Resolves #5934


lie112:
Is it possible to get this pulled. I have the markdown update code (absolute path links) ready to pull, but it's in the same branch and really needs to go through close to this pull for functionality.

PR #5945, opened 2020-12-11
CLEM fixed example graphs to reflect new transaction details

### LINKED ISSUE #5939: Selling a proportion of last harvest not working
No crop is sold when selecting the option to sell a proportion of the last harvest

lie112:
Resolves #5939


lie112:
Is it possible for us to see any status of the upgrade manager from Git?
Would be good to know if it is running or has stopped.
I need to get the latest changes to users as soon as possible.

hol430:
Jenkins is not really related to git, so it's not possible to check its status using git. To view the status you'll need to go to jenkins' [dashboard](http://apsimdev.apsim.info:8080/jenkins).

lie112:
Thanks. cool. That's very useful.
Seems it is the documentation of every merge since 8/12 that is failing with the create installation pipeline.

lie112:
I can't find the issue with the installation and doumentation except it still seems to be in the run tests examples section. Can You tell me if it is still the CLEM examples causing this issue?

hol430:
I think it's a problem in wheat. Will see if I can reproduce it locally, but it will be a while as I'll have to run the whole file first

PR #5947, opened 2020-12-13
Adding MaintenanceRespiration to IOrgan

### LINKED ISSUE #5928: Unable to merge with recent version
I used to use the following code to sum maninanance respiration and fixation of all organs. But I cannot use it when I merge with the latest version of trunk. May I please know what has changed in Arbitrator and how I can get this working again? It seems **MaintenanceRespiration** does not exist anymore.

```
                    List<IArbitration> Organs = new List<IArbitration>();
                    foreach (IOrgan organ in Plant.Organs)
                        if (organ is IArbitration)
                            Organs.Add(organ as IArbitration);

                    double grossAssim = Organs.Sum(organ => organ.DMSupply.Fixation);
                    double respiration = Organs.Sum(organ => organ.MaintenanceRespiration);
```

BrianCollinss:
Resolves #5928

### DIFF
--- a/Models/PMF/Interfaces/IOrgan.cs
+++ b/Models/PMF/Interfaces/IOrgan.cs
@@ -18,6 +18,6 @@ namespace Models.PMF.Interfaces
         /// <summary>
         /// Maintenance respiration.
         /// </summary>
-        double MaintenanceRespiration { get; set; }
+        double MaintenanceRespiration { get; }
     }
 }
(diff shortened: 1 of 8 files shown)


hol430:
Does maintenance respiration actually need a public setter? It seems a bit odd, especially since setting this won't actually do anything on some (most?) implementations.

BrianCollinss:
Actually I tried to copy what was already in the code, assuming there was a reason for that public setter. Just let me know what I need to change and I will. Cheers.

hol430:
Ok, well in that case I'd suggest removing the setter from the interface, and from those implementations which use a calculated value (HIReproductiveOrgan and Leaf). For the rest, a private setter would make sense to me.

BrianCollinss:
Oh! I just realised what you are referring to! I thought you were talking about 'public' or 'private'! I am on it. Thanks.

BrianCollinss:
@hol430 could you please explain for me how you decide what needs to be public or private? Cheers.

hol430:
I tend to leave them as private by default unless they really need to be public (e.g. if they're a user input).

BrianCollinss:
@hol430 can this be merged now?

Issue #5615, opened 2020-09-01
Croptimiser issues


hut104:
1) Croptimiser needs to pull code down from github but the user can exceed the rate limit from github causing installation to fail.
2) RTools needs to be installed but installation process is different depending on R version
3) When installing github packages can fail when there is a cyclic dependency

hol430:
When passing simulation names and observed data to CroptimizR, we need to limit the simulation names to only those present in the observed data.

hol430:
To be implemented:

- remove option for output path. always copy results to same directory as .apsimx file
- actual english in crit function description
- fix tolerance criterion tooltip
- graph vary by on numeric column
- put initial values next to final values variable-wise in datastore
- bar chart shows repetition + .4 when clicking on a bar
- add column for optimal-ness (bool)
- create docs page in netlify site

PR #5948, opened 2020-12-14
CroptimizR Improvements

hol430:
Working on #5615


hol430:
@hut104 - as part of this, I've added a help page about croptimizr to the website. Do you think it's time to add an example into the examples directory? If so, do you think we should use the one that currently lives in prototypes, or one of the sensibility tests you've been using recently?

hut104:
I think we can make a good example using the files we were working on.  We will need to add a memo to explain the example.

Issue #4982, opened 2020-03-31
Upgrades are not being created


hol430:
Recent changes to lifecycle.apsimx have highlighted a bug in the documentation system wherein replacements are not applied to a simulation if there is a model in replacements whose name is different to the name of the .apsimx file.

PR #5949, opened 2020-12-14
Hopefully fix wheat autodocs

hol430:
Working on #4982


hol430:
@HamishBrownPFR some of the graphs in the wheat test set take a long time to render. An example is the daily graphs inside the latrobeCE experiment. Each of these graphs seems to take >5 minutes to load on my pc. This is slowing the autodocs down - it's not a huge issue but I'm just wondering if these graphs are deliberately or accidentally included in autodocs. If it's deliberate - no big deal, we can leave as is.

HamishBrownPFR:
None of the graphs in the controlled environment folder need to be included in auto docs at the moment

PR #5956, opened 2020-12-15
Added more explicit error messages to factorials

### LINKED ISSUE #5944: More explicit error messages from experiment
- When a factor contains multiple composite factor children, the descriptor should be left blank. In the event that it's not blank, an appropriate error should be shown
- When a composite factor's number of paths is not equal to its number of children it should throw an error

hol430:
Resolves #5944


hol430:
@mpandreucci - I've added some more explicit error checking to the experiment/factorial code, and it's turned up a few oddities in the agpasture test set. There's one that I'm not sure about. The error message is `Error in composite factor WinchmoreIrrig: Unused child models found: IrrigationSchedule`. This occurs because of the following factor setup:

image

There is a child of this factor called IrrigationSchedule, but there's no instruction in the factor to override the IrrigationSchedule operations node inside the base simulation. For now, I'm going to maintain the current behaviour by not overriding the IrrigationSchedule, but let me know if you think it should be overriden, and I'll go ahead and fix it.

mpandreucci:
Hi Drew.
Can you give me a couple of days to have a look at this please?
I don't think it should be using both automatic irrigation and irrigation schedule. I'll check that and I'll get back to you. Is that ok?

hol430:
Yep, sure thing - thanks.

hol430:
Hi @mpandreucci, have you had a chance to look into this yet? Would be good to get this merged one way or the other.

mpandreucci:
Hi Drew.
I'm sorry I took so long for this.
I'm just looking into this now. I think the irrigation schedule was used as a test. The paper from Frank Li mentioned that this was border dyke irrigation, so I think I used this to test the irrigation in a schedule (75 mm is the 100 mm that Frank mentioned in his paper with a correspondent efficiency for the system of border dyke irrigation) in comparison to the automatic irrigation.
In the validation, the irrigation is being done by the automatic irrigation anyway, so we can ignore the irrigation schedule.
Again, sorry it took me so long to get to this.

hol430:
Awesome, thanks for looking into that.

mpandreucci:
I'm sorry it took me so long.

hol430:
@hol353 this is showing slight regressions in the agpasture performance tests because I've changed the agpasture validation file and removed some extra irrigations (see above discussion). The previous outputs were the result of an ambiguous factor setup which now causes an explicit error rather than silently doing nothing. I think the changes could probably be accepted.

PR #5970, opened 2020-12-18
Improvements to markdownview

### LINKED ISSUE #5937: Markdown view improvements - tables
I can see that APSIM is generating its own markdown view in GTK for displaying markdown text.
I assume this will be maintained and improved in future. I assume it also means we can improve functionality over time.

Are the [markdown instructions](https://apsimnextgeneration.netlify.app/usage/memo/) correct in that should the final display look like the examples given on the webpage or is this an example taken from online and not what APSIM NextGen produces?

image

I have built a table but the cell contents do not get markdown rendered and there is no/incorrect formatting of the table. There is no centring of contents or shading of headers correctly. Is this because I have incorrectly provided the markdown text?

image

I assume all the markdown tools online are converting to html/css and thus look professional but not what we will get for now.

image

hol430:
Resolves #5937 

- Changed markdown parser used by the markdownview (to markdig)
- Changed the way tables are rendered, so that the text is selectable
- Implemented text alignment for table columns
- Allow inline/block content (ie emphasis) inside table cells


lie112:
I have a linebreak problem, that I recall may have been discussed recently. Here are some notes for your future reference if needed as I encountered a few problems with Markdig and all the new dynamic markdown code generated in CLEM to avoid HTMLview.

- Linebreaks is a well discussed area of markdown parsers. All my single \n breaks were ignored even when I added the required space before them. I have replaced my \n with \r\n in dynamically created markdown text and all html code in CLEM. **But even with spaces I cannot get my linebreaks to show up in markdown display! without creating a double line break in the text ** which leaves an extra blank line in view rather than just shifting to the next line which I can live with, but needs a replace to double the original \r\n

- The MarkDown standard requires a space after title tags # ## etc. I have added this lazy ignored space in places as it was not necessary in the previous MarkDown parser 

- I have now wrapped all local file paths in links with < and > to allow spaces to be handled as per CommonMark standard.

hol430:
Good catch on the linebreaks. It looks like we weren't considering the case of hard vs soft line breaks. I think I've fixed this - if you post a specific example of what you have vs what you would expect it to look like, I can double check that it works for your case. Agreed on other points - yes, headings require a space after the hashes and putting the link destinations in < > is a good idea.

PR #5971, opened 2020-12-18
After clicking save in the UI, the right hand panel should be redrawn.

### LINKED ISSUE #5968: Right hand panel does not get redrawn when saving/running the model
When the save or run button is clicked, the presenter for the right hand panel is detached so that any changes to the view are applied. However, the right hand panel does not get redrawn after this procedure, and the user needs to click away to another node and back again to continue making changes.

hol430:
Resolves #5968


hol430:
retest this please jenkins

hol430:
retest this please jenkins

Issue #5979, opened 2020-12-21
-ve value for solute was passed to SWIM   Total NO3(  3)


cranju:
Hi.. I was trying to run SWIM model for Potato in next gen... However I got the error as:
-ve value for solute was passed to SWIM   Total NO3(  3) =     -1.24296

error information:
FATAL ERROR: ERROR in file: C:\Users\ranjuc\Desktop\PhD_workfolder\APSIM_example files\APSIM_X_2models_potato_eg_3.apsimx
Simulation name: SWIM
System.Exception: -ve value for solute was passed to SWIM   Total NO3(  3) =     -1.24296
   at Models.Soils.Swim3.ConcWaterSolute(Int32 solnum, Double[]& concWaterSolute)
   at Models.Soils.Swim3.GetSoluteVariables()
   at Models.Soils.Swim3.OnProcess()
   at Models.Clock.OnDoCommence(Object sender, CommenceArgs e)
   at Models.Core.Simulation.Run(CancellationTokenSource cancelToken)

Can anybody help me with this?

Regards
Ranju

hol430:
Hi @cranju, if you upload your .apsimx/.met files I'll take a look at it.

cranju:
Hi @hol353, I couldn't attach my .apsimx/.met files files as it says the file type is not supported. I have sent you an email with my APSIM file.

BrianCollinss:
@cranju Compress all files in a Zip file. Or, as not a clean idea, if it is just one file, change the extension to txt.

hol430:
Hi @cranju, this is happening because you have both soilnitrogen and nutrient present in the simulation (under the soil). As I understand it, these two models essentially serve the same purpose and should not both be present in the same simulation. I removed one from your simulation and I no longer get the -ve number error. I then get another error about an invalid report variable - you're reporting `[soil].SoilWater.ESW` which fails because there is no SoilWater model in the simulation - you're using swim instead. Changing this to `[soil].Swim3.ESW` fixes the problem and the simulation runs without error.

There should really be an explicit error message when both nutrient and soiln are both present in the simulation - I'll see what I can do.

cranju:
Thanks @hol430...Much appreciate it.

PR #5982, opened 2020-12-21
Ambiguous scoped links should cause a fatal error.

hol430:
Working on #5979


hol430:
@hol353 I've changed the links infrastructure to always throw if there are multiple potential targets for a given [Link]. Previously, we only threw in this scenario if the link was not a scoped link. The problem is that it's caused jenkins to light up like a christmas tree (which I guess may be appropriate given the time of year). The reason I'm considering this change in the first place is that the other day I was debugging a simulation which had both soilN and nutrient. I only noticed this because I was getting a -ve solute value from one of them, but I wasn't getting an error due to having them both in the same simulation. The same could be said for having both swim and soilwat present. Currently, I don't think that this will give you an obvious error message.

A lot of the failures on jenkins actually seem to be legitimate ambiguities in the simulations, so maybe it's worth perservering. Let me know what you think.

hol353:
I like the idea of ensuring unambiguous links so I would prefer we try and fix all the 'legitimate' ambiguities. Can you find a 'legitimate' one and show us why it shouldn't be an error?

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
@hol353 one example is:

```
System.Exception: ERROR in file: C:\Jenkins\workspace\PullRequest_v2\ApsimX\Prototypes\MultiSpeciesPasture\MultiSpeciesPastures.apsimx
Simulation name: FRNLRuakuraMix1NRate100
System.Exception: : Found 2 matches for link Leaf in model .FRNLRuakuraMix1NRate100.Field.WhiteClover !
   at Models.Core.Links.ResolveInternal(Object obj, ScopingRules scope, Boolean throwOnFail)
   at Models.Core.Links.Resolve(IModel rootNode, Boolean allLinks, Boolean recurse, Boolean throwOnFail)
   at Models.Core.Simulation.Run(CancellationTokenSource cancelToken) ---> System.Exception: : Found 2 matches for link Leaf in model .FRNLRuakuraMix1NRate100.Field.WhiteClover !
   at Models.Core.Links.ResolveInternal(Object obj, ScopingRules scope, Boolean throwOnFail)
   at Models.Core.Links.Resolve(IModel rootNode, Boolean allLinks, Boolean recurse, Boolean throwOnFail)
   at Models.Core.Simulation.Run(CancellationTokenSource cancelToken)
   --- End of inner exception stack trace ---
   at Models.Core.Simulation.Run(CancellationTokenSource cancelToken)
   at APSIM.Shared.JobRunning.JobRunner.RunActualJob(IRunnable job, IJobManager jobManager)
FRNLRuakuraMix1NRate100 (C:\Jenkins\workspace\PullRequest_v2\ApsimX\Prototypes\MultiSpeciesPasture\MultiSpeciesPastures.apsimx) has finished. Elapsed time was 3.3 seconds.
```

Which is rather alarming. This simulation has ryegrass and whiteclover in the same paddock and whiteclover's link to leaf can resolve to either its own leaf, or the ryegrass component (pasturespecies implements ICanopy). What's happening on the master branch is that whiteclover is finding its own leaf, as the scoping rules add the descendants of the model before looking elsewhere. The problem is that you could delete leaf from under whiteclover in this simulation and it would still run, as the infrastructure would find ryegrass and use it instead.

There will be many similar scenarios in the other files.

hol353:
Ok this is alarming. What's the process for fixing this? We make the change to the code to throw an exception when it find ambiguity, raise a PR and then we fix all the broken sims on jenkins? Let's do it.

hol430:
@hol353 - I think the ManageStock script in the stock example (WheatSheepFeedlot simulation) needs revision. There are 3 paddocks in this simulation, but in the ManageStock script (which exists at the god level), the criteria to move animals to/from a field/feedlot seems to use an expression function which references the wheat model. Unforunately, there are 3 wheat models in this simulation (one per paddock). So I think that this script is actually not doing what you think it's doing. I think it's looking at the wheat crop in the first zone - not the zone into which it's considering moving the stock. If that all doesn't make sense, you can just checkout this branch and run the the WheatSheepFeedlot simulation in the stock example.

I'm not sure about the best way to fix this while retaining the generic ExpressionFunction that you're using. The expression function doesn't give enough flexibility to specify *which* wheat crop it's talking about. Perhaps the best way would be to have an expression function in each paddock and the godlevel script references these functions.

hol430:
@hut104 - the TreeProxy model has a link to IWeather but in the tree belt example, we have several zones which each contain a LocalMicroClimate model (LocalMicroClimate implements IWeather). This is a problem because all of these local micro climates are in scope of the tree proxy model, which is not inside a zone. In this pull request I've changed the infrastructure to throw if there are multiple targets found for a given link, so the tree belt example is now throwing. We will need to rethink how the treeproxy model accesses the weather component to get this PR over the line.

hol430:
@hol353 - the PatchyMcPatchFace simulations will also need some revision if we want this to go through. The NutrientPatchManager seems to dynamically add the nutrient model as a child while the simulation is running. The problem is that the CarbonFlow models (which are children of nutrient) have scoped links to ISolutes (NO3 and NH4). Normally, these solutes are children of nutrient, and the carbon flows are also descendants of nutrient - so they're cousins once removed, so they need a scoped link. In this case though, the NutrientPatchManger also has some child solutes called NO3 and NH4, which causes the infrastructure to throw because the link could resolve to the solutes in the patch manager, or they could also resolve to the solutes in the nutrient model.

hol353:
PatchyMcPatchFace and the Agroforestry simulations have been built using the previous scoping rules of 'find closest match'. This PR changes that rule to 'find exact match'. As a result it is very hard to find way to make them work with this new scoping rule. We need to have a group think with whiteboard about this in the near future.

hol430:
Agreed, I haven't been able to think of a good solution. Would be good to get this sorted soon, before we get any more merge conflicts on this branch.

PR #5990, opened 2021-01-04
Tentative fix for wheat autodocs

### LINKED ISSUE #5978: Observed values for Soil.SoilWater.SW(7) in wheat test set
There are two simulations in the wheat test set (Lincoln2014IrrigLate and Lincoln2014IrrigNil) which have a zero observed value for Soil.SoilWater.SW(7). These should be removed.

hol430:
Working on #4982 
Resolves #5978


hol430:
@hol353 I've removed some zeroes from the observed SW in the wheat test set for the Lincoln2014 experiment. This has slightly changed the stats, but due to the bug addressed in APSIMInitiative/APSIM.PerformanceTests#40, these are showing up as failures in the POStats - note how RMSE has gone down but it's calling this a failure. We need to merge this to get the upgrades going again so I've accepted the stats here.

image

PR #5996, opened 2021-01-05
Improvements to herd management for small herds

### LINKED ISSUE #5995: Problems encountered in herd management with very small herd
A number of bugs and missing functionality was identified when managing smallholder herds.

lie112:
Resolves #5995

Big CLEM update for new year.

Also adds over Christmas improvements to summary descriptive saving and display
Also does a lot of CLEM code clean up.


hol430:
What's the motivation for moving chart.min.js into Models?

lie112:
There is a need to create the full descriptive summary (html) when the model runs for transparent reporting of CLEM model setup, irrespective of whether the user accesses this through the UI (by clicking on a CLEM component). Therefore, I had to move the html creation code from the CLEMSummaryPresenter to Model namespace in the CLEMModel so this process is now independent of the UI or even need a user interface. This is because while the ApsimNG UI uses Models, the models namespace does not have links to any UI. The creation of graphs in the CLEM summary descriptions uses the chart.min.js script. I could hard code this in  the creation of css, but it's easier to have it as an embedded resource, only added to the html string if needed.

lie112:
Can I get the latest CLEM updates approved now that everyone's back from holidays?

PR #6007, opened 2021-01-13
Improve datastore support of table names containing whitespace

### LINKED ISSUE #6004: Excel input doesn't work when multiple sheet names contain spaces
When using the excel input component with more than one sheet name, if any of the sheet names contain spaces then the component will fail with sql syntax error.

### LINKED ISSUE #6006: DataStore doesn't work well with table names which contain spaces
The DropTable command fails if the table name contains spaces. There may be other parts of the datastore which similarly fail in this scenario, which can occur in user defined tables such as reports or observed tables.

hol430:
Resolves #6004 Resolves #6006


hol430:
retest this please jenkins

PR #6071, opened 2021-01-27
Include multi-harvest crop feature

### LINKED ISSUE #6070: Allow multi harvest of crop
We need the ability to allow multiple harvests from an input data file between the first harvest (with associated tasks such as preparing and planting) and the last harvest of a sequence before doing post harvest tasks or rotating to the next crop.

lie112:
Resolves #6070


lie112:
Is the current problem with APSIM.POStats.Shared my problem?
I am getting a couple warnings that FireBirdClient in Shared has different target framework
Also Rest.ClietRuntime, System.Buffers, Azure.Batch, Sqlite.Core SQLite.EF6, Sqlite.Linq, System.memory and CompilerServices.Unsafe. 
Should I be updating these, or will these be updated automatically in future?

PR #6079, opened 2021-01-27
Markdownview needs to account for hard vs soft line breaks.

### LINKED ISSUE #6073: MarkdownView doesn't correctly render paragraphs containing single newlines
If a paragraph contains a single newline, the markdownview renders this incorrectly - it correctly discards the newline, but it needs to insert a space character when it does so.

hol430:
Resolves #6073


hol430:
retest this please jenkins

Issue #6106, opened 2021-01-29
Remove old property UI


hol430:
The new property UI has been tested and is ready to go. We can phase out the old propertyPresenter/GridView and replace it with the new one.

lie112:
I will have to update a couple CLEM presenters to point to the new one unless it is a direct name substitution. I am moving house today so can address this next week.

hol430:
No problem. If it's easier for you then we can do a direct name substitution I think. @hol353 any preference here

hol353:
I don't have any preference for which way you do this.

lie112:
I'm looking into how Shaun created the CLEM PropertyTreePresenter, PropertyTablePresenter and PropertyTreeTablePresenters in Presenters/CLEM as I think he made copies of the old PropertyPresenter and added additional changes. These are used to create

![image](https://user-images.githubusercontent.com/19215637/106422358-fce24200-64a5-11eb-8bec-0fd7a06720ea.png)

and 

![image](https://user-images.githubusercontent.com/19215637/106422421-16838980-64a6-11eb-9494-a14acc4de457.png)

I was never really happy with this approach as it is not very flexible and wont take advantage of any improvements to the copied original presenter as we're seeing now.

What is the name of the new PropertyPresenter? 

I will have a look at whether I can just substitute the new property presenter or whether the modifications to the presenter would make it work for all. The standard CLEM wrapper with a property grid will be fine and we can always just leave the other detailed presenters there for now until I update them. If the new presenter no longer uses a grid we might be stuffed and a legacy of what we had to work with in early APSIM Next Gen development.

Likewise there's some duplication of Reports into CLEM reports based on Shaun's initial development. I wanted to make the Standard report a little more inheritable to create custom reports (e.g. dynamic creation of properties and events), but haven't had the time.

hol430:
The new presenter is called SimplePropertyPresenter and the view is called PropertyView. The way they're used should be essentially the same as the old PropertyPresenter/GridView combo (just call .Attach() on the presenter and add the view's MainWidget to the parent view).

The first picture above looks like it's using the PropertyPresenter/GridView so I would imagine you could just change the type names of the presenter and view and it should just work. The second image won't work with the new property UI, as it doesn't look like it's using the PropertyPresenter. It might be possible to adapt it to use the PropertyView with a different presenter - or perhaps with some minor changes to the model's attributes it might be possible to use the SimplePropertyPresenter.

lie112:
[...]

PR #6126, opened 2021-02-04
Implement SimplePropertyPresenter in CLEM

lie112:
This replaces the old PropertyPresenter in all CLEM derived UI presenters
This applies the new SimplePropertyPresenter to the TreePropertyView offered by CLEM
The table property presenter still needs work to update, so will maintain current look and feel for now.

Working on #6106


ApsimNG/Classes/Property.cs
```diff
@@ -310,6 +312,49 @@ public Property(object obj, PropertyInfo metadata)
                     if (Value is IEnumerable enumerable && metadata.PropertyType != typeof(string))
                         Value = string.Join(Environment.NewLine, ((IEnumerable)metadata.GetValue(obj)).ToGenericEnumerable());
                     break;
+                case DisplayType.CLEMCropFileReader:
```
hol430, line 315: Do we actually need these extra DisplayTypes? It's possible to have a dropdown populated with custom values only known at runtime by using DisplayType.DropDown and setting the Values property of the attribute. There's an example at Examples/Tutorials/PropertyUI.apsimx.

ApsimNG/Presenters/SimplePropertyPresenter.cs
```diff
@@ -133,7 +133,7 @@ protected virtual PropertyGroup GetProperties(object obj)
         /// <param name="obj">Object whose members will be retrieved.</param>
         private IEnumerable<PropertyInfo> GetAllProperties(object obj)
         {
-            BindingFlags flags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.DeclaredOnly | BindingFlags.FlattenHierarchy;
+            BindingFlags flags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.FlattenHierarchy;
```
hol430, line 136: Why do we need the DeclaredOnly flag? In most cases we will want to show inherited members.

ApsimNG/Views/PropertyView.cs
```diff
@@ -100,7 +105,16 @@ public void DisplayProperties(PropertyGroup properties)
                     }
                 }
             }
-            box.Label = $"{properties.Name} Properties";
+
```
hol430, line 108: I think we can just get rid of this frame altogether.

ApsimNG/Classes/Property.cs
lie112: These were used before Values was provided.
I see that Values will replace them
Is values a comma delimited list as it's a single string. Why didn't you make it a string[] from the attribute? 
I am switching to Values, but it's a heap of changes to all attributes and so won't come through until next big commit which includes getting all the table views to work, but I need to pass the logic of that passed you as I needed to make some changes to SimplePropertyPresenter and PropertyView to handle the new logic. Good thing is it's all clean and no repeated code and gets rid of all the old CLEM presenters

ApsimNG/Presenters/SimplePropertyPresenter.cs
lie112: Didn't I remove it? It was present in your code and so the new Property Grid was not displaying any inherited properties.

ApsimNG/Views/PropertyView.cs
lie112: We can leave the logic there for now and get rid of all the displayFrame concept completely.
I can't edit glade files.
I thought you needed it to name all your sub groups or something.

lie112:
Want me to push all the following changes that get the multicolumn property grid working with SimplePropertyPresenter?
It was pretty easy in the end. I have added a List<object> models to SimplePP to hold list if needed. The refresh method now gets this group if the multi view has been specified. There is now an additional GetProperties(List<object> obs) that simple steps through all models and builds upon the propertymap ad submodelproperties returning a List<PropertyGroup>.
The PropertyMultiView accepts this List<PropertyGroup> into the 2nd display properties method and handles sizing the table and placement or widgets.
Otherwise there is no change to any of the SimplePropertyPresenter and PropertyView code and logic.
There is also a PropertyTreeMultiView (in name only as inherits from PropertyTreeView but allows us to specify multi column version from attributes) which uses the property tree wrapper with the multi column view.
I am going to rename PropertyTreePresenter -> SimplePropertyCategorisedPresenter etc to get rid of word tree in future.

ApsimNG/Classes/Property.cs
hol430: Makes sense. Yeah, the values property actually contains the name of a method in the same class which returns the valid options. The return type of the method should be a generic IEnumerable type such as string[], so it lets you fill the dropdown with values known only at runtime. There's an example in PredictedObserved.cs.

ApsimNG/Presenters/SimplePropertyPresenter.cs
hol430: Good point, I must have misread this last week

ApsimNG/Views/PropertyView.cs
hol430: I think it may just be a visual thing, but I'll double check and remove at a later date.

hol430:
I'd rather not have two modes of operation for the presenter, we've done that in other places and it makes things more difficult to understand/debug. That being said, it's hard to comment further without seeing the code - might as well push it.

lie112:
I'm not sure how to address these conflicts.

lie112:
I forgot the release master. I have pulled the release master to local and there are lots of conflicts I need to fix
A lot is implementing the dot net core stuff.
And there's a lot of #if netframework #end if
Just to clarify, I get rid of all the <<<< Head and >>>> commit id  rows in files don't I and accept all the #if code?

hol430:
It will really depend on the context. In some cases, you may want to accept the incoming changes (the #if s), but in most(?) cases you'll probably want to accept your changes, and address any problems that appear as a result. E.g. those #if directives are usually there to address differences between the gtk2 and gtk3 APIs. So if you remove them, you might see compilation warnings/errors, or potentially differences which only show up at runtime.

lie112:
So many things seem to have changed to the PropertyView since the version I was working with.
Can you explain:
I have never seen the #if netframework in the code
Propertytable is defined as grid or table depending on netframework (I was working with the new pure table)
What is using the grid? I'll have to write new methods for the grid version of the multi column presenter
is the new .net core implemented?
The APSIMNG.csproj is hugely different
You obviously had to do some work on the ActivityLedgerGridView
Does child.Cleanup replace the old child.Destroy
what is global::UserInterface.extensions doing

hol430:
> I have never seen the #if netframework in the code

These were added because when building against .net core, we target gtk3, and when building against .net framework, we target gtk2. The gtk2 and gtk3 APIs are a bit different, so some of our code needs to be conditional on whether we're running on .net core or .net framework. These preprocessor directives could be read as #if gtk2 and #if gtk3. Ultimately that would probably be clearer - but we're not really planning on target both .net framework and .net core in the long run anyway. This is basically an interim solution.

> Propertytable is defined as grid or table depending on netframework (I was working with the new pure table)

The table widget is considered deprecated in gtk3. It's basically replaced with the Grid widget. Therefore in gtk3 (.net core) builds, we use the grid widget and in gtk2 (.net framework) builds, we use a table. The APIs are somewhat similar but subtly different (e.g. they both have an Attach method with the same arguments but the meaning of the arguments is different).

> is the new .net core implemented?
> The APSIMNG.csproj is hugely different

Each project file will build against .net core and .net framework. When resolving conflicts in ApsimNG.csproj, you probably want to accept incoming/upstream changes rather than your own. Moving forward, I would expect that we won't often need to modify the project file.

> You obviously had to do some work on the ActivityLedgerGridView

Probably. I had to modify lots of views to get the UI to build on both gtk2 and gtk3. I don't really recall the details of this class but the git logs should show what I've changed.

> Does child.Cleanup replace the old child.Destroy
> what is global::UserInterface.extensions doing

In the gtk3 world, we need to call Widget.Dispose() rather than Widget.Destroy(). Therefore we've added a Cleanup(0 method which handles calling the appropriate method based on the target framework/runtime:

```csharp
        public static void Cleanup(this Widget widget)
        {
#if NETFRAMEWORK
            widget.Destroy();
#else
            widget.Dispose();
#endif
        }
```

> what is global::UserInterface.extensions doing

This will be pulling in a bunch of extension methods, I'm not sure which file you're talking about but most likely it's the extensions for the gtk widgets (see WidgetExtensions.cs) which improve compatibility between the gtk2 and gtk3 builds.

lie112:
Thanks. 
I assume we'll move to pure gtk3 when all is transferred to .net core. I hadn't come across the table widget until the PropertyView and worked with that thinking it was the new standard.
This is the most confusing conflict I have come across to resolve as I had deleted the old unneeded PropertyTablePresenter and PropertyTreeTablePresenter that you worked on. I'm almost there
I also noticed you have fixed up the incorrect summary info for some of these old classes. 
Are you removing tabs from in front of #region tags in the code?  Is this a standard for APSIM NG? Just like the #if Netframework
Can you send me your version of the ApsimNG.csproj? I have broken mine with a HEAD remaining and I don't know what should have been deleted.

hol430:
I don't think I was deliberately changing tabs vs spaces. In fact we generally try to avoid using regions in the code. So I'd say resolve that one whichever way you want. For ApsimNG.csproj, you can find it [on github](https://github.com/APSIMInitiative/ApsimX/blob/master/ApsimNG/ApsimNG.csproj).

Agreed about these merges being painful...I have quite a few older branches which I'm not looking forward to updating.

lie112:
It seems that the changes had no spaces or tabs before #regions and so were left aligned as were all the #if statements.
I had come across #if previously, but never needed to use it.
I did read that APSIM standard was not to use #regions, but I fond them everywhere so used them very sparingly in CLEM.
I forgot about getting code form git. I'll get all the changes fixed for your .net core dev and get the changes to you.

lie112:
Re #if NETFRAMEWORK
Where are we with the move to .net core? Will all the if netframework become redundant as soon as the move is complete?
I have installed the VS workload. and my code is still considering the NETFRAMEWORK statements. 
Which should I be aiming to get right and how do I run to see the .net core version working?

hol430:
Yep, all of the .net framework -specific code will become redundant as soon as we move to .net core for good. The projects are setup to run the .net framework version by default. If you want to try the .net core version, in visual studio there should be a little dropdown next to the green run button near the top/middle of the screen (I think the button says ApsimNG?). In the dropdown there should be an option to switch between .net framework (called net472) and .net core (called netcoreapp3.1). If you change to netcoreapp3.1, then visual studio should change which parts of the #if directives are highlighted/greyed out.

lie112:
Do I add gtk3 from nuget? I get an error
Error	MSB3923	Failed to download file "https://github.com/GtkSharp/Dependencies/raw/master/gtk-3.24.20.zip".  A task was canceled.	ApsimNG	C:\Users\lie112\.nuget\packages\gtksharp\3.22.25.128\build\GtkSharp.targets	12

hol430:
That's safe to ignore. Somewhere in your build log you should see a `Ignore Gtk extract errors, bug msbuild/issues/3884`. This will only happen the first time you build a project which references this package

lie112:
**DESCRIPTION OF CHANGES TO IMPLEMENT TABLE AND TREE/CATEGORY PROPERTY PRESENTERS USING SIMPLEPROPERTYPRESENTER**

The implementation of the SimplePropertyPresenter and PropertyView as the new standard in APSIM Next Generation required the previous Tree- and Table- PropertyPresenters used in CLEM to be updated. These were initially built by copying and repeating the code of an early version of the original propertyPresenter so did not automatically take advantage of updates and fixes. This new approach of applying this functionality for all APSIM users implements all the previous functionality by wrapping the SimplePropertyPresenter and only adding a little logic in this presenter to determine what is provided based on the views selected in attributes and a PropertyTableWrapper presenter.

**_Presenters_**
_SimplePropertyPresenter_
The new APSIM Next Generation table-based (grid in gtk#3) property presenter. This provides the ability to use all the new widgets and interaction code for special properties.

_PropertyTreePresenter (may be renamed PropertyCategorisedPresenter)_
This presenter uses a tree structure to allow filtering of the properties displayed in the SimplePropertyPresenter (right hand pane). The Categories and Sub-categories are provided as attributes of the properties of the attached model. This presenter changes the rule filter of the SimplePropertyPresenter based on the item selected in the tree structure of categories/sub-categories.

image

**_Views_**
_PropertyView_
The standard view for the property presenter.

_PropertyMultiView_
If the user specifies this view as the view attribute the property presenter will assume that the properties of all the children of the attached model are to be displayed, assuming all are of the same type. For example, if a parent model or folder contained a number of children (e.g. soil layers, individuals etc) then the properties are displayed for the type and each child is provided in a column. This allows the users to see and edit the properties of all children on a single screen (needs left and right arrow navigation added to move between columns)

image

_PropertyTreeView (May be renamed PropertyCategorisedView)_
The view used by the PropertyTreePresenter to include a treeview of categories with the simplePropertyPresenter in the right-hand wrapped panel.

_PropertyTreeMultiView_
Specifying this view with the PropertyTreePresenter will allow the tree presenter to wrap the PropertyMultiView and allow selection of categories for modifying the properties of all children of the attached model. This is inherited from the PropertyTreeView and so provides no additional functionality except to let the user specify that this style is needed.

lie112:
These commits gets APSIM NG working for .net framework
Some work is still needed to get these changes to work for .net core.

hol430:
> This allows the users to see and edit the properties of all children on a single screen (needs left and right arrow navigation added to move between columns)

Do we really want this? If I'm editing text, I expect the left/right arrows to move the caret left/right in the textbox

lie112:
I guess I was thinking more akin to excel where arrows move you around until you hit enter to go into edit mode. Up and down arrows don't matter as currently used as they are not used to move around inside entry unless multi line text box.

ApsimNG/Presenters/SimplePropertyPresenter.cs
```diff
@@ -76,9 +84,23 @@ public void RefreshView(IModel model)
         {
             if (model != null)
             {
+                // if multi model view specified get child models of model
+                if (view is PropertyMultiView)
```
hol430, line 88: I don't really like having multiple 'modes of operation' here. This is how problems started in other areas (such as autodocs and the old property presenter), where we would gradually add more and more use cases over time. In this instance, I think it might be better to create a new presenter, inherit from the simple property presenter, add on the extra bits you need, and call into the base class' code where necessary. That way the simple property presenter maintains a single purpose.

ApsimNG/Views/PropertyView.cs
```diff
@@ -149,10 +166,11 @@ public void DisplayProperties(PropertyGroup properties)
 
 #if NETFRAMEWORK
             uint nrow = 0;
+            AddPropertiesToTable(ref propertyTable, properties, ref nrow, 0);
```
hol430, line 169: If this is the same in both .net framework and .net core builds, it shouldn't be inside this conditional section.

ApsimNG/Views/PropertyView.cs
```diff
@@ -170,16 +188,26 @@ public void DisplayProperties(PropertyGroup properties)
             }
         }
 
+        /// <summary>
+        /// Display properties and their values to the user.
+        /// </summary>
+        /// <param name="properties">Properties to be displayed/edited.</param>
+        public virtual void DisplayProperties(List<PropertyGroup> properties)
```
hol430, line 195: I'm guessing this is here only to satisfy the requirements of an interface? If so, I would suggest perhaps it shouldn't be part of the interface.

ApsimNG/Views/PropertyView.cs
lie112: Yeah I know. This line was missing from the commit I got so was never adding the properties (so I couldn't get to change the "use new presenter" in system settings. I stuck it in here to see if it worked and never moved both outside. I thought in the past I had them both outside and it wasn't working. I also figured the # if would be removed soon and it would get cleaned up then.

ApsimNG/Views/PropertyView.cs
lie112: I just need a way of turning off the frame and label for my views. I figured based on the styling this should be a property we can set for the presenter when defined in the code that passes the settings to the view. Open to ideas here as I assume you will be keeping the frame and label. There does need to be some more padding in places for your controls that I will then turn off the padding I have used to address the problem.

ApsimNG/Presenters/SimplePropertyPresenter.cs
lie112: I really didn't want to muck around too much with your presenter, so did this workaround as I figured the presenter always knows it's associated view and I was only checking the type.
I will create a SimplePropertyMultiPresenter that uses the PropertyMultiView and handles creating the List< model>right now!
We can also move these Presenters and Views out of the CLEM folder as they are now more generic
Do you want "tree" renamed "categorised" in naming to better reflect functionality

lie112: I am trying to remove all multiple modes of operation in the new presenters. This means there are now 4 presenters and 4 views to perform the task. While I had everything working well, I have now broken the logic somewhere. When I click on the nested treeview for the treepropertypresenter it fires the explorer presenters onNodeSelected event. I cannot work out how this happens as the category tree presenter is hooked up to it's own OnNodeSelected and other than the explorerPresenter passed with the Attach has no knowledge of the treeview in the explorer. I am tracking down whether I use an incorrect view somewhere. I have spent a day on this task and might have to put this update on hold for a bit.

ApsimNG/Views/PropertyView.cs
hol430: We're planning on getting rid of that frame and its label anyway. If it makes things easier for you, feel free to remove it as part of this PR. Would that allow this method to be removed from the base class/interface?

lie112:
I have removed all multiple modes of operation from the CLEM-based property presenters (no 4 presenters [inc SimplePP] x 4 views to cover range
I still have a few formatting issues I am struggling to solve in GTK3. I have copied the explorer.glade to propertyTreeView.glade so I can make changes for my case, but editing without the interface is hard.
image
In the image above I would like

- The tree window to dynamically resize with a border padding at 2. This helps fit most on the screen for smaller windows
- I would like to get rid of the background of the explorer view at 1 and have no padding so it all moves up
- Is there a min size of boxes defined or can we have the entry boxes a little more responsive to screen size at 3.

After this commit I will do the name changes to get rid of "tree" and replace with "categorised" in my presenters and views as I think it is a better description of what they do.

I'll also address the DisplayFrame so it's not yet ready to accept.

hol353:
Is this ready for merging?

hol430:
Sorry, I haven't had a chance to look at these extra changes for the gtk3 version yet. Will try and get onto it later today or tomorrow.

If it helps, it's possible to install and run the glade editor on windows, and it makes it a lot of easier to edit the .glade files.

lie112:
Yes, all changes checked and working.
Is ready as soon as @hol430 has a final look over.

I noticed a desire to remove views in future and have only presenters and a glade files #6129. I have been able to make boxes with packing a labels in codes to do everything I needed. I should convert them to glade. I used to have the glade editor, but wasn't doing much UI at the time and haven't reinstalled on new computer.

ApsimNG/Properties/Settings.Designer.cs
```diff
@@ -12,7 +12,7 @@ namespace ApsimNG.Properties {
     
     
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "16.7.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "16.8.1.0")]
```
hol430, line 15: Why has this been changed? There's also a corresponding change to the csproj which we probably don't need.

ApsimNG/Classes/Property.cs
hol430: Are we able to get rid of these display types and replace them with uses of DisplayType.DropDown?

ApsimNG/Presenters/PropertyMultiModelPresenter.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using UserInterface.Classes;
+using UserInterface.Interfaces;
+using UserInterface.Presenters;
+using UserInterface.Views;
+
+namespace UserInterface.Presenters
+{
+    public class PropertyMultiModelPresenter: SimplePropertyPresenter
```
hol430, line 15: Could you add a description to describe what this class does

ApsimNG/Presenters/ReportPresenter.cs
```diff
@@ -123,7 +123,8 @@ public void Detach()
             this.view.EventList.TextHasChangedByUser -= OnEventNamesChanged;
             this.view.GroupByEdit.Changed -= OnGroupByChanged;
             explorerPresenter.CommandHistory.ModelChanged -= OnModelChanged;
-            dataStorePresenter.Detach();
+            if(dataStore != null)
```
hol430, line 126: Just out of interest - did you encounter a scenario where dataStore was null at this point?

ApsimNG/Views/PropertyView.cs
```diff
@@ -170,16 +174,26 @@ public void DisplayProperties(PropertyGroup properties)
             }
         }
 
+        ///// <summary>
```
hol430, line 177: Can this be deleted rather than commented out

ApsimNG/Views/PropertyView.cs
```diff
@@ -197,18 +211,22 @@ private void AddPropertiesToTable(ref Grid table, PropertyGroup properties, ref
...
+                {
+                    Label label = new Label(property.Name);
+                    label.TooltipText = property.Tooltip;
+                    label.Xalign = 0;
 #if NETFRAMEWORK
-                propertyTable.Attach(label, 0, 1, startRow, startRow + 1, AttachOptions.Fill, AttachOptions.Fill, 5, 0);
+                    propertyTable.Attach(label, 0, 1, startRow, startRow + 1, AttachOptions.Fill, AttachOptions.Fill, 5, 0);
 #else
-                propertyTable.Attach(label, 0, startRow, 1, 1);
+                    label.MarginRight = 10;
```
hol430, line 226: The MarginRight property is deprecated as of gtk3 - it's replaced with MarginEnd. On a side note, is this extra spacing not required in the gtk2 builds?

ApsimNG/Views/PropertyView.cs
```diff
@@ -197,18 +211,22 @@ private void AddPropertiesToTable(ref Grid table, PropertyGroup properties, ref
                         propertyTable.Attach(box, 0, 3, startRow, ++startRow, AttachOptions.Fill | AttachOptions.Expand, AttachOptions.Fill, 5, 5);
 #else
                         propertyTable.Attach(box, 0, startRow, 2, 1);
-                        startRow++;
 #endif
+                        startRow++;
```
hol430, line 215: Hmm, might want to double check the implications of this in gtk2 builds - pretty sure this will result in a double increment of this variable (it's also incremented on L211 in .net framework builds). If this does indeed cause problems it could be moved back into the .net core-specific section - or you could change line 211 to not increment the variable (ie replace `++startRow` with `startRow + 1`).

ApsimNG/Properties/Settings.Designer.cs
lie112: I don't know. It just happened. If we reject this change I guess my system will start up correctly the next time without it.
I'm note sure what the version update was all about.

ApsimNG/Presenters/PropertyMultiModelPresenter.cs
lie112: I can no longer get back to this branch. I get a whole bunch of pages that cannot be added. The same is true trying to revert to my previous CLEMWorking branch.. I am now working on a newly created branch from the latest release master commit.
Didn't realise I missed this. I can add this when all is up and running.

ApsimNG/Presenters/ReportPresenter.cs
lie112: Yes. This happens if there is a problem in the Attach process before the datapresenter is attached. It occurred in one of the CLEM reports

ApsimNG/Properties/Settings.Designer.cs
lie112: I have reversed this, but can't change the csproj change in sourcetree.

ApsimNG/Presenters/ReportPresenter.cs
lie112: It happens in a CLEM report that does not have any Events defined as it creates them dynamically. Therefore the view.EventList.Lines line of code fails in Attach and the code never gets to set up the datastore. I could move the datastore code to the top and then this would  not be needed. I could also have trapped adding the EventList lines of the report.EventNames was empty. Either way trappng it in the detach seemed the best approach. This must have been an update added that now affects CLEM reports with no Events as it previously worked. I am pretty sure this fixes the problem

ApsimNG/Views/PropertyView.cs
lie112: Oops, missed that one in the process of deleting this functionality.

ApsimNG/Views/PropertyView.cs
lie112: changed!. I haven't been too concerned with keeping the gtk2 stuff working as I figure we're not far of switching to gtk3. It's not critical and can be lived with in gtk2 and I think the table handled it better than the gtk3 grid implementation.

ApsimNG/Views/PropertyView.cs
lie112: thanks. missed that from original code.
Things have gone haywire in .Netframwork version. That I also need to fix

ApsimNG/Views/PropertyView.cs
lie112: The Table handles all padding much better than the grid. 
I will also need some left padding of labels or the entire simplePropertyPresenter in my CategorisedPresenter.

lie112:
These new presenters will only use the new SimplePropertyPresenter and are therefore ready for the move to this style. They will no longer work with the old presenter for these CLEM entries.
I am unable to get the right hand presenter to display in the Categorised view (simple or multiple model). I can step through and see all the properties being read and widgets created, but nothing displays. I am not sure what is different to the NetCore version in this respect that works fine.

hol430:
> These new presenters will only use the new SimplePropertyPresenter and are therefore ready for the move to this style.

Which style are you talking about here?

> I am unable to get the right hand presenter to display in the Categorised view (simple or multiple model) UPDATE in .Net4.7. I can step through and see all the properties being read and widgets created, but nothing displays. I am not sure what is different to the NetCore version in this respect that works fine.

I can look into this- is there a specific node in a specific file I can click on to see the problem?

lie112:
Sorry for any confusion. All these presenters previously used for CLEM (tree and multi) will now only use SimplePropertyPresenter as I assume everyone is moving to that when bugs are all ironed out and we go to gtk3

You can use the Examples\CLEM\CLEM_Examples_Grazing.aspimx
Navigate through the tree to Activities\Manage Herd\Manage Numbers and you'll see the treeview get created but no simple property presenter is displayed in the right hand panel based on the tree selection. This will work in CoreApp3_1 but not in .Net4.7
Navigating to Resources\Ruminants\Bos Indicus\Initial Cohorts will show the multi model presenter and view working.
But the Ruminants and Bos indicus levels do not work as they are Categorised presenters (single and multi versions)

lie112:
Any ideas how to fix this?
I can't find a difference between core and framework in the code, but I get no display in Framework.

hol430:
Ok, looks like there are a few problems here. The main problem isn't that the properties grid isn't being displayed - it is being displayed, but the 'splitter' in the GtkHPaned widget inside the PropertyCategorisedView is all the way to the right in .net framework builds. If you move your mouse around near the right side of the window, you should see the cursor change to show the drag icon and you can drag the splitter to the left, revealing the properties UI. I would assume that the reason it's working better in the .net core builds is that the initial size arbitration algorithm in the Paned widget has been improved in gtk3.

There are a few properties you can play with the glade file to change this. What I did was change the horizontal scrollbar policy for scrolledwindow1 to 'Never', and I changed the child properties (aka the properties in the packing tab if you're using the glade editor) of both scrolledwindows, setting resize and shrink to false on scrolledwindow1, and resize to true and shrink to false for scrolledwindow2. The end result of these changes is that the horizontal scrollbar will never appear in the tree on the left panel, but the left panel can't be resized to be smaller than its preferred width - so the horizontal scrollbar is never needed. This may be a problem on an extremely narrow display such as a phone but I think most of the time it should be ok, and it's much simpler than the alternative. When we move away from gtk2 builds we can re-enable 'shrink' mode on scrolledwindow1 and enable the horizontal scrollbar.

I've also noticed some other gtk errors/warnings in the console output:

- `Gtk-CRITICAL **: gtk_box_pack: assertion '_gtk_widget_get_parent (child) == NULL' failed`

This one is caused by adding the same widget to its parent twice - in ModelDetailsWrapperView.cs, we add modelTypeLabel into vbox1 twice, on lines 94 and line 98. We don't need to do this - solution is to just delete one of these lines.

- `Gtk-CRITICAL **: gtk_notebook_get_tab_label: assertion 'list != NULL' failed`

This is caused by ClemView.cs L122. Calling GetTabLabelText() on a widget which is not a child of the notebook is apparently not correct API usage. The fix is to change this check to `if (!nbook.Children.Contains(newViewport))`.

If it helps, here is all of the changes I made:

[clem.patch.txt](https://github.com/APSIMInitiative/ApsimX/files/6026841/clem.patch.txt)

lie112:
This is now working and ready for merging if the changes are approved.

ApsimNG/ApsimNG.csproj.rej
```diff
@@ -0,0 +1,21 @@
+diff a/ApsimNG/ApsimNG.csproj b/ApsimNG/ApsimNG.csproj	(rejected hunks)
```
hol430, line 1: Looks like an accidentally-committed file?

ApsimNG/Classes/Property.cs
```diff
@@ -310,6 +312,49 @@ public Property(object obj, PropertyInfo metadata)
...
+                case DisplayType.CLEMPastureFileReader:
+                case DisplayType.CLEMResourceFileReader:
```
hol430, line 317: Can we refactor these to use DisplayType.DropDown with the Values property? That would allow us to remove these extra display types.

ApsimNG/Properties/Settings.Designer.cs
hol430: Looks like the revert didn't make its way into a commit(?)

ApsimNG/Views/PropertyView.cs
hol430: Can this be moved out of the conditional section?

ApsimNG/Classes/Property.cs
lie112: This all relates to cleaning up old code to meet the new approach. I only worked out yesterday that the values property of Display attribute represents a method and see it can have no arguments. I am thinking through how this would be implemented as I don't want to replicate the method in every class that uses it but can move it up my inheritance tree.  My first approach is to move the logic to three individual methods (for the data reader lists). I could also think about removing CLEMResources DisplayType but it's expecting to have a list of models and a list of extra values to be supplied in order to build the list. If I could have one extra array of strings (string[] AdditionalDetails) to replace ResourceGroups in DisplayAttribute I can remove all reference to CLEM and work with the standard properties. I could use this array to hold both the names of components to search for, the resource groups and extra identified values that are all  handled by the various methods called, but I'd need to add some logic to the method calling to add the arguments from the defined AdditionalDetails as arguments is specified. What do you think?

ApsimNG/ApsimNG.csproj.rej
lie112: I get lost in strange commits. How do I find and resolve this issue.

ApsimNG/Views/PropertyView.cs
lie112: done. in next commit

ApsimNG/ApsimNG.csproj.rej
lie112: Do I reverse hunk this in a previous commit, or go and find the file and delete it, raising a new commit?

hol430: You can always just delete this file - it will then show up as deleted in your git client, you stage the deleted file and commit.

ApsimNG/Classes/Property.cs
hol430: Maybe add an Args property of type object[] to display attribute. These args could be passed in when calling the method specified by Values. It's starting to get a bit convoluted but I can see why you might need this capability, and it might be useful in other contexts as well. If you're wanting to pass a string[] into the method, then your Args property would be an object[] of length 1, where the first item is your string array.

lie112: I am implementing the required changes. How far off moving to only the SimplePropertyPresenter are we? All these requested changes break the logic used in the old PropertyPresenter.cs to display these drop downs as the new approach uses a DisplayType = Dropdown and the Values property to specify the method to give labels, while the old method still needs to use the DisplayType = CropFileReader and the coded logic in FormatGrid(). I cant evensolve the problem with # if as I don't know whether the user has turned on "use new property presenter" even in NETFRAMEWORK

hol430: We're ready to remove the old property presenter. It's not the highest priority for me at the moment but it won't take long so I can do it as soon as this is merged. That being said, I thought the old property presenter already supported the Values property with DisplayType.DropDown?

lie112: Thanks. Didn't realise the Values was implemented in old Presenter. I had already been using a Values property in DisplayAttributes that was replaced still there incorrectly when not a dropdown. My current problem is my object[] containing type[] as first item is ok when obtained from the new ValuesArgs, but after Method.Invoke calls method in code the Args passed are Type[] and it has stripped the surrounding object[]. Have you come across this anywhere? Any ideas?

hol430:
> Thanks. Didn't realise the Values was implemented in old Presenter. I had already been using a Values property in DisplayAttributes that was replaced still there incorrectly when not a dropdown. My current problem is my object[] containing type[] as first item is ok when obtained from the new ValuesArgs, but after Method.Invoke calls method in code the Args passed are Type[] and it has stripped the surrounding object[]. Have you come across this anywhere? Any ideas?

I believe this is the expected behaviour. When we call MethodInfo.Invoke(), the object array isn't passed in as a single argument - each element of the array is itself treated as an argument to the method that's being called. So if your method has three arguments, you'd need to pass an object array with three elements into the call to MethodInfo.Invoke().

lie112:
Thanks, I thought I deleted this response after a bit of thought. I figured that the object array would be passed as each individual argument in the method. I think it's all good now. I have deleted all CLEM peroperties form the Display attribute and deleted all the code throughout moving everything to Values and ValuesArgs. It should be right to go. My only assumption is that if ValuesArgs is not defined it is returned as null and passed to the Invoke call as null and thus they way the code was.

hol353:
@hol430 Are you ok with this PR now?

Examples/CLEM/CurrentDescriptiveSummary.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
```
hol430, line 1: Did you mean to commit this file? I had thought this was the file that gets generated when you click on a clem node in the UI. If so, maybe delete it and add it to the gitignore file.

ApsimNG/Properties/Settings.Designer.cs
hol430: Can this change be reverted?

ApsimNG/Classes/Property.cs
```diff
@@ -8,6 +8,8 @@ namespace UserInterface.Classes
     using System.Reflection;
     using APSIM.Shared.Extensions.Collections;
     using APSIM.Shared.Utilities;
+    using Models.CLEM;
```
hol430, line 11: Looks like these two using lines aren't required - if so, could they be removed?

ApsimNG/Properties/Settings.Designer.cs
lie112: I'm not sure what you want here. I thought I got ride of the changes.

Examples/CLEM/CurrentDescriptiveSummary.html
lie112: Removed.

lie112:
This should be right to accept. I have addressed all of @hol430 suggestions and requests.

hol430:
Yep, looks good to me

Issue #6139, opened 2021-02-10
Problems identified with move to .Net Core GTK#3


lie112:
I thought it might be worth having this issue so we can all flag issues and fixes for problems identified as we start to run with netcoreapp3.1

lie112:
There's a known bug with core when trying to to open a url from System.Diagnostics.Process.Start() that we have been using.
[Here's the fix](https://brockallen.com/2016/09/24/process-start-for-urls-on-net-core/), but we should have this somewhere everyone can access rather than duplicate code wherever we need to open a url from the UI.

`        public static void OpenBrowser(string url)
        {
            try
            {
                System.Diagnostics.Process.Start(url);
            }
            catch
            {
                // hack because of this: https://github.com/dotnet/corefx/issues/10361
                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                {
                    url = url.Replace("&", "^&");
                    System.Diagnostics.Process.Start(new System.Diagnostics.ProcessStartInfo("cmd", $"/c start {url}") { CreateNoWindow = true });
                }
                else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
                {
                    System.Diagnostics.Process.Start("xdg-open", url);
                }
                else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
                {
                    System.Diagnostics.Process.Start("open", url);
                }
                else
                {
                    throw;
                }
            }
        }`

hol430:
Is this fixed by setting just UseShellExecute on the Process object to true?

(For code formatting, there's a handy cheatsheet for [github markdown](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet))

lie112:
Must say I do like the look and feel of the core version. Much more professional.

lie112:
I'm not sure how to set the UseShellExecute on the ProcessInfo when we don't know the default app being used to open the url.

lie112:
Shall I create a new issue for each problem I am facing or happy to address all related here?

hol430:
Probably put them all here, I expect there will be quite a few problems :). By UseShellExecute, I mean setting `Process.StartInfo.UseShellExecute = true;`

lie112:
Yep. Once I found out how to get the ProcessStartInfo for a url, changing UseShellExecute to true fixed the problem.
IGNORE fix above.

lie112:
[...]

PR #6147, opened 2021-02-12
Improved manager editor appearance in gtk3 builds

hol430:
Working on #6139 and #5344


hol430:
@hol353 this is actually green even through it appears yellow in the pull requests page.

Issue #158, opened 2015-06-19
New Chickpea model


hut104:
The APSIM Chickpea model has been developed from the original APSIM Classic Chickpea model described by Robertson et al (2002) using the Plant Modelling Framework of Brown et al (2014).

Functionally, it remains a HI model that partitions N to grain (via target N concentration) and root nodules, simulating N fixation. Efforts to model phenology intend to represent field observable characteristics such as buds, that are in turn precursors to flowers.

The work is mostly being undertaken by Allan Peake and should be presented for review in early 2021.

hut104:
Happy for this, but the work is mostly being undertaken by Allan Peake, who may or may not be on GitHub.

sarahcleary:
@hut104 - just following up the concept note.  For background, please refer to https://github.com/APSIMInitiative/ReferencePanel/issues/57#issue-595681585

PR #6148, opened 2021-02-12
Chickpea Validation Cleanup

hut104:
working on #158


hut104:
@hol353 the stats are only different because I added LAI to the PO Table

Issue #6151, opened 2021-02-12
Limited transpiration rate module


BrianCollinss:
Details in: Limiting transpiration rate in high evaporative demand conditions to improve Australian wheat productivity 
https://doi.org/10.1093/insilicoplants/diab006

hol353:
@APSIMInitiative/cereals @APSIMInitiative/legumes @APSIMInitiative/other-crops

hol353:
@kchenu This has now been merged into the release. It is now an option that can be turned on if the user is interested. We have created an example simulation where this demonstrates how to turn it on:  Apsimx/Prototypes/LimitedTranspirationRateWheat/LimitedTranspirationRateWheat.apsimx

hol430:
I guess that would depend on your definition of "in the release" since the prototypes aren't distributed with the official builds.

PR #6152, opened 2021-02-12
Testing limited transpiration rate module

BrianCollinss:
Working on #6151


BrianCollinss:
@hol430 I can successfully run **WhiteClover.apsimx** on my laptop. Could you please help me get this run?

hol430:
Try running wheat test set, it's the one with problems:

```
System.Exception: ERROR in file: C:\Jenkins\workspace\PullRequest_v2\ApsimX\Tests\Validation\Wheat\Wheat.apsimx

Simulation name: YanYeanTOS4CvBolac

System.Exception: Error in report DailyReport: Invalid report variables found:

Wheat.Phenology.PTQ: Error while locating variable '[Leaf].Photosynthesis.FT' in variable reference '.YanYeanTOS4CvBolac.Field.Wheat.Phenology.PTQ.Radn_mol_per_m2.RUE Temp factor'
```

BrianCollinss:
@hol430 Thanks mate. Could you please help me with this Unit test as well? Does it have to with TEC?

BTW, can we still have performance statistics if the unit test fails?

hol430:
Ok yeah, looks like the unit test is failing because it's trying to use the edit file feature to change `[Wheat].Leaf.Photosynthesis.RUE.FixedValue`, which is not working (in that sense, it's not really a proper "unit" test). Anyway, I'm guessing this is not working because you made some changes to do with the rue function in wheat? I guess the best way to fix this would be for me to change the way this test works so that it doesn't assume a particular implementation of rue in wheat. That being said, the croptimizr example also depends upon this behaviour but won't fail if it doesn't work this way (I think). So we may need to change that as well.

We can have performance tests if the unit test fails, but we can't have performance tests if the validation tests failed. In this case it's failing on nutrient:

```
System.Exception: ERROR in file: C:\Jenkins\workspace\PullRequest_v2\ApsimX\Tests\Validation\Nutrient\Nutrient.apsimx
Simulation name: PendletonTreatment0_nBN0
System.InvalidOperationException: Sequence contains no elements
```

BrianCollinss:
Instead of all this hassle, do you think I can temporarily create a folder, WheatTE, in Validation and implement these changes in the **Wheat.apsimx** inside that folder to just get the stats for modified wheat?

hol430:
Well there wouldn't be any old/accepted values for WheatTE.apsimx. You would also still need to look into the problem with the nutrient test set, as the PO stats won't be generated unless all of the validation files run without error. (You can ignore the problem with the unit tests for now, it won't stop the stats from being generated).

hol353:
This message is from @kchenu

Thanks a lot for this. The model gave better results when working from the Canopy model (coded by Bangyou) and Enli’s photosynthesis model – both the canopy and transpiration models are the results of a GRDC project. Now, when Brian tried the transpiration model on its own directly with the release version (using hourly instead of daily and changing the order of how the different processes are run), it doesn’t improve the main outputs - this may due to a lack of proper calibration, but it still has some value in terms of the science we are looking at. Anyway, putting it as an optional module (as was done for the canopy model) may be the way to go. I’ll wait for your assessment. Happy to discuss.

Karine

jbrider:
@BrianCollinss I can't see what has been changed in the wheat json file as it comes up as all being changed - do you just replace one of the existing modules with the Limited Transpiration or does it change more than that?
@hol353 It's showing me that every line in wheat.json file was changed - is that what you are seeing too? Is that because it handle json files differently?

BrianCollinss:
I put the new Photosynthesis module in Wheat.json and also added the module to SupplyFunctions. The Wheat.json is not supposed to be changed, eventually as this module is going to be an option, not replace the RUE module. B
-------- Original message --------From: Jason <notifications@github.com> Date: 19/2/21  09:52  (GMT+10:00) To: APSIMInitiative/ApsimX <ApsimX@noreply.github.com> Cc: Brian Collins <brcollins2020@gmail.com>, Mention <mention@noreply.github.com> Subject: Re: [APSIMInitiative/ApsimX] Testing limited transpiration rate module (#6152) 
@BrianCollinss I can't see what has been changed in the wheat json file as it comes up as all being changed - do you just replace one of the existing modules with the Limited Transpiration or does it change more than that?
@hol353 It's showing me that every line in wheat.json file was changed - is that what you are seeing too? Is that because it handle json files differently?

hol430:
Closing this as it's not going to be merged, as it's causing failures in unit tests and it makes the wheat stats significantly worse. I've created a new PR (#6182) which incorporates these changes in a single .apsimx file under the Prototypes directory, without changing Wheat.json.

PR #6162, opened 2021-02-16
Many fixes for gtk3/netcore builds

hol430:
Working on #5344

@hol353 - this fixes all of the problems you've mentioned so far in #5344 - so feel free to keep posting anything else you've found.


hol430:
retest this please jenkins

lie112:
There are three other instances of the ResourceType in other Transmutation classes (poorly included in Transmutation.cs) for CLEM that need the NonSerialized() attribute. I assume [field: NonSerialized()] is the preferred approach to use. I think I have some [JsonIgnore] attributes performing the same task. I'll convert these soon.

hol430:
Ah ok, thanks for the tip, yeah I'll need to add the NonSerialized attributes to the other classes as well. Just to double check - these are basically holding state and are set during the simulation runs, and they aren't inputs from the user?

lie112:
That's right. I have applied a NonSerialized option to most whenever I got the serialisation error. They will never have a Description attribute and therefore are not user entries, but public properties of the class/model.

hol430:
retest this please jenkins

PR #6166, opened 2021-02-17
Sorghum/micro

jbrider:
working on #572


jbrider:
retest this please jenkins

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -40,6 +40,139 @@ namespace Models.PMF.Organs
...
+        [Serializable]
+        public class LeafParameters : Model
+        {
+            /// <summary>
+            /// Minimum Number of Leaves
+            /// </summary>
+            [Description("Minimum Number of Leaves")]
+            [Link(Type = LinkType.Child, ByName = true)]
+            public IFunction MinLeafNo = null;
+
```
hol353, line 55: Links should be private and with a lowercase first variable

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -40,6 +40,139 @@ namespace Models.PMF.Organs
...
+            public IFunction NewLeafSLN = null;
+
+            /// <summary>Input for SenescedLeafSLN.</summary>
+            [Link(Type = LinkType.Child, ByName = true)]
+            public IFunction SenescedLeafSLN = null;
+
+            ///// <summary>Input for TargetSLN</summary>
+            //[Link(Type = LinkType.Child, ByName = true)]
+            //public IFunction TargetSLN = null;
+
```
hol353, line 115: Remove all commented code.

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -40,6 +40,139 @@ namespace Models.PMF.Organs
...
+            /// <summary> Leaf Initiation Rate </summary>
+            [Link(Type = LinkType.Child, ByName = true)]
+            public IFunction LeafInitRate = null;
+
+        }
+
+        /// <summary>The leaf cohort parameters</summary>
+        [Link] LeafParameters Parameters = new LeafParameters();
 
         #region Canopy interface
```
hol353, line 178: Do you really need #region statements? I don't like them.

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -110,6 +243,12 @@ public double PotentialEP
         /// </summary>
         public const double smm2sm = 1.0 / 1000000.0;      //! conversion factor of mm^2 to m^2
```
hol353, line 245: smm2sm should be private.

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -110,6 +243,12 @@ public double PotentialEP
...
 
+        /// <summary>
+        /// Flag to test if Microclimate is present
+        /// </summary>
+        [Description("Use MicroClimate")]
+        public bool UseMicroClimate { get; set; } = false;
+
```
hol353, line 251: I think the default for UseMicroClimate should be true.

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -40,6 +40,139 @@ namespace Models.PMF.Organs
     public class SorghumLeaf : Model, IHasWaterDemand, IOrgan, IArbitration, ICustomDocumentation, IOrganDamage, ICanopy
     {
         //IHasWaterDemand, removing to see if it's necessary
+        /// <summary>
+        /// Collated Parameters
+        /// </summary>
+        [Serializable]
+        public class LeafParameters : Model
+        {
```
hol353, line 48: Why have a parameters class? Why not just put all parameters straight into the SorghumLeaf class? I know the regular Leaf has a parameters class as well - which I also don't like.

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -360,45 +453,58 @@ public double CoverTotal
         #region States and variables
 
         /// <summary>Gets or sets the k dead.</summary>
+        [Description("Extinction Coefficient (Dead)")]
         public double KDead { get; set; }                  // Extinction Coefficient (Dead)
+
+        /// <summary>The maximum cover</summary>
+        [Description("Max cover")]
+        [Units("max units")]
+        public double MaxCover = 1.0;
```
hol353, line 462: Don't have public fields. Convert them to public properties e.g.
public double MaxCover { get; set; } = 1.0;

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -360,45 +453,58 @@ public double CoverTotal
...
         [Description("This is the intercepted radiation value that is passed to the RUE class to calculate DM supply")]
         public double RadiationIntercepted
         {
             get
             {
-                return CoverGreen * MetData.Radn;
+                if (UseMicroClimate && LightProfile != null)
+                {
+                    double TotalRadn = 0;
+                    for (int i = 0; i < LightProfile.Length; i++)
```
hol353, line 491: Local variables should start with lower case.

Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -623,21 +719,24 @@ public double SenRadnCrit
...
+                return Parameters.SenRadnCrit.Value();
             }
         }
 
         /// <summary>sen_light_time_const.</summary>
-        public double SenLightTimeConst { get { return senLightTimeConst.Value(); } }
+        public double SenLightTimeConst { get { return Parameters.SenLightTimeConst.Value(); } }
+
+        /// Temperature threshold for leaf death, when plant is between floral init and flowering.
+        public double FrostKill { get { return Parameters.FrostKill.Value(); } }
```
hol353, line 731: Keep all class properties together and before any methods.

hol353 (commented):
This PR needs some tidying up to meet coding standards.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: I can remove them - I'm not sure why you dislike them so much though - I find them useful at times.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: Disagree. The sorghum model was not calibrated using Microclimate, preferring to use a different method to calculate radiation and waterdemand.  Considerable effort has been made to enable users to choose to use Microclimate if they wish - particularly for intercropping situations, without having large effects on the results but we're not comfortable making that the default.  Happy to talk about better ways to let the user choose this functionality, particularly as we want to enable DCaPS to perform this functionality in certain situations as well.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: The Leaf component has a long list of defaults that have been added to allow people to potentially extend their implementation rather than having them as hidden constants. It makes the tree harder to read so I would prefer they were available but not immediately visible. The best way to avoid that would be to have optional IFunctions which we also don't like.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: will do.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: will do.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: will do.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: will do.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: will do.

Models/PMF/Organs/SorghumLeaf.cs
jbrider: will do.

Models/PMF/Organs/SorghumLeaf.cs
hol353: Ok. I don't feel strongly about this one.

Models/PMF/Organs/SorghumLeaf.cs
hol353: This will need discussion at the RP. I'm trying to have all models behave the same i.e. they all (by default) can be intercropped, can be eaten, will respond to CO2 etc). APSIM Classic is terrible like this. I got sick of responding to user queries as to why model 'abc' can't do 'xyz' when model 'ghi' can. Trying to avoid this in Next Gen.

jbrider:
retest this please jenkins

hol430:
retest this please jenkins

PR #6173, opened 2021-02-18
Use BruTile to obtain basemaps

### LINKED ISSUE #6171: Mapping should be improved
The current map only works at a world scale, has outdated national boundaries, and indicates points a bit inaccurately. It should be improved.

zur003:
Resolves #6171


zur003:
It looks like I don't quite have this working properly when generating auto-documentation - it's not waiting for BruTile to fetch all the map tiles (although I think it works if the tiles are already in the cache).

hol430:
Nice work Eric, I've just tested this out locally and it looks much better than the current map implementation. The only problem I see is, as you mentioned in the comments, the relatively long wait while fetching the maps from the openstreetmap API. I haven't explored the SharpMap API too much, but I wonder if it's possible to use a TileAsyncLayer to load the image asynchronously, and stitch together the bitmaps for each tile as they are loaded. I can't tell if this is supposed to be handled internally or not, but there is a MapNewTileAvailable event which seems to return small chunks of the overall image as they are loaded, so we could potentially do it ourselves if need be. It also has a NumPendingDownloads property so perhaps that could be used for autodocs purposes, so that we know when it's finished loading the image?

zur003:
The TileAsyncLayer approach should be possible, but my first try at using it didn't work very well. I was trying to update the display as each new tile arrived, but ended up with problems of recursion and thread conflicts. I did get things partly working, but not well enough to commit. I decided to leave that as something for the future. 

Also for the future, I'd like to let the user have the ability to choose the basemap, which could be OpenStreetMap, Bing, ESRI or whatever.

PR #6174, opened 2021-02-18
Grape vine development

HamishBrownPFR:
working on #1822


Models/MicroClimate/MicroClimate.cs
```diff
@@ -298,15 +298,15 @@ private void CalculateStripZoneShortWaveRadiation()
...
 
                 if ((c.Canopy.Height - c.Canopy.Depth) > 0 && c.Canopy.Width <= Wt)
-                    TallestIsVine = true;
+                     TallestIsVine = true;
             }
             if (TallestIsTree)
                 DoTreeRowCropShortWaveRadiation(ref tallest, ref shortest);
-            if (TallestIsVine)
+            else if (TallestIsVine)
                 DoVineStripShortWaveRadiation(ref tallest, ref shortest);
```
hol353, line 310: This highlights that we need to refactor MicroClimate at some point. I don't like if statements that have specific sections for whether a tree is the tallest or the vine is the tallest. There has to be a better way. A job for the future.

hol353:
Is this really grape vine? There is an orchid apsimx file in this PR.

junqi108:
> Is this really grape vine? There is an orchid apsimx file in this PR.

yeah, the grapevine model needs this urgently, as the error makes the transpiration for the leaf canopy too low.

HamishBrownPFR:
The Orchard model has a "GrapeVine" cultivar in it and I have fixed some things in there to make sure both Grapevine an the SimpleFruitTree "GrapeVine" cultivar are working well.

Models/MicroClimate/MicroClimate.cs
HamishBrownPFR: Yeah, we are aware of this one.   It should be possible to harmonize the three methods code so a single method works well for all three situations.  Currently the StripCrop model works for crops where there is no horizontal overlap, the TreeRowCrop works for situations where there is horizontal overlap and the VineRowCrop works for situations where the tall crop does not fully cover its own row.   If not we could have some component we drop into the simulation to select which strip crop model is used.

PR #6176, opened 2021-02-18
Consolidate IFunctions under sorghum grain organ into a single object

hol430:
Working on #572


jbrider:
@hol430 That will work much better - looks really good.
It's probably easier to merge this one first, and I will merge the other sorghum one early next week.

jbrider:
@hol353 This is right to merge.

hut104:
@jbrider I note that the GrainNumberFunction is very specific to the Sorghum model. Parameters are hard-wired within the code, as are the names of plant growth stages, and paths to other variables within the plant model.  It is unlikely that this code would be appropriate for other crops.
Is it possible to adapt this code to allow it to be more readily used in other crops.  I can see this approach being useful for other crop models.  One of the design principles of PMF is to make classes usable across crop models.

jbrider:
@hut104 Fair comment - we're having a bit of trouble finding the balance between readability and extendibility. I'll have a look at it keeping in mind what you mentioned.

jbrider:
retest this please jenkins

jbrider:
Getting an **_apsimdev.apsim.info sent an invalid response_** error when trying to find out why the build failed.

jbrider:
The changes can be accepted if you're happy with the code changes.
@hut104 @hol353

PR #6188, opened 2021-02-22
Fixed paths to met files in whiteclover validation file

### LINKED ISSUE #6154: Paths to Met and Observation files in WhiteClover.apsimx are relative

hol430:
Resolves #6154


hol353:
Why is this PR necessary? If no path is specified to the met files then it should assume the same directory as the .apsimx file.

hol430:
With this change, you can open the file and save it somewhere else and run it without having to change the .met file paths (similar to the example files). If this isn't the approach we want to take, then feel free to close the PR and #6154.

PR #6191, opened 2021-02-23
Implemented DirectedGraphView.Export() in gtk3 builds.

hol430:
Working on #6139


hol430:
@hol353 I leave it up to you if we want this. Ultimately this will be obsolete when we rework how the autodocs get generated, but for now it will prevent the 'create documentation' feature in the UI from throwing when run under .net core.

PR #6193, opened 2021-02-23
Put soybean into release

### LINKED ISSUE #2731: New Soybean Model

hol353:
Resolves #2731


hol353:
@hol430 Why is this failing? The Jenkins Example project is throwing an R error and the Validation project is looking for:

Tests\UnderReview\Soybean\ObservedFACTS.7z

This PR moves this from the UnderReview directory to the Validation directory. I can't find where this is being referenced.

hol430:
This is part of the headache caused by having datasets which are restricted-access. You'll need to edit ApsimX/Jenkins/runTests.bat, and change the part which extracts ObservedFACTS.7z. You should just need to fix the soybean path on line 65.

Issue #6129, opened 2021-02-04
APSIM Next Gen Infrastructure Redesign


hol353:
After several days trying to implement changes to the APSIM infrastructure and encountering many strange issues and coding design patterns, I think its time to plan for a infrastructure redesign. I propose we incrementally move to this rather than down tools and do a rewrite. I'm posting this here as a way to capture my thoughts but equally importantly to get ideas and suggestions from others. Please feel free to comment.

## Need better separation between GUI and models.

* I'm not sure GUI should talk directly to Models.
* There is a lot of code in the GUI to do non-GUI things. Have a look in ContextMenu (e.g. ExportDataStoreToEXCEL) or ExplorerPresenter (e.g. IsValidName) or ProfileGridPresnter (lots of code using reflection to discover properties etc.) 
*  GUI should talk to an API that has methods for each GUI function.
* This API should be unit tested.
* If the API was in another assembly, then we could remove the Models assembly reference from APSIMNG project - complete separation of user interface from the models.
* The API would need to have methods to get and set a property of a model, add a model, delete a model, enable a model, rename a model, get intellisense items, .... This would easily allow the detection of when simulation is dirty and speedup closing of a file in GUI #5536 
* The API method to return intellisense items should call SimulationGeneration.ToSimulation() to get a simulation and use reflection on that.
* The API could be callable from other applications e.g. command line or something else.
* Name of assembly: **APSIM.Services**?

## User Interface changes

* Can we remove all View classes (except the widget wrapper classes e.g. ButtonView) and replace with a Glade file and a presenter (e.g. AddModelPresenter and AddModelView.glade)?
* Allow multi-select in the tree view #2763 
* Name of assembly: **APSIM.UserInterface**?

## Create an APSIM kernel (assembly) that provides APSIM infrastructure core services but contains no models.

* This API should be unit tested.
* Would contain what is in the 'Core' folder of Models assembly.
* Separate assembly would allow us to use 'internal' on classes, properties, methods to hide detail from other assemblies.
* Tighten up when OnCreated() gets called in models. It is called in many places currently.
* Rename SimulationDescription class to SimulationGenerator. Currently it generates simulations but also runs them. Rework the ToSimulation method to build a simulation by applying replacements in the correct order. Should just be a few lines of code. It should do model replacements before property replacements. 
[...]

PR #6214, opened 2021-03-02
Added post simulation tool to send a datastore table to SQLServer

### LINKED ISSUE #6213: Add a post simulation tool to send a datastore table to SQLServer
This is useful when running APSIM on compute clouds where each nodes writes the important data to a central database.

The tool needs to allow the user to specify one or more DataStore table names to send to the database.

hol353:
Resolves #6213


Models/Main.cs
```diff
@@ -77,6 +77,11 @@ public static void Run(Options options)
                 else if (options.ListSimulationNames)
                     foreach (string file in files)
                         ListSimulationNames(file, options.SimulationNameRegex);
+                else if (options.ListReferencedFileNames)
```
hol430, line 80: This is related to #6129 -  but I wonder if it's worth (in the long run) having a separate assembly for interaction with .apsimx files - e.g. editing them, querying them etc. This addition doesn't really seem any more egregious than the existing option to list simulation names but it does seem odd having them sit next to all of these options for running simulations - they really serve a different purpose.

Models/PostSimulationTools/WriteTableToSqlServer.cs
```diff
@@ -0,0 +1,209 @@
...
+    [ValidParent(ParentType=typeof(DataStore))]
+    [Serializable]
+    public class WriteTableToSqlServer : Model, IPostSimulationTool
+    {
+        [Link]
+        private IDataStore dataStore = null;
+
+        /// <summary>Connection string.</summary>
+        [Description("Connection string to connect to database.")]
+        public string ConnectionString { get; set; }
```
hol430, line 29: Is this really something we want to store in the .apsimx file?

Models/Main.cs
hol353: Yep I agree. I suspect we're going to have lots of command line options in the future. Agree with a separate assembly for this stuff. In https://github.com/APSIMInitiative/ApsimX/issues/6129 I was suggesting the same thing. We should probably have a model services assembly for all 'APSIM logic'. In there we could have a class for editing, querying of .apsimx files. 

ListSimulationNames and ListReferencedFileNames are also very slow to run for large files like the wheat validation. All I really need for my cloud work is to have the CLI return the number of simulations in the .apsimx file and then the ability to tell the Models CLI to run simulations 10 through 20 i.e. run a subset. This should be quicker than trying to work out simulation names.

Models/PostSimulationTools/WriteTableToSqlServer.cs
hol353: No we don't. This is a similar to our vault discussion the other day. Where do we want to store our secrets?

hol353:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

Issue #6163, opened 2021-02-17
Many of our .apsimx files are not culture-friendly


hol430:
Many of the .apsimx files in this repository fail when run under culture settings that are not en-AU.

lie112:
What aspects of the apsimx file cause the issues? Number, date formatting? If I know I'll have it in my head for further development.

hol430:
Yep, date and number formatting mainly. The problems that I found today have all been with manager scripts which are converting strings to dates, or reading data (including dates) from a separate csv file. The easiest way I've found to test files is to just modify `ApsimNG/Main.cs` and add something like this (this will use a german locale but you could use others):

```csharp
System.Threading.Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo("de-DE");
```

hol430:
Re-opening this issue. Many of our validation files give different results when run under the invariant culture:

- Barley
- Chicory
- Maize
- Plantain
- RedClover
- Wheat
- WhiteClover

In most cases this is due to converting strings to DateTimes without specifying a culture. Please don't do this.

hol430:
Closing as fixed (as far as I can tell).

PR #6215, opened 2021-03-02
Fixed some culture-related bugs in some of the validation files

hol430:
Working on #6163


hol430:
@hol353 I've finished all of the changes I want to make on this branch, so it's ready for merging if you're ok with the changes. See the message in the second commit for details (all of these files had a copy of the defective manager script). This is related to the changed PO stats when running the validation tests in a docker container. It fixes about half of the diffs we're seeing - the rest don't seem to be culture-related.

Issue #6107, opened 2020-08-26
2020 Infrastructure replacement


peter-devoil:
The tri-ennial issue of replacing ageing hardware is again with us in 2020, as machines that perform APSIM's continuous integration tasks reach past their end of life. It's now clear that several viable cloud options are available to the initiative, and comparisons between each hosting organisation's native services and equivalent commercial operations are required before a recommendation can be made. These comparisons are informed by several years usage of each platform, providing a perspective not only of price, but reliability and service quality.

Discussions are expected in the latter half of 2020, involving @hol353, @peter-devoil and interested participants.

hol430:
Current plan is to use jenkins with google compute VMs provisioned on an ad-hoc basis to be used as build agents for jenkins jobs.  Provisioning of the build agents is currently working well. Tasks that need to be completed before we can move over to this system:

- Get unit tests running on linux
- Delete multi process job runner
- Leave release builds on dizzy for now, once autodocs have been refactored (see #6129) and we have removed the gtk dependency we can move this into the main build phase
- Delete ui tests - there aren't many and their dependency on gtk is a deal breaker. Need to rethink how we test the UI
- Create new jenkins instance which supports https(!). Plan is to commission a new linux server and slowly migrate existing web services over. Planned domain name is dev.apsim.info (dev.apsim.info/jenkins for the jenkins instance)

In the longer run:

- Run the jenkins builds inside docker containers
- Have some official apsim docker images. These could be built as part of the jenkins builds

@hol353 - I just realised that we're still using Muck for old apsim builds. If we still care about old apsim then we'll need to move this to be cloud-based

hol430:
The remaining tasks mentioned above are now covered by separate issues.

PR #6219, opened 2021-03-03
Fixed a floating point error in nutrient constrained allocation method

hol430:
This was causing some simulations to be quite sensitive to their initial conditions, and we were also getting different results depending on the runtime and OS on which they were run.

Working on #6107


hol430:
@hol353 this is red because it's changed the stats, as we expected. The changes are all quite minor, and given that this fixes a bug I think it could be accepted.

PR #6231, opened 2021-03-07
N uptake error in AgPasture

rcichota:
Working on #5356

I (and others) have noticed something strange with AgPasture, with it producing less than expected and favouring clover over ryegrass too much. After some checks I found that the issue is with how the model is calculating N available in the soil, it limits way too much the ability of plants to take N up.  This probably emerged when Dean cleaned up the code and defined the uptake routine to be that of DefaultApsim. The original code for that routine was simplified a bit; there was this weird multiplication of no3 times no3ppm, which was removed. However, the code that was correct, if messy. That multiplication does exist, or rather there should be a 'concentrationFactor', which is kno3×no3ppm (basically the greater the concentration the easy is to uptake), this then multiplies no3. I am now fixing that.
I also noticed that the root length density is also wrong, I believe this was my mistake, converting m/m3 to mm/mm3, instead of m/m2 to mm/mm2. This variable and code is not used anywhere, so should have no consequence...

Firstly, I am correcting the length density (and also adding an output that is needed for checking the root distribution).
I also am removing two parameters that were left from when Dean removed the superfluous routines for N uptake.
Finally here I am making public a parameter that is part of the N uptake routine (control the response to water deficiency), which I think ought to be public (PMF has a similar function, and different crops do have different values).


MarkLieffering:
Hi Rogerio - last year I found the growth of clover was way too dominant. I tried to get around this in a crude way by changing the kNO3 to 0.1 (default was 0.02) and the kNH4 to 0.02 (default was 0.01) via a script

It helped a bit by dropping the clover percentage from the default of c. 40% to c.20%

Cheers

Mark

rcichota:
Thanks Mark, that trick would help. After digging a bit in the code (and of PMF) I found out that the value of KNO3 should be multiplied by the concentration of NO3 in the soil; this will give a concentration factor that then dictates the amount available to plants. Without that, AgPasture was (is!) limiting too much the amount available to plants, favouring clover. I have fixed that, clover growth is more limited, but overall the model is producing too much DM, need to find out why...

rcichota:
I have found another error in the model. The initialisation procedure was doubling the initial DM amount in roots.
This was probably due to the somewhat confusing way that the roots are initialised, and related to the reset method. I have cleaned that up, hopefully making a bit easier to understand/maintain the code.
I don't think this will solve the issue that now AgPasture is growing too much...

rcichota:
@hol353 , do you remember if any changes were made to improve growth in the model last year?  I am assuming that when you removed the extra routines for soil N availability/uptake and made DefaultApsim the ... default, the overall growth was reduced (this is what I got when comparing with classic), if no changes were made/needed, then the model was already producing less than in classic and I need to check older versions to find out why...

hol353:
I can't remember what was changed in AgPasture last year in terms of growth. The PR where we changed the default N availablility/uptake to DefaultApsim is #5372. Perhaps @mpandreucci remembers more than I do about what we changed in AgPasture last year.

mpandreucci:
I can help tiding up the validation if you guys want. Just let me know.

mpandreucci:
Sorry, pushed the wrong button

hol353:
retest this please jenkins

hol353:
retest this please jenkins

PR #6234, opened 2021-03-09
Rebuilt PivotTable into ReportPivot

### LINKED ISSUE #6233: CLEM PivotTable is outdated and needs redesigning
The CLEM PivotTable model has not been touched in several years and needs to be completely overhauled.

MikeStower:
A new pivot table model has been built from the ground up, so that it is more robust. It uses existing views and should work nicely with .NET Core. It should also work with reports outside the context of CLEM, if necessary.

The old PivotTable model has been marked as deprecated, but left in place so as not to break existing work.

Resolves #6233


hol430:
Would be good if we could get rid of the deprecated model altogether. I had a quick look at the source code and it looks like it may be possible to do so with a converter function(?). It should be possible to convert from the indices used by the old model to the strings shown in the UI. From there, it should be possible to initialise an instance of the new class which is functionally equivalent to the old model?

Models/CLEM/Reporting/ReportPivot.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// The report data
+        /// </summary>
+        private DataTable report = null;
+
+        /// <summary>
+        /// Tracks the active selection in the value box
+        /// </summary>
+        [Description("The column to aggregate values for")]
+        [Display(Type = DisplayType.DropDown, Values = nameof(GetValueNames))]
+        public string[] Value { get; set; }
```
hol430, line 32: Is this supposed to be an array? I don't think the dropdowns in the UI will let you select multiple items.

Models/CLEM/Reporting/ReportPivot.cs
```diff
@@ -0,0 +1,274 @@
...
+        /// </summary>
+        /// <param name="source">The source data</param>
+        /// <param name="pivot">The table to populate</param>
+        private void AddPivotRows(DataTable source, DataTable pivot)
+        {
+            // Rescale the time over the rows if required
+            var date = source.Columns[Row[0]].DataType == typeof(DateTime);
+
+            var names = source.AsEnumerable()
+                    .Select(r => date ? FormatDate(Convert.ToDateTime(r[Row[0]])) : r[Row[0]].ToString())
```
hol430, line 220: What happens here if the first column isn't a date column (e.g. if the user doesn't report Clock.Today)?

MikeStower: In both cases, we are returning a string representation of the row data. 'date' is a bool that tests the type of the data. If it is true, the string is formatted for dates first. Otherwise, we just convert the value directly to a string. It's a LINQ friendly way of concisely writing

```
if (date)
    return FormatDate(Convert.ToDateTime(r[Row[0]));
else
    return r[Row[0]].ToString();
```

So it will work even if there is no date data.

Models/CLEM/Reporting/ReportPivot.cs
MikeStower: That's me not quite figuring out how the drop down display worked. I thought it had to be an array to populate the display with items, and you simply had to access the first element to get the users selection (which worked, it just looks ugly). I'll do a quick commit to clean all those properties up.

hol430:
By using a converter function you'd be able to automatically remove the old pivot table and add in an equivalent ReportPivot, allowing us to delete the defunct class without causing type load exceptions, and without the user having to manually patch up their file. If we don't do a converter function, we'll probably never be brave enough to delete the PivotTable class.

There should be some examples in the existing converter functions similar to what we're trying to do here...if you look in Converter.cs, UpgradeToVersion105 does something similar, adding in variable references under all LeafAppearancePhase models.

lie112:
The old PivotTable only needs to stay for a month or so while users get used to, and migrate to, the new ReportPivot. I don't think it's worth writing a converter for the current userbase. I will give them a months warning at one of our CLEMinars (and maybe a warning before the final change happens in the what's new section of the website) and then we'll delete the old model. Deleting it now means that the simulation file will just lose all their current info on pivots when loaded leading to confusion and difficulty migrating.

PR #6264, opened 2021-03-17
Enable manager script debugging under .Net Core

zur003:
Working on #6139


hol430:
As you mentioned in the issue, it would be nice if this could be a user option, ideally on a per-manager basis. It's easy to add a property to the manager itself, but the question would be how do we expose the option to the user? I don't think that properties of class Manager are shown in the properties tab of the manager UI, although perhaps they should be. I don't think the manager class currently has any properties with a `[Description]` attribute, so we haven't run into this issue before. Perhaps the best way forward is just to merge this for now and log a feature request to make this a user-controllable option.

hol353:
Perhaps the simplest is to add an option to debug manager scripts in the GUI's options screen.

zur003:
It looks like we could just use the System.Diagnostics.Debugger.IsAttached property to see whether we're running under the IDE debugger, and only generate debugging information when that is the case.

For what it's worth, my initial analysis about needing to write external files wasn't quite correct. The .pdb can be a stream rather than a file, provided the stream is specified when the assembly is loaded. Also, the source can be embedded (somewhere) and, if needed, gets extracted to a temporary file automatically by the debugger. This all works without having to worry too much about cleaning up temporary files.

PR #6266, opened 2021-03-17
N Uptake Underflow

### LINKED ISSUE #6265: N uptake equations can provide NAN values
We have a simulation during which uptake is estimated to be NAN, and thereafter all the nutrient model state variables also become NAN.  It was caused by an underflow problem. The code should be fixed to avoid this.

hut104:
resolves #6265

### DIFF
--- a/Models/PMF/Arbitrator/Methods/NitrogenUptakeMethod.cs
+++ b/Models/PMF/Arbitrator/Methods/NitrogenUptakeMethod.cs
@@ -63,6 +63,7 @@ namespace Models.PMF.Arbitrator
             }
 
             double NDemand = (N.TotalPlantDemand - N.TotalReallocation) / kgha2gsm * zone.Area; //NOTE: This is in kg, not kg/ha, to arbitrate N demands for spatial simulations.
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@hut104 I always resisted putting these limit overwrites in because the equation above should not return zeros and if it does there is an error further up the line that should be fixed.  Problem is there could now be an error giving zero uptake values but it is not apparent in the code. Perhaps raising an exception would be a better option?

hut104:
In this case it was simply -ve due to rounding error in the 19th decimal place (they were essentially the same).

Do we have mass balance checks in the arbitrator that look for the sorts of errors you mention? If so, perhaps it’s best to check there for real errors, and perhaps the check for rounding should be moved to where the reallocation is calculated.  Floating point arithmetic is a pain sometimes.

HamishBrownPFR:
There are mass balance checks in the arbitrator but they have a small tolerance to account for floating point errors.  Can't remember off the top of my head but I am pretty sure it is not 19dp so this one would have passed the arbitrator test.

PR #6288, opened 2021-03-22
Meet OpenStreetMap usage requirements.

zur003:
This entails providing a UserAgent, and caching downloaded files.
Working on #6171

### DIFF
--- a/ApsimNG/ApsimNG.csproj
+++ b/ApsimNG/ApsimNG.csproj
@@ -54,8 +54,8 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.7.0-1.final" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Features" Version="3.7.0-1.final" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.7.0-1.final" />
-    <PackageReference Include="SharpMap" Version="1.2.0-pre.185420022" />
-    <PackageReference Include="SharpMap.Layers.BruTile" Version="1.2.0-pre.185420022" />
+    <PackageReference Include="SharpMap" Version="1.2.0" />
+    <PackageReference Include="SharpMap.Layers.BruTile" Version="1.2.0" />
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="5.0.0-preview.1.20120.5" />
     <PackageReference Include="System.Resources.Extensions" Version="5.0.0-preview.1.20120.5" />
     <PackageReference Include="Webview.Core" Version="0.0.3" />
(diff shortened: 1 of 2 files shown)


zur003:
Jenkins seems to be having problems with Apsim.PerformanceTests.Service. Can this be fixed, please?

hol430:
retest this please jenkins

zur003:
Still to do - add attribution to OpenStreetMaps.org somewhere....

Retest this please Jenkins

hol430:
retest this please jenkins

zur003:
@hol430, @hol353 - Looking for a bit of advice here. If we continue to use OpenStreetMap maps from their servers, we should abide by the usage policy given at [https://operations.osmfoundation.org/policies/tiles/](https://operations.osmfoundation.org/policies/tiles/). They're starting to enforce this stuff, so I found I had to add an HTTP User-Agent attribute to be able to let APSIM download tiles from them. We also should display license attribution somewhere, but I don't really know where the best place would be. Any thoughts? Sometimes attribution is overlaid on the map image, and I think with Gtk3, this could be done using a GtkOverlay approach, but that isn't part of Gtk2. I suppose it could go into the "dialog" above the map, where the scale and map centre values are displayed.

This also raises the question of where we should be obtaining our base map data. SharpMap/Brutile has support for Bing, ESRI and a few others as well as for OSM, but with all there is a question of the right to use these maps. Obviously the providers don't want their servers to be heavily over-loaded, so even though OSM's data itself is "free", use of their servers is not entirely so. Do you have a preference as to which map data and sources we use? I suppose there is even the possibility of setting up our own server, using the OSM data. What do you think?

hol353:
> I don't really know where the best place would be. Any thoughts?

Perhaps the simplest would be to create a new 'acknowledgements' page accessible from a button on the 'Home' screen of the GUI. We could then acknowledge all the open source software we rely on from nuget with a description of what we use it for and a link to their home page. I've seen this sort of thing in other open source products.

> Obviously the providers don't want their servers to be heavily over-loaded

This might be a problem as we hit their servers once for each auto-doc created every time we merge a PR that closes an issue. I don't have a preference for which map data source we use. I'm not in favour of setting up our own server though - too much work.

hol430:
From their terms of use page:

> Heavy use (e.g. distributing an app that uses tiles from openstreetmap.org) is forbidden without prior permission from the Operations Working Group.

This covers our usage, right?

zur003:
@hol353 - the auto-docs shouldn't be a problem if the tiles are cached on the build machine. This pull request implements file caching.

@hol430 - yes, I suppose we should go through their Operations Working Group, even though we're a relatively light-weight set of users. How many people are using Apsim Next Generation? And how many of them use the mapping facility at all?

hol353:
> How many people are using Apsim Next Generation? And how many of them use the mapping facility at all?

Probably less than 50 people use Next Generation and almost none of those will click on the map component in the GUI. Probably not worth bothering the Operations Working Group?

hut104:
Funny thing.  In a demonstration today for APSIM NG, the first thing demonstrated was the world map showing where the simulations came from...
We've created a monster.

PR #6275, opened 2021-03-18
Fixed compiler warnings

hol430:
Working on #6139

@lie112 just fyi, I've touched a few clem files, to fix some compiler warnings - mainly I've just removed the odd unused variable.

PR #6293, opened 2021-03-23
Fixed some compiler warnings which are now errors.

hol430:
Working on #6139

### DIFF
--- a/Models/CLEM/Common.cs
+++ b/Models/CLEM/Common.cs
@@ -560,6 +560,9 @@ namespace Models.CLEM
     }
 
 
+    /// <summary>
+    /// 
+    /// </summary>
     public enum TransactionStyle
     {
         /// <summary>
--- a/Models/CLEM/Reporting/ResourceTransaction.cs
+++ b/Models/CLEM/Reporting/ResourceTransaction.cs
@@ -54,8 +54,14 @@ namespace Models.CLEM
             }
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
         public TransactionStyle Style { get; set; }
 
+        /// <summary>
+        /// 
+        /// </summary>
         public double Amount { get; set; }
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


hol430:
@hol353 - in #6275 I changed the project files so that the compiler treats warnings as errors. As part of this, I "fixed" all existing warnings. Unfortunately, some other pull requests were merged after #6275 ran on jenkins but before #6275 was merged. These other pull requests introduced some extra compiler warnings which I've fixed in this pull request. All builds will therefore fail until this is merged.

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
Duplicate of #

hol430:
retest this please jenkins

hol430:
retest this please jenkins

lie112:
I'll update the CLEM summary details when we have a full compiling version.
Does this include all warnings, or just selected ones that will stop the the build?
I have noticed in the past lots of warnings for other components I come across.

hol430:
This includes all warnings except CS1591 and CS1572 in ApsimNG, and several others in the unit tests (you can search for <NoWarn> element in the csproj files). These particular warnings will be ignored and will not show up even as warnings. Everything else will be treated as an error, rather than as a warning. Would be good to turn it on for all warnings, but there are hundreds of failures in ApsimNG/UnitTests which I don't have time to fix right now.

lie112:
Great. I appreciate it is very hard, or almost impossible, to make these sweeping decisions. Just like if you wanted to enforced a program style into the mix, everything would blow up until all was fixed and applying changes across entire codebase is very time consuming unless everyone was on the task. I recall most of the warnings I have seen were in Gtk. It would be nice if the .Net Core compile could have it's own rules so we didn't have to worry about the .Net framework in getting the code up to spec.

Has anything changed in the Summary.WriteWarning? I've seen some things that used to be only warnings coming up as red and nasty for users, but maybe it's associated with a real error also being thrown.

hol430:
Don't think the summary code has changed recently. If something is showing up as red (with a more info button) in the GUI, then chances are it's an exception that was thrown during the simulation.

PR #6295, opened 2021-03-23
Allow custom sale flag to be set when marking ruminants in CLEM

### LINKED ISSUE #6294: Allow user to define SaleFlag when Marking individuals for sale
At present CLEM uses the MarkedForSale flag when using the Mark for Sale Activity, but the user should be able to select from a range of reasons.

lie112:
Resolves #6294


hol430:
retest this please jenkins

hol430:
retest this please jenkins

PR #6303, opened 2021-03-25
Fix toggling enabled on resource models in UI.

### LINKED ISSUE #6301: Re-enabling a crop model generates and object reference error
I've tried this with AgPasture models and with PMF (Barley) with the same result. To replicate, pull up an example simulation, disable a crop model, enable the crop model -> "Object reference not set to an instance of an object." See the additional information below. Disabling and reenabling manager scripts does not generate the same error.

   at UserInterface.Views.TreeView.Delete(String nodePath) in C:\aaUnderVC\apsimNextGen\ApsimNG\Views\TreeView.cs:line 284
   at UserInterface.Presenters.ContextMenu.Enabled(Object sender, EventArgs e) in C:\aaUnderVC\apsimNextGen\ApsimNG\Menus\ContextMenu.cs:line 772

hol430:
Resolves #6301

### DIFF
--- a/ApsimNG/Menus/ContextMenu.cs
+++ b/ApsimNG/Menus/ContextMenu.cs
@@ -761,15 +761,21 @@
                     explorerPresenter.CommandHistory.Add(command);
 
                     // Now call OnCreated() for all changed models. Note ToList() to force greedy evaluation.
+                    bool showModelStructure = ShowModelStructureChecked();
                     if (model.Enabled)
                     {
                         model.OnCreated();
                         foreach (IModel descendant in model.FindAllDescendants().ToList())
                             descendant.OnCreated();
                         if (model is ModelCollectionFromResource)
+                        {
                             foreach (IModel child in model.Children)
-                                if (child.IsHidden)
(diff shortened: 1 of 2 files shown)


zur003:
This fixed the original problem, but introduced some other subtle issues related to interactions between enabling/disabling and showing/hiding model structure. For example, start with an enabled PMF model; show the model structure; disable and re-enable the model. Everything looks OK for child nodes, but if you expand sub-models, they are displayed as disabled, although right-clicking on them indicates they are enabled. Now go back to the PMF model, hide its structure, then re-show its structure. Only direct children now appear in the tree, not more distant descendants.

hol430:
Thanks Eric, that is an interesting one. I notice also that clicking undo/redo doesn't seem to update the models in the tree in this case - this will be due to my recent changes to the command code. It's probably possible to hack around this but I think it might just be better to question why we're even calling OnCreated() here from the ContextMenu code in the first place. This appears to be a workaround for #5634, but it might be possible to reconsider how we approach this issue.

PR #6308, opened 2021-03-28
Changed soils in slurp validation to use nutrient

### LINKED ISSUE #6307: Slurp validation uses SoilN
@hol353 - do we want to change these to nutrient? This will probably have an impact on the stats.

hol430:
Resolves #6307


Models/Soils/Nutrients/NFlow.cs
```diff
@@ -57,6 +57,25 @@ public class NFlow : Model
         [Description("Name of destination pool")]
         public string destinationName { get; set; }
 
+        [EventSubscribe("StartOfSimulation")]
+        private void OnCommencing(object sender, EventArgs args)
+        {
```
hol353, line 62: Why is this method necessary? No other simulation in our test suite needs it.

hol430: See the most recent commit message for details. Basically, I was getting a null reference exception on the first day of the simulation. This was occurring because we were accessing certain properties of nutrient on the first day of the simulation before it had a chance to initialise itself - they were being accessed from the DoManagement event in an ExtraVariables manager script. I would hazard a guesss that nothing else in the test set accesses these nutrient properties in the DoManagement event.

hol430:
@hol353 did you accept stats? I'm surprised this is green.

hol353:
Nope I didn't accept the stats. Nutrient should give very similar results to SoilN.

hol430:
@HamishBrownPFR fyi - looks like changing slurp validation to use nutrient didn't cause any change in the stats.

HamishBrownPFR:
Great, thanks

PR #6345, opened 2021-04-14
Don't draw a box around graph annotations

hol430:
Working on #6139


hol430:
@hol353 - this will also remove the box around the regression equations, as it uses the same code. Is this what we want?

PR #6347, opened 2021-04-14
Speed up graph display by improving read from db speed.

### LINKED ISSUE #6313: Graphs are slow to load with many series
Graph loading time seems to increase linearly with the number of series definitions (not series) generated by the graph. The attached file demonstrates the problem. [slowgraph.zip](https://github.com/APSIMInitiative/ApsimX/files/6225769/slowgraph.zip)

hol353:
Resolves #6313


hol430:
@hol353 this has broken graph filters which reference simulation name. Clicking on any such graphs now yields the following error:

```
Internal error: Don't pass SimulationName in a filter to DataStoreReader.GetData. Use the SimulationNames argument instead.
```

Is this intentional?

PR #6351, opened 2021-04-14
Increased icon size in treeview in gtk3 builds

hol430:
Working on #6139


lie112:
Are all treeview icons now 24x24 across all platforms and will this be the new standard or might we move to multiple resolution ico sometime? I might start upgrading some clem treeview icons from 16x16 to 24x24 as they will look a lot better, but only if it's not going to change again in future.

hol430:
The icons are already rendered 24x24 in gtk2 builds. This was achieved via a line in the .gtkrc file which is a gtk2-specific style/theming file. The extra API call I've added here should have no effect in gtk2 builds because we've already made an equivalent change to the icon sizes via the .gtkrc file.

No idea if this will change again in the future. Most of our icons are low quality so it would be nice to upgrade them at some point, at which time we might consider other options for icon sizes.

PR #6372, opened 2021-04-20
Allow annotations on graphs to be positioned by user

### LINKED ISSUE #6125: Can we have an option to change size of regression equation on plots?
An option would be helpful to:

Change size of regression equation
Change size of regression statistics.

hol353:
Resolves #6125


hol430:
retest this please jenkins

PR #6373, opened 2021-04-20
Changed output directory structure

hol430:
Working on #6139 

This changes the output location when building apsim. Previously, it was built to ApsimX/Bin (or ApsimX/NetCoreBin for .net core builds). In this PR I've changed this so that the output location is dependent on the configuration (ie debug vs release) and the target framework version (ie .net core vs .net framework). So after building apsim you will find the output assemblies in places such as `ApsimX/bin/Debug/net472`, `ApsimX/bin/Release/netcoreapp3.1`, etc. When using the installed version, the binaries will all be inside a directory called bin (e.g. `%userprofile%\AppData\Local\Programs\APSIMxxxx.xx.xx.xxxx\bin\Models.exe`, `/usr/local/lib/apsim/xxxx.xx.xx.xxxx/bin/Models.exe`).

@HamishBrownPFR @frank0434 @femiguez - not sure if this will affect your work, but I'm pinging you guys just in case. Some of your scripts may need slight tweaks if you upgrade to a version of apsim containing this change.


byzheng:
@hol430 I use models.exe under HTCondor. In which, I copy models.exe and other dependency dll files into a folder in the condor computers. 

the process in the condor computers can only access one folder (i.e. no user profile folder). Whether does installed version modify any folders in appdata of user profile?

hol430:
The appdata location is the default apsim install location on windows (if you run the installer without administrator privileges). So it' won't matter for running apsim on the cluster. This change might affect the way that you copy Models.exe - you just need to copy it from its new location if/when you pull down this change (probably `ApsimX/bin/Release/net472`).

femiguez:
I'm testing this new version on Mac and it does not seem to work. (I'm not using R at all.  Just testing APSIM.2021.04.19 using the GUI.). Trying to run the "Wheat.apsimx" example does not produce a report table and instead it says "Report does not exist" after running the simulations. I suspect that the issue is that I need to update mono, but not sure if that will solve the problem. I'm currently installing APSIM.2021.04.19 on Windows and it asked me to update .NET Framework, so if this is the issue, then the Mac App should also have a message that asked to update mono. This is what I'm running at the moment: Mono JIT compiler version 5.18.1.3. I'll update later when I try this on Windows. Also, thanks for tagging me. :)

hol430:
Well, this hasn't been merged yet, so any versions out there at the moment will not contain these changes. That being said, I suspect your version of mono may be the problem here. I was doing some experiments with a docker container a few days ago which came with mono 5.18 installed, and I was running into a variety of problems running apsim. Upgrading to a more recent version fixed all of these problems. I am currently using 6.12 on my pc without any issues, although I know @peter-devoil has had some success with 4.8.X, so that may also be an option.

femiguez:
Is it possible that I actually did test these changes? I downloaded APSIM from the website and the version stamp was APSIM.2021.04.19.6139 (the issue you reference above). This version is not available at the moment, because they were not actually merged after testing, but I happen to get the version before the checks. I will most definitely have to update mono when these changes are implemented. Thanks!

hol430:
I don't think you've been testing these changes....the issue #6139 has had dozens of pull requests made against it, so I would guess that your version comes from an older pull request (possibly #6366). Plus the timestamp in your version number (2021.04.19) indicates a version from 19 April, and I didn't open this pull request until the 20th.

Was apsim previously working fine for you with mono 5.18?

hol430:
retest this please jenkins

PR #6376, opened 2021-04-21
Map component should only display dots for enabled weathers

### LINKED ISSUE #6370: Map component draws dots for disabled weather files
The map component adds a dot to the map for each weather file in scope, regardless of whether they are enabled or not. Further, it doesn't handle any parser errors while reading coordinates from the files, so if one file throws, then the map generation will fail.

hol430:
Resolves #6370


hol430:
retest this please jenkins

PR #6377, opened 2021-04-21
Fixed converter function 127 when memo text is not set

### LINKED ISSUE #6357: Error on import
System.Exception: Error during Import: Exception has been thrown by the target of an invocation. Value cannot be null.
Parameter name: input
   at UserInterface.Presenters.MainPresenter.Import(String fileName) in C:\Users\her123\Projects\ApsimX\ApsimNG\Presenters\MainPresenter.cs:line 1119
   at UserInterface.Presenters.MainPresenter.OnImport(Object sender, EventArgs e) in C:\Users\her123\Projects\ApsimX\ApsimNG\Presenters\MainPresenter.cs:line 1092

[test_imports.zip](https://github.com/APSIMInitiative/ApsimX/files/6314567/test_imports.zip)

hol430:
Resolves #6357

### DIFF
--- a/Models/Core/ApsimFile/Converter.cs
+++ b/Models/Core/ApsimFile/Converter.cs
@@ -3348,9 +3348,12 @@
             foreach (JObject memo in JsonUtilities.ChildrenRecursively(root, "Memo"))
             {
                 string text = memo["Text"]?.ToString();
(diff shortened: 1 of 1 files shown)


hol430:
retest this please jenkins

PR #6382, opened 2021-04-22
Implement find/replace in .Net Core editing

zur003:
Working on #6139

### DIFF
--- a/ApsimNG/Utility/FindAndReplaceForm.cs
+++ b/ApsimNG/Utility/FindAndReplaceForm.cs
@@ -8,6 +8,7 @@ using GtkSource;
 #endif
 using Cairo;
 using UserInterface.Views;
+using UserInterface.Extensions;
 
 namespace Utility
 {
@@ -72,11 +73,7 @@ namespace Utility
 
         public void Destroy()
         {
-#if NETCOREAPP
-            window1.Dispose();
-#else
-            window1.Destroy();
-#endif
+            window1.Cleanup();
         }
 
         private void Window1_DeleteEvent(object o, DeleteEventArgs args)
@@ -89,19 +86,14 @@ namespace Utility
         }
 
 #if NETCOREAPP
-        SearchContext context;
-        SearchContext Context
-        {
-            get { return context; }
-            set { context = value; }
-        }
+        private SearchContext context;
 
-        SourceView editor;
-        SourceView Editor
+        private SourceView editor;
+        public SourceView Editor
 #else
 
-        TextEditor editor;
-        TextEditor Editor
+        private TextEditor editor;
+        public TextEditor Editor
 #endif
         {
             get { return editor; }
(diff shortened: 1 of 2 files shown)


ApsimNG/Utility/FindAndReplaceForm.cs
```diff
@@ -69,17 +72,37 @@ private void Window1_Destroyed(object sender, EventArgs e)
 
         public void Destroy()
         {
+#if NETCOREAPP
+            window1.Dispose();
+#else
             window1.Destroy();
+#endif
```
hol430, line 79: This conditional code could be replaced with a call to `window1.Cleanup()`, which is an extension method in the `UserInterface.Extensions` namespace:

```suggestion
            window1.Cleanup();
```

ApsimNG/Utility/FindAndReplaceForm.cs
```diff
@@ -69,17 +72,37 @@ private void Window1_Destroyed(object sender, EventArgs e)
...
+#if NETCOREAPP
+        SearchContext context;
+        SearchContext Context
+        {
+            get { return context; }
+            set { context = value; }
+        }
+
+        SourceView editor;
+        SourceView Editor
```
hol430, line 100: Could we have some explicit accessibility modifiers here? We also probably don't need to have a manual backing field for the Context property.

```suggestion
#if NETCOREAPP
        public SearchContext Context { get; set; }

        private SourceView editor;
        public SourceView Editor
```

ApsimNG/Utility/FindAndReplaceForm.cs
```diff
@@ -88,26 +111,67 @@ TextEditor Editor
...
         /// </summary>
         public void ShowMsg(string message)
         {
             MessageDialog md = new MessageDialog(Editor.Toplevel as Window, DialogFlags.Modal, MessageType.Info, ButtonsType.Ok, message);
             md.Run();
+#if NETCOREAPP
+            md.Dispose();
+#else
             md.Destroy();
+#endif
```
hol430, line 126: This conditional code could be replaced with a call to `md.Cleanup()` (in the `UserInterface.Extensions` namespace).

hol430:
Thanks Eric - looks good. This will be a good addition to the .net core gui. I think it's one of the few features which haven't been brought across yet.

ApsimNG/Utility/FindAndReplaceForm.cs
zur003: Ah, right. I thought there was something like that.

zur003:
I also fixed a bug in which repeated undos would totally clear the editing buffer.

One problem remains - use of ctrl-f and  ctrl-h accelerator keys isn't set up until the first time the popup is populated, so if you start editing and press crtl-f, nothing happens. After you've activated the popup menu once, it works as expected.

PR #6389, opened 2021-04-27
Fixed cultivar names returned by Plant

### LINKED ISSUE #6387: Euc cultivar selection fails dt inclusion of memo
Using the TreeManagement Tool in the EucalyptusRotation example (at least - maybe in other examples too), selecting a cultivar causes the simulation to fail as the assoicated memo is included - see screenshot
image

hol430:
Resolves #6387

### DIFF
--- a/Models/PMF/Cultivar/Cultivar.cs
+++ b/Models/PMF/Cultivar/Cultivar.cs
@@ -38,11 +38,6 @@
         private List<object> oldPropertyValues = new List<object>();
 
         /// <summary>
-        /// Gets or sets a collection of names this cultivar is known as.
-        /// </summary>
-        public string[] Alias { get => FindAllChildren<Alias>().Select(a => a.Name).ToArray(); }
-
-        /// <summary>
         /// Gets or sets a collection of commands that must be executed when applying this cultivar.
         /// </summary>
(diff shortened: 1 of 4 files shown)


hol430:
retest this please jenkins

hol353:
Won't this remove all the alias names from the returned list of cultivar names?

hol430:
Yes. How are the alias names supposed to work? Or rather, what are they supposed to do?

hol353:
Alias are just different names for the same cultivar. The wheat model makes extensive use of these where a cultivar is known by many different names. It will allow the user to sow the cultivar that they know and behind the scenes it is mapped to another name. We need to keep this feature.

hol430:
Ok, I've tweaked the way that this works. The Alias property is no longer referenced by any of our code outside the Cultivar class - I'd like to remove this property altogether, but do we need to allow for manager scripts which reference it?

hol353:
You can remove the alias property. It should be an internal thing only. I don't think any manager scripts will use it (famous last words!)

hol353:
Do you want to get rid of :

```c#
public string[] Alias
```

from the Cultivar class before I merge this?

hol430:
Yep, I'll do that.

PR #6398, opened 2021-04-27
Created a ReportQuery model to replace CustomQuery

### LINKED ISSUE #6397: CLEM CustomQuery is outdated and needs redesigning
The CLEM CustomQuery model has not been touched in several years and needs to be completely overhauled.

MikeStower:
Resolves #6397

As per the PivotTable overhaul, the old model is being left in place for a short grace period to allow users to change over without breaking their simulations.


hol430:
In the future it would probably be better to write a converter function to replace instances of the old model type with the new one. That way we could delete the old code immediately (without breaking people's files).

btw - what is the status on removing the old PivotTable model?

Issue #6410, opened 2021-05-05
Why is N2O loss from nitrification (in Nutrient) set to zero?


sno036:
Kia ora @hut104 - looking at N2O losses from Nutrient, there is a comment that losses from nitrification are set to zero. Was there a reason for this? And if there was a reason can we revisit? 

My thoughts - set the default value to 0.01 (1% of N nitrified, I need to verify but I think that is the generally no-better-information value) and allow users to override (at present it is read-only).  

Your thoughts?

sno036:
PS - there is a constant on the DayCent function that Thorburn et al. identified as likely needing site-specific values. The current is fine as a default but it should be user settable.

hut104:
The flow is set to zero because that is the approach used in SoilNitrogen.
Nutrient was added to reproduce the existing model.  Testing will be required for changes such as this.

Almost all N from the decomposition of N from plant material passes into the NH4 pool first and is nitrified from there to accumulate as NO3 until taken up (generally - though some plant models take up NH4 it is only usually after some period of time after the previous crop and NO3 pools end up larger than NH4).
That being the case, all mineral N starts as NH4.  Adding a 1% loss to this flow may impact on the overall N budget.  The impact of this change will need to be assessed on the complete APSIM test set prior to adoption.  Would require tests for long term simulations and checking of relative contribution of nitrification and denitrification to N2O.


________________________________
From: Val Snow ***@***.***>
Sent: Wednesday, 5 May 2021 10:47 AM
To: APSIMInitiative/ApsimX ***@***.***>
Cc: Huth, Neil (A&F, Toowoomba) ***@***.***>; Assign ***@***.***>
Subject: [APSIMInitiative/ApsimX] Why is N2O loss from nitrification (in Nutrient) set to zero? (#6410)


Kia ora @hut104<https://github.com/hut104> - looking at N2O losses from Nutrient, there is a comment that losses from nitrification are set to zero. Was there a reason for this? And if there was a reason can we revisit?

My thoughts - set the default value to 0.01 (1% of N nitrified, I need to verify but I think that is the generally no-better-information value) and allow users to override (at present it is read-only).

Your thoughts?

sarchontoulis:
@sno036 @hut104 in some datasets from Iowa the 1% N20 from NH4 helped but not in all cases. Another challenge I see is the N20 production by depth in wet soils (with shallow watertables). Something to be considered if changes/tests are to be made.

sno036:
[...]

PR #6418, opened 2021-05-05
Test enabling N2O from nitrification

sno036:
working on #6410 

Modified nutrient.json for 1% loss of nitrified N to N2O to see what the flow on effects would be.

### DIFF
--- a/Models/Resources/Nutrient.json
+++ b/Models/Resources/Nutrient.json
@@ -1441,7 +1441,7 @@
                 },
                 {
                   "$type": "Models.Functions.Constant, Models",
-                  "FixedValue": 0.0,
+                  "FixedValue": 0.01,
                   "Units": null,
                   "Name": "NLoss",
                   "Children": [],
@@ -1451,13 +1451,13 @@
                 },
                 {
                   "$type": "Models.Functions.Constant, Models",
-                  "FixedValue": 0.0,
+                  "FixedValue": 1.0,
                   "Units": null,
                   "Name": "N2OFraction",
                   "Children": [
                     {
                       "$type": "Models.Memo, Models",
-                      "Text": "Currently, no N2O is assumed to be produced during nitrification processes.",
+                      "Text": "Testing to see the carnage if N2O from nitrificaiton is enabled. Change here and in N loss aboe - Currently, no N2O is assumed to be produced during nitrification processes.",
                       "Name": "Memo",
                       "Children": [],
                       "IncludeInDocumentation": true,


sno036:
Scanning the checks, all the changes were minor and there were more improvements (44) than worsenings (27).

The advantage of having the N2O losses turned on is that this is a known, but often minor, contributor to N2O emissions and including this makes APSIM both more compliant with current understanding and in agreement with several other models.

Can this be merged please?

hut104:
Can I suggest that we look at N2O outputs and a few other things prior to merging?

There are currently no data for N2O or NAtm outputs and so nothing on this front will appear in the tests. The thing we are testing does not appear in the tests.

Many of the crop models have minimal N test data (e.g. not even fertiliser trials) and so optimal N rate trials are unlikely to respond to this change.

Perhaps some sensibility tests or something?  For example, the suggestion that the N losses should be minor contributor can be tested.  There should be papers that can tell us something about the expected levels.  Perhaps this could be used.  I have previously asked researchers working in this area to submit data/tests in their possession and they have failed to do so.  This was going to be the next step for testing this part of the model prior to changing things.  In the absence of this perhaps tests on emissions factors or the like.

The effect may also be context dependant.  I think the oil palm model has some annual stats in its reporting in the example file.  That could be used to look for changes in emissions for high input tropical agroforestry systems.  The review by Pardon might be useful to check N2O levels in these systems.  This sort of information will be in the APSIM Oil Palm N2O paper.  Peter thorburn's sugar work or the work on wheat in WA are other diverse environments.

It might be interesting to look at low input systems (Africa or Australia) that rely on mineralisation.  A long term simulation (e.g. 100 years) might show a cumulative impact if we have an added N loss pathway.  ie does N rundown accelerate.  The Brigalow catchment study might a possible candidate.

All this might depend on the initial test on whether we have significantly impacted N losses or not.




sno036:
In Classic this algorithm was one that was optionally turned on and it was used in several publications (see below a non-exhaustive list in which it was used to my knowledge). Given this history, that Nutrient is replicating SoilNitrogen, and that turning it on has led to minimal changes in the test suite I think it is time to turn it on in NextGen rather than sending a lot of time gathering simulations not currently in the test suite - particularly as not all those crops will be available in NextGen or that they might be reliant on SWIM (which is not quite there yet). If future insertions into the test suite reveal an issue then the setting can be always revisited. 

https://doi.org/10.1016/j.agee.2009.12.014 - Thorburn et al.
https://doi.org/10.1016/j.agee.2009.12.016 - Huth et al.
https://doi.org/10.1111/gcb.13965 - Ehrhardt et al.
https://doi.org/10.1029/2020GB006561 - Fuchs et al.
https://doi.org/10.1029/2019JG005261 - Fuchs et al.
https://doi.org/10.1016/j.envsoft.2017.01.005 - Snow et al.

hol353:
Closing this without merging in favour of #6438.

Issue #6422, opened 2021-05-06
Mapping problems


zur003:
I'm encountering two different problems with mapping:

1) If running a debug version from the IDE, the program is no longer able to find the countries.shp file (the path isn't what is used to be), and fails with an exception. Note that use of the counties.shp file is no longer really needed. It's just there as a backup in case the OpenStreetMap layers can't be obtained.

2)The other problem is a bit more odd. Maps are working OK for me with my "custom" ApsimNG builds, but when I use the "installed" version, I get a map with a big red "X" over it, and no data points. Attaching a debugger indicates a problem in ProjNET - an exception indicating that it can't convert between the specified co-ordinate systems. Now it does need to convert between lat/long and the map projected system, but I have no clue as to why this all works as expected with "custom" builds but fails with the "installed" version.

zur003:
Further clarification on the second problem - it seems to occur if I "install for all users", but not if I install for myself only. So it may have something to do with the protections slammed onto contents of the ProgramFiles folder. Still seems rather weird to me, though.

PR #6423, opened 2021-05-06
When mapping, check to be sure the countries.shp file actually exists…

zur003:
… before trying to use it. Working on #6422


hol430:
This is probably caused by the changed output path. `Directory.GetParent(bin).FullName` will never return the path to the ApsimX directory in debug builds. In other places I've done something like `PathUtilities.GetAbsolutePath("%root%", null)` to get the path to the ApsimX directory. That way, we can keep the directory structure logic in one place. Maybe it's even worth having a method called something like `PathUtilities.GetApsimXPath()` or something to that effect which calls GetAbsolutePath() passing in the %root% macro.

PR #6432, opened 2021-05-07
Construct only the two mapping coordinate systems we need, rather tha…

### LINKED ISSUE #6422: Mapping problems

zur003:
…n attempt to pull in all known systems. Resolves #6422

### DIFF
--- a/ApsimNG/Views/MapView.cs
+++ b/ApsimNG/Views/MapView.cs
@@ -66,6 +66,11 @@
         /// with units of decimal degrees. Note that both are based on WGS84, so they have the same underlying
         /// model of the shape of the earth, but use vastly different units. The two transformation objects defined 
         /// below handle coordinate transformation.
+        /// 
+        /// An earlier version of this unit made a call to SharpMap.Converters.WellKnownText.SpatialReference.GetAllReferenceSystems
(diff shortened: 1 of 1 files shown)


zur003:
Just a note of explanation - mapping wasn't working when ApsimNG was installed into Program Files because it was attempting to generate a file there, named "SpatialRefSys.xml". This 3 MByte file (I think it was being downloaded behind the scenes from spatialreference.org) holds descriptions of 3912 different mapping coordinate systems. We only make use of 2 of them, and those two are readily generated "manually". Doing so should make things faster, as well as eliminate the problems of trying to cache a file in the read-only Program Files space, and unnecessarily chewing up 3 MBytes of disk space.

PR #6438, opened 2021-05-09
N2O Sensibility Tests

### LINKED ISSUE #6410: Why is N2O loss from nitrification (in Nutrient) set to zero?

hut104:
resolves #6410


hut104:
@hol353 @sno036 This can be merged.  The differences are small, and only really impact predictions for which stats were already marginal (ie more variance in predictions than data).  I have also checked the autodoc and ensured that new tests are documented for readers.

Issue #6448, opened 2021-05-10
No failure for manager script compilation error


hut104:
The Nutrient sensibility test has a manager script to test stats for N2O losses.  It uses an extension of C# for string handline (ie $). If this is compiled using .NET framework the compilor complains.
However, there is no error when I run everything.  It simply says that the simulations are complete.

I was unsure whether there should have been a failure somewhere given that the script may not have compiled.  I would have thought that script compilation failure should throw a fatal error when run.

hol430:
If you have a manger script with compilation errors inside a simulation, it will cause an exception when you open the file in the GUI, when you run it in the GUI, and when you run the file from the command-line.

If you have a manager script which is not inside a simulation (as is the case in the nutrient test set), you will get an exception when you open the file in the GUI, and when you run the file from the command line, but NOT when you run the file from the GUI.

This all comes back to the manager compilation rules. We compile manager scripts when we:

- open a file (all enabled managers are compiled)
- perform GUI actions involving the manager (edit code, enable it, copy it, etc)
- run a simulation containing the manager

I'm not sure what, if anything we should do about this. It's a bit of an edge case, but should only be a problem if you're in the habit of ignoring the glaring error messages when you open a file in the GUI. So I guess this may be a problem for a lot of users...

hol353:
The script in question compiles correctly. As hol430 above says, if you modify the script you will see compile errors. Likewise, if you open the file you will see compile errors. Closing...

PR #6451, opened 2021-05-11
Remove string interpolation from nutrient test script.

hol430:
As this has once more broken the releases.

Working on #6448


hol430:
retest this please jenkins

Issue #6428, opened 2021-05-07
Many models do not zero themselves correctly at start of simulation


hol430:
Many models fail to initialise themselves correctly at the start of simulation. This is made obvious when running the same simulation object serially without cloning it in-between.

hol430:
Re-opening, as there are more problems that show up with different simulation configurations.

ric394:
This is very old and will be closed for now, unless more models are found with the same issue.

PR #6478, opened 2021-05-20
Fixed some models which don't zero their variables at start of simulation

hol430:
Working on #6428 

@hol353 please don't merge this yet. I'm just pushing this up here to make sure I haven't broken anything with these changes.


lie112:
Hi @hol430, I'm just getting up to speed on latest changes after getting back from holidays
Are these changes only for manager scripts or are they needed for the new server process? Does the server process only use the initial simulation loaded in memory so no need to worry about correct initialisation of things such as counters between experiments?
I have some lazy initialisation in CLEM that I might need to update.

hol430:
These changes will be needed for all models in order for them to work properly with this server feature. The server process will reuse the same simulation object for multiple simulation runs without cloning or otherwise modifying it between runs. So if you've got a counter variable, you'll need to make sure to reset it to its initial value either at the end of a simulation run or at the start.

I've added a test for this in #6529 but it has its limitations - for now it's just checking the example files, so if there's a model that's not in an example then it's not tested. I need to give it a bit more thought before that pull request is merged in.

PR #6497, opened 2021-05-25
Added ReferenceHeight test to MicroClimate

### LINKED ISSUE #6476: Reference Height in MicroClimate
It is possible for people to have Reference Height set to zero in MicroClimate if they have not thought about the value.

If Reference Height is zero, wind profiles become extreme and ET becomes very high (e.g. twice what they should be).

MicroClimate should check that Reference Height has some sensible value.

hol430:
Resolves #6476


hol353:
Why change a bunch of .apsimx files instead of writing a converter?

hol430:
Well if we write a converter then we might be disguising some actual problems with people's microclimate configuration. Setting it to 2m in all cases might not be the best approach. Of course, that would be consistent with how we've fixed similar problems in the past - so I guess I'll just change this PR to use a converter approach.

PR #6499, opened 2021-05-25
Initial go at addressing the layering issue in SWIM

### LINKED ISSUE #6498: The calculation of node depths in SWIM is not tika
Part of the problem was that MathUtilities.Sum(dx,0,0) was not doing what it was thought to have. Fixed that by replacing it with a calculated cumulative depth

Another problem was that the indices were pointing to the wrong layers and always skipped the second layer.

Fundamental problem was the when this was converted it is looking like the person converting this wanted the same number of nodes as there were layers. There are of course many ways to assign nodes if layers already exist (it is easier to go from nodes to layers but APSIM doesn't work that way). In this conversation where the desire seemed to be to have #nodes = #layers, if N is the number of layers, the scheme is configured to completely bypassing layers N-2 and N-1 and place the Nth node at the bottom the the Nth layer. This leaves a thicker (by about 2.5 times if there are uniform layers at the base) than expected internode distance at the bottom of the profile and could well be the cause of the numerical instabilities. Putting several (at least 3) thin (1 mm perhaps) layers at the end of the profile might assist with the instability but this (#nodes = #layers) really isn't standard for numerical solutions.

Are we sure that the hydraulic properties are being got sensibly? Even though #nodes = #layers It should not be assumed that the UI properties for layer j should be assigned to node j.

swim.cs circa L1675
            x[0] = 0.0;
            x[1] = 0.0 + dx[0] / 2.0;
            double cumDepth = 0.0;
            for (int i = 2; i < n; i++)
            {
                cumDepth += dx[i-2]; 
                x[i] = cumDepth + dx[i-1] / 2.0;  
            }
            x[n] = MathUtilities.Sum(dx);   // jumps straight to the bottom of the profile - can end up being a very thick node in a plce that can be highly dynamic
            // There should be 2 more nodes than there are layers (one node in the middle of each layer, one at the surface, and one at the base)
            // This scheme has the same number of nodes as layers
            // If N (so there should be N+2 nodes) is the number of layers then what it is doing is skipping the two layers completely 
            // It would be interesting to know where/how it is getting the soil properties

sno036:
resolves #6498 

(well sort of but enough to get going)


hol353:
@sno036 You mentioned that the Buckworth files weren't supposed to be included as part of this PR. I've merged the layering fix as part of PR #6564 so I'm closing this PR.

Issue #5946, opened 2020-12-13
More sensibility tests


hol430:
While debugging a simulation recently, I came across an instance of a negative soil water content. This took a while to figure out because the actual error from the simulation was about N minerlisation. We go through the same pain whenever we see NaNs in a simulation - they tend to propagate rather quickly and it's hard to trace where they are coming from. In an ideal world, these sorts of errors would be trapped close to (or at) the source. Pre- and post-condition checks for sensibile-ness of data would go a long way towards addressing this. Current plan is to start adding assertions for this sort of thing which only get run if apsim is built in debug mode.

hol430:
As part of this we could check for tmin>tmax as in #6153

hol430:
As part of this, we should automatically check that bounds attributes on inputs are not exceeded.

PR #6501, opened 2021-05-25
Check model parameter bounds before a simulation is run

### LINKED ISSUE #5955: Better display of composite exceptions in the gui
Wrapping exceptions inside each other can be a useful way to provide additional context in error messages. Unfortunately, the gui will only display the contents of the inner-most exception in the status panel. We could change this so that the inner exceptions' messages are concatenated to the error text shown in the status panel.

hol430:
Working on #5946 
Resolves #5955


lie112:
CLEM uses IValidatableObject on models for custom logic and setup testing and property validation attributes to allow full testing of property values types and bounds. These test are managed by the CLEM model, but could be equally applied to all ASPIM models (maybe managed by Simulation). A switch to turn off validation would also be useful to reduce CPU overhead when not needed on HPC, but never sure where simulation wide properties should live. You often need to perform this validation after the model has had a chance to set things up. In CLEM there are events to initialise resources and then initialise activities (After StartOfSimulation) followed by the validation checks (After FinalInitialise)
So APSIM validation would need to go after the Finalize event and walk through every component in the tree.  CLEM uses a static validation error and warning log class (Models.CLEM.WarningLog) that any model can write to and these are checked and reported before starting the simulation run.

lie112:
I have been looking at the code changes to implement validation but trying to work out the approach taken. Why did you choose not to use the .Net validation with property attributes? This allows much of the range testing and required values and can also provide additional attributes such as the clem written attribute that can check a value against another property value set. This would allow a check to ensure end date is after the start date (an issue that took me ages to find the other day). My approach would be for Model to be IValidatableObject and an empty base.Validate method can be overridden to provided for any model for custom validation not covered by the properties and after any initial setup occurs.

hol430:
The purpose of this pull request was to check how many bounds attributes (which are already in the code) are being violated (turns out it's a lot). If I were to ever revisit this PR, it would be to try and fix these violations and get this check merged in, but that would be a huge job. Refactoring the entire codebase at the same time to use the builtin validation attributes  would just make the changes that much bigger. As to why the builtin attributes weren't used in the first place - no idea. Maybe they didn't exist when this code was written (large parts probably came from old apsim, so you'd need to ask the 'lifers).

zur003:
@hol430 - A few words on behalf of the "lifers" (I guess I'm now an "after-lifer"): I think IValidatableObject was introduced with .Net 4.0 in 2010, and the .Net work done for ApsimClassic pre-dates that. If I recall correctly, the initial work for ApsimX targetted .Net 3.5 rather than version 4, since it adapted a few things from Classic. And of course there was also the problem that we were all still learning this stuff - it was quite a significant change from Fortran and C++.

hol430:
@hol353, @hut104 - do you think this is worth pursuing? It would certainly be nice if our bounds attributes were actually checked somewhere, but getting to that point will be a lot of work. If you don't think it's worth the effort, I'll just close this. Same question applies to #5982 - I don't see much point leaving these dead PRs open for the rest of eternity.

hol353:
Close this PR and other long dead PR's.The associated issue will still point to the PR so that at some point in the future we can find the above discussion from the issue.

PR #6504, opened 2021-05-27
Set ref height to 2.0m in MicroClimate in Canola simulations

her123:
working on #6476


her123:
retest this please jenkins

PR #6513, opened 2021-05-31
Changed MicroClimate to use the soil's albedo

### LINKED ISSUE #6500: Soil Albedo in two places
The soil albedo is currently set in two models (Soil and Microclimate).  MicroClimate should use the values in the soil, and in doing so allow for variations in soil albedo between zones.

hol430:
Resolves #6500


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
@hol353, @hut104 - the stats have changed because micro climate is now using the soil's albedo rather than its own value and in some files these values were obviously not the same. Do we want to accept these changes?

I've also had to modify a few simulations (such as the prawn simulation) to add in a soil. Microclimate now will not run without a soil (previously it would).

PR #6550, opened 2021-06-16
Fixed a few soil-related bugs

### LINKED ISSUE #6341: Disabled soils causing problems
1. The attached simulation has two disabled soils. If I run this simulation I get an index out of range exception. This is coming from one of the disabled soils. If I delete the disabled soils it runs ok.
2. If I click on 'Greenethorpe_soil' Physical node (or the CanolaSoil node under that), I get a nice exception which tells me the problem "Number of soil layers in SoilWater is different to number of layers in SoilWater.Crop" but I can't edit or see the soil in the GUI right hand view.

[canola.zip](https://github.com/APSIMInitiative/ApsimX/files/6300384/canola.zip)

hol430:
- Disabled soils shouldn't cause errors when the simulation is run
- Need to be able to edit faulty soils in the gui

Resolves #6341

### DIFF
--- a/ApsimNG/Presenters/ProfileGridPresenter.cs
+++ b/ApsimNG/Presenters/ProfileGridPresenter.cs
@@ -196,7 +196,20 @@ namespace UserInterface.Presenters
                     continue;
 
                 // Ensure that we have enough rows to display all items in this array.
-                Array array = property.Value as Array;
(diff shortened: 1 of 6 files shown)


hol430:
retest this please jenkins

PR #6559, opened 2021-06-21
Speed up graphs in GUI

### LINKED ISSUE #6558: GUI graphs are slow for complex series
Many of the graphs in the wheat validation are very slow because of the number of factor values in some of the experiments. See the controlled experiment as an example. This is made worse by having pages of graphs.

hol353:
Resolves #6558


hol430:
retest this please jenkins

PR #6573, opened 2021-06-24
Improve scrollbar appearance in gtk3 builds

hol430:
Working on #6139


hol430:
retest this please jenkins

lie112:
It seems the default scroll bar position upon opening is close to the bottom
Can we frames open scrolled to the top?
If, for example, you have a large simulation tree, the simulation does not open up with the Simulations model showing which is not really logical.

hol430:
Yes, the scroll position really ought to be at the top of the simulations tree when we open a file.

PR #6578, opened 2021-06-25
Add the ability to run post-simulation tools in parallel

### LINKED ISSUE #6554: Add the ability to run post-simulation tools in parallel
Currently, all post-simulation tools are run serially. Some of these can be quite slow (e.g. merging large datasets), so having the entire run pause while waiting for a single thread to read some data from a spreadsheet is obviously not ideal. Some of these tools are dependent on other tools (e.g. PredictedObserved cannot be run until its corresponding ExcelInput component has finished reading the observed data). A simple solution for this might be to add a model which implements IPostSimulationTool and just runs all child IPostSimulationTools in parallel.

hol430:
Resolves #6554


hol353:
@hol430 This is now conflicted.

Issue #6577, opened 2021-06-25
Problem with simulation count in status area


zur003:
When running a single simulation (e.g., the Wheat example simulation), the status area at the bottom of the screen indicates that there are 2 simulations, with the first completing almost immediately, so that one sees "Simulations running (1 of 2 completed)" when the (single) simulation is running.

lie112:
This may also relate to the problem I am having where the results for datastore._Messages get written every time I run a simulation (duplicated) by clicking the simulation and selecting run. It seems to forget what simulation run is performed or no longer clears the datastore between manual simulations.

hol430:
I'm having trouble reproducing the problem with the _Messages table - does the problem occur every time for you? I notice that the CleanCommand could run concurrently with the simulation which makes me wonder if this is a race condition...

lie112:
@hol430 I am able to repeat the issue with the Messages table, by doing multiple runs of examples\clem\grazing.
I get a terminated successfully message each time it finishes with the correct clock, so it seems the table is not cleared at the start of a simulation.

hol430:
Yeah there have been a number of changes to the database cleaning code over the last few weeks. I haven't used apsim much this week, but when I have, I've observed some suspicious happenings related to the datastore. Will look into this when I get the chance.

hol430:
@lie112 - could you check if this is still an issue for you? I have a feeling it may have been fixed by #6669.

lie112:
The incorrect counter display in GUI problem is fixed

HOWEVER
If the simulation has a long post simulation process the message displayed during this stage (something like "searching for simulations") is inappropriate and should be "finalising 1 of 1" or something

I will discuss the message table in database in the more appropriate #6586

PR #6581, opened 2021-06-27
Add the ability for DataStoreWriter to run a clean command internally

hol430:
Working on #6577.

@hol353 is there any reason in particular why you added the CleanCommand to the job manager's list of jobs, rather than the datastorewriter's internal list of jobs?

### DIFF
--- a/Models/Core/Run/SimulationGroup.cs
+++ b/Models/Core/Run/SimulationGroup.cs
@@ -237,8 +237,7 @@
                                         names.Add(name);
                                 }
                             }
-                            if (names.Count > 0)
-                                jobs.Insert(0, storage.Writer.Clean(names));
+                            storage.Writer.StartClean(names);
                         }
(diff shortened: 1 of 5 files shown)


hol353:
No particular reason why it is a job manager list of jobs. Probably better if it was an datastore writer internal job.

hol430:
I wonder if it should be allowed to run in parallel with the simulations. Seems like there's potential for it to delete data written by a simulation, depending on the timing. Of course, blocking while it runs seems inefficient as well.

hol353:
Yes it should run in parallel with simulations. Simulations will create and add WriteTableCommand instances to the writers 'commands' queue. The writer works through this queue (the first command will be the clean command) in order so there is no chance that the clean command deletes data written by a simulation.

Do you want this merged?

hol430:
Ok - in that case, this should be fine to merge.

PR #6589, opened 2021-06-28
Don't leave the DataStoreWriter idling after performing a clean.

hol430:
Working on #6577

### DIFF
--- a/Models/Storage/DataStoreWriter.cs
+++ b/Models/Storage/DataStoreWriter.cs
@@ -446,24 +446,18 @@
         /// Create a db clean command.
         /// </summary>
         /// <param name="names">A list of simulation names that are about to run.</param>
-        public IRunnable Clean(IEnumerable<string> names)
+        public void Clean(IEnumerable<string> names)
         {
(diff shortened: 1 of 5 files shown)


Models/Storage/DataStoreWriter.cs
```diff
@@ -446,24 +446,18 @@ private void ReadExistingDatabase(IDatabaseConnection dbConnection)
...
-        /// Initiate a clean of the database.
-        /// </summary>
-        /// <param name="names">Simulation names to be cleaned.</param>
-        public void StartClean(IEnumerable<string> names)
-        {
             Start();
             lock (lockObject)
-                commands.Add(Clean(names));
+                commands.Add(command);
+            Stop();
```
hol353, line 460: Won't this call to Stop block and wait until the clean finishes?

hol430: Yeah, I don't think we necessarily want this change. I was more interested in seeing its effect on jenkins' runtime, which seems to be negligible so I'm just going to close this.

Issue #6584, opened 2021-06-28
Recent changes to graphing code have broken a number of existing graphs


hol430:
A number of graphs in the validation sets have been broken by the recent changes to the graphing code. The releases will be broken until this is fixed.

hol430:
This has been fixed, so I am going to close this issue.

PR #6598, opened 2021-06-30
Stop graph errors from stalling auto-doc creation

hol353:
Working on #6584

### DIFF
--- a/ApsimNG/ApsimNG.csproj
+++ b/ApsimNG/ApsimNG.csproj
@@ -51,7 +51,6 @@
     <PackageReference Include="SharpMap.Layers.BruTile" Version="1.2.0" />
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="5.0.0-preview.1.20120.5" />
     <PackageReference Include="System.Resources.Extensions" Version="5.0.0-preview.1.20120.5" />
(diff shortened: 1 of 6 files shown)


hol430:
I'm pretty sure that we've deliberately set things up this way so that autodocs *do* break if a graph fails to draw. Otherwise you end up with empty/missing graphs in the autodocs. As time goes on (and no one checks the logs for successful CreateInstall jobs), more and more of the graphs in the autodocs become empty.

hol353:
Hmmm Perhaps you're right. The problem we need to stop is installations not being created when the auto-doc creation fails. Does the 'Create Installation' jenkins job actually create installations anymore? If not, then I will undo part of this PR to allow exceptions to propagate again and can you rename the 'Create Installations' job on Jenkins.

hol430:
The CreateInstallation job on jenkins will build the installers for the .net framework version but not the .net core version (those are generated at the same time that we run the validation tests for a pull request). However, the installers generated by this job are only released (ie show up as upgrades on the downloads page and in the GUI) if the CreateInstallation job passes, which requires that the documentation stage runs without error.

Ultimately, I think the only way to "fix" this problem once and for all is to build the autodocs when testing a pull request.

PR #6602, opened 2021-07-01
Refactor TreeView, to avoid calls to TreeIter.Equals()

### LINKED ISSUE #6599: Error upon opening .apsimx file in the gui
On some systems, attempts to open an .apsimx file in the gui result in the following exception: `Method not found: 'Boolean Gtk.TreeIter.Equals(Gtk.TreeIter)'.`. It would appear that an older version of gtk-sharp has been stored in the GAC, and this older version is missing a few methods which we are currently attempting to call. It should be possible to work around this problem by refactoring the TreeView class to avoid calls to `TreeIter.Equals()`.

hol430:
Resolves #6599.

### DIFF
--- a/ApsimNG/Views/LegendView.cs
+++ b/ApsimNG/Views/LegendView.cs
@@ -141,23 +141,6 @@ namespace UserInterface.Views
             }
         }
 
-        /// <summary>Returns the index of an item.</summary>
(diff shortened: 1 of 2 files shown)


hol430:
retest this please jenkins

Issue #6603, opened 2021-07-01
Create new grid control for GUI


hol353:
We need a new, **simple**, grid control to replace the existing GTK grid (which is really just a tree view).

The grid should:
- allow multiple rows and columns to be frozen,
- be theme aware,
- allow cell editing,
- allow multiple cell selection,
- allow simple cell formatting (bold, italics, text left/right justification),
- NOT allow cells to contain drop downs, images, checkboxes etc.,
- NOT allow different fonts in cells.

hut104:
It is good to see clear indication of things OUT of scope.  Less is more.  Grids have caused issues for a long time.

________________________________
From: Dean Holzworth ***@***.***>
Sent: Thursday, 1 July 2021 2:46 PM
To: APSIMInitiative/ApsimX ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [APSIMInitiative/ApsimX] Create new grid control for GUI (#6603)


We need a new, simple, grid control to replace the existing GTK grid (which is really just a tree view).

The grid should:

  *   allow multiple rows and columns to be frozen,
  *   be theme aware,
  *   allow cell editing,
  *   allow multiple cell selection,
  *   allow simple cell formatting (bold, italics, text left/right justification),
  *   NOT allow cells to contain drop downs, images, checkboxes etc.,
  *   NOT allow different fonts in cells.

lie112:
I would like to be able to include some additional functionality or use the grid as a base class
I need to be able to provide an image (or font-icon) in each cell for a status report. Link image/icon to value for display.
At present this report has become slow to render and not very responsive to move around so look forward to grid and an upgrade to justify more girded display.

example of the current activity status report using the old treeview based grid
![image](https://user-images.githubusercontent.com/19215637/124206770-266dcb00-db23-11eb-9936-48f79cc11087.png)

Grid also needs to be more responsive for large datasets, display on demand as previously discussed.

hol430:
You should be able to subclass the sheet widget, and just override the DrawCell() function. That would give you complete control over what gets rendered in each cell.

> more responsive for large datasets, display on demand as previously discussed.

Data retrieval is hidden behind an interface (`ISheetDataProvider`), and the current implementation used for the DataStore loads data lazily, so the UI remains responsive and loads quickly.

> read only switch to disable cell editing
> allow select cells/rows to be disabled

[...]

PR #6604, opened 2021-07-01
New grid widget for GUI

hol353:
Working on #6603. Currently the grid is only used on the DataStoreView. Doesn't yet support cell multi-select.


hol353:
@hol430 Can you pull this branch and check new grid is ok on Linux?

hol430:
Noticed a couple of oddities. Not sure if they're linux-specific, and some of them may be intentional:

mono (net472) version:

- Control + scroll up scrolls to the top, but control + scroll down doesn't scroll to the bottom
- Shift + scroll up/down to scroll left/right doesn't work
- Can't copy data
- Colour scheme in dark mode isn't right:

image

- Assorted gtk warnings:

```
Gtk-WARNING **: 08:22:24.188: Unknown property: GtkLabel.halign
Gtk-WARNING **: 08:22:24.188: Unknown property: GtkLabel.margin-bottom
```

.net core version:

- Cells seem very tall:

image

- Shift + scroll up seems to scroll up much farther than shift + scroll down (scroll deltas may be incorrect?)
- Can't figure out how to view a different table or checkpoint (actually I can't even see the checkpoint column, is that deliberate?)
- If I click in the area beneath the cell area but above the horizontal scrollbar, I get the following exception:

```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'index')
   at System.Linq.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument)
   at System.Linq.Enumerable.ElementAt[TSource](IEnumerable`1 source, Int32 index)
   at UserInterface.Views.SheetWidget.CellHitTest(Int32 pixelX, Int32 pixelY, Int32& columnIndex, Int32& rowIndex) in /home/drew/code/ApsimX/ApsimNG/Views/Sheet/SheetWidget.cs:line 238
   at UserInterface.Views.SingleCellSelect.OnMouseClickEvent(Object sender, EventButton evnt) in /home/drew/code/ApsimX/ApsimNG/Views/Sheet/SingleCellSelect.cs:line 97
   at UserInterface.Views.SheetWidget.OnButtonPressEvent(EventButton evnt) in /home/drew/code/ApsimX/ApsimNG/Views/Sheet/SheetWidget.cs:line 351
```

hol353:
> Control + scroll up scrolls to the top, but control + scroll down doesn't scroll to the bottom
    Shift + scroll up/down to scroll left/right doesn't work

I haven't implemented control and shift mouse wheel scrolls. It aiming for a minimum viable grid widget at this stage. This might be a version 2 feature.

> Can't copy data

Version 2 feature. There is no multi-select yet so no need for copy.

> Colour scheme in dark mode isn't right:

Hmm. Really hard to get GTK theming right. My dark mode looks ok. Can you help with this one?
image

> Assorted gtk warnings:

I don't get any warnings but I've just fixed the glade file so maybe that removed the warnings.

> Cells seem very tall:

Hmm. I had to fix cell height to a number of pixels. I choose a number that matches my simulation tree node height. I notice in your screen shot that your tree is very compressed compared to mine. Not sure how to fix this. Can you help with this one?

> Shift + scroll up seems to scroll up much farther than shift + scroll down (scroll deltas may be incorrect?)

Not implemented yet.

> Can't figure out how to view a different table or checkpoint (actually I can't even see the checkpoint column, is that deliberate?)

Fixed.

>    If I click in the area beneath the cell area but above the horizontal scrollbar, I get the following exception:

I don't have an area above the horizontal scrollbar so can't reproduce this one. Can you help with this one?

hol353:
I just noticed you're running the .net framework version. The look in it isn't perfect. Can you use the .net core version please? I haven't put much time into look and feel of .net framework as we're switching to .net core soon.

hol430:
My first screenshot is the .net framework version, the second screenshot is the .net core version. I'll look into these other problems, sounds like they may be specific to linux if you're not seeing them.

hol430:
Another Q - are you still parsing the global.css? The idea is that this file will contain style changes which is applicable in both light and dark modes. Things like the progress bar height, scrollbars always being visible, etc. I see that you've created a light.css and dark.css which is fine, but we don't really want to have these other style changes duplicated in both of these files.

hol353:
> Another Q - are you still parsing the global.css? 

Hmm not sure. Maybe I'm not anymore although the scroll bars (which is in global.css) on the tree view seem ok.

hol430:
Ok, I've fixed most of the issues outlined above, although you might want to double check that colours in gtk2 builds look ok on windows in dark mode. I notice that some of the cell text doesn't fit inside the allocated cell width. I'm guessing this is because column width is calculated once, based on he length/width of the column title? This will cause problems in some places, such as the following (from barley validation):

image

I'm not sure what we should do about the cell height. For now I've just vertically centered the text so it doesn't look so abnormal.

image

zur003:
Looks pretty good. I'm impressed, Dean.

However, I'm not seeing correct vertical positioning in the cells (Windows, .Net Core).

image

hol430:
Wow, your simulations tree is very bold Eric :).

Is your branch up to date? I made some changes to the text's vertical alignment about an hour ago which may affect this. If you're running with these changes, the problem is probably windows-specific as it looks normal for me on linux.

jbrider:
It looks pretty good on mine - Win11 using netcore build.
Selecting a header cell ca have a long delay but the grid below is very responsive.

I assume resize doesn't exist yet.

zur003:
@hol430  - Yes, I'm using the latest code from hol353\NewGridWidget. I've only tried on Windows so far, but will try on LInux shortly.

jbrider:
The header selection seems to be tied to an onExit event that is only triggered when moving the mouse left or below the grid bounds.. at least on mine it does.

hol430:
I didn't think you could click on the column headers. At least, it doesn't do anything on my machine, and I can't see any code which handles that scenario.

zur003:
@hol430  - When I click on a column header, nothing seems to happen immediately, but if I then shift focus to a different widget, the column header appears as selected. This is on Windows; Linux may behave differently.

hol353:
Ok I have fixed (hopefully) the above problems. Drew, your changes to center text vertically in a cell didn't work on Windows. My screen looked like Eric's screenshot above. I've changed it to use the vertical text extents that Cairo reports. Hopefully this works on Linux as well - can you please check it. Eric, I've also fixed the infinite loop you reported. I'm not sure why this PR is running red - I'll look into it.

hol430:
Text vertical alignment looks good for me. However I've just noticed some oddities with the horizontal scrollbar. When I first click on the datastore, the horizontal scrollbar is in the middle of the scrollbar range. If I move the scrollbar to the left, the displayed cells don't change. However, if I move it back to its starting position (one scroll increment to the right), then it scrolls 1 cell to the right, as expected.

hol353:
My scrollbar isn't in the middle of the range. It is slightly further right than it should be though so I've committed that. Can you check and see if it fixes yours as well.

My scroll bars in the grid don't look right though:

image

If I comment the code below from global.css then they come good. What does the below css do?

`scrollbar, scrollbar button, scrollbar slider {
	min-width: 8px;
	min-height: 8px;
}`

hol430:
Yep, your commit fixed the scrollbar problem for me. That part of the css makes the scrollbars a bit wider. Some people were complaining about them being too narrow under the stock adwaita theme. You can remove it if you want. Out of interest, which file/model are you looking at in your screenshot?

PR #6621, opened 2021-07-07
Avoid trying to read data values when all we really want are the unit…

### LINKED ISSUE #6612: Annoying exceptions
If one tries to run the Wheat Validation simulations under the IDE debugger, with settings to stop on all exceptions (whether handled or not) one sees a lot of exceptions (well, mostly Null Reference exceptions when trying to get a value for root depth, I think). Things seem to run correctly, but stopping for all these exceptions make debugging problematic. The triggering condition should be handled in some other way, rather than relying on silent exceptions.

zur003:
…s of measurement of a variable. Resolves #6612

### DIFF
--- a/Models/Core/ILocator.cs
+++ b/Models/Core/ILocator.cs
@@ -27,6 +27,13 @@ namespace Models.Core
         /// <returns>The found object or null if not found</returns>
         IVariable GetObject(string namePath);
 
+        /// <summary>
+        /// Get the properties of the underlying variable object for the given path.
+        /// Unlike the GetObject method, this does not return the data value of the object.
+        /// </summary>
+        /// <param name="namePath">The name of the variable to return</param>
+        /// <returns>The found object or null if not found</returns>
+        IVariable GetObjectProperties(string namePath);
 
     }
 }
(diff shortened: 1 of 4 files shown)


hol430 (commented):
Thanks Eric - looks good to me. Out of interest, what's the idea behind the break statements you've added to the locator? If I'm reading this correctly, we're about to exit the loop anyway when `j == namePathBits.Length - 1` evaluates to true(?).

zur003:
@hol430 - No particularly good reason for the "break" statements, other than it being one way to prevent the immediately following statements from being executed. "If" statements could have been used to similar effect. I know that many coders are philosophically opposed to the use of "break" statements, but in this instance I personally find it slightly easier to read (it being a case of "we've now got all the information we need; let's get out of here"). Feel free to modify it, if you wish.

hol430:
Yeah, I don't really have a strong preference either way.

her123:
Avoid break statements where possible.

hut104:
Agree.

hol353:
I prefer to use if/else statements but am ok with break statements if they improve readability (i.e. sometimes the if/else statements can lead to complex code). In this case I see no harm in the breaks.

Issue #5317, opened 2020-06-18
We should use official sqlite package


hol430:
Is there any particular reason that we directly interact with the native SQLite functions? Seems like it would be more reliable (and simpler) to use the official sqlite packages from nuget.

hol353:
Purely historical reasons probably. Can probably change to the Microsoft package.

her123:
Just check that it will work with dotnetcore

hol430:
Appears to be working fine on .net core as far as I can tell. Still trying to fix some issues related to closing the connection after closing a tab in the UI, but that's an issue with the .net framework version as well.

I wonder if it's worth keeping the database connection closed at all times except when we're actively using it. Might be a bit of work to implement though, so maybe it's not worth it...

zur003:
Which of the SQLite nuget packages are you considering?

zur003:
Probably wouldn't hurt to test this on Linux and OS X - it probably needs to load a native binary, and we need to be sure it works cross-platform.

lie112:
I've used System.Data.SQLite.Core in other projects without trouble. 
I wasn't able to get it to work in a .Net Standard library but works in Core
I always find it best to keep SQL connections open to a minimum and always dispose of Commands otherwise the database is locked if you run the simulation again.
Looking forward to MAUI with .Net 5 and 6. That will solve some of the cross platform UI issues.

zur003:
Thanks @lie112 - I wasn't aware of MAUI. It looks like it doesn't support Linux, though. @hol353 - do we need GUI support on Linux platforms?

hol353:
Yes we need GUI support on Linux. MAUI is still a way off. We could spend some time evaluating Xamarin (which is where MAUI is evolving from) for our needs. Or we could try the MAUI preview later this year.

hol430:
Yes, [System.Data.SQLite.Core](https://www.nuget.org/packages/Microsoft.Data.Sqlite.Core/) is the package I was considering. Haven't tested it on Linux yet but I will.

Migrating the UI to gtk3 is going to be a big job, but I expect that converting to another toolkit such as maui will be an even bigger job, so I don't really see much point unless it does something way better than gtk.

her123:
I also found that the libsqlite3.so in the repository wouldn't work on Linux. I had to update it locally.

hol430:
Where is your libsqlite3.so located? As far as I can see we don't have it anywhere in our repo

her123:
[...]

PR #6627, opened 2021-07-09
Implement access to datastore views for plotting in graph series

### LINKED ISSUE #6626: Graphs stopped accepting database views for series plots
It seems that graphs no longer accept views in the database for x and y plots

lie112:
Resolves #6626 

- Added new test for SQL validity that returns bool rather than exception
- Check SLQ validity before adding a view to datastore otherwise errors thrown when trying to get view columns for creating datastore on opening simulation
- Fixed checking of CLEM ReportQuery to report bad sql before updating


APSIM.Shared/Utilities/SQLite.cs
```diff
@@ -457,6 +457,25 @@ public void ExecuteNonQuery(string query)
...
+
+            // do not allow testing of modifying sql
+            if(new List<string>() { "drop", "create" }.Any(query.ToLower().Contains))
+            {
+                return false;
+            }
+
+            //prepare the statement
+            IntPtr stmHandle = Prepare(query);
+            int code = sqlite3_step(stmHandle);
```
hol430, line 475: Need to finalize the prepared statement after we're done with it, or we will leak memory.

However I'm not sure if this function belongs here, it seems beyond the scope of this class, which is supposedly to wrap the native SQLite API. Admittedly, I'm not really sure of  the use case here - are the clem users really writing SQL to create the view? Most apsim users I know wouldn't be up for that. The approach we've taken in other places is to have the user specify the inputs and have the SQL generated by apsim (e.g. in the PredictedObserved model). In the longer run we'll need to rethink this anyway as we will be adopting the nuget version of sqlite.

lie112: I had to get down to this level to avoid all the exception handling and just report the SQLITE_DONE code. The readers and datastore classes didn't give me this level of checking.
I've added a Finalize but couldn't use the Method as it throws an exception that I am trying to avoid

I was looking for a neat way to test an sql (select) and have boolean return if ok but everything threw an exception.
We have a ReportQuery that creates a view based on an SQL select statement provided by the user to summarise database for analysis and graphing. We will also be providing SQL to do this low level analysis for model summary and checking before moving to other data analysis packages. 

We have a ReportPivot which in time will generate it's own SQL for a view, but the need for validating this is probably less important. This report will be the case where the data table, row and column variables are reported by the user and the code will generate the SQL to create the view. We have moved (are moving) away from generating additional summary tables in the database as sql views are so much quicker to access.

In the past errors were simply ignored but after I tried to put the views back into the graph series (they used to be there) if there was a bad SQL it would throws an exception before the simulation even starts (getting the column names for tables array)

hol430: That seems fair, although again this is going to break when we implement #5317, as we'll lose access to these low level routines. Out of interest, what's stopping you from just calling `dataStore.Reader.ExecuteNonQuery()` in your code and catching the exception to detect an error? By doing so, you'd also have some extra info (the exception message) about what specifically has gone wrong. (Admittedly, this message is not always helpful.)

lie112: I looked for ages the other day trying to learn how to "cancel" an exception once in the catch code, but didn't find a solution online. I tried what you suggest by returning the bool false in the catch statement, but it still throws the exception and crashes out of the code and then because the view is created the simulation won't open because datastore loading fails on the broken view not able to return columns. I wasn't able to avoid all the exceptions thrown and continue with my code as exceptions are also thrown in finalize method. I thought you would prefer handling the problem at the source rather than using try catch statements to do logical decisions. There is no SQL statement to check the validity of SQL, but SQLite suggests using the return code, so I went to where that is in APSIM. Looks like it might have to be a try catch in Microsoft.Data.SQLite. By nuget version of Sqlite, do you meant start using Microsoft.Data.SQLite?

hol430: You don't need to do anything to "cancel" an exception, as long as you catch it and don't rethrow it, it will stop propagating. There are a few types of exceptions (StackOverflowException, AccessViolationException) which can't be caught as far as I know...is it possible your exception was an AccessViolationException?

Agreed about not using a catch to do logic. On the other hand I do think that having the exception message here is useful. I also question whether checking for "drop" and "create" in the query is the right thing to do here. This seems like application-layer logic, and if any of the column/table names contain either of these words, then it will be flagged as an invalid query, which seems a bit harsh.

According to my notes, I was trialling System.Data.SQLite, although I'm not totally sure what the differences to Microsoft.Data.SQLite are, other than the microsoft version apparently being more "lightweight".

lie112: I can implement this. It seemed like I am starting to duplicate functionality and this doesn't help if the user supplies a csv file for the data input the the TextFile reader automatically handles for me. So, I'd also have to write or use the csv reader.  Maybe I could add a OpenSimple method for the excel open.

lie112: I'll look into this further, but I think there's more than one place where exceptions are thrown .

lie112: I must have been muddled and stuck in another problem the other day heading down the wrong rabbit hole. I have the try-catch working to prevent writing bad SQL to view to avoid future loading failure. I have also added this logic to the DataStore.AddView just in case the passing model has not performed the checks.

hol353:
@hol430 Can you relook at this PR please?

hol430:
Well I don't really like the swallowing of exceptions in the AddView() function- I think it might be better to allow the caller to handle the error, as the datastore doesn't really have the context to make that sort of decision. In some cases we may not want to continue if the command didn't execute correctly, and if this function just swallows the error like this, the caller has no way of knowing if it ran successfully or not.

On the other hand, this function is only used by clem, so if Adam's happy then it could probably be merged.

lie112:
One cannot assume the UI has handled a bad sql statement being written to the database and the consequences of saving this bad sql as a view to the SQL is that the simulation file will not open. The user is presented with an error that doesn't really tell them that they need to get into the json aspimx file and fix a SQL property entry, delete the database and try open APSIM and the simulation file again, upon clicking open file. This can be avoided by 

- Preventing bad SQL being written to the datastore as a view in DataStore.AddView. This will result in the user not seeing the view in the datastore, and hopefully a warning is provided to alert them to the problem. This is the approach I took. Bad sql should simply fail to create a view as the last line of logic. The UI should trap for this and provide warnings or exceptions as required, but this cannot be assumed.
- Handling the error in the DataStore initialisation when get columns names is called on the bad view causing an exception. But here we'd need to not throw the exception, but simply skip adding the view to the tables list and continue on with opening the simulation. This would then allow bad tables to also be skipped and was harder to implement without putting a try catch in the DataStore code where hard to provide correct logic.

Upon more thought I don't care if the user puts drop, delete, update etc in the SQL text as at worst they are only destroying their own stand-alone SQL database that will be recreated upon the next simulation run.

hol430:
> One cannot assume the UI has handled a bad sql statement being written to the database

Currently, if the UI doesn't handle all exceptions, we will get a crash. Therefore, the UI must handle all exceptions. We make this assumption every time we throw an exception in the gui.

If a dodgy sql view is causing the .apsimx file to fail to load in the gui, that's a bug. Ideally, if the user tries to add a view but the sql is dodgy, the view should not be added to the db. However, the user should also be able to open the .apsimx file in the GUI even if the .db file contains a dodgy view. If this isn't working, we're probably trapping the exception too late...could you post either exception message including stack trace, or instructions on how to reproduce it? (Or just upload the .apsimx + .db file if you have an example at hand.)

lie112:
I thought there were places where exceptions are thrown even if not handled at the UI (e.g. DataStoreReader). In this case if the UI does not look for an error the DataStore.AddView could be called without check and getting bad Sql. Therefore, I was doing the same quick check before adding the bad sql as a view to the database in AddView. 

I am currently unable to get to my code base that can reproduce the error on opening the model. The problem happens in the DataStorereader.Refresh when we again allow views to be added at line 182 (GetTableAndViewNames) that hen calls tables.Add with the GetTableColumns method as an argument. This method assumes the table is valid and will return column names which is not the case for the view with bad SQL and so a crash at this point fails to allow the simulation apsimx file to open.

Thus, I was ensuring that a bad sql statement was never written to the database to avoid this issue. In the past all my SQL was good and I never came across this issue. Something must have changed in the latest version as the code to add views to the x and y series of graphs used to work fine.

lie112:
Regarding @hol430 comment from 20 days ago.
I now have an aspimx file that reproduces the error.

The problem is that during setup on opening the datastore.Open(), dbReader.SetConnection (), Refresh(),  for each table in database, Connection.GetTableColumns(tableName), GetTableColumns(), ExecuteQuery(), PrepareQuery -> throws exception back to datastore.Open and crashes out rather than open the apsimx simulation with an empty table of columns for a bad view. This is because the SQL contains a table that doesn't exist in the database and was never a problem while APISM only considered good tables.

We decided that there was no scope to use try catch to check validity of sql statement in an execute query in SQLite.cs and SQLite engine is changing. The only way to check valid sql is with try catch as SQLite doesn't have a return for success.

We moved the requirement to test SQL back to the UI as requested, which is fine, as long as the SQL remains valid after further user changes, such as deleting a table (or disabling) the report that is referenced in the SQL. If the table is deleted then the SQL considered valid is in fact bad and wont let the apsimx file open up if the db has been emptied.

The offending snippet of code is where the columns are defined for a table and so we could skip adding any table where this issue happens.

Line 181 DataStoreReader
`            // For each table in the database, read in field names.
            foreach (var tableName in Connection.GetTableAndViewNames())
            {
                try
                {
                    tables.Add(tableName, Connection.GetTableColumns(tableName));
                }
                catch
                {
                    // ignore error or report somehow
                }
            }
`
Is this an acceptable solution?

hol430:
No, the datastore reader doesn't have enough context to be able to report the error (if we caught the error here, what would we do with it?). Whatever is calling into the datastore api needs to be prepared for a failure to open the DB and should be responsible for handling the error. Ideally, we should also still be able to open the DB even if we are unable to retrieve table/view names but I can't really suggest the best way to achieve this without seeing the problem in action. Could you post a full stacktrace for this error and/or your faulty apsimx file?

lie112:
@hol430 , here's the aspimx file and the database that causes the problem
You won't be able to get the run working as it's missing other files, but the issue is in opening the simulation file anyway.

The simplest solution is probably to allow get tablecolumns to return an empty datatable if the executeQuery fails.
This is the logical place to trap the error. In that way the table/view is still added to the table list, but just has no columns as would be expected. I am not sure if no columns creates problems later on.

[invalid sql test.zip](https://github.com/APSIMInitiative/ApsimX/files/6934979/invalid.sql.test.zip)

PR #6628, opened 2021-07-12
When saving files, apply the filter's file extension

### LINKED ISSUE #6618: File Extension not set
"Save As" does not appear to add the file extension (.apsimx) to the file name.

If you open the sobol example, then choose run, then type a file name in the file save as dialog, even though .apsimx is selected, the file name in the folder does not have that extension.

zur003:
Resolves #6618

### DIFF
--- a/ApsimNG/Utility/FileDialog.cs
+++ b/ApsimNG/Utility/FileDialog.cs
@@ -160,9 +160,10 @@
 #endif
             fileChooser.SelectMultiple = selectMultiple;
 
+            string[] specParts = null;
             if (!string.IsNullOrEmpty(FileType))
             {
-                string[] specParts = FileType.Split(new Char[] { '|' });
+                specParts = FileType.Split(new Char[] { '|' });
                 for (int i = 0; i < specParts.Length; i += 2)
                 {
                     FileFilter fileFilter = new FileFilter();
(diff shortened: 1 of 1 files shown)


hol430:
Does this problem occur on the .net framework build on windows? I believe the GenericFileDialog() function is only called in the .net core builds, and in the linux .net framework (mono) build.

zur003:
@hol430 - I think the original problem was only with .net core builds. The frameworks version appears to work as expected, at least on Windows.

However, I just noticed that there is another thing I should probably address while we're looking at this: the Frameworks version asks for confirmation if you're asking to overwrite an existing file; the Core version doesn't.

zur003:
@hol430 - I'm not entirely happy with this, as there's one case where things don't work well: if the user has an extension filter active, but doesn't provide the extension in the filename (so the extension is being applied as a default) **and** a file with that (extended) name already exists. With FileChooserNative, there's no way to interact with the dialog components while the dialog is running, so we can't "correct" the lack of an extension until the dialog is closed, so the normal overwrite test can't be applied. I've written a sort of hack which just reopens the dialog. It's not an ideal solution, but it's the best I've been able to come up with.

lie112:
This merge should have raised the update/upgrade builder
Do we need an open issue to report updates not being released

Seems 08/07/2021 Annoying Exceptions was the last successful build of upgrade today (19/7/2021)
That's the last one showing up in my Upgrade form but Jenkins suggests there's a 17th July built.

hol430:
This was only merged 45 minutes ago (the release builds take longer than 45m). However, there has indeed been a failed build (on the 16th). I will investigate.

PR #6631, opened 2021-07-14
Small tidy up of gotophase

### LINKED ISSUE #6630: Why does GoToPhase have separate properties for End and PhaseNameToGoTo
The GoToPhase has an End property (which is not used) as well as a PhaseNameToGoTo property. The End property looks as if it only exists to satisfy the requirements of the IPhase interface. I think it would make sense to remove the PhaseNameToGoTo property altogether, and just use the End property. This would allow us to remove some special conditional code from Phenology.cs.

hol430:
Resolves #6630 

See the change to Phenology.cs for the rationale behind this change.

I'm not sure if the GotoPhase's end stage name should be its PhaseNameToGoto (which is what it's set to in all existing instances of this class), or the start stage name of the target phase, which would make more sense to me based on what this class actually does.

### DIFF
--- a/Models/PMF/Phenology/Phases/GotoPhase.cs
+++ b/Models/PMF/Phenology/Phases/GotoPhase.cs
@@ -30,9 +30,14 @@ namespace Models.PMF.Phen
         [Description("Start")]
         public string Start { get; set; }
 
-        /// <summary>The end</summary>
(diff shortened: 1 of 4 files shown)


hol430:
retest this please jenkins

Issue #6632, opened 2021-07-14
Additional functionality needed for economic analysis of beef enterprises


lie112:
Research economists have provided a list of additional functionality needed for full economic analysis

- Price schedules from external data/streams
- Fully customisable category for reporting transaction types
- Breakdown of ruminant purchases and sales by class/pricing groups
- Present amount and/or value in resource balance report

PR #6633, opened 2021-07-14
Add price management from input file

lie112:
working on #6632


lie112:
I thought this had been accepted and was about to do the next pull.
We can either leave the last 8 commits in this request or remove them and put them in as another pull request
They do relate to adding economic functionality to CLEM but not the price management

PR #6645, opened 2021-07-20
Updates the balance report for economic output

lie112:
Provide a number of new properties to the ReportResourceBalanaces for economic reporting
Add memo to summary description and clean up Ruminant individual as added to Help system.

working on #6632

### DIFF
--- a/Models/Report/Report.cs
+++ b/Models/Report/Report.cs
@@ -67,14 +67,14 @@ namespace Models
         /// </summary>
         [Summary]
         [Description("Output variables")]
-        public string[] VariableNames { get; set; }
+        public virtual string[] VariableNames { get; set; }
 
         /// <summary>
         /// Gets or sets event names for outputting
         /// </summary>
         [Summary]
         [Description("Output frequency")]
-        public string[] EventNames { get; set; }
+        public virtual string[] EventNames { get; set; }
 
         /// <summary>
         /// Date of the day after last time report did write to storage.
@@ -298,7 +298,7 @@ namespace Models
 
 
         /// <summary>Called when one of our 'EventNames' events are invoked</summary>
-        public void DoOutputEvent(object sender, EventArgs e)
+        public virtual void DoOutputEvent(object sender, EventArgs e)
         {
             DoOutput();
         }
--- a/Models/CLEM/Activities/RuminantActivityGrow.cs
+++ b/Models/CLEM/Activities/RuminantActivityGrow.cs
@@ -503,7 +503,7 @@ namespace Models.CLEM.Activities
                 // limit calf intake of milk per day
                 energyMilkConsumed = Math.Min(ind.BreedParams.MilkIntakeMaximum * 3.2, energyMilkConsumed);
 
-                energyMaintenance = (ind.BreedParams.EMaintCoefficient * Math.Pow(ind.Weight, 0.75) / kml) * Math.Exp(-ind.BreedParams.EMaintExponent * ind.AgeZeroCorrected);
+                energyMaintenance = (ind.BreedParams.EMaintCoefficient * Math.Pow(ind.Weight, 0.75) / kml) * Math.Exp(-ind.BreedParams.EMaintExponent * (((ind.Age == 0) ? 0.1 : ind.Age)));
                 ind.EnergyBalance = energyMilkConsumed - energyMaintenance + energyMetablicFromIntake;
                 ind.EnergyIntake = energyMilkConsumed + energyMetablicFromIntake;
                 ind.EnergyFetus = 0;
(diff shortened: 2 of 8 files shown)


Models/Report/Report.cs
```diff
@@ -67,14 +67,14 @@ public class Report : Model
         /// </summary>
         [Summary]
         [Description("Output variables")]
-        public string[] VariableNames { get; set; }
+        public virtual string[] VariableNames { get; set; }
```
hol353, line 71: I'm unsure why you need to make these virtual properties. In your derived report components can't you just set the value of VariableNames and EventNames?

lie112: Its all about controlling what is displayed in PropertyPresenter. The problem is I wanted to give them a new DescriptionAttribute for Variables or no attribute for Events to ensure they show up (or don't given the no inheritance option) in the property presenter. I can't just use new properties to handle this as the old descriptions will be used and show up.

lie112: The full changes to Report.cs and mu custom reports are in the next big pull request
This was enough to get the BalancesReport working ages ago.

hol353: This is a hack to get around the way PropertyPresenter looks for properties to display to the user. Yes, currently it looks up the inheritance chain for description attributes. Does it need to though? We don't use inheritance much so maybe we could turn that off and see what happens. That way you don't need to make these properties virtual. @hol430 can you comment?

hol430: My opinion is that if the base class' properties aren't relevant to the derived type, perhaps inheritance is the wrong tool. I think that the property presenter should indeed use `BindingFlags.FlattenHierarchy` and display the base class' properties.

However, what we could do is modify DescriptionAttribute's AttributeUsage attribute, and set Inherited to false. That way, if you override the property in a derived type, the overriding property won't automatically inherit the description attribute from the base class' property. That way it would be a conscious decision on the part of the model developer to not show that particular derived property in the GUI, even though the base class *would* show that property.

I think that would be reasonable in these special reporting types, although in most cases I think you would want to think carefully about whether inheritance is really the right solution to your problem.

lie112: I have have already made all the changes @hol430 suggests for you to look at in the next pull request that I can submit after this one is accepted as they are on my same branch (more protected privates). The binding flag worked like a charm but only works on classes and can't be defined for a property
I could use public properties local to the inherited class and work with the VariableName and EventNames in the background.
But, I need a way of disabling the DescriptionAttribute from the base VariableNames and eventNames so they don't show up in my property presenter. I could use property new VariableAttribute and then work with base.VariableNames but this approach makes the code harder to understand with two versions of VariableNames being used. At present the only way is to have them in your class with the description attribute (or non) relating to the class

lie112: There is an assumption that the Description Attribute is the switch to display in the Property Presenter. Creating another attribute to ignore the property in the property presenter doesn't work either as we will need an instance of the property in the derived class to assign the class specific display requirements, thus the need for public virtual

I can outline the other places where I have used inheritance for approval. I didn't realise it was not favoured even though I have read all the coding guidelines. Another is not to use #Regions but they seem to be used elsewhere and I have used this to hide the code for various aspects (e.g. descriptive summary creation). Somewhere I cam across (os someone told me) that you wanted curly braces for every if statement to avoid confusion and for readability, but this doesn't seem to be the case throughout APSIM code. I have the system giving me errors when they are missing. 

We could apply an APSIM code standard file which would force everyone to clean up coding practices. But just turning this on would cause chaos for a while.

CLEM component is an APSIM Zone. Because I didn't want the area, slope, aspect angle and altitude, I made them new properties in CLEM and turned display off. This is effectively to solve the same problem as here.

hol430: Just remove the description attribute from the variable names and event names properties in report. The report presenter doesn't look at these attributes anyway.

hol353: > Its all about controlling what is displayed in PropertyPresenter. 

Yes actually report models don't use the property presenter. I'm confused.

hol430: Yeah, I'm also not sure why it's there, but it's probably not the only case of a model misusing this attribute, as its meaning has changed several times. It's also not really documented anywhere - there should probably be a doc in the "contributing" section of the website about all of this.

lie112: Good point. I run CLEM reports through a wrapper to display the additional properties in a property presenter for the user.

I have reverted back to public string[] for VariableNames and EventNames in my code but it's mixed with all the other changes to clean up the inherited reports.

Can we accept the public virtual now an my next commit and pull will get it to how you 'd like it?

PR #6651, opened 2021-07-25
Remove duplicated report code from CLEM reports

### LINKED ISSUE #6632: Additional functionality needed for economic analysis of beef enterprises

### LINKED ISSUE #6640: Inheritance of models in APSIM development (e.g. Report)
I am checking on the policy and best practice for inheritance of models in APSIM next gen.

There are clear advantages to inheriting classes and polymorphism in that there is a single, reusable source of code logic and inherited classes can add to this functionality or provide their own implementation. A clean implementation of inheritance also avoids the problem of code changes not being propagated when duplicate code is added to the system to replicate functionality. There can also be disadvantages in keeping the software clean if not done properly.

I have added some inheritance within the CLEM model (individual RuminantFemale is inherited from Ruminant) and there is also need to build on the functionality of some of the APSIM components (e.g. Reports). There are some problems when trying to inherit some models.

Problems

- Models have not been implemented for inheritance with virtual or abstract classes and logic
- The binding of EventSubscribe does not consider base class methods
- You cannot ignore a property of the inherited model in the property presenter if the base class has description attribute (can check for empty description as well as description present to allow inherited description text to determine outcome). Can also be solved by using "new" if the property is not needed, but this is sloppy.

Allow inheritance of Report.cs
CLEM provides users with reports that automatically generate the variables and events list and require some additional properties and presenters. This has been achieved by inheriting the Report model and providing a new "Commencing" method. Some additional changes are needed to fix the need to duplicate a range of methods in the inherited class (e.g. DoOutput, static SortColumnsOfDataTable, FindVariableMembers) which work with the base private variables that are not accessible by the derived class.

lie112:
Implement inheritance for the Report model to remove duplicated text in the custom reports in CLEM that only need an OnCommencing event method to propagate the VariableName and EventNames arrays for the user. They use all other Report logic after initialised. ReportBalances also allows for a CLEM Timer to determine frequency of reporting by overriding the DoOutput event.
Also fixes display of report properties when needed in property presenters of inherited reports

resolves #6640
resolves #6632


lie112:
resolves  #6640

lie112:
I do not know how to fix this issue with the version string in the build.

hol430:
You just need to edit your initial comment and add "resolves xxxx"

lie112:
retest please jenkins

lie112:
How do I fire a rebuild after moving the resolves statement from it's own entry to the initial entry here?

hol430:
Any comment containing the phrase "retest this please jenkins" will do the job (editing an existing comment will have no effect). Github lets you save comments which I find useful for this sort of thing.

image

lie112:
retest this please jenkins

lie112:
Not sure how to get it to retest to fire.

hol430:
Looks like the webhook didn't reach its destination, likely due to some changes I was making to the website this morning. I've rerun the build manually from within jenkins but it didn't update the link on this pull request. You can find the manual build [here](http://apsimdev.apsim.info:8080/jenkins/blue/organizations/jenkins/apsimx/detail/apsimx/670/pipeline).

lie112:
Not sure what's wrong here. Something in POStats and a full disk
_Microsoft.Data.SqlClient.SqlException (0x80131904): Could not allocate space for object 'dbo.VariableData'.'PK_VariableData' in database 'APSIM.POStats' because the 'PRIMARY' filegroup is full. Create disk space by deleting unneeded files, dropping objects in the filegroup, adding additional files to the filegroup, or setting autogrowth on for existing files in the filegroup._

hol353:
retest this please jenkins

Models/CLEM/Reporting/ReportPasturePoolDetails.cs
```diff
@@ -33,31 +33,31 @@ public class ReportPasturePoolDetails: Models.Report
...
-        /// <summary>The columns to write to the data store.</summary>
-        [NonSerialized]
-        private List<IReportColumn> columns = null;
-        [NonSerialized]
-        private ReportData dataToWriteToDb = null;
+        ///// <summary>The columns to write to the data store.</summary>
+        //[NonSerialized]
+        //private List<IReportColumn> columns = null;
+        //[NonSerialized]
+        //private ReportData dataToWriteToDb = null;
```
hol353, line 40: Need to remove commented code

Models/CLEM/Reporting/ReportResourceLedger.cs
```diff
@@ -79,26 +73,6 @@ public class ReportResourceLedger : Models.Report
         [JsonIgnore]
         public new string[] EventNames { get; set; }
```
hol353, line 75: Is this still needed? I don't like using the 'new' keyword like this. Usually indicates that you're overriding something in the base class that wasn't meant to be overridden.

Models/Report/Report.cs
```diff
@@ -31,30 +31,30 @@ public class Report : Model
...
 
         /// <summary>Link to a locator service.</summary>
         [Link]
-        private ILocator locator = null;
+        protected ILocator locator = null;
 
         /// <summary>Link to an event service.</summary>
         [Link]
-        private IEvent events = null;
+        protected IEvent events = null;
```
hol353, line 57: Hmmm. A whole bunch of privates in Report that are now less private. Do these all need to be made protected?

Models/Report/Report.cs
```diff
@@ -368,7 +366,7 @@ private void AddExperimentFactorLevels()
         }
 
         /// <summary>Store descriptors in DataStore.</summary>
-        private void StoreFactorsInDataStore()
+        protected void StoreFactorsInDataStore()
         {
```
hol353, line 370: Do all these methods need to be less private?

Models/Report/Report.cs
```diff
@@ -104,25 +104,23 @@ private void OnFinalInitialise(object sender, EventArgs args)
...
 
-            // Tidy up variable/event names.
-            VariableNames = TidyUpVariableNames();
+            //// Tidy up variable/event names.
+            //VariableNames = TidyUpVariableNames();
 
-            // Locate reporting variables.
-            FindVariableMembers();
+            //// Locate reporting variables.
+            //FindVariableMembers();
```
hol353, line 115: Remove commented code

lie112: Oops, I left this to show what I had moved around in Report only to now realise you know that from Git commits and the commented code isn't then removed. This should now be a clean Report model. I have learned some things along the way,

lie112: Would be great if this could be approved today as I have some fixes in these commits that users are waiting for.

PR #6700, opened 2021-08-11
Fixed 'x is empty' error in p/o graphs

### LINKED ISSUE #6697: Checkpoints not working if not displayed
Create a checkpoint.
Click on predicted/observed graph - has the error message ```x is empty``` and nothing is displayed.
If you then click on the checkpoint and change 'Show on graphs' and go back, it seems to display ok.

Looks like it is creating an error if there are checkpoints that aren't displayed?
Possibly related: Clicking on the datastore and selecting the checkpoint and there is no data displayed in the table.

### LINKED ISSUE #6699: Predicted-observed graphs not displaying
The predicted-observed tables are populated, but the graph has an error saying "x is empty"

hol430:
Resolves #6699
Resolves #6697

This is basically a bandaid. We really need more informative errors in the graphing pipeline, however I don't have time to do that right now.

### DIFF
--- a/ApsimNG/Views/Sheet/PagedDataProvider.cs
+++ b/ApsimNG/Views/Sheet/PagedDataProvider.cs
@@ -224,7 +224,7 @@ namespace UserInterface.Views
             if (string.IsNullOrEmpty(filter))
                 filter = checkpointFilter;
             else
-                filter = $" AND {checkpointFilter}";
+                filter += $" AND {checkpointFilter}";
 
             if (simulationNames != null)
             {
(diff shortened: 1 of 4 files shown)


hol430:
retest this please jenkins

jbrider:
@hol430  That hasn't worked unfortunately. 
On line 99 of the Regression file the checks against the X being null stop it from populating the x and y lists.
I added an additional check to ensure there was data in the x and y lists at the showOneToOne test (line 146) but then I never see any checkpoint data at all.
I think the checkpoint data may be getting cleared out mistakenly - maybe during postprocess?
If I remove the checkpoint and then add a different one then checkpoint data is displayed again.
If I then run the simulations - the checkpoint data disappears.

hol430:
@jbrider - this fixed things for me, which might indicate that we're having different problems. I followed your steps in #6697 and was able to view the graph in wheat example when another checkpoint was present, and also the p/o graphs in the sorghum validation. What problem are you experiencing exactly? Is it just the regression line not showing up? Because that seems to be working for me, with and without extra checkpoints.

image

Issue #6692, opened 2021-08-05
Refactor and implement new list filtering and sorting by properties and attributes


lie112:
After the initial release and user requested improvements made to CLEM it needs a code clean up and some improvements with respect to memory and performance.  This includes better management of individuals (ruminant and labour) throughout in preparation for the new implementation of generic filtering and sorting of lists for activities that allows the user to filter and sort on any property or attributes tags assigned to individuals.

PR #6703, opened 2021-08-11
Refactor generic transmutation to include labour and ruminants

lie112:
working on #6692


Models/Storage/DataStoreReader.cs
```diff
@@ -180,7 +180,17 @@ public void Refresh()
...
             foreach (var tableName in Connection.GetTableAndViewNames())
-                tables.Add(tableName, Connection.GetTableColumns(tableName));
+            {
+//                tables.Add(tableName, Connection.GetTableColumns(tableName));
+                try
+                {
+                    tables.Add(tableName, Connection.GetTableColumns(tableName));
+                }
+                catch
+                {
```
hol430, line 190: Again, we can't just swallow the exception and bury our head in the sand. This could fail for lots of reasons - e.g. if the database is corrupt, or disk IO fails for some other reason, then we need to handle the failure gracefully, and most likely report it to the user. In order to do that, the exception needs to be allowed to propagate up the call stack. This is the same problem as #6627, which also shouldn't have been merged as it also implemented this pattern in the datastore.

If this exception is being caught too late (resulting in a failure to open the file in the GUI), that's a bug and should be corrected.

lie112: Fully agree, this wasn't meant to be committed. I thought I had discarded this chuck in an accidental inclusion of this file.
Actually the change that I have still sitting in SourceTree is the full removal of this block. So lets try again.
I think @hol430 had taken this issue on to find a better solution using the example files provided that generate the error.

hol430 (approved):
Fine to merge @hol353

PR #6704, opened 2021-08-11
Add default save file extension only on Windows

### LINKED ISSUE #6660: Unable to overwrite .apsimx file after clicking "save as"
1. Open any .apsimx file in the gui
2. Click save as
3. Select an existing .apsimx file
4. When prompted, click "replace" to overwrite the existing file

At this point, apsim should overwrite the specified file and close the file selection dialog, however, the file dialog seems to reopen, and the selected file is not changed.

zur003:
Resolves #6660 

Things still don't work as well as I'd like, but I haven't been able to come up with anything better.

### DIFF
--- a/ApsimNG/Utility/FileDialog.cs
+++ b/ApsimNG/Utility/FileDialog.cs
@@ -202,7 +202,7 @@
 
                 
                 tryAgain = false;
-                if (Action == FileActionType.Save && fileChooser.Filter != allFilter && specParts != null)
+                if (ProcessUtilities.CurrentOS.IsWindows && Action == FileActionType.Save && fileChooser.Filter != allFilter && specParts != null)
                 {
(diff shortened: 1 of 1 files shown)


hol430:
@zur003 - perhaps I'm misunderstanding the problem, but when I just tried this on linux; I opened the wheat example, clicked save as, changed the filter to "all files", clicked save, and `fileChooser.Filter.Name` was set to "All files", as one might expect. Is this different to what you're seeing?

(For what it's worth, I'm running the gtk3 version, the mono version may well be different, as I don't think we use FileChooserNative there).

zur003:
@hol430 - that's interesting. For me, fileChooser.Filter is null, unless I set it programmatically (in the .Net core /Gtk3 version).

PR #6708, opened 2021-08-12
Files should open in GUI despite errors in .db file

### LINKED ISSUE #6707: Error in database should not cause .apsimx file to fail to open in GUI
If reading the database causes an exception, the .apsimx file should still be opened and usable in the GUI. Currently, any such errors will cause the file opening operation to fail.

hol430:
Resolves #6707


hol353 (approved):
Looks good to me.

Issue #6529, opened 2021-06-07
Develop an apsim file runner/server tool


hol430:
Running a parameter optimisation/sensitivity analysis typically requires an .apsimx file to be run many times serially. Currently this can be done only by repeatedly invoking Models.exe, with all of the overhead that entails (disk IO, model deserialization, connection of links/events, factor initialisation, etc). The proposed solution to this problem is to implement a "server" application which reads an .apsimx file once, and performs all of this initialisation once. It will then hold the simulations in memory, ready to be run on-demand with minimal overheads.

lie112:
Is this also applicable to the experiment with factors simulations, or is it the same thing to perform sensitivity analysis?

hol430:
This will work fine if you have an experiment in your file - it will rerun the whole file each time it receives a "run" command.

BrianCollinss:
@hol430 Hi Drew. What does the user need to do to be able to use this server thing? Does APSIM dceicde when to use this when it faces many simulations?

BrianCollinss:
Thanks. That's what I meant, e.g. running a factorial experiment with one simulation as the base and changing the parameters many times. Is there any instruction on how to use this?

hol430:
Yeah, there are some instructions [here](https://apsimnextgeneration.netlify.app/usage/server/). The command-line arguments section may be slightly out of date, I'm not sure. Running with `--help` should give you a full list of options with help text, otherwise feel free to ask here if anything doesn't make sense.

PR #6710, opened 2021-08-13
Working on Fast APSIM Runner Tool (FART)

hol430:
Working on #6529


hol353 (approved):
This is fine. Very socket centric which I guess you need for your use-case.

hol430:
Yeah - sockets were really the only use case I had in mind; that being said, the relevant connection and message protocol code is all hidden behind interfaces, so it shouldn't be too hard to refactor this to add in another connection type or message protocol.

PR #6713, opened 2021-08-16
Get a working mac installer for the .net core builds

hol430:
Working on #6139 

This installer is entirely self-contained, so the user doesn't need to install gtk+3, gtksourceview4, or the .net runtime. The resultant .dmg installer is about 150MB compressed, or ~400MB once installed (by comparison, the windows and linux installers are about half this). Later down the road, we could perhaps implement a lightweight version of this installer, which would require the user to install the dependencies themselves.

The native gtk dependencies, along with instructions for rebuilding them, are committed [here](https://github.com/hol430/apsim-mac). This repo contains a tarball of the native libraries (30MB compressed) which is downloaded by the mac install script. The .net runtime is automatically bundled with the application if you don't supply the `--no-self-contained` switch when running a `dotnet publish` command.

The launcher script included in this PR is what will be run when a mac user runs their installed apsim. It initialises the runtime environment so that apsim is able to locate the bundled gtk libraries.

### DIFF
--- a/Models/Properties/AssemblyVersion.cs
+++ b/Models/Properties/AssemblyVersion.cs
@@ -1,5 +1,5 @@
 using System.Reflection;
-[assembly: AssemblyVersion("1.2.3.4")]
-[assembly: AssemblyFileVersion("1.2.3.4")]
+[assembly: AssemblyVersion("0.0.0.0")]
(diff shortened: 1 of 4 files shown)


hol353 (approved):
looks good to me.

PR #6715, opened 2021-08-16
Change Summary class to write messages to storage immediately

### LINKED ISSUE #6714: Ability to write to Summary in OnCompleted
Relates to #6586  at end of conversation

https://github.com/APSIMInitiative/ApsimX/issues/6586#issuecomment-892269363

I would like to be able to trap errors generated while trying to create SQL queries in OnCompleted and present them in the Summary and keep going ignoring the bad sql and reporting all issues at the end. This is because I need to be assured that all reports have been created and written before creating SQL views that may include other report tables. Unfortunately, the summery cannot write in OnCompleted

I don't mind which approach suggested by @hol430  is used.

hol430:
(rather than in bulk in the Completed event).

@hol353 let me know what you think of this one. I'm not 100% sold on this approach, but am not sure what the alternative is.

This fixes #6714.


Models/Summary.cs
```diff
@@ -154,6 +154,7 @@ public void WriteMessage(IModel model, string message)
                 newRow[3] = message;
                 newRow[4] = Convert.ToInt32(Simulation.ErrorLevel.Information);
                 messages.Rows.Add(newRow);
+                storage.Writer.WriteTable(messages.Copy(), false);
             }
```
hol353, line 158: Ok this seems strange. When summary wants to write a message to the data store, it deletes the previous messages, it writes the new message to a DataTable, then sends a copy of the DataTable (why a copy?) to the DataStoreWriter. Yes? Perhaps we can get rid of the messages table from Summary and introduce a new method into the DataStoreWriter that summary just calls when it wants to store a message. The method needs to be generic in that it could be used to write to any table in the .db. I'm not sure what the method arguments would be. Perhaps it just uses the overload of DataStoreWriter.WriteTable that accepts a ReportData?

hol353 (changes requested):

Models/Summary.cs
hol430: Yep, that would be better - I'll go and do that. Only reason I didn't do that initially was to keep this self-contained.

The reason behind `messages.Copy()` is because the `WriteTable()` function actually mutates the table you pass in - it adds in a SimulationID column(!). I guess this is the price we pay for our (over)use of mutable/reference types...

PR #6717, opened 2021-08-16
Add more helpful error message for invalid event handler delegate types

### LINKED ISSUE #6716: Useless error message when event callback has incorrect number of arguments
Writing an event handler (ie in a manager) with an event handler with 0 arguments leads to this error message, which is correct but less than helpful to most users:

```
Cannot bind to the target method because its signature is not compatible with that of the delegate type.
```

hol430:
Resolves #6716


hol353 (approved):
This looks fine.

PR #6719, opened 2021-08-16
adding facts data 

sarchontoulis:
working on #6725

PR #6724, opened 2021-08-17
Full refactor of all CLEM Activities to clean code and remove duplicate code

lie112:
Working on #6692


Models/CLEM/Activities/ActivityTimerPastureLevel.cs
```diff
@@ -25,7 +25,7 @@ namespace Models.CLEM.Activities
     [Description("This activity timer is used to determine whether a pasture biomass (t/ha) is within a specified range.")]
     [HelpUri(@"Content/Features/Timers/PastureLevel.htm")]
     [Version(1, 0, 1, "")]
-    public class ActivityTimerPastureLevel : CLEMModel, IActivityTimer, IValidatableObject, IActivityPerformedNotifier
+    public class ActivityTimerPastureLevel : CLEMModel, IActivityTimer, IActivityPerformedNotifier
     {
         [Link]
         ResourcesHolder Resources = null;
```
hol353, line 31: Private fields should start with lowercase

Models/CLEM/Activities/CropActivityManageCrop.cs
```diff
@@ -29,6 +29,9 @@ public class CropActivityManageCrop: CLEMActivityBase, IValidatableObject, IPast
         [Link]
         Clock Clock = null;
 
+        private bool gotLandRequested = false; //was this crop able to get the land it requested ?
+        private int CurrentCropIndex = 0;
+
```
hol353, line 34: Start with a lowercase letter

Models/CLEM/Activities/CropActivityManageProduct.cs
```diff
@@ -28,10 +28,14 @@ public class CropActivityManageProduct: CLEMActivityBase, IValidatableObject
     {
         [Link]
         Clock Clock = null;
-
         [Link]
         Simulation Simulation = null;
```
hol353, line 33: And some more privates that start with upper case.

Models/CLEM/Activities/LabourActivityFeedToTargetsV1.cs
```diff
@@ -355,33 +355,33 @@ public override List<ResourceRequest> GetResourcesNeededForinitialisation()
...
+        //public override event EventHandler ActivityPerformed;
+
+        ///// <summary>
+        ///// Shortfall occurred 
+        ///// </summary>
+        ///// <param name="e"></param>
+        //protected override void OnActivityPerformed(EventArgs e)
+        //{
+        //    ActivityPerformed?.Invoke(this, e);
+        //}
```
hol353, line 385: Remove commented code

Models/CLEM/Activities/PastureActivityCutAndCarry.cs
```diff
@@ -29,6 +29,10 @@ public class PastureActivityCutAndCarry : CLEMRuminantActivityBase
         [Link]
         Clock Clock = null;
```
hol353, line 31: Another private starting with upper case. Please check all CLEM files for this.

hol353 (changes requested):

Models/CLEM/Activities/ActivityTimerPastureLevel.cs
lie112: How does [Link] work? Does it search for the Type up the tree. I thought SV told be it was searching for a Named component of type and therefore the name had to match. I thought it was odd there were caps elsewhere. I'll add this to my list of things to fix up as I go through the rest of the code.

hol430: The link attribute has a Type property which determines how it searches for a model. By default, it's set to LinkType.Scoped, so it will search for a model in scope. No matter which link type you use however, the target must always be of a matching (or more derived) type. The name of the property is only used if `ByName` is set to true on the link attribute. So

```csharp
// Resolves to the first model of type ResourcesHolder in the simulations tree (in a depth-first search)
[Link]
// [Link(Type = LinkType.Scoped, ByName = false)] // Equivalent
ResourcesHolder Resources = null;

// Resolves to the first model of type ResourcesHolder in the simulations tree (in a depth-first search)
// which has a name of hello
[Link(ByName = true)] ResourcesHolder hello = null;
```

hol353: In your second source line I think it is case insensitive so it will find a ResourcesHolder with name hello or Hello.

hol430: Or hElLo :). I guess this should all probably be documented on the netlify site. In fact there are quite a few comments on github which could probably be useful as a page on that site...

lie112:
Seems there's a validation error to do with Tests/Validation/System/FACTS_CornSoy

hol430:
There may be similar recurring problems until #6719 is merged...just ping me if you spot it happening again.

retest this please jenkins

lie112:
This is ready to go and needed to the next release of improved filtering

Models/CLEM/Activities/ActivitiesHolder.cs
```diff
@@ -227,17 +209,37 @@ private void OnCLEMEndOfTimeStep(object sender, EventArgs e)
...
+        [EventSubscribe("Completed")]
+        private void OnSimulationCompleted(object sender, EventArgs e)
         {
-            // clear the activity performed status at start of time step
-            foreach (CLEMActivityBase child in FindAllChildren<CLEMActivityBase>())
-                child.ClearAllAllActivitiesPerformedStatus();
+            UnBindEvents(FindAllChildren<IModel>());
         }
 
+        #region validation
```
hol430, line 221: Can the regions be removed?

Models/CLEM/Activities/ActivitiesHolder.cs
```diff
@@ -227,17 +209,37 @@ private void OnCLEMEndOfTimeStep(object sender, EventArgs e)
             OnActivityPerformed(ea);
         }
 
-        /// <summary>An event handler to allow us to initialise ourselves.</summary>
-        /// <param name="sender">The sender.</param>
-        /// <param name="e">The <see cref="EventArgs"/> instance containing the event data.</param>
-        [EventSubscribe("CLEMStartOfTimeStep")]
-        private void OnCLEMStartOfTimeStep(object sender, EventArgs e)
+        /// <summary>
+        /// Overrides the base class method to allow for clean up
```
hol430, line 213: This is kind of minor, but this description seems to be outdated, as this method doesn't override anything. (Also doesn't contain <param> tags for arguments)

Models/CLEM/Activities/CLEMActivityBase.cs
```diff
@@ -22,6 +22,16 @@ namespace Models.CLEM.Activities
     [Version(1, 0, 1, "")]
     public abstract class CLEMActivityBase: CLEMModel
     {
+        /// <summary>
+        /// A public link to the CLEM resource holder
+        /// </summary>
+        [Link]
+        public ResourcesHolder Resources = null;
```
hol430, line 29: This isn't really an issue per se, but why do we need a public link here? We usually try to avoid this elsewhere - usually it's not necessary.

Models/CLEM/Activities/CropActivityManageProduct.cs
```diff
@@ -27,10 +27,14 @@ namespace Models.CLEM.Activities
     public class CropActivityManageProduct: CLEMActivityBase, IValidatableObject
     {
         [Link]
-        Clock Clock = null;
-
+        Clock clock = null;
```
hol430, line 30: These should be explicitly declared as private.

Models/CLEM/Activities/ActivitiesHolder.cs
lie112: I think this has arisen from very early on and is a 3 year copy and paste error from SV original code. This is likely to be the case for every Subscribed event handler in CLEM. This will be a big job to fix that I don't really have time to do right now. Can I put it on my list of things to do?

Models/CLEM/Activities/ActivitiesHolder.cs
lie112: I guess so. I have used these two regions #validation and #descriptive summary throughout CLEM to separate the validation and description logic from the core model. They are more to do with the UI and experience. I didn't realise there was a aversion to #region

Models/CLEM/Activities/CLEMActivityBase.cs
lie112: This allows one Link here, and all Activities that inherit this base have access to this Model used widely in CLEM. It saves many other [Link] statements being required in every activity Model. I think there are 3 or so cases where this happens in CLEM. Another is the Herd Model that all CLEMRuminantActivityBase classes can see.

Models/CLEM/Activities/PastureActivityBurn.cs
```diff
@@ -25,6 +25,11 @@ namespace Models.CLEM.Activities
     [HelpUri(@"Content/Features/Activities/Pasture/BurnPasture.htm")]
     public class PastureActivityBurn: CLEMActivityBase
     {
+        private GrazeFoodStoreType pasture { get; set; }
```
hol430, line 28: These should be fields since they're private.

Models/CLEM/Activities/PastureActivityCutAndCarry.cs
hol430: These should also be declared private explicitly - there are quite a few other instances of this which I've seen while skimming but didn't comment on - they should also be fixed.

Models/CLEM/Activities/PastureActivityCutAndCarry.cs
```diff
@@ -27,7 +27,11 @@ namespace Models.CLEM.Activities
     public class PastureActivityCutAndCarry : CLEMRuminantActivityBase
     {
         [Link]
-        Clock Clock = null;
+        Clock clock = null;
+
+        private GrazeFoodStoreType pasture { get; set; }
+        private AnimalFoodStoreType foodstore { get; set; }
```
hol430, line 33: More private properties which should be fields.

Models/CLEM/Activities/RuminantActivityControlledMating.cs
```diff
@@ -252,45 +249,16 @@ public IEnumerable<RuminantFemale> BreedersToMate()
         }
 
         /// <summary>
-        /// Private method to determine resources required for this activity in the current month
+        /// Provate method to determin resources required for this activity in the current month
```
hol430, line 252: Typos?

Models/CLEM/Activities/RuminantActivityGrazeAll.cs
```diff
@@ -143,13 +136,8 @@ private void GrazeAll_ResourceShortfallOccurred(object sender, EventArgs e)
...
         {
-            ActivityPerformed?.Invoke(sender, e);
-        }
-
-        /// <inheritdoc/>
-        public override List<ResourceRequest> GetResourcesNeededForActivity()
-        {
-            return null;
+            OnActivityPerformed(e);
+            //ActivityPerformed?.Invoke(sender, e);
```
hol430, line 140: Commented code - please remove

Models/CLEM/Activities/RuminantActivityGrazePasture.cs
```diff
@@ -196,7 +196,8 @@ private void Paddock_ResourceShortfallOccurred(object sender, EventArgs e)
 
         private void BubbleHerd_ActivityPerformed(object sender, EventArgs e)
         {
-            ActivityPerformed?.Invoke(sender, e);
+            OnActivityPerformed(e);
+            //ActivityPerformed?.Invoke(sender, e);
```
hol430, line 200: Commented code -please remove

In fact, if this is duplicated from the above class, might be worth thinking about how to reuse, rather than duplicate code, although it's probably beyond the scope of this PR.

Models/CLEM/FileSQLitePasture.cs
```diff
@@ -33,11 +33,8 @@ namespace Models.CLEM
     [HelpUri(@"Content/Features/DataReaders/PastureDataReaderSQL.htm")]
     public class FileSQLitePasture : CLEMModel, IFilePasture, IValidatableObject
     {
-        /// <summary>
-        /// A link to the clock model.
-        /// </summary>
         [Link]
-        private readonly Clock clock = null;
+        private  Clock clock = null;
```
hol430, line 37: Very minor - but this double space should be single

Models/CLEM/Reporting/ReportQuery.cs
```diff
@@ -57,14 +55,17 @@ public DataTable RunQuery()
         [EventSubscribe("Completed")]
         private void OnCompleted(object sender, EventArgs e)
         {
-            if(!SaveView(dataStore))
+            string errorMsg;
+            if(!SaveView(dataStore, out errorMsg))
             {
-                summary.WriteWarning(this, $"Invalid SQL: Unable to create query report [{this.Name}] using SQL provided \r\nIf your SQL contains links to other ReportQueries you may need to run this Report after the others have been created.");
+                throw new ApsimXException(this, $"Invalid SQL: Unable to create query report [{this.Name}] using SQL provided\r\nError: {errorMsg}\r\nIf your SQL contains links to other ReportQueries you may need to run this Report after the others have been created by disabling it in the first run and then enabling again.");
+                //summary.WriteWarning(this, $"Invalid SQL: Unable to create query report [{this.Name}] using SQL provided\r\nIf your SQL contains links to other ReportQueries you may need to run this Report after the others have been created.");
```
hol430, line 62: Commented code - please remove

Models/CLEM/Activities/RuminantActivityGrazeAll.cs
lie112: I wasn't sure if some big changes here were going to come back and bite me. Gone now

Models/CLEM/Activities/RuminantActivityGrazePasture.cs
lie112: I realise these two activities use some shared code to create versions of RuminantActivityGrazePastureHerd for running the simulation. It is only in these two Models.

Models/CLEM/FileSQLitePasture.cs
lie112: Thanks, I reckon CLEM might now have the cleanest Model code. I would love to use the apply to all in Folder, but it only exists for Document, Project and Solution, which doesn't help me.

Models/CLEM/Resources/SpecifyRuminant.cs
```diff
@@ -66,13 +62,17 @@ public class SpecifyRuminant : CLEMModel, IValidatableObject
...
         /// <summary>
         /// The ruminant type for this specified ruminant
         /// </summary>
         public RuminantType BreedParams { get { return ruminantType;} }
 
+        /// <summary>
+        /// Records if a warning about set weight occurred
+        /// </summary>
+        [JsonIgnore]
+        public bool WeightWarningOccurred = false;
```
hol430, line 74: Public - should be a property.

Models/CLEM/Reporting/ReportQuery.cs
lie112: This line will be reinstated when @hol430 fixes the datastore error bubbling and allows immediate Summary message writing. I don't want an error here, but report to summary so all similar errors are reported rather than stopping here.

Models/CLEM/Resources/WaterType.cs
```diff
@@ -22,6 +22,9 @@ namespace Models.CLEM.Resources
     [HelpUri(@"Content/Features/Resources/Water/WaterStoreType.htm")]
     public class WaterType : CLEMResourceTypeBase, IResourceWithTransactionType, IResourceType
     {
+        private double amount { get { return roundedAmount; } set { roundedAmount = Math.Round(value, 9); } }
```
hol430, line 25: If private - should be a field. If you need the rounding logic, perhaps change that to a method.

Models/CLEM/Activities/ActivitiesHolder.cs
hol430: https://apsimnextgeneration.netlify.app/development/software/codingstyle/#region-usage

Models/CLEM/Activities/ActivitiesHolder.cs
hol430: Feel free.

Models/CLEM/Activities/CLEMActivityBase.cs
hol430: Perhaps make it protected then, if it's only accessed from derived types?

Models/CLEM/Reporting/ReportQuery.cs
hol430: Makes sense - perhaps add a comment to that effect?

Models/CLEM/Activities/ActivitiesHolder.cs
lie112: I think it came from the original copy of the Method in Clock.OnSimulationCommencing

lie112:
I assume the .vs folder is not part of the repo. I had to delete my .vs folder to fix issues with Visual Studio. Now it wants to commit the deletes. Can I add ignore .vs folder to the Git checklist? Is that a shared file, or local to my Git repo?

hol430:
The .vs directory shouldn't be part of the repo, although someone must have committed some files inside it at some stage. To undo any local changes so that they don't show up in sourcetree, you should be able to delete the .vs directory from your computer, then go to sourcetree and those files will show up as deleted. I think you can right click them and choose "discard" (or similar), and you should see that the .vs directory exists once more (and sourcetree will show no changes).

Models/CLEM/Activities/ActivitiesHolder.cs
lie112: Yeah, I know. What about all the other models in APSIM (Utilities, ApsimNG, Pasture, Aqua, Functions, GRange.......) all using # region? Just do a quick find all. It's widely used so I thought it was accepted. Happy for them to go when you do a remove all from Solution.

lie112:
Hmm, error to be addressed by @hol430 .

hol430:
retest this please jenkins

hol430 (approved):
@hol353 I think this is good to merge. This PR makes the clem code much more compliant with the code standards.

hol353 (approved):

Issue #6727, opened 2021-08-18
expand QPriority allocation method to work for N as well as DM


HamishBrownPFR:
Fodder beet model needs alternative way of arbitrating N to improve partitioning.  QPriority method seems to be a good approach so implement this for N.  The approach to passing QPriorities could be tidied up and generalised also.

HamishBrownPFR:
This was finished with Arbitrator priorities #6739

PR #6728, opened 2021-08-18
Add priority factors into Biomass pool type

HamishBrownPFR:
working on #6727


Models/PMF/Interfaces/IArbitration.cs
```diff
@@ -90,16 +90,28 @@ public class BiomassPoolType : Model
...
+        public double QStructuralPriority { get; set; }
+        /// <summary>Gets or sets Storage Priority.</summary>
+        /// <value>The Storage Priority.</value>
+        [Description("Initial Storage biomass priority")]
+        public double QStoragePriority { get; set; }
+        /// <summary>Gets or sets the metabolic biomass priority.</summary>
+        /// <value>The metabolic.</value>
+        [Description("Initial Metabolic biomass priority")]
+        public double QMetabolicPriority { get; set; }
+
```
hol353, line 113: Hmmm. I'm unsure why you need 3 more properties in this organ arbitrator class. Is there another way to do this? @hut104 Can you have a look at this? Is it worth a quick chat about how you're going to implement this?

jbrider:
@HamishBrownPFR 
We have priority/order considerations included within the methods for sorghum arbitration.
Would be interested in working together on something more generic.

HamishBrownPFR:
Sorry, should have put a do not merge on this as it is far from complete. 
@hol353.  The intention here is not to add new properties it is to move the DMDemandPriorityFactor (In IArbitraton) into the BiomassPoolType.    This PR was just testing to see what would happen if these properties are added to BiomassPoolType but are not set.   Seems it causes no problems.  This will mean I can tidy the implementation of the DMDemandPriorityFactors up quite a bit.  At the moment all classes implementing IArbitration implement the DMDemandPriorityFactor in their code but only a few have a function to set its value and this is an optional function.  Now I will be able to get rid of DMDemandPriorityFactor from IArbitration interface and pass the same variables in DMDemand which is a BiomassPoolType.  This tests confirms this will not cause problems for other uses of BiomassPoolType that do not need to set the QPriorityFactors.  Net result will be less code and removal of an optional function.  It will also declutter the GUI layout of Organs as these parameters will no longer be at the top level (for organs that use them) but tucked away under the DMDemand Function.
@jbrider Keen to work together on this but very rusty at the moment so just doing some basic refactoring to get my head back into the arbitrator.  Intent at this stage is not to add any new science code, just to improve the implementation of the QPriorityThenRelativeAllocation arbitration method that @junqi108 added for grapevine as I am keen to test this for some other models.

@jbrider @hol353 @hut104 @hol430 I don't think we have got the design right for the organ arbitrator.  From a software point of view it is great because it allows flexibility and provides plenty of code reuse.  However, it is really hard to follow the flow of calculations through all the different classes when debugging and nearly impossible for non programming model developers to get their head around how it works.  Suggest we have a VC to discuss this and see if there is a better way to do it

jbrider:
@HamishBrownPFR I know what you mean by rusty! Getting back into it has taken a little while.
I agree about the arbitrator - it works but it's not easy to follow.

HamishBrownPFR:
retest this please Jenkins

HamishBrownPFR:
@hol353 @hol353 Do either of you guys understand the error I am getting here.?  Something to do with the performance test system

hol353:
retest this please Jenkins.

@HamishBrownPFR We fixed the problem shown in this PR a short while ago. Just need to retest this and it should run.

HamishBrownPFR:
thanks @hol353.  It is giving an error I can do something about now.

HamishBrownPFR:
@hol353 this is ready for review now

hol353 (approved):

Issue #6677, opened 2021-08-03
Get unit tests running on .net core


hol430:
The unit tests project should be ported to .net core in anticipation of dropping support for the .net framework builds.

PR #6740, opened 2021-08-24
Refactor sheet view

hol353:
Working on #6677


ApsimNG/Views/Sheet/DefaultCellPainter.cs
```diff
@@ -36,66 +41,67 @@ public bool PaintCell(int columnIndex, int rowIndex)
...
+        public (int Red, int Green, int Blue) GetForegroundColour(int columnIndex, int rowIndex)
         {
 
             if (Utility.Configuration.Settings.DarkTheme)
             {
 
 
                 if (rowIndex < sheet.NumberFrozenRows)
-                    return new Cairo.Color(1, 1, 1); // white
+                    return (255, 255, 255); // white
```
hol430, line 70: We probably shouldn't be hardcoding colours in the source like this. Although I guess fixing this is probably beyond the scope of this PR.

ApsimNG/Views/Sheet/Sheet.cs
```diff
@@ -0,0 +1,534 @@
...
+
+                    cr.Rectangle(cellBounds.ToClippedRectangle(Width-20, Height));
+                    cr.Clip();
+
+                    cr.SetLineWidth(lineWidth);
+
+                    cr.Rectangle(cellBounds);
+                    if (!CellPainter.PaintCell(columnIndex, rowIndex))
+                    {
+                        //cr.SetSourceColor(CellPainter.GetForegroundColour(columnIndex, rowIndex));
```
hol430, line 482: Commented code should be removed.

ApsimNG/Views/Sheet/SheetWidget.cs
```diff
@@ -20,242 +20,39 @@ namespace UserInterface.Views
...
         /// <summary>The width of the grid lines in pixels.</summary>
         private const double lineWidth = 0.2;
 
         /// <summary>The number of rows to scroll (up or down) on mouse wheel.</summary>
         private const int mouseWheelScrollRows = 10;
 
-        /// <summary>A backing field for NumberHiddenColumns property.</summary>
-        private int numberHiddenColumnsBackingField;
+        ///// <summary>A backing field for NumberHiddenColumns property.</summary>
+        //private int numberHiddenColumnsBackingField;
```
hol430, line 33: More commented code

ApsimNG/Views/Sheet/SheetWidget.cs
```diff
@@ -359,7 +143,24 @@ protected override bool OnKeyPressEvent(EventKey evnt)
         {
             try
             {
-                KeyPress?.Invoke(this, evnt);
+                var keyParams = new SheetEventKey();
+                if (evnt.Key == Gdk.Key.Left)
```
hol430, line 147: I wonder if this would be better off as a utility/extension function somewhere which translates a gdk key into a Keys value.

ApsimNG/Views/Sheet/SingleCellSelect.cs
```diff
@@ -58,29 +63,29 @@ public void GetSelection(out int columnIndex, out int rowIndex)
         /// <summary>Invoked when the user presses a key.</summary>
         /// <param name="sender">Sender of event.</param>
         /// <param name="evnt">The event data.</param>
-        private void OnKeyPressEvent(object sender, EventKey evnt)
+        private void OnKeyPressEvent(object sender, SheetEventKey evnt)
         {
             if (Editor == null || !Editor.IsEditing)
             {
-                if (evnt.Key == Gdk.Key.Right && evnt.State == ModifierType.ControlMask)
+                if (evnt.Key == Keys.Right && evnt.Control)
```
hol430, line 70: I don't think it's worth changing this in this PR, but we use this pattern in a few other classes in ApsimNG, I wonder if we should look for a better way to achieve what we want here (e.g. a `Dictionary<Keys, Action>`). This is difficult to read/parse (for me, anyway).

hol430 (changes requested):

Issue #6058, opened 2021-01-21
Pinus


sme016:
Improving the Pinus model

sarahcleary:
@APSIMInitiative/forestry-and-agroforestry - FYI

sno036:
@sme016  - @hol353 and I just discussed this to get it moving for you. Can you disable (leave it in the file but get it crossed out) the grazing test for Pine and do a pull request for it? Dean is a little way away from having the new way of specifying grazing sorted and the grazing test can be reenabled after that.

@hol430  - see above. Once Philip has disabled the grazing test the simulation should run to completion and then can you move it into release (Philip doesn't know what to do there). If you need additional information @hol353 will assist. Some urgency as there is a training workshop soon. 

OK everyone?

hol430:
If you want, I can disable the grazing test and move this into the release in one fell swoop.

sme016:
I'd be happy to remove grazing, but that is unnecessary as that (pre-)test didn't make it into the the Pinus validation that I committed. The Pinus validation and example simulations currently released are good for now, as far as I am concerned. I'll be happy to put a stock sensibility test into the validation after we get that working properly. Thanks

jbrider:
Apologies @sme016 I told you via email that I would follow this up and I got sidetracked.
Thanks @sno036 for following up.
As all of the requests have been addressed and people have had time to comment I think this is now right to be put into release.

PR #6746, opened 2021-08-26
Pinus modifications to address review comments and finalise for release

sme016:
Copied TreeThinning and modified as TreeThinningFromAboveOrBelow and included it unused in the BFG simualtion for the validation and used as a modified Control treatment in the example.

Working on #6058

### DIFF
--- a/Examples/Pinus.apsimx
+++ b/Examples/Pinus.apsimx
@@ -1,8 +1,8 @@
 {
   "$type": "Models.Core.Simulations, Models",
   "ExplorerWidth": 300,
-  "Version": 136,
-  "ApsimVersion": "2021.7.6.6551",
+  "Version": 139,
+  "ApsimVersion": "2021.8.25.6741",
   "Name": "Simulations",
   "Children": [
     {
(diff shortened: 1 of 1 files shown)


hol430 (changes requested):
@sme016 - Hi Phil, I see you've committed the Pinus.pdf autodocs file. Normally we don't commit this - you'll need to delete this (e.g. using your file explorer), then it will show up in sourcetree as deleted. You can then stage/add this change, commit and push. Let me know if it gives you any problems.

sme016:
Now removed.
I put it there, as Jason was surprised it wasn’t there for the review. Anyway, it’s not needed now and anyone can generate it when needed.

Thanks,
Philip

hol430 (approved):

hol353:
@sme016 Why have the stats changed? Were you expecting changes?

sme016:
I extended (Clock) the simulations for the full set of observations available for the few sites where I had previously (for the review) cut them short due to issues of disease, or very low stocking – as requested by Hamish. This led to a deterioration in some stats, but I didn’t think enough of a deterioration to warrant filtering some OvP graphs to the former date range.

I also included comments that warn folk about this issue for the sites concerned, i.e. IMPAC and Puruki.

Hopefully that is ok with you and for review purposes, but let me know if more work is needed.

PR #6754, opened 2021-08-30
Performance improvements for aggroforestry sims

### LINKED ISSUE #6753: Perennial simulations are slow
Simulations with tree model (e.g. eucalyptus) are slow to run. Profiling and refactoring are needed in order to increase execution speed.

hol430:
...in order to avoid re-calculating these every time the Live or Dead property is accessed.

Resolves #6753.


hol430:
@hol353 - I don't think that using SleepJobs instead of EmptyJobs is going to work. Can discuss this with you tomorrow, but basically the post-simulation tools are only run if `InvokeJobCompleted` is called. If the last job that gets run happens to be a sleep job, the post-simulation tools don't get run  (this occurs in one of our unit tests, and also when you refresh the datastore in the gui). The other problem is that NumJobsComplete is incorrect when using SleepJobs, as it doesn't get incremented. This is easier to solve, but does require us to acquire an extra lock.

I think we will need to find another approach to this issue.

hol353:
Hmmm this is too messy. We shouldn't have either SleepJob or EmptyJob. Both are hacks. Time for a redesign of the JobRunner?

hol430:
@hol353 this is now green and could probably be merged. Do we want to close the underlying issue, or were there more problems you wanted to address?

APSIM.Shared/JobRunning/JobRunner.cs
```diff
@@ -177,10 +177,9 @@ protected virtual void RunActualJob(IRunnable job, IJobManager jobManager)
                     lock (runningLock)
                     {
                         SimsRunning = SimsRunning.Add(job);
+                        var startTime = DateTime.Now;
```
hol353, line 180: What is this? Surely it isn't needed.

Prototypes/Canola/Canola.apsimx
```diff
@@ -8920,8 +8920,6 @@
               "Propertys": [
                 "[Root].Live",
                 "[Root].Dead",
-                "[Nodule].Live",
-                "[Nodule].Dead"
               ],
```
hol353, line 8923: Why has canola.apsimx changed? Why has nodule been removed?

hol353 (changes requested):

hol353:
Let's close the underlying issue once this is merged. Can you also create a new issue to rethink how a JobRunner is kept 'awake'. In other words can we get rid of EmptyJob and SleepJob?

APSIM.Shared/JobRunning/JobRunner.cs
hol430: I think I added a few lines into this lock but that change has been partially reverted somewhere along the line. I'll revert this particular hunk as well.

Prototypes/Canola/Canola.apsimx
hol430: Canola doesn't have a nodule. Previously these were being ignored. With the new biomass they cause errors.

hol353 (approved):

Issue #6763, opened 2021-09-01
Tidy up Arbitrator and Organs


HamishBrownPFR:
The OrganArbitrator and associated classes as well as all the classes implementing IArbitration (organs) hold a lot of code that provides a lot of flexibility for configuring biomass supplies, demands and partitioning.  However, most of the options provided are never used, there is a lot of repetition and a lot of things that should be done consistently between different plants and organs are not.  
Having built and worked with others to build many different plant models in PMF I am pretty sure we can tidy up and improve our plant models if we have a much simpler (in its programming not functionality) arbitrator and a single organ type that it communicates with.  Alternative arbitration approaches could be achieved by writing alternative arbitrator classes which may duplicate code but minimizes the amount of code that any specific plant model is using so makes it easier to understand, configure, document and debug biomass partitioning.

HamishBrownPFR:
I have completed programming a prototype arbitrator and organ class (https://github.com/PlantandFoodResearch/ApsimX/tree/SimpleOrganArbitrator). It still needs lots of testing and debugging but I have gone far enough to know the structure will work and get some more review and opinion on the approach.
Quick summary:
* A prototype implementation of the approach is in \ApsimX\Prototypes\SimplifiedOrganArbitrator\FodderBeetTest.apsimx
* A new arbitrator class (BiomassArbitrator.cs) has been created that applies only the QPriority partitioning approach but existing alternatives could be substituted in with relative ease.
* A new organ class has been created (Organ.cs) that is used to represent all organs and contains code and data structures for representing each organs biomass states and deltas.
* Any other specific organ functionality is added through additional child classes.  For instance a EnergyBalance.cs child added to the organ named "Leaf" so it has a canopy to interface with Microclimate.  The organ named "Root" has a RootNetwork.cs child which interacts with the SoilArbitrator to do Water and N uptake.  
* Other functionality is created by adding IFunction children.
* BiomassArbitrator.cs and Organ.cs currently deal with C and N but have been programmed in a way that P and K can be easily added following the same pattern as N.

HamishBrownPFR:
@jbrider @hol353 @hol430 @hut104.
I have got this working well enough now to know it will work.  There are still some bugs that need tracking down but I think it is ready for doing a code review.  
We should take this opportunity to get naming conventions and the namespace and folder locations of classes right.  Once this is done we will need a converter.
[...]

PR #6764, opened 2021-09-01
Simple organ arbitrator

HamishBrownPFR:
working on #6763.

This will fail as I have renamed some types but not written a converter.   Testing to see if anything else fails first


hol430:
@HamishBrownPFR / @hol353 - do you want me to go through this and flag code style/implementation details? Or are we not at that stage just yet?

HamishBrownPFR:
have moved this onto a different branch.  will reopen from there soon

PR #6768, opened 2021-09-06
Update and fixes to AgPasture

rcichota:
Working on #6767 
Starting with changes aimed at tidying up the code a bit, fixing type, deleting superfluous comments and updating doco tags. Also deleted a few variables/parameters no longer used.
Few minor changes to variable names to come

### DIFF
--- a/Models/AgPasture/GenericTissue.cs
+++ b/Models/AgPasture/GenericTissue.cs
@@ -85,25 +85,44 @@
 
         //----------------------- Public methods -----------------------
 
+        [EventSubscribe("Commencing")]
+        private void OnSimulationCommencing(object sender, EventArgs args)
+        {
+            ClearDailyTransferredAmounts();
+        }
+
         /// <summary>Preparation before the main daily processes.</summary>
         public void OnDoDailyInitialisation()
         {
             DMRemoved = 0;
             NRemoved = 0;
-            ClearDailyDeltas();
+            ClearDailyTransferredAmounts();
         }
 
-        /// <summary>Updates the tissue state, make changes in DM and N effective.</summary>
-        public void Update()
+        /// <summary>Sets the biomass of this tissue.</summary>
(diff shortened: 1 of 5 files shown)


hol430 (approved):
Looks good to me.

rcichota:
@hol353 
These were the changes I made that do not alter the results in AgPasture validation. If OK, could you please merge these and I will start adding the bits that do change the results. Thanks

rcichota:
@hol353 or @hol430 
I am not sure whether, or how much/where, this has failed as I cannot access the link to Jenkins...
In the last commit I reverse some variable setting that Drew added (maybe only last week?). Those variables do have a default value in the code, so I thought the re-setting was redundant... APSIM run successfully in my PC, maybe it fails somewhere else in Jenkins or unit test...?? Or it is something completely different... Any help much appreciated

hol430:
You should be able to access the jenkins link...what error do you get when you click on the link?

The problem that jenkins has found is that agpasture no longer resets all of its variables. The context for this is that I'm currently working on a tool which can hold an .apsimx file open in memory and rapidly re-run the file without reading or cloning the simulations in between each run. This requires us to be very thorough about resetting/zeroing variables at the start and end of simulations, and I've added some tests to ensure that variables are indeed being reset correctly. It's one of these tests that's failing. I'm guessing the change that I made and that you reverted is probably one of the fixes I made, to ensure that variables are being reset correctly.

rcichota:
Thanks @hol430 
I will have a look at reverting it. I think its odd that it would need to set these variables twice on initialisation, but maybe I am not seeing some of the workings... 
I get the error below when I click on Jenkins link:
image

rcichota:
Hi @hol430 
I managed to get through the Jenkins link. I have no problems if I use FireFox, probably some security thingy in Edge (or PFR) that was blocking... I will now have a good look at the messages there

hol353 (approved):

PR #6783, opened 2021-09-08
Eliminate SizeAllocated response functions, as one was unused and the…

### LINKED ISSUE #6702: Resizing of property presenter widgets with window resize
The resizing of property presenter entry boxes cannot keep up with a quick resize of the window under windows.
Entry box right hand side resize works if you slowly shrink the window in windows with the mouse, but stops resizing as you start to change the window size quickly, or one rapid shrink.

zur003:
… other interferred with window resizing. Resolves #6702


hol430 (approved):
Looks good, thanks Eric. @lie112 - if you end up needing to add similar functionality back in, you can probably achieve the resizing by changing the packing of the widget, and you should be able to achieve an 8px padding by changing the margin properties of the widget. (I think.)

PR #6784, opened 2021-09-08
Implement CLEM filtering and sorting of individuals

### LINKED ISSUE #6355: Add herd genomics to CLEM
New research projects will use CLEM to investigate considering genetics in herd management
The initial functionality to track genetic state (index) of individuals in the herd is required

### LINKED ISSUE #6615: Apply smart breeding
Smart breeding will allow the timing of controlled mating to be managed as well as providing the individuals for mating to attempt to provide continuous lactation for milking in dairy systems.
Implementation of this functionality will also involve model validation and the correct implementation of the upcoming filtering and sorting models of ienumerable lists of resource individuals

### LINKED ISSUE #6692: Refactor and implement new list filtering and sorting by properties and attributes

lie112:
resolves #6692 
resolves #6355
resolves #6615

Implements the new filter and sort of individuals, both with full generic reflection and a a faster hard coded method for large herds
Cleans up the remainder of CLEM code
Fixes bugs identified from previous refactoring around generic resource location and economic reporting


lie112:
I realise this is a bit to go through, but can I get this into an upgrade soon
There shouldn't been too many code issues like wayward commented code.

Models/CLEM/Groupings/FilterByProperty.cs
```diff
@@ -0,0 +1,540 @@
...
+        {
+            if (useSimpleApporach)
+            {
+                Func<T, bool> simple = t => {
+                    var simpleFilterParam = Expression.Parameter(typeof(T));
+                    var thisOperator = Operator;
+
+                    // Try convert the Value into the same data type as the property
+                    var simpleCompareVal = Expression.Constant(propertyInfo.PropertyType.IsEnum? Convert.ChangeType(Value.ToString(), typeof(string)) :Convert.ChangeType(Value ?? 0, propertyInfo.PropertyType));
+                    //var simpleCompareVal = Expression.Constant(Convert.ChangeType(Value ?? 0, propertyInfo.PropertyType.IsEnum? typeof(string): propertyInfo.PropertyType));
```
hol430, line 81: Commented code

Models/CLEM/Groupings/FilterByProperty.cs
```diff
@@ -0,0 +1,540 @@
...
+        /// <summary>
+        /// Convert filter to html string
+        /// </summary>
+        /// <returns></returns>
+        public string ToHTMLString()
+        {
+            return filterString(true);
+        }
+
+        private string filterString(bool htmltags)
```
hol430, line 448: Function name should start with upper case letter.

Models/CLEM/Groupings/OtherAnimalsFilterGroup.cs
```diff
@@ -17,25 +13,10 @@ namespace Models.CLEM.Groupings
     [Serializable]
     [ViewName("UserInterface.Views.PropertyView")]
     [PresenterName("UserInterface.Presenters.PropertyPresenter")]
-    [Description("This other animal filter group selects specific individuals from the other animals using any number of Other Animal Filters.")]
+    //[Description("Selects specific individuals from the other animals")]
```
hol430, line 16: Commented code - could be removed

Models/CLEM/Groupings/SortByProperty.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <summary>
+        /// Convert sort to html string
+        /// </summary>
+        /// <returns></returns>
+        public string ToHTMLString()
+        {
+            return sortString(true);
+        }
+
+        private string sortString(bool htmltags)
```
hol430, line 70: Function name should start with upper case letter.

Models/CLEM/Groupings/TakeFromFiltered.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// <summary>
+        /// Convert sort to html string
+        /// </summary>
+        /// <returns></returns>
+        public string ToHtmlString()
+        {
+            return takeString(true);
+        }
+
+        private string takeString(bool htmltags)
```
hol430, line 78: Function name should start with upper case letter.

Models/CLEM/Resources/CLEMResourceTypeBase.cs
```diff
@@ -247,18 +218,18 @@ public double ConversionFactor(string converterName)
...
-                case "FinanceType":
-                case "HumanFoodStoreType":
-                //case "WaterType":
-                //case "AnimalFoodType":
-                //case "EquipmentType":
-                //case "GreenhousGasesType":
-                case "ProductStoreType":
+                case FinanceType _:
+                case HumanFoodStoreType _:
+                //case WaterType _:
```
hol430, line 225: Commented code - please remove if not needed

Models/CLEM/Resources/IndividualAttribute.cs
```diff
@@ -61,14 +61,10 @@ public float Value
         {
             get
             {
-                if(float.TryParse(storedValue.ToString(), out float val))
-                {
+                if(float.TryParse(StoredValue.ToString(), out float val))
```
hol430, line 64: Probably  best to explicitly specify the culture rules when parsing numbers from strings like this. Using the invariant culture is usually best, otherwise your simulations can give different results depending on the language/locale settings of the computer on which they're run - e.g. some cultural settings use commas for decimal place separators and periods for thousands separators.

Models/CLEM/Resources/IndividualAttribute.cs
```diff
@@ -79,18 +75,13 @@ public float MateValue
         {
             get
             {
-                if (float.TryParse(storedValue.ToString(), out float val))
-                {
+                if (float.TryParse(StoredMateValue.ToString(), out float val))
```
hol430, line 78: See above comment re cultural settings

Models/CLEM/Resources/Labour.cs
```diff
@@ -393,7 +395,7 @@ public double PayRate(LabourType ind, LabourFilterParameters property, string va
                 if (matchCriteria == null)
                 {
                     // report specific criteria not found in price list
-                    string warningString = "No [Pay] rate entry was found meeting the required criteria [" + property + "]" + (value.ToUpper() != "TRUE" ? " = [" + value + "]." : ".");
+                    string warningString = "No [Pay] rate entry was found meeting the required criteria [" + property.Name + "]" + (value.ToUpper() != "TRUE" ? " = [" + value + "]." : ".");
```
hol430, line 398: Usually best to avoid lots of string concatenation like this as it can result in a lot of allocations, especially if this is called daily (or multiple times per day). This isn't really a huge issue - but usually it's preferable (and more readable!) to use string interpolation or a StringBuilder.

Models/CLEM/Resources/Ruminant.cs
```diff
@@ -83,30 +85,19 @@ public RuminantFemale Mother
...
         /// <summary>
         /// Sex of individual
         /// </summary>
-        public Sex Sex { get { return (this is RuminantFemale)?Sex.Female:Sex.Male; } }
-
-        /// <summary>
-        /// Gender as string for reports
-        /// </summary>
-        public string GenderAsString { get { return Gender.ToString().Substring(0,1); } }
+        public Sex Sex { get { return ((this is RuminantMale) ? Sex.Male : Sex.Female); } set {  } }
```
hol430, line 100: What's the idea behind the empty setter? Is this required by an interface? Would it be better/feasible to throw a not implemented exception?

In fact, looking closer, I see that this is an abstract class. If this property comes from an interface, perhaps it would be better to implement this property only in the concrete types derived from this abstract class - I'm guessing `RuminantMale` and `RuminantFemale`?

Models/CLEM/Resources/SetAttributeWithValue.cs
```diff
@@ -78,33 +83,37 @@ public class SetAttributeWithValue : CLEMModel, IValidatableObject, ISetAttribut
...
-        /// </summary>
-        public IndividualAttribute GetRandomSetAttribute()
+        /// <inheritdoc/>
+        public IndividualAttribute GetAttribute(bool createNewInstance = true)
         {
-            double value = Value;
-            double randStdNormal = 0;
-
-            if (StandardDeviation > 0)
+            if (createNewInstance | lastInstance is null)
```
hol430, line 94: Bitwise OR - probably a typo?

Models/CLEM/ZoneCLEM.cs
```diff
@@ -372,7 +371,7 @@ public string GetFullSummary(IModel model, bool useFullDescription, string htmlS
                     htmlWriter.Write(this.ModelSummaryOpeningTags(formatForParentControl));
                     htmlWriter.Write(this.ModelSummaryInnerOpeningTagsBeforeSummary());
                     htmlWriter.Write(this.ModelSummary(formatForParentControl));
-                    htmlWriter.Write(CLEMModel.AddMemosToSummary(this));
+//                    htmlWriter.Write(CLEMModel.AddMemosToSummary(this));
```
hol430, line 374: Commented code

hol430 (changes requested):

ApsimNG/Presenters/CLEM/CLEMPresenter.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using APSIM.Shared.Utilities;
 using Models.CLEM;
+using Models.CLEM.Interfaces;
```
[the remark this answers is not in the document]
lie112: This is needed for ICLEMUI on line 19, 38. It is not greyed out in my code. Shall I remove all unused usings from other models and presenters etc?

ApsimNG/Presenters/CLEM/CLEMReportResultsPresenter.cs
```diff
@@ -42,47 +42,47 @@ public void AttachExtraPresenters(CLEMPresenter clemPresenter)
         {
             try
             {
+                bool parentOfReport = false;
                 Report report = clemPresenter.model as Report;
+                if(report is null)
+                {
+                    report = (clemPresenter.model as Model).FindChild<Report>();
```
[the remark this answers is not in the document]
lie112: The clemPresenter.model is an object. report is a private local variable only used in this method. It is the same as in ReportPresenter to identify simulations, simulation, experiment and paddock for the datastorePresenter.ZoneFilter

ApsimNG/Presenters/CLEM/CLEMPresenter.cs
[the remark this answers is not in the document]
hol430: I wouldn't bother going out of your way to fix existing ones - I usually just give them a quick glance whenever I add or remove one while working on a class.

Examples/CLEM/CurrentDescriptiveSummary.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
lie112: Yes. These are annoying. I have updated my GitIgnore file to include these, but it doesn't seem to pick them up and ignore them. I didn't use **/ but just the filename as Git doco said a filename alone will be ignored anywhere in the source code.

Models/CLEM/Activities/PastureActivityManage.cs
```diff
@@ -242,10 +245,19 @@ private void OnCLEMInitialiseResource(object sender, EventArgs e)
...
+
+                    double firstMonthsGrowth = 0;
+                    if (pastureDataList != null)
+                    {
+                        PastureDataType pasturedata = pastureDataList.Where(a => a.Year == clock.StartDate.Year && a.Month == clock.StartDate.Month).FirstOrDefault();
+                        firstMonthsGrowth = pasturedata.Growth;
+                    }
+
+                    LinkedNativeFoodType.SetupStartingPasturePools(Area * unitsOfArea2Ha, firstMonthsGrowth);
+                    //SetupStartingPasturePools(StartingAmount);
```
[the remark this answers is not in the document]
lie112: yeah, missed all those. It was the last thing I was working on

Models/CLEM/Activities/RuminantActivityFeed.cs
```diff
@@ -355,28 +401,30 @@ public override void DoActivity()
                 }
 
                 // feed animals
-                if(feedRequest == null || (feedRequest.Required == 0 | feedRequest.Available == 0))
+                if(feedRequest == null || (feedRequest.Required == 0 | feedRequest.Available == 0) || feedLimit <= 0.00001)
```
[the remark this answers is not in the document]
lie112: Yes, thanks. That was mentioned in last weeks APSIM catch-up. I was confused with previous && to avoid moving on if feedrequest was null. Solved with feedRequest?.Required

Models/CLEM/Groupings/OtherAnimalsFilterGroup.cs
lie112: This was work in progress relating to another problem.

Models/CLEM/Resources/Ruminant.cs
lie112: We actually had this implemented the way you suggest, but it won't work with Report. If a user wants to report Individual.Sex (where individual is type Ruminant), Report will set the PropertyInfo for this based on the first individual applied to (e.g. Female, RumiantFemale.Sex), then when report tries to report Sex for a Male later in the reporting an exception is thrown as RuminantFemale.Sex PropertyInfo cannot be found  on male. I think the null set was to get this working as it didn't seem to pick up get only properties in Report. I thought about a fix for report but couldn't come up with a generic solution for this specific issue. In the end implementing Sex for Ruminant was the easiest fix.

ApsimNG/Presenters/CLEM/CLEMReportResultsPresenter.cs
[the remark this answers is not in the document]
hol430: Ok - clemPresenter.model should have an uppercase M (so clemPresenter.Model). I would probably be inclined to change its type at least to IModel, or possibly something more derived if you think that would make more sense. I can't think of a situation where the model passed into the presenter isn't an IModel - in fact, the `Attach()` method's definition in `IPresenter` should really have an IModel, rather than an object, for the model property...but I digress. Feel free to change this one or leave it, whatever you think makes sense.

Models/CLEM/Resources/SetAttributeWithValue.cs
lie112: Ors still do my head in a bit. It really doesn't matter in this case does it? It doesn't matter if both are evaluated, but might be a smidgen quicker if we didn't need to evaluate the null check.

Models/CLEM/Resources/Ruminant.cs
lie112: I had a large reply for this but it never sent. We originally had the classes working as you suggested. The only problem is Reports don't like it. The way you suggest, if you want to report on Sex of a ruminant in a report the PropertyInfo is taken from the first individual in the list (which may be a Female, so PI is RuminantFemale.Sex), then later when it tries to report on a ruminant male if cannot find the property (RuminantFemale.Sex) on the object (RuminantMale) and throws a fatal error. I looked into ways to resolve this but it got quite messy and I figured the current implementation was the easiest for a single situation and we didn't want reflection calls at every write to the report which is already a intesnive part of the model. In trying to resolve the issue I thought the Report was not accepting {get; } only properties so I added the empty set {} and it started working, but it may have been related to something else and so might be able to be deleted. I thought I'd come across this before so tried my same old approach to fix it and keep exploring options.

ApsimNG/Presenters/CLEM/CLEMReportResultsPresenter.cs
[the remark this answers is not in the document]
lie112: This is a hangover from using the private variables already defined by APSIM in most of the presenters and needing to access it without changing its name. I missed that my internals were actually public. Their types was a carryover from propertypresenter 
line 53 PropertyPresenter,cs        public virtual void Attach(object model, object view, ExplorerPresenter explorerPresenter)
I have changed my internal objects for CLEM presenter  to capital letters. This may be the case elsewhere

Models/CLEM/Resources/Ruminant.cs
lie112: I would like to fix this while cleaning this all up. I don't like the type check (is RuminantMale) every time we access Sex (a common call). I could have a localised version on each inherited type that returns the enum value back to the general base property. Am I missing something fundamental in inheritance here? I'm not sure if the problem was the abstract class not allowing me to assign a property for the report, or just the fact that it asked for the property info from the RuminantFemale as the first object not from the female's base Ruminant class that would have worked for both males and females. Unless the base property is not valid. Any ideas here?

lie112: It worked today. Must have just been some fresh coding.

hol430: This works in simple cases, but I've found the bug you mentioned. It happens when the property we're reporting is of the abstract class' type. As you said, it's caching the PropertyInfo of the concrete type which leads to an exception if the value of the property (containing an instance of the abstract type) changes.

hol430: This is fixed by #6795.

lie112: ok. It all looked good but the example I was working with only had females so didn't introduce the issue.
The problem is visible in the current Jenkins build when it tries to use the clem grazing example where there are males and females. I would have thought this was using the #6795 fix, but isn't working.

Models/CLEM/Resources/Ruminant.cs
```diff
@@ -83,30 +85,19 @@ public RuminantFemale Mother
...
         /// <summary>
         /// Sex of individual
         /// </summary>
-        public Sex Sex { get { return (this is RuminantFemale)?Sex.Female:Sex.Male; } }
-
-        /// <summary>
-        /// Gender as string for reports
-        /// </summary>
-        public string GenderAsString { get { return Gender.ToString().Substring(0,1); } }
+        public virtual Sex Sex { get { return this.Sex; } }
```
hol430, line 100:
```suggestion
        public abstract Sex Sex { get; }
```

Then in the derived types:

```csharp
public override Sex Sex { get { return ... } }
```

Models/CLEM/Resources/Ruminant.cs
hol430: Well, this jenkins build ran half an hour ago, whereas I only created #6795 a few minutes ago...we haven't implement a time travel feature in jenkins yet :).

Once #6795 is merged, we can retest this and it should (hopefully) run green. I'm just hoping I haven't  broken anything with #6795...

lie112: How do you implement this. I have tried a public virtual and abstract on the base Ruminant with the inherited models providing a new Sex Sex { get { return Sex.Male }} etc. But I still get an error in Report. 
Invalid report variables found: Sex: Object does not match target type.

lie112: oops, should be override?

hol430: Yep, you will want abstract in the base class (Ruminant), and override in the derived class (RuminantMale/Female). The "object does not match target type" error should be fixed by #6795.

lie112: I should have checked the date of #6795, I assumed it was already included. I'll await it to be accepted and try again.

lie112:
This is right to go. Just need to retest jenkins after #6797 merges as I am awaiting that Reports fix.

lie112:
@hol430 @hol353 Can I please get the outstanding #6795 merged so I can re-run Jenkins to allow this pull request to be included and an upgrade created. I have a number of users awaiting this upgrade for projects.

hol430:
@hol353 - see above. I think if you're happy with #6795 it could be merged, then we can get this PR going again.

hol430:
retest this please jenkins

hol430 (approved):
Looks good to me.

lie112:
@hol430 seems the last upgrade was 7/9 --6775 
Should I be able to see the master branch including our latest changes? SourceTree has masterbranch sitting at this previous CO2 not read from met file merge, but git repo online says master is at our latest merge. What am I missing?

hol430:
Well, it depends on which master branch you're referring to. There's a master branch on this remote repository, there's a master branch on your forked repository, and there's also (presumably) a local master branch sitting on your computer. The master branch on this repo definitely contains the changes in this PR - if it's not showing that in sourcetree, try doing a fetch (of this repository). If the problem is with the local master branch on your computer, you can checkout that branch, then do a pull from this repo's master branch. If the problem is with the master branch on your forked repository, you'll need to perform the previous step to bring your local master branch up to date, and then push up to the master branch on your forked repo.

[This page](https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches) has some background info and diagrams which may be more helpful than my explanation.

lie112:
I was looking at releasefork/master on the apsimx repo but couldn't get a refresh.
I restarted SourceTree and all is is good. The usual refresh wasn't working.
Time to ditch sourcetree I think. What are you using for source control?...command line I assume!

Is the update generator currently working? The last upgrade is 2021.09.07.6675
Did you once give me a link to see the latest progress on upgrade.
I have the jenkins progress link which I use a lot.

hol430:
Yeah, I tried a few git guis but had (minor-ish) problems with all of the ones I tried, so I just use the command-line git these days.

Upgrades seem to be broken...again...I'm hoping that the problem fixed by #6803 is the only problem. If so, we may see a new upgrade later today.

You can see jenkins building upgrades by going to the [jenkins homepage](http://apsimdev.apsim.info:8080/jenkins/) and clicking on the CreateInstallation job. Or I suppose you could just use [this link](http://apsimdev.apsim.info:8080/jenkins/job/CreateInstallation/) and cut out the middle man, so to speak.

PR #6785, opened 2021-09-08
Further updates and fixes to AgPasture

rcichota:
working on #6767 
Fixing a few bugs and updating the model as needed. This will change the performance of the validation...
Doing it in small steps to avoid/minimise unseen consequences.


rcichota:
@hol430 
Could you help me here please. The changes I am here do not alter any of the outputs, I was just tidying up a bit the methods that zeroes some of the variables every day, which I did not expect would change any resetting. Clearly it did as the commit fails on your check for zeroing all variables. The message on Jenkins does not give me a clue of which variable is the problem and I don't know how/whether I can test these things in my PC before pushing up... Thanks

hol430:
My guess would be that the removal of the `OnSimulationCommencing()` event handler in GenericTissue.cs is to blame, although I'm not certain. You can always run the failing unit test locally - it's in SimulationResetTests.cs. The test is quite slow to run, so it may be worth commenting out the lines corresponding to the .apsimx files which aren't failing (which is all of them except AgPasture.apsimx, so you could comment out lines 85-103). There are a few different ways to run the unit test. From the command-line, you can run `dotnet test --filter SimulationResetTests`. Most IDEs will let you run/debug unit tests from inside the IDE as well. I think visual studio has a test explorer window (debug -> windows -> test explorer?) which you can use to selectively run unit tests, or you might also be able to right click inside the test method and click "run test" or "debug test" or something along those lines. Perhaps someone who uses visual studio can comment here.

rcichota:
Thanks @hol430 I managed to get the unit test running using the command line.
Indeed the problem was the removal of OnSimulationCommencing() event in GenericTissue; which, in retrospect, I sort of understand why. However why it was not explicitly needed for the RootTissue is a bit baffling. I have added something similar in case it will be needed in the future.

rcichota:
retest this please Jenkins

rcichota:
@hol430 @hol353 
Hi guys, I am a bit stuck here. The unit tests are failing, apparently one or more variables is not zeroing properly. I wonder how can one debug this. I have run the UnitTest from the command line in my PC (as per Drew's instructions early in this thread), but I cannot do it in debug mode to track the correct spot (tried a couple of tricks to run the executable from the debug tab in the solution properties, no success)... is there a way to do it? or I have to use 'brute force' and review all the changes I made in the [second to] last commit. I did quite a bit of changes to how to the processes transferring values between organs and tissues work. I tried to be careful and make them very similar for both above and below ground organs (as they were a bit of a mess and there was a few errors there - mass balance now is better than before). Is there a 'typical' culprit I could have a look? I did add a ClearVariables on DailyInitialisation, but no success...
Thanks for you help

hol430:
You can open the file in the GUI, right click on the simulations node and choose "check server compatibility", which I believe should at least let you reproduce the problem with a debugger.

Not sure about a typical culprit, but the cause of the problem is not resetting variables to their initial state at the end of the simulation. Even if you're clearing your variables on DailyInitialisation, they could be modified by the end of the day which could cause this problem.

rcichota:
Thanks @hol430 
The variables are indeed reset early in the day only (as they could be used for reporting). So if I understood you, they should also be reset at the end of the simulation?
I managed to run the 'Verify server compatibility' thingy, I will see how I go. I do notice quite a few errors now on models that did not fail before (e.g. manager scripts), is it expected?

hol430:
I could be misremembering, as it's been a while since I wrote the test, but I think the unit test only reports the first problem it finds, whereas the tool in the GUI will try and report all problems it detects in a simulation.

rcichota:
HI @hol430 
I don't think this hack from the GUI is working as expected, or to replicate the command line test. I do get something going, and in debug mode, but I only get error messages related to other models (manager and Nutrient), not from AgPasture... And nothing like those I get from running the test from the command line.
I have tried quite a few changes to the code this afternoon, but got nowhere really. As I don't understand what I am trying to fix, I don't know what to look for...
I noticed that the variable the message is saying is not being zeroed (at least the bottom of the message) comes from the RootTissue, but it is exactly the same implementation as in GenericTissue (aboveground) and that was working fine before... I tried to revert some of the changes I made yesterday, tried to change the variable from a property to a field, I also have added a variable cleanup at the end of the simulation (subscribing to OnEndOfSimulation), but the very same error message keep showing up. If you have any hints or leads i could follow it would be much appreciated.

rcichota:
Hi guys @hol430 @hol353 
Sorry to bother, but I need to get the error with these variables in UnitTest sorted to see if I can finally get a pass on AgPasture validation, I think I am nearly there with the latest fixes to N remobilisation (see one of the comparisons below)...
Any hints on what to look for on those tests would be much appreciated, as I said I don't know what those tests are checking so it is difficult for me to try to debug, and the error message does not provide me with clues as to why it is failing on some variables but not others, which I think are the same... Thanks in advance
image

hol430:
I think what's happening here is that the root's dead biomass is not being reset at the end or start of the simulation. PastureSpecies will, inside `SetInitialState()` call `roots[0].SetBiomassState(...)`. In `PastureBelowGroundOrgan.SetBiomassState()`, we set the live biomass (`tissue[0].SetBiomass()`), but not the dead biomass. There's a comment on this line to this effect - the assumption seems to be that there's no existing dead material. If we were to remove any dead biomass here, the issue should go away.

The other way would be to subscribe to EndOfSimulation and manually clear the biomass:

```csharp
[EventSubscribe("EndOfSimulation")]
private void OnEndOfSim(object sender, EventArgs args)
{
    biomass = new AGPBiomass();
}
```

The reason this isn't an issue in GenericTissue/PastureAboveGroundOrgan, is that PastureAboveGroundOrgan seems to explicitly set the biomass state in all tissues.

While debugging this, I found a few issues with the state checker, as you've noted. The biggest problem is that it's not serializing private fields, which makes it quite difficult to diff the json it spits out. I'll also tweak the unit test so that it's running exactly the same code as from the gui.

In case you're wondering what the tool is actually checking for: we need to be able to run a simulation, and have the state of the simulation (all objects, properties, fields) be identical to what they were before the simulation was run. So clearing variables every day is not really necessary, but they do need to be reset to their initial values either at the end or the start of the simulation.

rcichota:
Thanks @hol430 
I will give another go shortly

rcichota:
@hol353 
I think I have gone as far as I can at this stage. Although the stats are not better than before, they are not too bad. I think there are tweaks in the validation simulation itself that can/could be done to improve things further (no longer in the model), but I don't have time right now to do that. I will work with Mariana (as she offered) and we'll see if we can improve the simulation outputs over the next weeks (the sim is quite a mess and I have a few more data to add). I want to also check how AgPasture will work with forage plantain in mixed swards *9that is what I am actually been paid to do right now).
I am confident the model is much better now, as there are (were) quite a few errors on the N uptake and allocation that have been fixed. If you are reasonably happy, could you please merge this changes, I worry that people are using the current version and it is really not a good thing...

rcichota:
By the way these are the agreement i am getting with the dataset I had more time to set up good simulation for:
image
image

hol353 (approved):

PR #6790, opened 2021-09-12
Fixed a problem with SWIM divide tolerance

hol353:
Working on #1254


hol430 (approved):
Looks fine. Might be worth decreasing the default tolerance at some point. 1e-5 seems rather high.

Issue #6706, opened 2021-08-12
Improve "add model" assessment of ValidParent and expanded state


lie112:
The Add Model Presenter tries to remove unavailable models from the list offered to the user based on the parent model selected. This is not quite working as expected for models in CLEM. 
I'm happy that some models are always present (Functions, Memo, and can live with Map), but given only the CLEM timers have the valid parent ResourcePricing (see example below), I would not expect any of the other model folders to appear. Are they also functions to be accepted anywhere in the simulation tree?

With respect to branches expanded upon opening, it would be nice if when only a small number of models are possible (say less than 20) and a folder contains < 10 models, those branches are expanded on opening. Therefore, when only a couple options are available to the user in a single folder they will all show up expanded and save a couple mouse clicks to find them.

Example of "add model" screen when selected from a ResourcePricing model in CLEM. All branches expanded manually

![image](https://user-images.githubusercontent.com/19215637/129129177-59722105-ce80-42a1-91f2-2ba0e1a61779.png)

hol430:
The idea is that IFunctions can be dropped anywhere, as they have `DropAnywhere = true` on their `ValidParent` attribute. Unsure how useful to this is - usually you will only need them in a PMF model. Perhaps others could comment here. Expanding all nodes when there are only a few visible seems like a good idea.

hol353:
IFunctions can be dropped anywhere. I know Nutrient uses them so they aren't just a PMF thing. Auto-expanding nodes in the AddModels tree when the node only as a few child nodes (e.g. < 20) might work ok.  

Some of the nodes in the screenshot shouldn't be displayed e.g. AccumulateCohortExpression, ColdUpRegVrn1, PTQPhyllochron, HeightFunction, CNReduction..., RunoffModel etc. They are probably missing ValidParent attributes.

hol430:
All of those models are showing up because they're IFunctions.

PR #6804, opened 2021-09-17
Add Exclude as ValidParent property to allow models (base) to be hidden

lie112:
working on #6706 

This commit adds an additional 'Exclude' property to the ValidParent attribute to do the opposite of 'DropAnywhere' property
This value can then be added to any model in its ValidParent attribute to ensure it hidden from the 'add model' presenter.
One new property and one if statement was able to implement this functionality in a clean consistent manner


hol430:
If the idea behind these classes is that they can't or shouldn't be instantiated, would it be feasible to make them abstract classes instead? That would seem to also solve the problem.

I'm just not sure what is the point of having a model which can't be added anywhere - which seems to be what this Exclude property does.

lie112:
There seem to be a heap of models in the 'add model' list not intended to be added by the user (e.g. Model and interfaces starting with I) and a switch to hide them would be useful. I was trying to clean up the list, at least for CLEM models not directly available for the user to add to the UI tree.
I have a number of abstract classes that still show up in "add model" list. Maybe this was a bug, and I didn't think to look for this as a fix to the problem.
Just trying to clean this list up and make it clearer for users.

lie112:
As APSIM grows with more variety of models, the argument could be made that DropAnywhere is not appropriate from some developer's perspective and you should almost be able to ignore this or specify InvalidParents(), but the second option would involve model specific values added to other developers' non related models which is sloppy. The drop anywhere functionality could equally be applied by adding a ValidParent(Interface) to which all models that allow that sort of drop anywhere can be dropped onto, such as PFG and Nutrient, could implement (e.g. IFunctionDroppable).

hol430:
Interesting, if there are interfaces and abstract classes showing up in the list, that's definitely a bug.

Agreed that there are a large number of useless models showing up in the list, and I also agree that DropAnywhere is not a good mechanism. Something like `[ValidParent(typeof(IModel)]` would be preferable in my opinion, although even that should really be used far more sparingly than it currently is. I wonder if the usage of ValidParent attributes is really the best pattern here.

I think that something like an InvalidParent attribute may not be desirable, as it would make the allowable model semantics even more complicated than they already are. I do think we need more flexibility in the GetAllowableChildModels() and IsChildAllowable() functions, but I'm not 100% sure what the best approach would be. Perhaps IsChildAllowable should be a (n overridable) method belonging to the Model class (or IModel)?

I don't think IFunctionDroppable (or equivalent) is the answer; having an empty interface just seems wrong to me, and this would need to be implemented by almost every model in apsim.

lie112:
oops, I thought this branch was not in use. It appears it holds the previous work being considered on Add Exclude as ValidParent property. I'm not sure how to fix this and pull the latest commits to 6962

hol430:
If you want to extract these commits to another branch, your best bet is probably to create a brand new branch, and then cherry-pick the individual commits that you want from this branch. Should be a way to do this in fork - I imagine you can probably select the range of commits you want, then right click -> cherry-pick.

lie112:
Can I just close this pull request as it is an unresolved question and it doesn't look like my code committed will be used?
I can then revert this change from my history and pull the branch with all my changes to the other issue raised.

lie112:
I am closing this pull request as this doesn't appear the best approach for solving #6706

PR #6810, opened 2021-09-21
Fix inherited attribute issue and update all descriptions

### LINKED ISSUE #6809: Issue with inheriting attribute in grow-out herd
User has an null object error when allowing grow-out herd to mature and breed

lie112:
resolves #6809


hol430 (approved):
Looks good to me

PR #6814, opened 2021-09-24
Allow property order to be set using DisplayAttribute.Order

### LINKED ISSUE #6806: Specify order of properties in property presenter
Would it be possible to allow developers to include the 
System.ComponentModel.DataAnnotations.Display attribute with associated Order property on any property to further define the order that properties are presented in the property presenter?
At present the properties are ordered as provided in the model using line number of the description attribute associated with the property.
This is not sufficient for properties inherited from a base classes and an additional order index would allow customisation of property order or some or all properties of a model.

propertyList.OrderBy(Display.Order).ThenBy(Description.LineNumber)
If no order properties were provided then this would revert to LineNumbers and all properties without a Order (0) would also be sorted by line number rather than the base alphabetical from GetProperties

lie112:
resolves #6806 
Provides the ability to provide primary order of properties in property presenter using Model.Core.DisplayAttribute.Order


hol430:
@lie112 you'll need to fix the merge conflicts before this can be merged.

lie112:
I kind of understand what you are saying, but support base classes in certain circumstances, I was not aware of the code style approach to not use base classes when developing CLEM and even worked from Shaun's initial code as he was my conduit to all APSIM Next Gen developers and code practices. I have used these sparingly, but they are a significant part of CLEM with CLEM models a special type of Model, and base ResourcesTypes and ActivityTypes providing the functionality to all inherited resource and activity models.
I prefer to provide my properties in code in alphabetical order, but using the code line number to define the order in property presenter dictates the order they must be defined (ok if you want alphabetical display). Isn't this equally confusing to developers potentially unaware of the link between property presenter and property line numbers?
Microsoft already provides an Order property of the Display attribute for this reason, and adding this to Core.DisplayAttribute does not affect anyone's code except the few CLEM base classes where it gets implemented.
Early on I was forced to do some work to stop the Zone Properties (Aspect etc) being presented in the CLEM model that inherits Zone. I had to start my journey by making CLEM a Zone as this was the only level it could be added and needed to be included in reports that need at least Zone.
This relatively clean change allows me to reorganise the placement of a few inherited properties in the property presenter to make it much clearer to the user using a ZIndex approach as used when ordering of objects is needed such as ccs where order is defined by the order items are added (akin to line numbers) but you can move objects to the top or bottom often using large arbitrary integers such as 1000 or -1000 as implemented here. 
I would have to add multiple properties to all my models just to provide the suggested approach and this leads to repetition in my code, but I agree it does show where properties are inherited.
There are other attributes or flags I need such as determining which properties are available for the user in filter property dropdowns, but I haven't been able to implement this to date.

lie112:
This can be addressed if a broader group need this functionality, but it is a somewhat consistent approach with other environments (css) and meets the requirement of the few CLEM properties where this really helps the user without a significant time coding for this functionality.

Issue #6824, opened 2021-09-28
SWIM3 - lower boundary condition needs


sno036:
- refactor in the code if necessary
- check that each option works sensibly, adding a test for each (need to discuss how to define pass/fail)
- does mass balance for each one?
- check alignment with Verburg and note updates etc.
- memo documentation of the tests
- manager example of how to use

sno036:
What is causing a mass balance error starting on about day 11?
Mass balance script could be more useful with:
- auto switching of weather data source
- auto finding of crops in the simulation
- warning and crash values for the water balance error - don't want a crash yet however

sno036:
Needs more work.

sno036:
Checking out issues with the set potential lower BC.

sno036:
Done. This issue will close once merged.

PR #6826, opened 2021-09-28
Start of a SWIM3 lower BC test etc.

sno036:
Worthwhile getting this simulation added/merged - there is a mass balance error that emerges at about day 11.

Working on #6824 

Merge this please Jenkins (or does this not work these days?) Does 'working on' end up being merged automatically?


hol353:
Merging is a manual process that happens once I or @hol430 do a review of it and am happy with it. The 'working on' phase is important to Jenkins to let it know which GitHub issue the PR is related to. It uses this info to version stamp the executables that it builds.

hol353:
If the simulation throws an exception (error) due to a mass balance problem then Jenkins will run 'red' and it won't be able to be merged. We only merge green PRs.

hol353:
@hol430 Can you review and merge if ok? This is a work in progress and there is a long way to go to get SWIM source code meeting coding standards and into the release.

hol430 (approved):
Looks good, although I'll note that if we're aiming to follow Microsoft's guidelines for code style, we should not have comments on the same line as code like in this line:

```csharp
int isbc = 0; // No storage of water on soil surface
```

Since this is a work in progress I'm happy for this to go through though.

PR #6844, opened 2021-10-11
Refactor the autodocs

### LINKED ISSUE #5606: The NG documentation website is broken
(again)

### LINKED ISSUE #6443: Refactor the Autodocs
Following discussion with @hol353 about an autodocs refactor:

Plant is to add a

```csharp
IEnumerable<ITag> Document();
```

method to class Model. The base implementation will just parse the markdown in the class' summary tag. The ICustomDocumentation interface will cease to exist (existing implementations will override the `Document()` method), and the tags will live in a new class library (APSIM.Services?) which has minimal dependencies. The export to pdf functionality will be separated from the auto-documentation and will live in its own assembly to remove the dependency on PDFSharp/Migradoc from the Models assembly. Ultimately Models will become a class library and will contain only the Models and should have minimal dependencies.

_Originally posted by @hol430 in https://github.com/APSIMInitiative/ApsimX/issues/6129#issuecomment-809031180_

Additionally: we need to remove the dependency on gtk so that the autodocs can be generated on a headless VM without any display server installed (ie as part of the new jenkins workflow).

### LINKED ISSUE #6793: Use VMs for autodocs
The autodocs should be built in the main jenkins test phase, and they should be built on VMs rather than a dedicated server in Toowoomba. This depends on #6443.

hol430:
- No longer requires the gtk stack
- Autodocs now built immediately after validation files are run
- Removed old CreateDocumentation project
- Added 2 new projects: APSIM.Cli and APSIM.Interop
  - APSIM.Interop sits between Models and ApsimNG, and contains mainly code to act as a go-between between Models/APSIM.Shared datatypes and those of upstream libraries (such as oxyplot, sharpmap, migradoc, markdig, etc)
  - APSIM.Cli could be removed if we don't want it. What I had in mind was a single command-line interface for all projects using verbs like `run` to run a file, or `document` to document a file (similar to `docker run`, `docker build`, etc)

Resolves #6443
Resolves #6793
Resolves #5606


hol430:
retest this please jenkins

hol430:
@hol353 - this is now starting to run green so I think we're getting close to being ready to merge. The autodocs are being generated as part of the main testing phase, which adds 5-10m onto the total build time. We may want to look into throwing more resources at the problem at some point.

You can see the generated documentation [here](http://apsimdev.apsim.info/ApsimX/Releases/2021.10.13.6443/index.html) (there should be a link in the build logs from now on). Is there anything you want to do/review/discuss before this is merged?

jbrider:
Soybean pdf has some empty graphs at the very end - not sure if that's from the file or an error.

hol430:
retest this please jenkins

Issue #6822, opened 2021-09-28
Need a Manager for checking mass balance errors


sno036:
Mostly as part of SWIM3 checking but likely more generically useful.

sno036:
@hol353  - when you need a break from the current engaging activity (!) would you mind looking at the Manger script "CalculateRollingWaterBalance"? It is in the simulation "SWIM3_Testing"I can easily add crops into the simulation and then the uptake components into the water balance calculation but it would be more useful if the component was able to find out what crops are in the simulation and do that automatically. 

This is more in the nice to have than absolutely needed now.

sno036:
This is complete for water (solutes are another issue) and now includes subsurface drainage.

PR #6860, opened 2021-10-14
Added uptakes to IPlant

hol353:
so manager scripts can be written with multiple plant models. Working on #6822


Models/Core/IPlant.cs
```diff
@@ -22,6 +22,12 @@ public interface IPlant : IModel
         /// <summary>Get above ground biomass</summary>
         IBiomass AboveGround { get; }
 
+        /// <summary>Daily soil water uptake from each soil layer (mm)</summary>
+        double[] WaterUptake { get; }
```
hol430, line 26: Is there a better way to do this? I don't really like having arrays in interfaces because even though you don't have a setter, you can still modify an element of the array from outside the class (or even do an `Array.Resize()` if you really wanted). This really causes problems and confusion in the soil water model.

I realise that you're probably just following existing conventions but I wonder if it would be better to use something like `IReadOnlyList<double>` here.

hol430 (approved):

PR #6878, opened 2021-10-18
Remove support for .net framework builds.

### LINKED ISSUE #6872: Stop targeting .net framework
Now that the autodocs can be built on .net core and without pulling in the gtk stack, we can stop targeting .net framework altogether. Which will reduce development overheads (particularly in the gui) and will speed up build times.

hol430:
Resolves #6872.


hol430:
Yep, what could go wrong!

PR #6884, opened 2021-10-19
'Fix' legends on graph

### LINKED ISSUE #6880: Graphs in new GUI are not showing legends

hol430:
This is a quick fix to get graph legends working again. It's more of a hack, but the bigger issue is that the graphing UI needs to be changed to take advantage of the graph -> oxyplot pipeline used for autodocs. I will create a separate issue for this, as it will be a bit of work to get it done, and in the meantime people probably just want their graph legends back.

Resolves #6880.


hol430:
retest this please jenkins

PR #6899, opened 2021-10-30
worldmodel url wants "." instead of ","

### LINKED ISSUE #6917: Weather downloader is not culture-sensitive
The weather downloader can fail under certain system-wide date/number formatting settings (e.g. if the default number format uses a comma for the decimal separator).

dexy58:
Fixes url when using NASA POWER w/ CHIRPS rainfall (World Modellers) radio button

When inserting latitude and longitude you have to use "," as decimal placement instead of "." (an error will pop up). However worldmodel webiste needs to have "." instead of "," in it's url.

Resolves #6917

### DIFF
--- a/ApsimNG/Utility/WeatherDownloadDialog.cs
+++ b/ApsimNG/Utility/WeatherDownloadDialog.cs
@@ -653,6 +653,10 @@ namespace Utility
                 ShowMessage(MessageType.Warning, "NASA/CHRIPS data end date can be no later than yesterday", "Invalid end date");
                 return null;
             }
+
+            entryLatitude.Text = entryLatitude.Text.Replace(',', '.');
+            entryLongitude.Text = entryLongitude.Text.Replace(',', '.');
+            
             string url = String.Format("https://worldmodel.csiro.au/gclimate?lat={0}&lon={1}&format=apsim&start={2:yyyyMMdd}&stop={3:yyyyMMdd}",
                             entryLatitude.Text, entryLongitude.Text, startDate, endDate);
             MemoryStream stream = WebUtilities.ExtractDataFromURL(url);
--- a/ApsimNG/Utility/WeatherDownloadDialog.cs
+++ b/ApsimNG/Utility/WeatherDownloadDialog.cs
@@ -1,4 +1,4 @@
-﻿
+
 namespace Utility
 {
     using APSIM.Shared.Utilities;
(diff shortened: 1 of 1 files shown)


ApsimNG/Utility/WeatherDownloadDialog.cs
```diff
@@ -653,6 +653,10 @@ public string GetNasaChirps()
                 ShowMessage(MessageType.Warning, "NASA/CHRIPS data end date can be no later than yesterday", "Invalid end date");
                 return null;
             }
+
+            entryLatitude.Text = entryLatitude.Text.Replace(',', '.');
```
hol430, line 657: Instead of doing a straight text replace here, I think it will be better to 

1. Convert the lat/long input strings to doubles (using the current culture) - this will account for things like ',' as decimal separator or '.' as thousands separator as may be the case on the user's system
2. Convert these parsed doubles back to strings (using the invariant culture) - this will generate a string in the american format (ie using '.' for decimal separator), which should be safe to pass into the API

ie something like this:

```csharp
double latitude = double.Parse(entryLatitude.Text, CultureInfo.CurrentCulture);
double longitude = double.Parse(entryLongitude.Text, CultureInfo.CurrentCulture);

string latitudeStr = latitude.ToString(CultureInfo.InvariantCulture);
string longitudeStr = longitude.ToString(CultureInfo.InvariantCulture);
// (and then use latitudeStr/longitudeStr as parameters to the climate API)
```

hol430 (changes requested):

ApsimNG/Utility/WeatherDownloadDialog.cs
dexy58: Hello :)

Do I have to commit this to the patch-1 branch on dexy58/ApsimX to make the change? I have never done this (this is the 1st time I am contributing to the open source project, so I don't know the procedure).
Thanks.

dexy58: Okay, I commited the changes in patch-1 branch on dexy58/ApsimX and it seems it has done the trick. It works, but I had to change a few other things because of those changes. Check them if you think we should revert it.

hol430:
retest this please jenkins

hol430 (approved):
Looks good - thanks for the contribution @dexy58!

PR #6900, opened 2021-10-30
Fixing if check

### LINKED ISSUE #6901: Weather download dialog doesn't check for longitude when clicking "ok"
You could have clicked "OK" without entering longitude (since it wasn't checking if anything was entered), which would result in an error.

dexy58:
Fixes if check when pressing "OK" button in download weather window.

You could have clicked "OK" without entering longitude (since it wasn't checking if anything was entered), which would result in an error.

Resolves #6901.

### DIFF
--- a/ApsimNG/Utility/WeatherDownloadDialog.cs
+++ b/ApsimNG/Utility/WeatherDownloadDialog.cs
@@ -153,7 +153,7 @@ namespace Utility
         {
             try
             {
-                if (!CheckValue(entryLatitude) || !CheckValue(entryLatitude))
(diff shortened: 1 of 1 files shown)


hol430:
add to whitelist jenkins

hol430:
Thanks for the contribution @dexy58! For future reference, your initial comment on the pull request needs to [reference an issue](https://apsimnextgeneration.netlify.app/contribute/sourcetree/pushandpullrequest/#2-jenkins-build-and-run-system).

hol430 (approved):

Issue #6112, opened 2021-02-01
Need options to suppress summary messages from several models


sno036:
including: supplement, fertiliser, irrigation. They insist on reporting everyday and it makes it nearly impossible to find actual information. I often use a 'verbose' setting on manager scripts to turn off messages like these. Can something like that be added to most of the models. By default verbose is on an all messages (as is at present) go to the summary. User can set verbose to off and only genuine warning messages and errors get sent? (or some way to filter out the dross?)

jbrider:
+1

peter-devoil:
There are many logging packages which would let us to do better than "Console.WriteLine...", though not much more difficult.The sooner we make that change, the better.

hol430:
Sounds like a good idea. That being said, there's already an option in the summary model to filter which messages are captured. Is this not sufficient?

peter-devoil:
At the moment, there are warnings, errors, and the informative messages that are overwhelming. We could replace these 3 with a severity index ranging from fatal (highest) to debugging (lowest) - informative being somewhere in the lower half. Then the filtering gets a bit more flexible, especially if we can filter whingeing components.

sno036:
The filtering that is there (actually, I am uncertain what it actually does - apparently nothing to the summary file I was looking at) does not filter out all the 'i did this today' messages that clutter up the summary file. It gets worse with multi-paddock simulations. A filter (in or out) by model (or field-model) would help. A filter that could also remove on a template (e.g. filter out all messages including the text "water added via irrigation") would be even better.

hol430:
Currently, all messages written to the summary have a log level associated with them (info, warning, or error). We could add a drop-down to the summary ui which restricts which messages are shown. We could also add more log levels (e.g. debug?). We could also add an option to the summary UI to filter which components' messages are shown. When running the simulation, all messages would still be recorded as normal - these changes would be mainly at the gui level.

Would that fix the problems? Or do we actually want to be able to control which messages are recorded when running the simulation?

jbrider:
So currently the options in the summary file only affect it the next time it is run?
Does that mean you have to change every summary file in files with multiple simulations?

I would assume that the option is there to be able to reduce the summary file size on large runs - so you would still need the option to restrict logging at runtime?

hol430:
[...]

PR #6920, opened 2021-11-03
Refactor summary and its UI

hol430:
Working on #6112

- Can filter messages based on severity (error, warning, info, diagnostic, all)
- Can control which messages are captured when sim is run (again based on severity)
- `ISummary` now has only a single method in which the caller specifies message severity (previously we had `WriteMessage()`, `WriteWarning()`, etc). I've written a converter to fix any manager scripts which use this API

Still to be implemented:

- Filter based on model name
- Tabular display of messages
- Control which messages are saved on a file-wide basis (currently can only control this on a per-summary basis)


lie112:
I think there should be a lot more "borked unit tests" (see last commit)

Can you explain what these new changes mean
- verbosity level
- just making sure all the links to summary in CLEM are still appropriate

hol430:
I've added a verbosity level in this PR, as suggested in #6112. `ISummary` now has only a single method:

```csharp
void WriteMessage(IModel model, string message, MessageType messageType);
```

Previously it had `WriteMessage()`, `WriteWarning()`, and `WriteError()`. Whenever you write to summary now, you provide the message type (error, warning, info, debug) as a parameter rather than as part of the method name. The idea being that we can easily add more verbosity levels in the future, and more verbosity levels means more fine-grained filtering of summary messages in the GUI. I could imagine some models in the future using a debug/diagnostic verbosity level to report their logic/decision making to users.

As part of this, I've done a find-replace over the repository to fix the summary API usage from all models (including clem). I've also written a converter function to fix manager scripts.

Not sure what you mean by "there should be more borked unit tests". Is there a test you'd like to see that would have failed on a change like this? I do agree though with the general principle of "there should be more unit tests"...

hol353:
Do you want this merged or is there more to do?

hol430:
This is good to go if you're happy with the approach and the implementation.

hol353 (approved):
I like it. I can see lots of grazing/cutting manager scripts using this to show diagnostic info rather than their current approach using custom manager code.

PR #6933, opened 2021-11-08
Fix NUptakeLayers

### LINKED ISSUE #6932: Root.NUptakeLayered is wrong
When summed is not the same as Root.NUptake

HamishBrownPFR:
resolves #6932


HamishBrownPFR:
@hol353 are you happy with this fix?  It is a reporting variable so has no impact on results.

hol353 (approved):

PR #6939, opened 2021-11-10
Forages

### LINKED ISSUE #6258: PMF doesn't have a mechanism to stop grazing of organs
Currently the Gliricidia plants in APSIM (and other tree models) will have their stems (trunks) grazed when livestock is in the same zone as the tree. PMF doesn't have a mechanism the model developer can use to stop this from happening.

Organs in the AgPasture model have a property specifying an ungrazable amount of biomass. This could be used to achieve non-grazing of trunks. This approach could be adopted by PMF. Is there a better mechanism?

hol353:
Resolves #6258.


hol353:
@hol430 I'm ok with the minor stats changes. This is ready for review / merge.

Models/AgPasture/PastureAboveGroundOrgan.cs
```diff
@@ -243,10 +249,11 @@ public void RemoveBiomass(OrganBiomassRemovalType biomassToRemove)
         {
             // The fractions passed in are based on the harvestable biomass. Convert these to
             // fractions of total biomass so that we can pass these to the tissue RemoveBiomass methods.
-            biomassToRemove.FractionLiveToRemove = MathUtilities.Divide(biomassToRemove.FractionLiveToRemove * DMLiveHarvestable, DMLive, 0);
-            biomassToRemove.FractionDeadToRemove = MathUtilities.Divide(biomassToRemove.FractionDeadToRemove * DMDeadHarvestable, DMDead, 0);
-            biomassToRemove.FractionLiveToResidue = MathUtilities.Divide(biomassToRemove.FractionLiveToResidue * DMLiveHarvestable, DMLive, 0);
-            biomassToRemove.FractionDeadToResidue = MathUtilities.Divide(biomassToRemove.FractionDeadToResidue * DMDeadHarvestable, DMDead, 0);
+            //biomassToRemove.FractionLiveToRemove = MathUtilities.Divide(biomassToRemove.FractionLiveToRemove * DMLiveHarvestable, DMLive, 0);
```
hol430, line 252: Is this still needed? If not, can it be removed?

Models/AgPasture/SimpleGrazing.cs
```diff
@@ -489,6 +503,11 @@ public void Graze(double amountDMToRemove)
         /// <summary>Add dung to the soil surface.</summary>
         private void AddDungToSurface()
         {
+
+            if (clock.Today == new DateTime(1995, 7, 8))
```
hol430, line 507: Debug code

Models/AgPasture/SimpleGrazing.cs
```diff
@@ -475,8 +484,13 @@ public void Graze(double amountDMToRemove)
...
             if (FractionPopulationDecline > 0)
             {
-                foreach (var forage in forages)
-                    forage.ReducePopulation(forage.Population * (1.0 - FractionPopulationDecline));
+                foreach (var forage in allForages)
+                {
+                    if ((forage as IModel) is IHasPopulationReducer populationReducer)
+                        populationReducer.ReducePopulation(populationReducer.Population * (1.0 - FractionPopulationDecline));
+                    else
+                        throw new Exception($"Model {(forage as IModel).Name} is unable to reduce its population due to grazing. Not implemented.");
```
hol430, line 492: Cast to IModel shouldn't be necessary, right? In fact how does this even work? I'm assuming that forage is a ModelWithDigestibleBiomass (I don't actually know because of the var keyword above...). ModelWithDigestibleBiomass doesn't implement IModel from what I can see so I would expect this to throw...

Models/ForageDigestibility/ModelWithDigestibleBiomass.cs
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        private class FractionDigestibleBiomass
+        {
+            public FractionDigestibleBiomass(DigestibleBiomass biomass, double frac)
+            {
+                Material = biomass;
+                Fraction = frac;
+            }
+            public DigestibleBiomass Material;
```
hol430, line 249: These should probably be properties (even though it's a private class). Might as well be explicit about the accessibility...

Models/PMF/Interfaces/IHasDamageableBiomass.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// Remove biomass from an organ.
+        /// </summary>
+        /// <param name="materialName">Name of organ.</param>
+        /// <param name="biomassRemoveType">Name of event that triggered this biomass remove call.</param>
+        /// <param name="biomassToRemove">Biomass to remove.</param>
+        void RemoveBiomass(string materialName, string biomassRemoveType, OrganBiomassRemovalType biomassToRemove);
+    }
+
+    /// <summary>A class to hold a mass of biomass and its digestibility.</summary>
+    public class DamageableBiomass
```
hol430, line 24: Can this class be moved into its own file? They can be hard to track down otherwise - and this class isn't even related to the interface mentioned by the file name.

hol430 (changes requested):

Models/AgPasture/SimpleGrazing.cs
hol353: Good catch. Yes the throw would have thrown (sounds weird).

hol430 (approved):

PR #6949, opened 2021-11-13
Remove inappropriate argument from call to GetMap, to get mapping wor…

### LINKED ISSUE #6891: Mapping is broken
Adding a map to the example Wheat simulation (and probably to anything else) gives garbage results.

zur003:
…king again on Windows. Resolves #6891.
I'm not sure why the overloaded version of GetMap with a "resolution" argument was ever called,
but it wasn't the right thing to do. When used, that argument is passed to System.Drawing.Bitmap.SetResolution, and attempts to set the DPI of the bitmap, which is clearly not what was intended.


zur003:
Hhmm. Maybe it was added for the maps generated for the auto-docs. This may need a bit more work....

hol353 (approved):

PR #6950, opened 2021-11-14
Turn off "UseTextShaping" when exporting plots to PNG.

### LINKED ISSUE #6871: 'Create Documentation' in gui causes a crash on linux
Appears to be caused by a segfault in harfbuzz:

```
#0  0x00007fff7c0ad6a8 in hb_shape_plan_create_cached2 () at /usr/lib/libharfbuzz.so.0
#1  0x00007ffa74343c34 in hb_shape_full () at .../ApsimX/bin/Debug/netcoreapp3.1/runtimes/linux-x64/native/libHarfBuzzSharp.so
#2  0x00007fff83ce3a36 in  ()
#3  0x0000000000c81e4a in  ()
#4  0x00007ffff7716728 in vtable for InlinedCallFrame () at /usr/share/dotnet/shared/Microsoft.NETCore.App/3.1.0/libcoreclr.so
```

zur003:
Resolves #6871, although I don't really understand why it needs to be done this way.

### DIFF
--- a/APSIM.Interop/Graphing/GraphExporter.cs
+++ b/APSIM.Interop/Graphing/GraphExporter.cs
@@ -60,8 +60,11 @@ namespace APSIM.Interop.Graphing
         {
             using (Stream stream = new MemoryStream())
             {
-
-                PngExporter.Export(plot, stream, (int)width, (int)height);
+                PngExporter exporter = new PngExporter();
+                exporter.Height = (int)height;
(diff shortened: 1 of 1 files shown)


zur003:
I think we should be able to remove the fonts from ApsimNG resources, as they are now in APSIM.Interop.

hol353 (approved):

PR #6964, opened 2021-11-19
Supply new svg icons for CLEM components.

### LINKED ISSUE #6945: New icons in APSIM ng
Congratulations on another GUI appearance upgrade with the new icons. 

What are the new icon standards in APSIM NG (#6855) ? I will start upgrading the CLEM suite of icons.

I realise icons can now be svg providing better scaling and display, but what is the target size of the various icon when displayed (recent list and tree view)? Once all icons are updated the treeview may need a larger gap between the new larger icons and the node label.

How does APSIM decide what icon to provide previously loaded simulations in the home tab. Some of my files seem to have been assigned icons (simulation icon and up arrow) based on the name of the icon found in the filename i.e.  "simulation" and "upgrade". This seems like a flimsy means of defining a recent file icon.

image

### LINKED ISSUE #6963: Update all CLEM icons to new svg standard
Need to recreate all CLEM icons as new 20x20 svg files
Some code changes are needed to ensure correct icons are displayed and UI looks good with 20px and old 16px icons

relates to #6855

lie112:
resolves #6963 
resolves #6945


ApsimNG/Presenters/ExplorerPresenter.cs
```diff
@@ -1120,12 +1121,10 @@ public static string GetIconResourceName(Type modelType, string modelName, strin
             }
             else
             {
-                (exists, resourceNameForImage) = CheckIfIconExists(modelType.Name);
+                string modelNamespace = modelType.FullName.Split('.')[1];
```
hol430, line 1124: I like the idea here of having the icons arranged in a similar directory structure to the model namespaces. However, I think it would be better to not hardcode index 1. This may work ok for namespace with exactly 2 parts (e.g. Models.CLEM) but many types live in either namespace with more or less parts (e.g. Models.Clock, most PMF types, etc). Perhaps it would be better to just remove the `Models.` prefix from the start of the string.

hol430 (changes requested):

ApsimNG/Presenters/ExplorerPresenter.cs
lie112: It was a little trickier as I had never assumed full namespace path after CLEM. I have removed the "Models." from the front and then used the first and last string split by "." 
This means all CLEM components are CLEM.[ModelType]. Others will be Core.[ModelType] and Memo etc that are under model work straight away as it strips "Models." and uses what is left, reducing the calls to CheckIfIconExists,
If you think the fuller namespace folder structure (e.g. CLEM.Activities.ActivitiesHolder) would be better than stripping out the intermediate parts let me know and I'll add folders to my icon resources and use the full namespace without the "Models" in this section.

hol430: - Probably better to just use string.Replace() instead of a regular expression. Much easier to read/understand that way
- If I'm reading this right, `modelNamespaceParts` actually includes the type name, which means that `modelNamespaceParts.Last()` will be the type name (e.g. Clock) - is this intentional? If so, this is not clear from the code
- Even if it were just fetching the last part of the namespace, the logic here is very hard to follow - `modelNamespaceParts.First()` means something different for types which live in the Models namespace to those which live in the Models.PMF namespace. I was about to type "the logic can be summarised as...." but I don't think I can summarise the logic without running this with a debugger.

What we want here is something simple to understand, and it would be nice if it resulted in a clean directory structure for the icons. If that means that we have directories like `.../TreeViewImages/CLEM/Activities/` and `.../TreeViewImages/PMF/Organs`, then I'm ok with that. We could check the `TreeViewImages` directory as a fallback, to support the existing icons. Thoughts?

lie112: I was ensuring "Models." is only ever removed from the start of the namespace. With .Replace we could lose a unintended icon path (Models.XXXX.QuickModels.Example). I thought that was as bad as assuming I wanted the second part of the namespace list using the [1] index to get CLEM that I previously had coded. We could see if the string starts with Models. and then remove the first 7 characters, but that requires a couple lines of code and two string operations.

Yes modelNamespaceParts includes the model type. This means models sitting directly under Models will only call the check icon method once. It got the functionality into a few lines of code, but can be more verbose if needed.  Memo is the only item in the part list as it sits under Model so would just look for a resource Memo.svg. For more complicated namespaces I was ignoring anything after the main model name and the type as that's how I threw all my icons in the resources\treeviewicons\clem folder. I had offered to put all the folders in and use the full namespace as the icon resource path. Given we now work with svg there is less need to have images and icons in different areas as they can be scales to be whatever we need them as.

I was trying to get what was previously working, working again given I am the only one separating the icons and wanted to avoid the incorrect allocation of some other apsim icons to clem models such as the orange ResourceHolder.

The cleanest approach would be for everyone to place icons in the namespace folder structure under resources, but I often have to work with what's already in place. I cant move all the icons as there are strange relationships back to the name people have given a node in the tree. I was also afraid there is a lot of other code the is not using this section to assign icons. There are some areas of code still assuming only png icons.

I need to get my current pull requests in as I have a couple other banking up as I get DARPA working and make more user validation to stop users putting in strange settings with changes I need to get out.  I have a pull coming to fix CLEM to work with the new messages as I previously had to write some clunky code to work with displaying messages cleanly and this all fell over and didn't display messages properly after the new improvements with MessageType. It would have been nice to assign index for the enum with a gap between them so additional types could be added in the sequence such ValidationError between error 0 and warning 1.

lie112:
I have applied all suggestions. CLEM icons are now in folders based on the namespace and there is a simple check of whether the namespace location of the icon (without Models.) is available, before searching using the other options of model type and model name only.

ApsimNG/Presenters/ExplorerPresenter.cs
lie112: namespace folder structure now applied to clem icons in resources

hol430 (approved):

PR #6966, opened 2021-11-22
Improves CLEM filtering performance

### LINKED ISSUE #6962: Filtering performance problems with very large herds
Large herds (10,000+ individuals) are taking a long time to run.
This should have been addressed with the recent refactoring of the filtering and use of IEnumerable throughout.
There are also a number of untrapped, invalid user entries that cause filtering to throw an error of little value to assisting the user to address the problem.

lie112:
Resolves #6962 
Updates CLEM to ensure the filtering expression rules are only created once at simulation commencing


lie112:
This has a revert of the exclude parent attempt, so now only has the changes relevant to this issue.

hol353:
I'm not completely comfortable with introducing a new property attribute [Filter] because what is to stop APSIM users and devs wondering what it is when they are writing manager scripts / code? It will appear in intellisense so they will see it and then start asking questions about what it is. Having an attribute that only works in certain CLEM classes and no where else seems wrong to me.

What's the reasoning for it? Is there another way to do this without the attribute?

lie112:
Ok. I have looked through all the other apsim and windows attributes to see if something already existed or I could use a current property. The only other option was to add a property to another attribute, but none really fit the task. I cannot think of another way of making particular properties in the code as being able to be used for filtering. The only other approach is to have a messy hard coded list that is difficult to update or apply to new models later. Would a different term be better? "filterproperty" "filterbyproperty" "useInFiltering"

lie112:
If the attribute is inside the CLEM namespace, will it not show up in intellisense unless the user has using.CLEM and would therefore have an understanding of the meaning. A rename would reduce the general confusion as well "FilterByProperty". At present I added it will all other APISM NG attributes.

hol353 (approved):

PR #6967, opened 2021-11-22
Wheat npi data

hut104:
working on #4773


hut104:
@hol353 I think this can be merged. @HamishBrownPFR  and I have worked to get this extra data and simulations into the test set as part of this project.
The results are different, mostly because we have much more data, and some new estimates of some cultivar phenology parameters.
Where results are red, we have generally have extra data.
Most statistics have been improved because of the wider testing.

PR #6975, opened 2021-11-23
Improved array support in expression parser

### LINKED ISSUE #6882: More mathematics options in Report possible?
Would it be possible to enable more mathematical functions in Report. Something like:

Log10(-1 * [Soil].Swim3.psi[1]) * -1

Perhaps also operating on arrays such as 

Log10(-1 * [Soil].Swim3.psi * -1   // this would apply the function element by element and produce a value for each layer

hol430:
Resolves #6882 

- Added (element-wise) array support to scalar report functions (like ln, sin, etc)
- Rename log() to log10() (+ converter function)
- Unit tests for expression parser


hol430:
retest this please jenkins

hol353 (approved):

Issue #5387, opened 2020-07-13
Wheat Validation is difficult to use in UI


hut104:
The wheat validation set is now sufficiently large that the UI is becoming slow to use, especially when accessing graphs.  There may be opportunities to speed up execution via improved indexing within the database for daily data.  Otherwise, a profiling exercise may provide insights on the best means for speeding up UI operations for this dataset.

PR #6983, opened 2021-11-29
Fix bug where the graph doesn't display due to exception

hol353:
Working on #5387

### DIFF
--- a/Models/Graph/SeriesDefinition.cs
+++ b/Models/Graph/SeriesDefinition.cs
@@ -310,6 +310,8 @@
                 {
                     var simulationIds = reader.ToSimulationIDs(simulationNameFilter);
                     var simulationIdsCSV = StringUtilities.Build(simulationIds, ",");
+                    if (string.IsNullOrEmpty(simulationIdsCSV))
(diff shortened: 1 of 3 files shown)


Models/Graph/SeriesDefinition.cs
```diff
@@ -310,6 +310,8 @@ public void ReadData(DataTable data, List<SimulationDescription> simulationDescr
                 {
                     var simulationIds = reader.ToSimulationIDs(simulationNameFilter);
                     var simulationIdsCSV = StringUtilities.Build(simulationIds, ",");
+                    if (string.IsNullOrEmpty(simulationIdsCSV))
```
hol430, line 313: What situation causes simulationIds to be empty here? Is this covered by the unit tests?

hol353: SimulationIds equals null when not all simulations have been run. The _Simulations table in the .db still has a record for the simulation but there is no data in the report table. I've added a unit test that reproduces this.

hol430 (approved):

PR #6986, opened 2021-12-03
Sorghum leaf

jbrider:
working on #572


hol353:
The stats are different. Can you provide advice on why they are difference and whether you're happy with them?

hol353:
I came across a PlantAvailableWaterSupply method in Root. It seems to be sorghum specific but isn't used anymore. Can you please remove it if not needed?

jbrider:
The changes are due to a fix made to the phenology water stress which resulted in changes to the phenology in some water stressed simulations - 2 of which no longer reach harvest.  This was an expected outcome that we are happy with. The same change will be made to Classic as well.

jbrider:
@hol353 This would be a good time to merge.
If it's only comments now, then I will be working on them and improving the documentation before it gets released.

hol353 (approved):

PR #6988, opened 2021-12-04
Added CLI run functionality for apsim.dll

### LINKED ISSUE #6987: Missing CLI run functionality for apsim.dll
Run method for apsim.dll CLI isn't implemented.

dexy58:
Fixes #6987

### DIFF
--- a/APSIM.Cli/Program.cs
+++ b/APSIM.Cli/Program.cs
@@ -66,10 +66,6 @@ namespace APSIM.Cli
             foreach (string file in files)
             {
                 Simulations sims = FileFormat.ReadFromFile<Simulations>(file, e => throw e, false);
-                if (Path.GetExtension(file) == ".json")
-                {
-                    sims.Links.Resolve(sims, true, true, false);
-                }
 
                 Runner runner = new Runner(sims);
                 List<Exception> errors = runner.Run();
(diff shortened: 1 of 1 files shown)


dexy58:
@hol430 Hello, can you please check why TestSendDataTable test fails? I can't seem to figure it out (I haven't even touched any of the methods this Test calls.

hol430:
Maybe a race condition in the test...

retest this please jenkins

dexy58:
@hol430 Thank you :)
All tests have passed now.

APSIM.Cli/Program.cs
```diff
@@ -53,7 +54,31 @@ private static void HandleParseError(IEnumerable<Error> errors)
...
+            {
+                files = options.Files;
+            }
+            foreach (string file in files)
+            {
+                Simulations sims = FileFormat.ReadFromFile<Simulations>(file, e => throw e, false);
+                if (Path.GetExtension(file) == ".json")
+                {
+                    sims.Links.Resolve(sims, true, true, false);
+                }
```
hol430, line 72: I don't think we need to resolve links for .json files here. I think we only do that in the `Document()` function as a bit of a hack, but it shouldn't be necessary for running files - only for documenting them.

```suggestion
```

hol430 (changes requested):
Thanks for the contribution, @dexy58!

APSIM.Cli/Program.cs
dexy58: Commited this change. Tested it and it works without this part.

hol430 (approved):

Issue #6984, opened 2021-11-30
Use docker to wrap calls to R


hol430:
Some of our upstream R packages are known to be unstable. This problem could be solved by having apsim use a docker container with known-stable versions of R packages to run CroptimizR/Morris/Sobol.

peter-devoil:
It's a bit late to mention this, but the good folks at microsoft invented a "[checkpoint](https://mran.microsoft.com/documents/rro/reproducibility)" package that allows the R user to specify packages at a snapshot in time.

hol430:
We tried using [packrat](https://rstudio.github.io/packrat/) which I think does the same thing. We still had problems - particularly with the version of R which is installed on the user's computer. It appears that what's stable in one version of R is not necessarily stable in another version (particularly around the 4.0.0 version). There were also some problems with packages disapparing from cran, with windows vs linux, etc. Docker solves most of these problems, though granted it does introduce some of its own with regards to file/path management.

hol430:
Closing as fixed.

PR #6989, opened 2021-12-07
Allow apsim to spin up docker containers to run CroptimizR simulations

hol430:
Working on #6984.

I've made docker the default R engine for CroptimizR simulations only, but the user can revert to old behaviour by setting environment variable `APSIM_NO_DOCKER=1`. Should this be the other way around?

I've created an `IR` interface to abstract away the comms with R. This is implemented by the new `RDocker` class, but not by the old `R` class (which runs R code "natively"). It probably wouldn't be too hard to refactor the original R class so that it can implement the interface, but things still wouldn't quite be transparent to the caller; you need to be really careful with regards to file management when using docker.

When running a croptimizr simulation via docker, all dependent files (such as .met, .xlsx, etc) need to be in the same directory as the .apsimx file, and cannot be referenced via the %root% macro.


hol430:
retest this please jenkins

hol353 (approved):

PR #7013, opened 2021-12-15
Fixed crash caused by adding initial water to GUI.

### LINKED ISSUE #7012: Crash after adding initial water in GUI
Adding an initial water node to the gui will cause a crash via a stack overflow exception while attempting to calculate PAW.

hol430:
Resolves #7012 

@hol353 - how is this supposed to work? Has this ever worked?


hol353 (approved):

hol353:
I thought this used to work. Maybe it didn't

Issue #6780, opened 2021-09-07
Property presenter drop-down pop-up. Autocomplete?


lie112:
The pop-up for the drop down widget in property presenter needs to have a max height and maybe a top at the bottom of the parent widget. At present it will go to screen height with large lists of options filling any extra room with white space (as seen in the image below on a portrait monitor). 

As it is only the  move up and move down arrow icons on the pop-up that allow navigation to the option you require, it is slow to select an option in large lists as there's no scroll bar handle.

Could I request someone with Gtk# skills create an auto complete selector that allows the user to start typing to select the value from the list. This is not a combo box allowing any value, but must select an item in the list

The pop-up should therefore start at the date entry widget and go down for a max height with maybe a scroll bar to make looking around the list faster for the user.

![image](https://user-images.githubusercontent.com/19215637/132419664-f52afa26-7dbf-48bd-ac19-a9d197a602b5.png)

par456:
At some point this was improved, dropdown goes from the box position to the bottom of the screen, has a scrollbar and mouse support. Closing this issue unless I missed something.

PR #7029, opened 2021-12-21
Graph panel improvements

### LINKED ISSUE #6942: Graph panel performance
The graph panel component is not taking full advantage of the GraphPage's ability to read data from storage for multiple graphs simultaneously. (Currently we create one GraphPage per graph, when we could/should create one GraphPage for all graphs.)

hol430:
- Make better use of new GraphPage API
- Speed up axis standardisation

Resolves #6942.
Working on #572.

### DIFF
--- a/ApsimNG/Menus/ContextMenu.cs
+++ b/ApsimNG/Menus/ContextMenu.cs
@@ -143,6 +143,8 @@
             {
                 GraphPanel panel = explorerPresenter.CurrentNode as GraphPanel;
                 panel.Cache.Clear();
+                explorerPresenter.HideRightHandPanel();
+                explorerPresenter.ShowRightHandPanel();
             }
             catch (Exception err)
             {
--- a/ApsimNG/Presenters/ExplorerPresenter.cs
+++ b/ApsimNG/Presenters/ExplorerPresenter.cs
@@ -245,27 +245,20 @@
 
         public bool FileHasPendingChanges()
         {
-            try
-            {
-                // Need to hide the right hand panel because some views may not save
-                // their contents until they get a 'Detach' call.
-                this.HideRightHandPanel();
+            // Need to hide the right hand panel because some views may not save
+            // their contents until they get a 'Detach' call.
+            this.HideRightHandPanel();
 
-                // Check the command history (beta feature - see comment in the
-                // property description for more details).
-                if (Configuration.Settings.UseFastFileClose)
-                    return CommandHistory.Modified;
+            // Check the command history (beta feature - see comment in the
+            // property description for more details).
+            if (Configuration.Settings.UseFastFileClose)
+                return CommandHistory.Modified;
 
-                // The fallback is to write the file to json then compare to the
-                // file on disk.
-                string newSim = FileFormat.WriteToString(ApsimXFile);
-                string origSim = File.ReadAllText(ApsimXFile.FileName);
-                return string.Compare(newSim, origSim) != 0;
-            }
-            finally
-            {
-                this.ShowRightHandPanel();
-            }
+            // The fallback is to write the file to json then compare to the
+            // file on disk.
+            string newSim = FileFormat.WriteToString(ApsimXFile);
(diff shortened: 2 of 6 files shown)


jbrider:
@hol430 I'm still getting crashes if I exit the program while I have a GraphPanel selected.
As it hasn't been crashing for you, I wonder if it might have something to do with checkpoints?

hol430:
I fixed *a* problem which occurred when closing the program while a graph panel was selected - whether it was the same one you were getting I don't know :). Are you still getting the crashes on this branch? If so, do you have a stack trace, or any Gtk messages in the stdout/stderr?

I did some testing with checkpoints - but wasn't looking for crashes at the time. Will give it another go now.

jbrider:
Just traced it back to a workerthread - looks like it was firing after the close - does that sound like the error?
I'll test this branch to see if it fixed it.

hol430:
Yep - that sounds exactly like the one I fixed on this branch.

jbrider:
That's fixed the crashing on mine.
It seems to have lost the ability to draw the legend above the graph - actually a lot of the properties are being ignored and I'm not sure if the clear cache is working (difficult to test as the properties don't seem to be working).

hol430:
I've just fixed the above.

After you clear the cache, would you expect it to immediately start regenerating the panels? Or would it be better to do nothing until the user explicitly clicks away and clicks back onto the graph panel?

jbrider:
I would have it immediately refresh - I can't think of a situation where you click that and don't want it to immediately update.

jbrider:
Is it possible to remove the 'Double click to add a caption' to get back a little more screen real estate?

hol430:
Ok, I've changed both of those. Let me know if you notice anything else you reckon could be improved.

I don't think the graphs load massively faster than before (though they around 20% faster on my machine) but the long pause after they've loaded should be gone.

jbrider:
That's not crashing now - so that's a big win.

The drop down on the series view show as full screen if there's a long list - is it easy to put a max height on it?

The performance within those dropdowns can be pretty slow - the graph panel ones seem to work harder than most. I assume they're looking up all the data in the sims - is there some way to short circuit/cache that do you think?

hol430:
Yep, it's possible to cache the data - although the first time load will still be pretty slow. But is it useful to render the data in the series views at all? It's not really legible - although I guess it tells you that your config is sensible (after a fashion).

jbrider:
Nope - not useful at all for GraphPanel - I figured the slow part was searching simulation names/variables not the data loading.
It might be useful for normal graph editing?

hol430:
Caching the series data in between edits of the series is going to be problematic. E.g. if you change the y-series, you've just invalidated the cache. What's worse, all of the existing graph caching code deals with SeriesDescription instances - which track colour, line type, etc in addition to the actual x/y data. So in other words, almost any change made to the series in the UI would invalidate the cache.

Which isn't to say that it would be impossible to improve on what it's doing currently - but it wouldn't be simple, and I'd argue that it would be beyond the scope of this PR. Actually there's a lot that could be improved in the graphing code - it probably needs another iteration, but that's a big job.

As for making it easier to edit the graphs under the panel - I've changed the graphs so that they don't display anything if they're under a graph panel, which should still make things easier for this particular use case.

jbrider:
Can you look at the height of the drop downs on the graph form - they can extend the full height of the screen and can make finding variables pretty slow. Can provide more info if needed/ open another issue.

hol430:
Yeah, that's a limitation of the GtkComboBox widget, which doesn't expose any options for controlling the max height (or tooltips or scrollbars!) for the drop-down menu. I guess the gtk folks would argue that ComboBoxes shouldn't really be used with so many dropdown items. Maybe they're right - but I'm not sure what the better input method would be in that case.

I agree this is a problem though - it's also relevant to the property UIs, such as surface OM. This problem is tracked by #6780.

hol353 (approved):

Issue #7030, opened 2021-12-22
about  the weather and soil input 


xiangyi-wang:
1. When I create the weather input(.met format), I need to add a blank line at the end so the model can read to the last day.  Otherwise, one less line will be read.
2. The layer depth input does not work, instead soil thickness recalculation depth is shown in the depth field. And is shown as 28.90000002 when the value is 28.9. Of course, this is just writing a small problem.It's just not pretty.
![image](https://user-images.githubusercontent.com/22881957/147049930-43ccb100-9da2-40b7-a9fe-e561c964ae64.png)
![image](https://user-images.githubusercontent.com/22881957/147050150-4dd207f5-cc35-4cb5-8a8c-f864b3980747.png)

I'm using APSIM 2021.12.6988.0

hol430:
The depths shouldn't be written to the .apsimx file at all - that's a bug, which I've fixed in #7031. The large number of decimal places I've fixed in #7032.

The requirement of an empty line at the end of met files appears to be intentional. @hol353 is this correct? This comes back to `StreamReaderRandomAccess.ReadLine()`, which returns an empty string if it reaches EOF before a LF character. This is consistent with other ReadLine() methods (such as `StreamReader.ReadLine()`), but may not be so convenient to the average user.

hol353:
There shouldn't be a requirement for a blank line at the end of a met file. Sounds like this needs to be fixed.

PR #7031, opened 2022-01-03
Soil depth should never be serialized to json.

hol430:
(Thickness is serialized instead.)

Working on #7030.


hol353 (approved):
Ahh I didn't see this PR before my recent comment in PR #7032.

PR #7032, opened 2022-01-04
Alleviate floating point arithmetic problems in soil depth strings

hol430:
Working on #7030

@hol353 - this isn't ideal, but it does help. Let me know what you think.

### DIFF
--- a/APSIM.Shared/Utilities/SoilUtilities.cs
+++ b/APSIM.Shared/Utilities/SoilUtilities.cs
@@ -133,6 +133,10 @@
                     double ThisThickness = Thickness[i] / 10; // to cm
                     double TopOfLayer = DepthSoFar;
                     double BottomOfLayer = DepthSoFar + ThisThickness;
+
+                    TopOfLayer = Math.Round(TopOfLayer, 3);
+                    BottomOfLayer = Math.Round(BottomOfLayer, 3);
+
                     Strings[i] = TopOfLayer.ToString() + "-" + BottomOfLayer.ToString();
                     DepthSoFar = BottomOfLayer;
                 }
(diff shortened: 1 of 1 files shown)


APSIM.Shared/Utilities/SoilUtilities.cs
```diff
@@ -133,6 +133,10 @@ public static string[] ToDepthStrings(double[] Thickness)
                     double ThisThickness = Thickness[i] / 10; // to cm
                     double TopOfLayer = DepthSoFar;
                     double BottomOfLayer = DepthSoFar + ThisThickness;
+
+                    TopOfLayer = Math.Round(TopOfLayer, 3);
+                    BottomOfLayer = Math.Round(BottomOfLayer, 3);
+
```
hol353, line 139: This will give depth strings 3 decimal places. Given that this code converts from mm to cm by dividing by 10 there will only ever be 1 decimal place. Change this to 1 decimal place.

hol353 (changes requested):

APSIM.Shared/Utilities/SoilUtilities.cs
hol430: Well, thickness is a double, so it can be fractions of a mm (see #7030). But if you think that's not really useful for UI purposes I could change it to 1 decimal place.

hol353: Soil thickness is never a fraction of a mm. In issue #7030, the Thickness property doesn't have fractions of a mm. Go for 1 dec place. As an aside, I notice the JSON in that issue shows we are serialising the 'Depth' property. No need to do this.

hol430: Ok, I've just changed it to only show 1 figure after the decimal point, although people will still be able to input numbers with more points if they want to.

Serialization of depth is fixed in #7031

hol353 (approved):

PR #7043, opened 2022-01-11
Fix graph annotation colours in dark mode

### LINKED ISSUE #5346: Get regression data by simulation as an output
I created a factorial between Nitrogen and Water levels which generates 9 simulations. I am comparing the predicted values to observed values. Then, I created an XY graph with regression data by SimulationName. I would like to export (e.g. txt/csv file) this data or see it in APSIM in some way that allows to identify which Simulation had the best fit to the 1:1 line. I can see the values in the graph but I cannot identify the corresponding SimulationName for each regression data. @hol430 
Untitled

hol430:
Resolves #5346.


hol430:
retest this please jenkins

hol353 (approved):

PR #7044, opened 2022-01-11
Allow summary to write messages during the Completed event

### LINKED ISSUE #6714: Ability to write to Summary in OnCompleted

hol430:
Resolves #6714.

### DIFF
--- a/Models/Summary.cs
+++ b/Models/Summary.cs
@@ -89,18 +89,6 @@
             CreateInitialConditionsTable();
         }
 
-        /// <summary>Invoked when a simulation is completed.</summary>
-        /// <param name="sender">Sender of the event</param>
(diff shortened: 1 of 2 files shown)


hol430:
retest this please jenkins

hol353 (approved):

PR #7046, opened 2022-01-11
Improve focus/selection handling in property UI

### LINKED ISSUE #7045: New property UI does not behave correctly after hitting the tab key
In most UIs, hitting tab will cause focus to shift to the next input element. In our property UI, hitting tab will cause the current input element's value to be saved, but focus does not move to the next input element - instead, the current element just loses focus. (Focus doesn't move to any input element.)

hol430:
Resolves #7045.


hol430:
retest this please jenkins

hol353 (approved):

Issue #6257, opened 2021-03-16
Run current simulation with F5 from any model


lie112:
A user has asked if a quick run F5 to run can be allowed from any node selected rather than having to select a Simulation or Simulations node before right clicking and selecting run or hitting F5. I guess a smart system would know what level of run is to be performed (e.g. if an experiment is present, run the experiment)

hol430:
Yep, it's not too high on the priorities list at the moment, but it would be nice. Related is #3810.

PR #7071, opened 2022-01-31
Global hotkeys

### LINKED ISSUE #3810: More hotkeys would be nice
Would be helpful to have hotkeys for some of the global actions - e.g.

ctrl + s - save
ctrl + shift + s or f12 - save as
ctrl + z - undo
ctrl + y - redo

Maybe something for split screen mode and to clear the status bar as well.

A pretty low-priority issue, just something which would be nice.

hol430:
Resolves #3810 (but not #6257).

I've added global hotkey support to all items in the main menu - these are the buttons along the top of the screen when a simulation is open (save, save as, undo, redo, ....). So you can, for example, press control + s to save your file from anywhere (ie while editing a manger script).

Note that this doesn't affect items in the context menu at all (ie run simulation). I think if we want f5 to run a simulation from anywhere, the easiest solution would be to move this button into the main menu, similar to old apsim. (The progress bar can stay where it is.)

This works fine in split screen mode, but the code to get this working is a bit hacky. Basically, when we receive an event from one of the accelerators (keyboard shortcuts), we determine which tab currently has focus, and send the event on to the corresponding ExplorerPresenter instance. I can't really think of a fundamentally better way of doing this which supports split screen mode.

Currently there's no option the user to customise the key bindings, so let me know if you want me to change any of the defaults.


hol353 (approved):

byzheng:
Would be nice if there is a documentation for all supported hotkeys.

PR #7087, opened 2022-02-10
Fixed server backlog CLI option initialiser

hol430:
Working on #6529.


hol430:
As part of this, I've also:

- Added apsimng-server to the list of docker images built in every release
- Added version metadata to the APSIM.Server assembly
- Removed the version component from AssemblyTitle attribute (this is why we were previously getting the version outputted twice when running `Models.exe --version`). AssemblyTitle is not intended to hold version info.
- Moved AssemblyTitle from AssemblyVersion.cs -> AssemblyInfo.cs
- Changed ApsimNG's AssemblyTitle from "APSIM" to "APSIM User Interface"

hol353 (approved):

PR #7096, opened 2022-02-12
Clem grid

### LINKED ISSUE #6358: Problem with ReportView rows not lining up.
This was found in the SimpleGrazing example:

image

zur003:
Resolves #6358, at least in the context of the CLEM ActivityLedgerGridView. Are the old Gtk Treeview based grids still in use anywhere else?

### DIFF
--- a/ApsimNG/Views/CLEM/ActivityLedgerGridView.cs
+++ b/ApsimNG/Views/CLEM/ActivityLedgerGridView.cs
@@ -741,15 +741,22 @@ namespace UserInterface.Views
         /// <param name="args"></param>
         private void RefColWidget_SizeAllocated(object o, SizeAllocatedArgs args)
         {
-            headerHeight = refColWidget.AllocatedHeight;
-            for (int i = 0; i < numberLockedCols; i++)
+            try
             {
-                Fixedcolview.Columns[i].Widget.HeightRequest = headerHeight;
+                headerHeight = refColWidget.AllocatedHeight;
+                for (int i = 0; i < numberLockedCols; i++)
+                {
+                    Fixedcolview.Columns[i].Widget.HeightRequest = headerHeight;
+                }
+                // We have the value we need, so we can clean up. There's no real need to return, as we don't expect
+                // the column headers to change.
+                refColWidget.SizeAllocated -= RefColWidget_SizeAllocated;
+                refColWidget = null;
             }
-            // We have the value we need, so we can clean up. There's no real need to return, as we don't expect
-            // the column headers to change.
-            refColWidget.SizeAllocated -= RefColWidget_SizeAllocated;
-            refColWidget = null;
+            catch (Exception err)
+            {
+                ShowError(err);
+            }
         }
 
         /// <summary>Get screenshot of grid.</summary>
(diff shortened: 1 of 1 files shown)


ApsimNG/Views/CLEM/ActivityLedgerGridView.cs
```diff
@@ -723,6 +732,26 @@ public void LockLeftMostColumns(int number)
...
+            headerHeight = refColWidget.AllocatedHeight;
+            for (int i = 0; i < numberLockedCols; i++)
+            {
+                Fixedcolview.Columns[i].Widget.HeightRequest = headerHeight;
+            }
+            // We have the value we need, so we can clean up. There's no real need to return, as we don't expect
+            // the column headers to change.
+            refColWidget.SizeAllocated -= RefColWidget_SizeAllocated;
+            refColWidget = null;
+        }
```
hol430, line 753: Might be worth wrapping all this in a try...catch block - otherwise, any exceptions thrown here will result in a crash.

hol430 (changes requested):

hol430:
I think the TreeView-based grids are still used for some CLEM components, the morris/sobol components, the soil profiles, XYPairs, TreeProxy, Weather, and maybe some others I've forgotten about. I'd like to see them drop out of use altogether in favour of the new SheetWidget control. We'll need to implement editable cells before that can happen, but that shouldn't be too difficult and once that's done it should be easy enough to migrate to SheetWidget.

zur003:
OK. Note that this PB, though I've flagged it as resolving, only resolves the problem for the CLEM ActivityLedgerGridView. Any others using the "locked" columns hack (are there any?) will continue to be at risk of this problem, so the Issue may need to re-opened.

@hol430  A quick check appears to indicate that TablePresenter and DualGridView are no longer in use anywhere. Can we delete them?

hol430:
I think the other usages should all be from GridView.cs - I *think* that's what all those other components are using.

TablePresenter and DualGridView aren't directly referenced from code as such - but they are referenced indirectly from the Models project - e.g. in Forages.cs, we have a `[ViewName("UserInterface.Presenters.TablePresenter"]`, which will be resolved at runtime via reflection, to create a TablePresenter instance.

lie112:
I believe "ActivitiesPerformedReport" is the only CLEM presenter using the old treeview grid as it is  a hangover from building the original version from the old grid presenter. I was aware of changes happening, but hadn't done an upgrade as it was low on my priority list while working.
All I need is the ability to fix x left hand columns, have the ability to replace the cell content with an associated image, and not make the columns too wide based on columns header names (columns the with if image with a vertically aligned heading text would be great).
It was a bit of a hack as I had to fix the image size and padding to make the table look ok and wasn't doing any modification of APSIM presenters at that stage. From memory, it was a copy of the apsim code so all the issues detected must have been in the old code as well and been missed in fixes over time. 
I have removed any such duplicated text in custom CLEM components by using approaches such as Report as a base class. 
I had thought about using a custom font to display the icons if I could specify a character and colour for each entry. This would also be scalable rather than working with the pixel based images to display. All the images are one colour.  Again, it was low in my priority list.
Once there is a sheet widget control version of grid view, I can easily move to this with a bit of added functionality or a few "new" methods and allow the removal the old treeview table from APSIM NG. I do not need editable cells.
Adam

hol430:
Well the SheetWidget is already available and working (it's used by report/datastore). Not sure if it supports all the features you need yet, but may be worth a look.

Anyway, I had a quick look and this bug doesn't appear to affect GridView, so I'll merge this now. Thanks Eric!

hol430 (approved):

PR #7133, opened 2022-02-23
Add ability to call overloaded methods

### LINKED ISSUE #7132: The operations model doesn't handle overloaded methods well.
The operations model only looks at the first matching method name and doesn't consider the arguments. e.g.

AgPasture.PastureSpecies has: 

public void RemoveBiomass(string organName, string biomassRemoveType, OrganBiomassRemovalType biomassToRemove)
public void RemoveBiomass(string type, double amount)
public Biomass RemoveBiomass(double amountToRemove)

If the user tries to call the second method from an operations schedule it will fail as it only looks at the first method in Agpasture.

hol353:
(e.g. RemoveBiomass in AgPasture) from operations.

Resolves #7132

### DIFF
--- a/Models/Management/Operations.cs
+++ b/Models/Management/Operations.cs
@@ -165,8 +165,9 @@ namespace Models
         {
             // convert arguments to an object array.
             ParameterInfo[] parameters = method.GetParameters();
+            int numMandatoryParameters = parameters.Where(p => !p.HasDefaultValue).Count();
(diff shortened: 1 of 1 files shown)


hol353:
retest this please jenkins

hol353:
@hol430 can you review and merge if ok?

hol430 (approved):

Issue #7111, opened 2022-02-18
"Object not set..." when clicking away from a model which uses an editor component


hol430:
When clicking away from a model which uses a CodeView in the UI (ie report, manager, cultivar, ...), a null reference exception is generated.

sno036:
@hol430  - is this in the highest priority list to fix?

zur003:
@hol430 - The problem here seems to involve some dark corners of Gtk# object destruction, but I think at least part of the problem is in IntellisenseView.cs where we have in the Cleanup routine:

>             if (completionForm.IsRealized)
>                 completionForm.Dispose();
>             completionView.Dispose();
>             completionForm.Dispose();
>             completionForm = null;


Note that this can result in completionForm.Dispose() being called twice, which doesn't seem like a good thing to do.

But that's not the whole problem, and there are bits I don't fully understand. I find that if I open the Standard Toolbox and scroll through the list of cultivars, that instances of EditorView are being created, but seemingly not destroyed. I find it all rather confusing.

hol430:
Yep - the problem appeared when we updated the version of the GtkSharp package (which also included an update of the native libs as well).

There may be multiple problems here - the one I observed was at CodeView.cs when we attempt to remove some event handlers from an object which has already been disposed of. I suspect something has changed in the object destruction in gtk as you say. It may be that we were relying on undefined behaviour in the past. Perhaps we need to, in general, detach event handlers from gtk objects sooner (ie before the Destroyed event is fired).

zur003:
And then there is the rather confusing distinction between "Dispose" and "Destroy". As Gtk evolves, it appears to be heading away from the use of "Destroy" (I think), but the relationship between the two is fuzzy and seems to sometimes change with different versions of Gtk. I may experiment a bit with overriding GObject.Dispose rather than relying on the Destroyed signal for doing memory cleanup, and see what happens.

hol430:
From the docs for [gtk_widget_destroy](https://developer-old.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-destroy):

> It's important to notice that [gtk_widget_destroy()](https://developer-old.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-destroy) will only cause the widget to be finalized if no additional references, acquired using g_object_ref(), are held on it. In case additional references are in place, the widget will be in an "inert" state after calling this function; widget will still point to valid memory, allowing you to release the references you hold, but you may not query the widget's own state.
[...]

PR #7136, opened 2022-02-24
Updated "DetachHandlers" logic to work with newer version of Gtk#.

zur003:
Working on #7111

### DIFF
--- a/ApsimNG/Utility/GtkUtilities.cs
+++ b/ApsimNG/Utility/GtkUtilities.cs
@@ -62,8 +62,9 @@ namespace Utility
         /// </summary>
         /// <param name="widget">The widget</param>
         /// <param name="signalName">Name of the signal</param>
-        public static void DetachHandler(this Widget widget, string signalName)
+        public static bool DetachHandler(this Widget widget, string signalName)
         {
+            bool result = false;
             PropertyInfo signals = typeof(GLib.Object).GetProperty("Signals", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static);
             FieldInfo afterHandler = typeof(GLib.Signal).GetField("after_handler", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static);
             FieldInfo beforeHandler = typeof(GLib.Signal).GetField("before_handler", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static);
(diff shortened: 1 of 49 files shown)


zur003:
Why did this fail? Retest this please, Jenkins

zur003:
@hol430 - Don't commit yet! I just noticed that although menu handlers are being cleared, the menu items themselves are not.

zur003:
Still a problem in the build system.
Retest this please, Jenkins

hol430:
Thanks Eric, looks good to me. Are you planning to push some more commits to this branch, or are you happy for this to be merged?

ApsimNG/Utility/GtkUtilities.cs
```diff
@@ -21,34 +21,69 @@ public static void DetachAllHandlers(this Widget widget)
...
+        /// Normally it's better to use the "-=" operator to do this, but it may be that we
+        /// don't actually know the method for the right side of the operator. This uses
+        /// reflection to find that method and detach it. As explained in the DetachHandlers 
+        /// method above, this routine may break in future versions of Gtk#.
+        /// </summary>
+        /// <param name="widget">The widget</param>
+        /// <param name="signalName">Name of the signal</param>
+        public static void DetachHandler(this Widget widget, string signalName)
+        {
+            PropertyInfo signals = typeof(GLib.Object).GetProperty("Signals", BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static);
```
hol430, line 67: Actually, I wonder if some of the duplicate code (between `DetachHandler()` and `DetachHandlers()`) could be extracted to their own methods. E.g. we could probably maintain `static readonly` fields for signals, afterHandler and beforeHandler, as they're not specific to a particular widget instance. We could also have extension methods for a GLib.Object which returns the delegeate for the after handler or the before handler. That might help to eliminate some of the duplicate logic.

zur003:
@hol430 - There will be a lot of additional changes to come, although mostly in the area of implementing IDisposable in ViewBase, and using _view_.Dispose to close views rather than _view_.MainWidget.Dispose or _view_.MainWidget.Destroy. I'd be happy to see this signal handler detachment logic committed separately, as they are somewhat separate (though closely related) issues.

zur003:
Although this set of changes should resolve the initial issue, it is worth noting that there still appears to be a lot of memory "leakage" in the GUI. This may be difficult to fix. Gtk# has its own reference-counting system beyond normal .Net reference counting. Letting both systems know when an object is no longer needed can be a challenge. Explicitly calling "Dispose" for GLib objects seems to help much of the time, but it's not particularly clear when this needs to be done, and being overly aggressive with use of "Dispose" seems to lead to its own set of problems. The complexity of the Apsim GUI, with views nested within views within views, can make it quite difficult to track down where leaks may be occurring. And as some of the problems might be regarded as "bugs" in Gtk#, it is quite possible that any approach taken to address issues under the current version of Gtk# might require revision in the future.

hol430:
Hi @zur003, this looks good and memory usage seems to grow much more slowly with these changes. I've been testing the branch locally and have found a few issues, which I think existed before your changes as well. I've pushed them to this branch - hope that's alright.

There's certainly still a lot of leaks in the GUI - I'm planning to fire up a memory profiler some time and see if I can use that to get to the bottom of this. Or at least fix some of them.

For my own edification, what sort of problems did you observe when being overly aggressive with calls to Dispose()?

zur003:
@hol430 - I don't recall exactly what problems I had - I was tinkering with a range of Dispose/Destory combinations and got the occasional crash. It _may_ have just been the double-Destroy problem that you described, but I thought I had something similar after too many calls to Dispose.
Gtk# doesn't handle cleaning stuff up very well. It's not at all clear when one needs to `Dispose` or `Destroy` object and widgets, and whether (or when) a call to DIspose will invoke Destroy or vice-versa. And top-level Windows seem to play by different rules from other WIdgets. It's a real mine field. Couple that with the complexity of the Apsim GUI, with it's frequent creation and destruction of nested interface elements, and you've got a real challenge getting things right.

hol430:
@zur003 are you happy for this to be merged, or were you planning to make more commits before this goes in? I'd be happy to accept this as-is, these changes seem to greatly improve the stability and rate of leaking memory in the GUI.

zur003:
@hol430 - You can merge it now. I probably won't get back to looking at memory leak problems for a while. Anyway, I believe this resolves the original issue (#7111), so it may be best if further work is treated as a new issue.

hol430 (approved):

PR #7150, opened 2022-03-04
Create an APSIM Server protocol for sending strings and dates

AJChapman:
Context: I have implemented a client, but couldn't receive variables that were dates or strings. This fixes that, at least for me.

Working on #6529

### DIFF
--- a/APSIM.Server/IO/NativeCommunicationProtocol.cs
+++ b/APSIM.Server/IO/NativeCommunicationProtocol.cs
@@ -204,7 +204,7 @@ namespace APSIM.Server.IO
 
         private void SendMessage(string message)
         {
-            byte[] buffer = Encoding.Default.GetBytes(message);
+            byte[] buffer = Encoding.UTF8.GetBytes(message);
             PipeUtilities.SendToPipe(connection, buffer);
         }
 
@@ -247,7 +247,7 @@ namespace APSIM.Server.IO
             byte[] buffer = PipeUtilities.GetBytesFromPipe(connection);
             if (buffer == null)
                 return null;
-            return Encoding.Default.GetString(buffer);
+            return Encoding.UTF8.GetString(buffer);
         }
 
         private void SendArray(Array data)
@@ -296,7 +296,7 @@ namespace APSIM.Server.IO
 
         private IEnumerable<byte> StringBytes(string s)
         {
-            return Encoding.Default.GetBytes(s);
+            return Encoding.UTF8.GetBytes(s);
         }
 
         private IEnumerable<byte> StringBytesWithLength(string s)
--- a/APSIM.Server/IO/NativeCommunicationProtocol.cs
+++ b/APSIM.Server/IO/NativeCommunicationProtocol.cs
@@ -250,28 +250,16 @@ namespace APSIM.Server.IO
             return Encoding.Default.GetString(buffer);
         }
 
-        private void SendInt(int value)
         {
-            byte[] buffer = BitConverter.GetBytes(value);
-            PipeUtilities.SendToPipe(connection, buffer);
         }
 
-        private void SendDouble(double value)
         {
-            byte[] buffer = BitConverter.GetBytes(value);
-            PipeUtilities.SendToPipe(connection, buffer);
         }
 
-        private void SendBool(bool value)
         {
-            byte[] buffer = BitConverter.GetBytes(value);
-            PipeUtilities.SendToPipe(connection, buffer);
         }
 
-        private void SendDate(DateTime value)
         {
-            // tbi - need to give this one some thought.
-            throw new NotImplementedException();
         }
 
         private void SendArray(Array data)
--- a/APSIM.Server/IO/NativeCommunicationProtocol.cs
+++ b/APSIM.Server/IO/NativeCommunicationProtocol.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Globalization;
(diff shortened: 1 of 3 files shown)


AJChapman:
Another option for transferring dates would be to use an int. This would be much more efficient, but is perhaps less robust at identifying an exact date between implementations. E.g. if we use Unix epoch then a 32-bit signed int will only go back to 1901 and forward to 2038. If we only want date resolution (and not a time within each date), then an integer-of-date representation might serve, e.g. 20220307 to represent 2022-03-07.

hol430:
add to whitelist please jenkins

APSIM.Server/IO/NativeCommunicationProtocol.cs
```diff
@@ -250,52 +255,69 @@ public string ReadString()
...
-            else if (arrayType == typeof(string))
-                return data.Cast<string>().SelectMany(i => Encoding.Default.GetBytes(i)).ToArray();
-            else
-                throw new NotImplementedException();
+            return StringBytesWithLength(DateToString(date));
+        }
+
+        private IEnumerable<byte> StringBytes(string s)
+        {
+            return Encoding.Default.GetBytes(s);
```
hol430, line 304: I wonder if it would be better to use a specific encoding here (e.g. Encoding.Utf8). Otherwise, the result of this method may depend on the system configuration of the machine on which the server is running, which may not be the same as the machine on which the client is running.

AJChapman:
I've changed the date representation to integer-of-date, as I mentioned previously, in the latest commit.

APSIM.Server/IO/NativeCommunicationProtocol.cs
AJChapman: Yes, that's true. The IntBytes and DoubleBytes are also susceptible to this, in that the endianness will change depending on the machine, and maybe the number of bytes in each as well.

hol430:
Looks like one of the recent commits on this branch has caused a deadlock in one of the unit tests (probably somewhere in NativeSocketConnectionTests.cs). These test methods really should have a timeout attribute but the problem also probably indicates either a problem with the changes on this branch, or a need for the test code to be updated to reflect these changes.

APSIM.Server/IO/NativeCommunicationProtocol.cs
AJChapman: I've added a commit to send strings as UTF-8.

The endianness issue is still there though, and extends to the use of PipeUtilities, which assumes that the other end of the pipe is talking the same endianness. If a client on a big-endian machine talks to a server on a little-endian machine then it will all go horribly wrong. By convention, network order is generally big-endian. So we could fix this by reversing the endianness when sending and receiving ints and doubles. This would be inefficient when running the client and server on the same machine, but maybe not materially so. Another way to fix it would be to have the client and server negotiate the endianness at the start of the session. Any preference, @hol430?

hol430: I'd be happy to always enforce a particular endianness, or to negotiate one at start of session as you say.

hol430:
retest this please jenkins

APSIM.Server/IO/NativeCommunicationProtocol.cs
AJChapman: With my latest changes, little-endian is always enforced.

AJChapman:
These changes are no longer backward-compatible: I changed sending of fixed-length types (int, double, bool, date) to no longer prefix the message with their length. My reasoning being that both ends must already know the type of the value being sent, otherwise the client will be confused regardless. And they both know the length of the value (4, 8, 1, 4). So this saves sending redundant information over the wire.

Is this a problem? I can change it back if backward compatibility is important. I'm not sure how much this is used in the wild yet.

I've also added in a protocol version command, so the client can check what version the server is running (for future breaking changes).

Another feature of my latest changes is that much of the protocol has moved into PipeUtilities from NativeCommunicationProtocol. This way both client and server can use a common encoding and decoding of each type. Some further refactoring might make this a bit neater, but it's working for now.

I also fixed the NativeSocketConnectionTests on Windows. It previously had a hard-coded /tmp path that Windows doesn't have. I'm hoping that my changes are portable to Linux -- I didn't use any Windowisms. Note that there are still a few broken unit tests on Windows, but at first glance these don't seem to be related to my changes.

Finally, please do comment if I'm not doing things the idiomatic C#/.Net way. This is my first time using either technology.

hol430 (approved):
Thanks @AJChapman, these changes will be good to have. I'll hold off on merging this until I update my client library (hopefully this week sometime), in order to keep the latest versions in sync.

Out of interest, what language is your client in? Is it on github?

Also worth noting, if you're planning on doing some simulation work with the server feature, you need to ensure your simulations are clearing their state correctly after each run. Normally apsim will handle this when running from the GUI/CLI, but when using the server, the assumption is that all of the models can reset/clear their state between runs. You can check if your simulations are doing this correctly by right clicking on the simulation node in the gui and selecting "check server compatibility".

AJChapman:
Thanks for the tip on clearing the simulation state.

My client is in Haskell: https://github.com/reinventingscience/apsim-params

I'm starting to think I may not need this though, if factorial simulations can do what I need.

hol430:
Yep, the server protocol isn't necessarily advantageous for most workloads, and the factorial feature is another way (the main way) to re-run a simulation with modified parameter values.

Nonetheless, thanks for your work on this, it's good to tighten up the protocol a bit.

PR #7156, opened 2022-03-07
Added conditional reporting frequencies to report model. 

### LINKED ISSUE #7155: Add conditional reporting ability to REPORT
To avoid having to create manager scripts to control REPORT, we need the ability to do conditional reporting. If we could put an if statement into the reporting frequency box then we can remove a reporting manager script from our SWIM simulations.

I also notice that you can't scale a x axis time series chart. You get an exception.

hol353:
and fixes bug in scaling date axes. Resolves #7155

### DIFF
--- a/ApsimNG/Interfaces/IAxisView.cs
+++ b/ApsimNG/Interfaces/IAxisView.cs
@@ -68,12 +68,6 @@
         bool CrossesAtZero { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether the axis is a date time axis.
-        /// This is not editable by the user.
-        /// </summary>
-        bool IsDateAxis { get; set; }
-
-        /// <summary>
         /// Sets the text in the minimum textbox.
         /// </summary>
         /// <param name="value">Value to display.</param>
--- a/Models/Report/Report.cs
+++ b/Models/Report/Report.cs
@@ -127,8 +127,8 @@ namespace Models
             foreach (string line in EventNames)
             {
                 if (!DateReportFrequency.TryParse(line, this, events) &&
-                    !ExpressionReportFrequency.TryParse(line, this, events, compiler) &&
-                    !EventReportFrequency.TryParse(line, this, events))
+                    !EventReportFrequency.TryParse(line, this, events) &&
+                    !ExpressionReportFrequency.TryParse(line, this, events, compiler))
                     throw new Exception($"Invalid report frequency found: {line}");
             }
         }
--- a/Models/Report/Report.cs
+++ b/Models/Report/Report.cs
@@ -26,12 +26,6 @@ namespace Models
     [ValidParent(ParentType = typeof(CLEMFolder))]
     public class Report : Model
     {
-        /// <summary>
-        /// A collection of functions that cause a line of output to be written when they
-        /// evaluate to true (1).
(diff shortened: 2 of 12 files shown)


ApsimNG/Interfaces/IAxisView.cs
```diff
@@ -67,12 +67,6 @@ public interface IAxisView
         /// </summary>
         bool CrossesAtZero { get; set; }
 
-        /// <summary>
```
hol430, line 70: Commited this on the wrong branch?

hol353: This PR resolves 2 problems as reported in the referenced issue.

Models/Report/Report.cs
```diff
@@ -136,7 +153,35 @@ protected void SubscribeToEvents()
...
+        private void AddFunction(string functionString)
+        {
+            var match = Regex.Match(functionString, "if (?<expression>.+)");
+            if (match.Success)
+            {
+                var expressionFunction = new CSharpExpressionFunction()
+                {
+                    Expression = $"Convert.ToDouble({match.Groups["expression"].Value})",
+                    Parent = this
+                };
```
hol430, line 178: I don't like this - does it really make sense to report on `if [Wheat].Leaf.LAI`?. I really think it would be better to use a boolean expression here. I get that it's been done this way because `CSharpExpressionFunction` is an `IFunction` and provides/assumes a double return value. I guess one way to improve this might be to extract a new class (`CSharpExpression`?) from `CSharpExrpressionFunction`, such that `CSharpExpression` doesn't assume any particular return type.

As more of a general observation, I think the report model has grown too complicated (not helped by hacks like this), and some modularity would really improve things here. Of course, that's probably beyond the scope of this change...

hol353: I guess you could do ```if [Wheat].Leaf.LAI``` with my code as it stands but my use case is for something like ```if Clock.Today >= new DateTime(2011, 2, 21)```. If could remove the if keyword if you like but that makes it hard to detect what is an expression and what is an event name or a date.

I could create a  CSharpExpression function that has this:
```c#
public CSharpExpression : CSharpExpressionFunction
{
        public CSharpExpression(string expression)
        {
                Expression = $"Convert.ToDouble(expression)";
        }

        public bool Value(int arrayIndex = -1)
        {
            return base.Value(arrayIndex);
        }
}
```

I'm not sure this is much better. I agree about the need for a refactor / rethink of report!

hol353:
@hol430 Any idea why this failed on Jenkins?

hol430:
retest this please jenkins

hol430 (approved):
This really highlights the need for a rethink of the reporting infrastructure.

lie112:
These changes appear to break previous functionality. 
I have a reporting frequency 
`[Resources].Ruminants.Bos indicus.ConceptionStatusChanged`
and when it is passed through EventReportFrequency.TryParse, the tokens are created based on splitting by space and it's not a date or expression report frequency. 
This now disallows models with a name containing a space (i.e. "Bos indicus") from subscribing their events.
Do I keep this thread going or is this a new issue?

hol430:
I'd say new issue - but something we definitely want to fix.

Issue #7207, opened 2022-04-01
Fix some issues in Plantain


rcichota:
Noted a couple of issues with the forage Plantain model while trying to simulated multi species sward
 - root initialisation is wrong, initial Wt is too high as it is multiplied by population twice (caused because root is initialised different than other organs)
 - the accumulation of harvested values is wrong in one set of simulations, summing one day after defoliations...

Furthermore, for simulation with AgPasture the height of plants cannot vary, so I am fixing here for now (should have none or very limited effect as height is not used for anything other than competition - we'll need to address the issue of height effect, and calculation of intercepted radiation by considering both AgPasture and MicroClimate)

rcichota:
Resolved by pull request #7209

PR #7209, opened 2022-04-01
Issue/fix plantain sim#7207

rcichota:
working on #7207


rcichota:
@hol353 
I have fixed a couple of errors in the model (a major one was the initial rootWt, which was two order of magnitude off!)
This has changed the results, as expected, but they are fairly small. I did look into it and made a couple of minor changes to the simulation, but could not get better than before (it is in one dataset but not in the others - there are only three datasets). I am happy with it as it is for now, for a fix in errors. If you agree you can please merge this. I will be working on the model over the next few months, adding more validation data, so hopefully the validation will actually improve soon.

rcichota:
Correction. when using the Report to type PlantainForage and then 'dot', I get access to a number of variables, but not all, missing the organs... And I cannot visualise the model structure...
If I paste the json file in the replacements I can see all the structure, and the simulation works...
Must be something little I am missing

hol353 (approved):

Issue #6435, opened 2021-05-07
Barley Phenology


hut104:
As part of the GRDC-funded National Phenology Initiative, new cultivar specifications, and an improved phenology model will be developed for Barley. 
The model will be based on controlled environment studies, field verefication, and genetic information.

hut104:
We need the first comment to link this to the issue I started in Github by adding the following

working on #6435

PR #7211, opened 2022-04-03
Barly ce

hut104:
working on #6435


hut104:
@hol353 These stats are failing simply because we have committed a lot of extra data to the test set.  I think you can merge these.

hol353 (approved):

PR #7214, opened 2022-04-04
Barley harvest at flowering

hut104:
working on #6435


hut104:
@hol353 This seems to be OK now.  The stats are different only because we have added more observed data.

hol353 (approved):

PR #7218, opened 2022-04-07
Classic import fix

### LINKED ISSUE #7238: Import of APSIM Classic file with more soil layers than N measurements fails
An import of an APSIM Classic file fails if e.g. the number of NO3 values does not match the number of soil layers. I encountered this in a whole series of files from an APSIM Classic power user, in which they had 11 soil layers but had only populated the top six layers with NO3 and NH4. These files failed to import, with no error message to give the user a clue as to why. In their research they did not need these deeper soil nitrogen values to be anything other than 0.

### LINKED ISSUE #7239: Import does not remember directory for next time
It is tedious to import multiple files from the same location because we have to do it one at a time and navigate back to that location each time. It would be much more convenient if the import saved the last opened location, as the 'open' feature does.

AJChapman:
This allows importing of APSIM Classic files in which the number of e.g. NO3 values does not match the number of soil layers. I encountered this in a whole series of files from an APSIM Classic power user, in which they had 11 soil layers but had only populated the top six layers with NO3 and NH4. These files failed to import, with no error message to give the user a clue as to why. In their research they did not need these deeper soil nitrogen values to be anything other than 0. I saw that there was already code to tolerate missing values in MapConcentration(), so have enabled this during the import.

Resolves #7238

I have also made a minor UI change, to set the last opened location after an import. This makes it much easier to import multiple files from the same location. Without the change, the user has to repeatedly navigate back to the location of the files they want to import.

Resolves #7239

### DIFF
--- a/ApsimNG/Presenters/MainPresenter.cs
+++ b/ApsimNG/Presenters/MainPresenter.cs
@@ -1141,12 +1141,13 @@
         {
             try
             {
-                string fileName = this.AskUserForOpenFileName("*.apsim|*.apsim");
+                string fileName = this.AskUserForOpenFileName("APSIM Classic Files|*.apsim");
                 this.view.ShowWaitCursor(true);
                 this.Import(fileName);
 
                 string newFileName = Path.ChangeExtension(fileName, ".apsimx");
                 this.OpenApsimXFileInTab(newFileName, this.view.IsControlOnLeft(sender));
+                Configuration.Settings.PreviousFolder = Path.GetDirectoryName(fileName);
             }
             catch (Exception err)
             {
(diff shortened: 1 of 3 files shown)


hol353:
@hol430 - how did this run without a resolves or working on in the top comment box?

@AJChapman - Can you create an issue with the text you have in the top comment box and then reference that issue with a 'resolves' or a 'working on'. See https://apsimnextgeneration.netlify.app/contribute/sourcetree/pushandpullrequest/

hol430:
retest this please jenkins

hol430:
Looks like a bug introduced during some of my recent refactoring of the jenkins scripts when I moved everything to the new jenkins server. I've fixed the problem now, although in the longer run it might be nice to have ApsimBot automatically post a comment on the pull request with some instructions.

Either way, I've just fixed the problem - the message/instructions are in the jenkins log. Let me know if you want the wording changed or anything like that.

AJChapman:
Created issues #7238 and #7239 that this PR resolves. Also rebased against the latest master.

hol430 (approved):
Looks good, thanks @AJChapman

Issue #7139, opened 2022-02-24
Current .Net version of APSIM NG


lie112:
Does the new approach adopted to move to .Net Core through the upgrade of gtk# version allow APSIM Next Gen to keep up with .Net releases?
I've noticed we're still in C# 8.0 using core 3.1, and I keep trying to use features of C# 9.0 and above.
Is it still a large job to move to the next release with gtk being the hold up or does the new compatibility of Core allow this easier?
We could benefit from the performance improvements in .Net 5 including to system.json.text serialisation, garbage collection and container image size.

hol430:
There is some work involved in changing framework versions, mainly related to the installers, and making sure that users have the required runtimes installed on their machines.

hol430:
I'm going to reopen this. Ubuntu 22.04 was released recently and does not officially support .net core 3.1 (only .net 6+). If we want to support the latest Ubuntu, we'll need to target .net 6.

zur003:
This may be a bit harder to do than would appear at first glance. With APSIM.Shared I tried a simple change of TargetFrameworks to net6.0, but compilation failed because use of BinaryFormatter to serialise/deserialise XML is now considered obsolete, as is some of the web access stuff like HttpWebRequest and FtpWebRequest. I tried changing BinaryFormatter to XMLSerializer, but that failed because it won't serialise Interfaces (like IModel). So there's going to be a bit of work in going to Net 6 beyond just worrying about the runtimes.

hol430:
I guess we do it iteratively. GtkSharp does support .net 6 though, so at least we don't have that headache again.

Moving to .net 6 would probably also force some people to update their dev environments. I'm not sure which older versions of visual studio support .net 6 development.

Another problem is related to building all projects into the same bin directory. This causes problems when publishing in self-contained mode, and it can also cause inconsistent warnings (and, potentially, errors, though I've never seen that) about locked files when building the solution. I'm unable to publish the solution in self-contained mode at all with the .net 6 sdk (whereas with the .net core 3.1 sdk, I can publish but there are problems with the resultant files).

zur003:
[...]

PR #7276, opened 2022-05-09
Small changes to assist with compilation under .Net 6 - handling of W…

zur003:
…indows-specific code and never-true condition.

Working on #7139

### DIFF
--- a/Models/CLEM/ZoneCLEM.cs
+++ b/Models/CLEM/ZoneCLEM.cs
@@ -448,15 +448,23 @@ namespace Models.CLEM
                     htmlWriter.Write("</div>");
                     htmlWriter.Write("\r\n<div class=\"defaultcontent\">");
                     htmlWriter.Write("\r\n<div class=\"activityentry\">This simulation runs from ");
-                    if (clk.StartDate == null)
-                        htmlWriter.Write("<span class=\"errorlink\">[START DATE NOT SET]</span>");
-                    else
+                    try
+                    {
                         htmlWriter.Write("<span class=\"setvalue\">" + clk.StartDate.ToShortDateString() + "</span>");
+                    }
+                    catch (Exception)
+                    {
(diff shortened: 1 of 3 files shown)


lie112:
Hi @zur003 
Can you explain why we need to use the try catch rather than a simple if else statement to trap the clock.StartDate and EndDate being null outside a running simulation? Is it because clock can be null, or because there's other issues with .ToShortDateString. Would clock?.StartDate have solved the issue. I've always avoided try catch for this operation and why doesn't this relate to other locations in the code?

hol430:
Clock.StartDate will never be null, because it's not a nullable type. Therefore `clock.StartDate == null` will never evaluate to true.

zur003:
Also note that the Clock StateDate and EndDate properties raise an Exception if any nullable values they attempt to reference are in fact null. Your "if-else" logic would never have worked, as the cases you were trying to catch would have raised exceptions that weren't being handled.

lie112:
This code is run at design time not during the simulation, thus the search in the tree for the first clock model.
Therefore if someone has just added a new Clock model to the simulation it will have null Start and End date until supplied in the property presenter. Therefore I was trapping this situation.

zur003:
@lie112 - But StartDate and EndDate can never be null; they are of type DateTime, not DateTime?. What you can do to check for unintialised values is, rather than test them against null, test them against the default value for DateTime.  So something like if (clk.StartDate = default(DateTime)) might be what you want. (The default value for DateTime is apparently DateTime.MinValue, which is 1 January 0001, and unlikely to appear in most simulations.)

zur003:
Thinking about this a bit more, and looking at how the StartDate and EndDate properties are implemented in Clock.cs, perhaps you just want to use the nullable Start and End properties rather than the non-nullable StartDate and EndDate properties when testing for unintialised values.

zur003:
Closing this, as I've included stuff that isn't going to work under 3.1.

Issue #7289, opened 2022-05-18
Error in setting tissue turnover rates in AgPasture


rcichota:
The turnover rate for the organ tissues is set up using an array with four values, but only the first is being used for all tissue ages. This has a considerable impact on biomass detachment when temperature is high but soil moisture is low (they should counter each other, but currently they add up)...

par456:
Was resolved by branch #7290

Issue #7225, opened 2022-04-12
Maintaining Carbon Balance


MarkLieffering:
There seems to be an issue with the reported values of GPP and NPP: GPP minus RespirationLossC should equal NPP

![image](https://user-images.githubusercontent.com/32279744/162863238-049e1c2c-9945-4f5a-a442-2aed428187a1.png)

The error comes from lines 1759 and 1766 of PastureSpecies.cs where GPP with units of kgC/ha is calculated from grossPhotosynthesis (also KgC/ha) but is then divided by CarbonFractionInDM (0.4) which converts the kgC to kgDM (which is incorrect). 

public double GPP
        {
            get { return grossPhotosynthesis / CarbonFractionInDM; }

Note that GrossPotentialGrowthWt has the same value as GPP - they use the same formula in the code.

For NPP there is a similar issue (note that respirationGrowth and respirationMaintenance are also in kgC/ha

public double NPP
        {
            get { return (grossPhotosynthesis - respirationGrowth - respirationMaintenance) / CarbonFractionInDM; }

hol353:
Some more context here. The above described problem is about trying to do a whole of simulation carbon balance using AgPasture. After fixing a defect in nutrient / soil organic matter (#7445), the daily carbon balance pseudocode looks like:

Daily change in carbon = start of day carbon (nutrient.TotalC.Sum() + surfaceOrganicMatter.C + plant.TotalC)
+ AGPRyegrass.NetPotentialGrowthAfterNutrientWt
- nutrientBalance.Catm.Sum()
- end of day carbon (same calculation as for start of day)

This should equal zero each day of the simulation.

Issue #7224, opened 2022-04-12
Kill AgPasture RespirationLossC


MarkLieffering:
In AgPasture, when ryegrass and white clover are killed through the EndCrop() operation (i.e. there is no pasture there) the output RespirationLossC is retained from the last day that the pasture was there. A screenshot probably explains it better:

![image](https://user-images.githubusercontent.com/32279744/162862602-91bad6c2-d4cf-4788-96ab-981b949838a7.png)

par456:
Was resolved by branch #7290

PR #7290, opened 2022-05-18
Fix a few AgPasture bugs

rcichota:
Working on #7224
working on #7225
working on #7289

Fixing a few bugs reported by @MarkLieffering plus a couple I identified (and created) in my last pull request
 - ensure respiration outputs are zero after killing a plant
 - correct units for GPP, NPP, NAPP, and NBPP
 - fix bug related to tissue turnover rates


MarkLieffering:
Thanks @rcichota !

hol353:
@rcichota Yesterday Neil made a fix to SWIM suction routine to try and improve SWIM's stability. I've merged that change and done a pull on your branch to include Neil's fix. I've then gone and put SWIM3_testing.apsimx back into your simulation and it now runs without SWIM crashing which is good.

This branch now seems to be failing in a manager example: 

Examples\ManagerExamples\RegressionExample.apsimx

This example has a sensibility test that does a regression between irrigation amount and harvested wt of agpasture. It throws if the slope of the regression line is < 0. I'm not sure why we have this test. It might just be an example of how to write a post simulation test in which case you can probably change it to something else to get Jenkins to pass.

rcichota:
Thanks @hol353 I got the error message for that slope test early in the week and was wondering what it was for...
I need to get this AgPasture stuff merged in, it is getting messy. I had done a few fixes and added outputs that your commit overwrote. No big deal (and the question about the use of carbon content in detachment rate made me check it. It is correct (from the original code - EcoMod) and I will reinstate it / rewrite as it is part of converting digestibility...).
Where I am strugling is that all my changes tend to make AgPasture (in the test simulation) produce too much. A lot of it is because the way the soils were set up, so I need to verify them... and there are many...

sno036:
@rcichota - I would be OK with temporarily taking the Radcliffe sites out of the test set and concentrating on the higher quality data at Winchmore and Ruakura. Would that assist progress?

MarkLieffering:
@sno036 , @rcichota. Would the long term Flock House data set be of use? There are a few gaps when we were maintaining the system and did not graze.

sno036:
retest this please Jenkins

rcichota:
@hol353 
This can now be merged. I still wat to add some datasets to the validation, but ran out of time this week. Will try get them early next week and be done with this. At least until Val and Mark find if there is an error in the C balance...
BTW, I have removed four simulations, as they will be added as 'detailed' simulations and test on their on right (I was using two of them - plus another set from a newer trial - as baseline in my changes, need a bit of clean up and I will add them back next week)
Resolves #7224
Resolves #7225
Resolves #7289
Resolves #7292

hol353:
Hi @rcichota This PR deletes SWIM3_testing.apsim. We don't want to do that! Can you undo the deleted file?

rcichota:
@hol353 
This issue with the SWIM simulation was an odd one, I did check and hope that was the only one...
I have updated the multispecies simulation as well as added a few more of the datasets I have been promising
Resolves #7325

rcichota:
Hi @hol353 
Are you happy with the state of this or do you see were I can make further amendments?
If you think things are ok, could you merge it, please. I would like to update our version of APSIM in PowerPlant and run some simulations with the fixes I've made...
Thanks

PR #7299, opened 2022-05-23
Maintain initial Plant population when using 'seeds'

### LINKED ISSUE #7298: Sowing parameter 'seeds' severely affects yield
Plant.Sow() has an optional "seeds" parameter which can be used in place of the normal "population" allowing the application of a seed mortality rate during the Emerging phase. While the seed mortality rate works, using this parameter seems to severely decrease crop yields. I believe this is caused by references to `SowingParameters.Population` in other components, which remains 0 when 'seeds' is used.

Below is a comparison between two basic Soybean simulations using the "population" and "seeds" parameters.

**Using the regular "population" parameter:**
image

**Same simulation, using the "seeds" parameter instead:**
image

codeandkey:
This PR adds an additional assignment in Plant.Sow() to keep SowingParameters.Population synchronized with the 'seeds' parameter when it is present.

Another fix could be to replace `SowingParameters.Population` references with `IPlant.Population`, however I have not been able to fix the issue doing this. I found relevant references in models:
- `PMF/Structure/Tillering/C4LeafAreaM.cs`
- `PMF/Structure/Tillering/C4LeafArea.cs`
- `PMF/Structure/Tillering/FixedTillering.cs`

Resolves #7298

### DIFF
--- a/Models/PMF/Plant.cs
+++ b/Models/PMF/Plant.cs
@@ -397,7 +397,7 @@ namespace Models.PMF
             if (population > 0)
                 this.Population = population;
             else
-                this.Population = seeds;
+                this.Population = SowingData.Population = seeds;
 
             // Find cultivar and apply cultivar overrides.
             Cultivar cultivarDefinition = FindAllDescendants<Cultivar>().FirstOrDefault(c => c.IsKnownAs(SowingData.Cultivar));


hol353:
add to whitelist jenkins

PR #7308, opened 2022-05-26
Added folders and updated scripts in manager toolbox.

hol353:
Also Added water table bottom BC to swim. 

Working on #1254


hol353:
retest this please jenkins

Issue #3467, opened 2019-01-25
New Canola Model


hut104:
A model is to be developed for _Brassica napus_ (Canola) for use in APSIM Next Generation.  This will use test data from APSIM version 7.10, but incorporate new phenological understanding from current research projects.  New data will be incorporated from current canola agronomic projects.  
The model is to be developed using PMF.
The phenology model will be very similar to the canola model in APSIM classic, but with the improvements included in the Brassica oleracea var. italica (Broccoli) model in APSIM classic.  
Datasets from APSIM Classic test set are being converted to NG and new datasets from existing CSIRO projects are also to be added.  The model will use the newer approaches currently in NG for canopy development etc.  Oil modelling will benefit from current developments for soybean.
Involvement of CSIRO canola project leaders will ensure that the model is developed for current emerging areas, such as links to genetic work.

sarahcleary:
@JackieOuzman

hut104:
Please find the attached presentation and auto-documentation file (both in PDF format) for the canola model review (7-June-2022).  Note that the autodocumentation software appears to have a few deficiencies at this point in time.

[APSIM NG Canola Model Presentation.pdf](https://github.com/APSIMInitiative/ApsimX/files/8840960/APSIM.NG.Canola.Model.Presentation.pdf)

[Canola.pdf](https://github.com/APSIMInitiative/ApsimX/files/8840970/Canola.pdf)

jbrider:
Little things...
- Grazing example has a full copy of the Model rather than referencing replacements?
- In Phenology in MidGrainFilling and LateGrainFilling there is a ThermalTimeNew.
  - As they are identical, would it be better to have a VariableReference to a ThermalTimeNew (or better name) that is located with the other ThermalTime related variables at the root of Phenology?
- In Leaf there are 3 Plastochron's - remove the 2 extra ones?
- In Leaf -> Branching Rate there is a memo that has "From Hammer Paper"
- In Stem there is a disabled DMDemandsOLD
- Quite a few empty Memo's.

jbrider:
In Validation
- Phenology Experiments - Omics Phenology - Gatton 2019 - Phenology
  - Selecting Phenology shows several blank charts
  - EmergenceDAS has -ve values?
  - GreenBudDAS has -ve values?

- Phenology Experiments - Omics Phenology - Leaf Appearance PvO
  - Leaf Appearance appears to have a hard limit of 8?

hut104:
[...]

PR #7313, opened 2022-05-31
fixed harvest date

hut104:
working on #3467


hut104:
@hol353 Not sure how a change to canola.apsimx caused a swim test to fail.

PR #7328, opened 2022-06-07
Canola

### LINKED ISSUE #3467: New Canola Model

hut104:
Resolves #3467


hut104:
@hol353 The recent changes improve many of the test data (e.g. LAI, yield, etc) with minor affects to stem wt and biomass.
However, the behaviour of winter cultivars is greatly improved.  I suggest that these results be accepted even though there is a slight impact on some predictions.

PR #7341, opened 2022-06-15
Fixed bug in debian installer generator script

### LINKED ISSUE #7339: Release builds are broken
When a file in examples directory (e.g. Canola Grazing.apsimx) has a space in it's name, Jenkins fails with:

```
rsync: [sender] link_stat "/tmp/ApsimX/Examples/Canola" failed: No such file or directory (2)
rsync: [sender] link_stat "/tmp/ApsimX/Setup/netcoreapp3.1/debian/Grazing.apsimx" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1333) [sender=3.2.3]
```

It appears to come from the debian build.

hol430:
Resolves #7339.

### DIFF
--- a/Setup/netcoreapp3.1/debian/buildDebianInstaller
+++ b/Setup/netcoreapp3.1/debian/buildDebianInstaller
@@ -35,7 +35,7 @@ then
 fi
 mkdir -p "$apsimx_inst"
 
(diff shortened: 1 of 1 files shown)


hol353:
Ah so that's how you fix it. Thanks for the PR Drew.

PR #7361, opened 2022-06-29
Soil initialisation refactor

### LINKED ISSUE #7170: We need a redesign of how solutes are specified and initialised in the GUI
We probably need a sprint to 

1. Redesign how solutes (Urea, NH4, NO3, CL and others) are initialised in the GUI. Currently they are specified in multiple places (sample, chemical). We need one way of specifying them. 

2. When SWIM is in a simulation, solutes appear in multiple places (with the same name). e.g. under Nutrient there is a NO3 node that contains a denitrification child node, there is a NO3 under SWIM with EXCO and FIP parameters and there are also 2 places (sample and chemical) where an initial value can be specified with NO3. Can we just have one NO3 for everything?

3. It is currently difficult to add another solute (tracer). When using SWIM, you need to add the solute under soil, add a SwimSoluteParameters under SWIM and then have a manager script to give the solute an initial value.

hol353:
Resolves #7170


hol353:
retest this please jenkins

PR #7367, opened 2022-07-03
Remove leaf tests

### LINKED ISSUE #7366: Remove leaf size data from Wheat Obs file
Only 6 value which are not very useful but always break the stats.  There is lots more leaf size data but needs to be added properly when we have a rethink of the leaf model.

HamishBrownPFR:
Resolves #7366


HamishBrownPFR:
@hol353 Can you please merge this.  Missing tests are the ones I intended to remove

PR #7385, opened 2022-07-19
Create override system for plant models, cultivars, factors

### LINKED ISSUE #7384: Create a system for deriving a pmf model from a generic base model.
There are a lot of plant models in APSIM that are (or should be) very similar to other models. e.g. barley and oats should be similar to wheat. At the moment we don't know what differences there are between wheat/barley/oats or between the different legume models. Some consistency is needed.

We would like to create the concept of a base (generic) model that other models can derive from. e.g. we could create a winter cereal model that wheat, barley and oats derive from. 

One way to implement this is to use the property overriding system that cultivars and factorials use. A plant model could then be the name of the base plant model and 1 or more property overrides. This would restrict derived models to simple re-parameterisation and not allow different model structures, which is one of the objectives for doing this. We already have code in APSIM to do parameter setting so the task should be straight-forward, although some refactoring will be needed first.

hol353:
Resolves #7384


hol353:
retest this please jenkins

APSIM.Server/Commands/RunCommand.cs
```diff
@@ -21,13 +17,13 @@ public class RunCommand : ICommand
         private bool runTests;
         private IEnumerable<string> simulationNamesToRun;
         private int numberOfProcessors;
-        private IEnumerable<IReplacement> changes;
+        private IEnumerable<(string name, object value)> changes;
```
hol430, line 20: I'm not a fan of the value tuple here. I think it makes the type name harder to parse (for a human) and you end up with some unsightly `.Item1` and `.Item2` code later on. I don't think you lose much by making this a bit more explicit and wrapping the name/value pair in a struct or class.

Models/Core/Run/SimulationDescription.cs
```diff
@@ -232,11 +220,8 @@ private void AddReplacements()
                 IModel replacements = topLevelModel.FindChild<Folder>("Replacements");
                 if (replacements != null && replacements.Enabled)
                 {
-                    foreach (IModel replacement in replacements.Children)
-                    {
-                        var modelReplacement = new ModelReplacement(replacement.Name, typeToFind:null, replacement);
-                        replacementsToApply.Insert(0, modelReplacement);
-                    }
+                    foreach (IModel replacement in replacements.Children.Where(m => m.Enabled))
+                        replacementsToApply.Insert(0, ($"Name={replacement.Name}", replacement));
```
hol430, line 224: This seems suspicious to me. I haven't really looked too closely at your new api, but a better approach might be to construct a parsed object. As opposed to creating a string only to (presumably) parse the string apart later on anyway. And if you go and refactor the way this works later on, finding and updating bits of code like this will be very difficult and time consuming.

Models/Core/Overrides.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        /// <summary>
+        /// Applies a collection of property sets to a model.
+        /// </summary>
+        /// <param name="model">The model to apply the changes to.</param>
+        /// <param name="factors">The property sets (keyword, value).</param>
+        /// <returns></returns>
+        public static void Apply(IModel model, IEnumerable<(string name, object value)> factors)
+        {
+            foreach (var factor in factors)
```
hol430, line 53: Ok, this one is really minor. But I think these 2 Apply() functions are the wrong way around. Having this one call the other one, and moving the actual replacement logic into the other one would be simpler. By the way, I'm still not convinced about this new `Name=Value` syntax that's been introduced in this PR - all of these calls to `.StartsWith()` and `.Replace()` seem very suspicious to me. But I'll admit I haven't really thought too much on the problem you're trying to solve.

Models/Core/Run/SimulationDescription.cs
hol353: Agreed. I've made the matching explicit with an enum instead of a magic string.

Models/Core/Overrides.cs
hol353: Agreed. I've moved the functionality from the second Apply into the first Apply - simpler. The MatchType enum removes the ```.StartsWith``` and ```.Replace``` calls. Much cleaner. 

Thanks Drew for doing a review of this PR. It has made it a lot neater.

PR #7391, opened 2022-07-21
Update values when units of soil properties are changed in the GUI.

### LINKED ISSUE #7388: Unit of initial values on NO3, NH4, Urea are not response after changing
apsimx master branch in the latest version, Windows 10 

Steps to reproduce this error

* Open wheat.apsimx in the validation folder
* Find the node: .Simulations.Validation.NPIField2019.WaggaWagga.WaggaWagga2019Base.Field.Soil.NO3
* Right click the unit of Initial values and change from ppm to kgha.
* The unit is not updated and the values are not changed.
* Switch to another node and come back. The unit is updated but values are still not changed

zur003:
Resolves #7388

### DIFF
--- a/Models/Soils/Chemical.cs
+++ b/Models/Soils/Chemical.cs
@@ -27,6 +27,12 @@
             CaCl2
         }
 
+        /// <summary>
+        /// Backing store for PHUnits property
+        /// </summary>
+        [JsonIgnore]
+        private PHUnitsEnum pHUnits;
+
         /// <summary>Depth strings. Wrapper around Thickness.</summary>
         [Summary]
         [Units("mm")]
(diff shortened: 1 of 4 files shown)


hol353:
This looks like it is going to convert the values after deserialisation which is not what we want. We want to leave the values entered by the user in what ever units they specify. If they change the units in the GUI, we don't convert the values. The values only get converted when the simulation starts running. This seems much simpler. The bug, I think, is that the units as shown in the GUI isn't changed when you select a new units from the popup menu.

zur003:
OK. That is simpler. I shall close this pull request and open a new one that doesn't overdo things.

PR #7413, opened 2022-08-02
Fixed add model function and icons in add model

### LINKED ISSUE #7400: APSIMNG GUI cannot save factor
Latest version of APSIM NG in the master branch with Windows 10

Step to reproduce this bug

* Open the attached apsimx file using APSIMNG.exe
* Right click the `Permutation` node to add a new model `factor`
* Save apsimx file and reopen it

The added new `factor` is lost.

[example.zip](https://github.com/APSIMInitiative/ApsimX/files/9194019/example.zip)

### LINKED ISSUE #7404: Issues with add new model displaying icons and adding children
The icons for models in Add new model seem to have disappeared
When I add a new model it shows up in the tree structure (without icon), but it is not added a a child of the parent in "save" and so isn't added to the apsimx file. 
Has something recently changed in this area. I couldn't find any code changes.

This is probably the issue behind #7400 as well.

hol353:
Resolves #7400
Resolves #7404

### DIFF
--- a/ApsimNG/Presenters/AddModelPresenter.cs
+++ b/ApsimNG/Presenters/AddModelPresenter.cs
@@ -175,7 +175,10 @@
                     { 
                         child = (IModel)Activator.CreateInstance(selectedModelType.ModelType, true);
                         child.Name = selectedModelType.ModelName;
-                        child.ResourceName = selectedModelType.ResourceString;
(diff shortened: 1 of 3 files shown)


hol353:
@lie112 - Can you pull this branch and see if this fixes your problems? It seems to work for me.

lie112:
Yes, that seems t o fix the problem of missing icons and saving newly added models.

PR #7414, opened 2022-08-02
Improved handling of weaners and orphans

### LINKED ISSUE #7410: Orphaned ruminants should not be weaned
A difference between various definitions of weaned means orphaned offspring should not be suckling, but should also not been weaned until passing through a weaning activity or reaching natural weaning age.

lie112:
Resolves #7410


lie112:
retest this please jenkins

PR #7420, opened 2022-08-03
Fix problem with drawing grids with large numbers of rows.

### LINKED ISSUE #7405: APSIM NextGen Freezing
APSIM NextGen is freezing at the drop of a hat.

zur003:
Resolves #7405 

@hol353 - Can you please review this? I think I've preserved the original intent here, but I don't understand Sheet.cs well enough to be certain. My assumption is that the original problem was ambiguity about whether the "RowCount" property provides the number of rows in the data table or the number of rows potentially displayable by the view.


hol353:
@zur003 Thanks for the fix. Looks good to me.

PR #7422, opened 2022-08-03
Several small scrolling fixes to Sheet grid control.

### LINKED ISSUE #7421: Scrolling quirks in grid control
The Sheet grid control has some scrolling quirks that could be improved. Specific points are:

1. After initial drawing, scrolling to the bottom of the sheet can take the user beyond the last data row, leaving a blank control. However, if the control is resized, scrolling to the bottom behaves as expected.
2. Resizing the control always takes the user to the top of the sheet. It would be better if, say, the cell positioned in the upper left corner would continue to be in that position after resizing. The current behaviour is a nuisance if you are resizing the control to see a bit more, as the current position is lost.
3. It would be good if the Home and End keys could be used to scroll to the top and bottom of the sheet, respectively (in addition to Ctrl-Up and Ctrl-Down, which do this currently).

zur003:
Resolves #7421


hol353:
retest this please jenkins

PR #7465, opened 2022-08-25
Fix ReflectionUtilities.Clone() when using a non-default AssemblyLoadContext

### LINKED ISSUE #7464: ReflectionUtilities.Clone() fails with "SerializationException: Unable to find assembly" when using non-default AssemblyLoadContext
### Summary

When ApsimX's ReflectionUtilities.Clone() method is called on an instance of a type that belongs to an assembly that was not loaded in the default AssemblyLoadContext, it fails with `SerializationException: Unable to find assembly '...'.`

(Another possibility is that the assembly will also be found in the default load context, and Clone() will succeed, but then there will be an InvalidCastException when the caller of Clone() tries to cast the result.)

A fix is proposed in #7465.

### Background and analysis

ApsimX's [ReflectionUtilities.Clone()](https://github.com/APSIMInitiative/ApsimX/blob/265679cee30b0355e222f6a5832272ecc16f98da/APSIM.Shared/Utilities/ReflectionUtilities.cs#L544) method is implemented by serializing and deserializing using .NET's [BinaryFormatter](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter).
When BinaryFormatter performs deserialization, it needs to be able to load assemblies.
.NET Core has something called "AssemblyLoadContext" ([documentation](https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext)), which allows assemblies to be loaded in isolated contexts (e.g. in order to load multiple versions of an assembly).
By default, BinaryFormatter.Deserialize loads assemblies using [Assembly.Load(AssemblyName)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.load#system-reflection-assembly-load(system-reflection-assemblyname)) (source code references: [[1]](https://github.com/dotnet/runtime/blob/v6.0.7/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/Formatters/Binary/BinaryObjectReader.cs#L1109), [[2]](https://github.com/dotnet/runtime/blob/v6.0.7/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/FormatterServices.cs#L319)). In the context of BinaryFormatter, this uses the [default AssemblyLoadContext](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext.default).
If the type that is being cloned belongs to an assembly from the default context, this works fine.
But if the type belongs to an assembly from a *different* context, this is problematic: the assembly might not be found in the default context, and if it *is* found it may later [cause an InvalidCastException](https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext#type-conversion-issues) (because loading the assembly in the default context results in a new Assembly instance and its types are distinct).

One situation where non-default contexts arise is when writing a COM (Component Object Model) server.
.NET Core loads the COM server in an isolated AssemblyLoadContext (reference: [COM hosting design doc](https://github.com/dotnet/runtime/blob/main/docs/design/features/native-hosting.md#loading-managed-code) and [notes in the .NET Core COM server docs](https://docs.microsoft.com/en-us/dotnet/core/native-interop/expose-components-to-com#additional-notes)).
The dependencies of the COM server (e.g. Models.dll) are automatically loaded in this isolated context when the COM server starts.
So if the COM server creates a Simulation instance that is later passed to Clone(), the call will fail:
If Models.dll is not found in the default context, there will be a SerializationException "Unable to find assembly 'Models ...'."
If Models.dll *is* found, there will be an InvalidCastException when the caller of Clone() tries to cast the result to Simulation, because the result's type will be from a different Assembly instance.

#### Example 1 (minimal, contrived)

This is a minimal (but contrived) example that manually loads an assembly in a non-default AssemblyLoadContext and calls Clone() on an instance of a type from that assembly.

1.  Create a class library project "DemoLibrary":

    ```
    dotnet new classlib -f netcoreapp3.1 -o DemoLibrary
    ```

    Create a file "DemoSimulation.cs" in this project with the following contents:

    ```c#
    using System;

    namespace DemoLibrary
    {
        [Serializable]
        public class DemoSimulation
        {
        }
    }
    ```

    Run `dotnet build` in this project (this generates bin\Debug\netcoreapp3.1\DemoLibrary.dll).

2.  Create a console application project "DemoApp" that loads the library and calls Clone():

    ```
    dotnet new console -f netcoreapp3.1 -o DemoApp
    cd DemoApp
    dotnet add reference ...\ApsimX\APSIM.Shared\APSIM.Shared.csproj # replace "..." with the actual path
    ```

    Modify Program.cs to have the following contents:

    ```c#
    using APSIM.Shared.Utilities;
    using System;
    using System.IO;
    using System.Reflection;
    using System.Runtime.Loader;

    namespace DemoApp
    {
        class DummyAssemblyLoadContext : AssemblyLoadContext
        {
        }

        class Program
        {
            static void Main(string[] args)
            {
                AssemblyLoadContext assemblyLoadContext = new DummyAssemblyLoadContext();
                Assembly demoLibrary = assemblyLoadContext.LoadFromAssemblyPath(Directory.GetCurrentDirectory() + @"\..\DemoLibrary\bin\Debug\netcoreapp3.1\DemoLibrary.dll");
                Type demoSimulationClass = demoLibrary.GetType("DemoLibrary.DemoSimulation", throwOnError: true);
                object sim = Activator.CreateInstance(demoSimulationClass); // dynamic equivalent of "DemoSimulation sim = new DemoSimulation();"
                object clone = ReflectionUtilities.Clone(sim); // fails with "SerializationException: Unable to find assembly 'DemoLibrary, ...'."
                Console.WriteLine("Clone() succeeded. IsInstance: {0}", demoSimulationClass.IsInstanceOfType(clone));
            }
        }
    }
    ```

    Run using `dotnet run`.

Output (on ApsimX at commit 4cc851718f8b):

```
Unhandled exception. System.Runtime.Serialization.SerializationException: Unable to find assembly 'DemoLibrary, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
   at System.Runtime.Serialization.Formatters.Binary.BinaryAssemblyInfo.GetAssembly()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.GetType(BinaryAssemblyInfo assemblyInfo, String name)
   at System.Runtime.Serialization.Formatters.Binary.ObjectMap..ctor(String objectName, String[] memberNames, BinaryTypeEnum[] binaryTypeEnumA, Object[] typeInformationA, Int32[] memberAssemIds, ObjectReader objectReader, Int32 objectId, BinaryAssemblyInfo assemblyInfo, SizedArray assemIdToAssemblyTable)
   at System.Runtime.Serialization.Formatters.Binary.BinaryParser.ReadObjectWithMapTyped(BinaryObjectWithMapTyped record)
   at System.Runtime.Serialization.Formatters.Binary.BinaryParser.ReadObjectWithMapTyped(BinaryHeaderEnum binaryHeaderEnum)
   at System.Runtime.Serialization.Formatters.Binary.BinaryParser.Run()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Deserialize(BinaryParser serParser, Boolean fCheck)
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream, Boolean check)
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream)
   at APSIM.Shared.Utilities.ReflectionUtilities.BinaryDeserialise(Stream stream) in C:\Users\User\ApsimX\APSIM.Shared\Utilities\ReflectionUtilities.cs:line 344
   at APSIM.Shared.Utilities.ReflectionUtilities.Clone(Object sourceObj) in C:\Users\User\ApsimX\APSIM.Shared\Utilities\ReflectionUtilities.cs:line 551
   at DemoApp.Program.Main(String[] args) in C:\Users\User\DemoApp\Program.cs:line 21
```

(If a reference to DemoLibrary is added to DemoApp, Clone() will succeed and the output will be `Clone() succeeded. IsInstance: false`.)

#### Example 2: COM Server

This is a larger example that uses a COM (Component Object Model) server. It is the real-life use case that led to the discovery of the issue.

1.  Download the repository [dotnet/samples](https://github.com/dotnet/samples) to get the COM Server Demo ([documentation](https://docs.microsoft.com/en-us/dotnet/core/native-interop/expose-components-to-com), [source subdirectory](https://github.com/dotnet/samples/tree/main/core/extensions/COMServerDemo)).

2.  Modify the COMServer project to use ApsimX:

    ```
    cd samples\core\extensions\COMServerDemo\COMServer
    dotnet add reference ...\ApsimX\Models\Models.csproj # replace "..." with the actual path
    ```

    Modify Server.cs to have the following contents:

    ```c#
    using Models.Core;
    using Models.Core.Run;
    using System;
    using System.Runtime.InteropServices;

    namespace COMServer
    {
        [ComVisible(true)]
        [Guid(ContractGuids.ServerClass)]
        public class Server : IServer
        {
            double IServer.ComputePi()
            {
                Simulations sims = Simulations.Create(new Simulation[] { new Simulation() });
                // based on APSIM.Cli/Program.cs
                Runner runner = new Runner(sims);
                List<Exception> errors = runner.Run();
                if (errors != null && errors.Count > 0)
                {
                    throw new AggregateException("Simulation ran with errors", errors);
                }
                return 3.14;
            }
        }
    }
    ```

    Build the server:

    ```
    dotnet build -p:RegFree=True
    ```

    (The above command also copies the COMServer output files to the COMClient project. I'm using registration-free mode because it does not require administrative privileges.)

3.  Build and run the COMClient project:

    ```
    cd ..\COMClient
    dotnet build -p:RegFree=True
    dotnet run --no-build
    ```

    (Note that `dotnet run` must be invoked with `--no-build`, otherwise it might try to automatically build without `-p:RegFree=True`. If that happens, the fix is to run `dotnet clean; dotnet build -p:RegFree=True`. Cleaning is necessary because the build system does not recognise the property change as a reason for a full rebuild.)

Output:

```
Unhandled exception. System.AggregateException: Simulation ran with errors (Error in file:  Simulation: Simulation)
 ---> System.Exception: Error in file:  Simulation: Simulation
 ---> System.Runtime.Serialization.SerializationException: Unable to find assembly 'Models, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
   at System.Runtime.Serialization.Formatters.Binary.BinaryAssemblyInfo.GetAssembly()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.GetType(BinaryAssemblyInfo assemblyInfo, String name)
   at System.Runtime.Serialization.Formatters.Binary.ObjectMap..ctor(String objectName, String[] memberNames, BinaryTypeEnum[] binaryTypeEnumA, Object[] typeInformationA, Int32[] memberAssemIds, ObjectReader objectReader, Int32 objectId, BinaryAssemblyInfo assemblyInfo, SizedArray assemIdToAssemblyTable)
   at System.Runtime.Serialization.Formatters.Binary.BinaryParser.ReadObjectWithMapTyped(BinaryObjectWithMapTyped record)
   at System.Runtime.Serialization.Formatters.Binary.BinaryParser.ReadObjectWithMapTyped(BinaryHeaderEnum binaryHeaderEnum)
   at System.Runtime.Serialization.Formatters.Binary.BinaryParser.Run()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Deserialize(BinaryParser serParser, Boolean fCheck)
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream, Boolean check)
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream)
   at APSIM.Shared.Utilities.ReflectionUtilities.BinaryDeserialise(Stream stream)
   at APSIM.Shared.Utilities.ReflectionUtilities.Clone(Object sourceObj)
   at Models.Core.Apsim.Clone[T](T model)
   at Models.Core.Run.SimulationDescription.ToSimulation()
   --- End of inner exception stack trace ---
   at Models.Core.Run.SimulationDescription.ToSimulation()
   at Models.Core.Run.SimulationDescription.Prepare()
   at APSIM.Shared.JobRunning.JobRunner.Prepare(IRunnable job)
   at APSIM.Shared.JobRunning.JobRunner.RunActualJob(IRunnable job, IJobManager jobManager)
   --- End of inner exception stack trace ---
   at COMServer.Server.IServer.ComputePi()
   at IServer.ComputePi()
   at COMClient.Program.Main(String[] args) in C:\Users\User\dotnet-samples\core\extensions\COMServerDemo\COMClient\Program.cs:line 12
```

### Proposed fix

#7465

### Alternative workaround

There is a simple (but ugly) workaround which can be implemented in the user's application code without modifying ApsimX. The idea is to add a handler to the event [AssemblyLoadContext.Resolving](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext.resolving) of the default context (or the event [AppDomain.AssemblyResolve](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.assemblyresolve)), and when it occurs load the assembly from the desired context instead. Note that this event only occurs when assembly resolution fails, so this workaround only works if the assembly is *not* found in the default context.

```c#
...
    public class Server : IServer
    {
        public Server()
        {
            // [Somewhere at the start of the program]
            AssemblyLoadContext.Default.Resolving += OnResolving;
        }

        public static Assembly OnResolving(AssemblyLoadContext originalContext, AssemblyName assemblyName)
        {
            // [Replace "Server" with the name of some class from the desired load context]
            AssemblyLoadContext customContext = AssemblyLoadContext.GetLoadContext(typeof(Server).Assembly);
            return customContext.LoadFromAssemblyName(assemblyName);
        }

        ...
    }
```

**Discovered by:** @HemdaLevyNZ.

tom93:
Resolves #7464. (In short: Clone() breaks on assemblies from a non-default AssemblyLoadContext, because by default .NET's BinaryFormatter can only load assemblies from the default AssemblyLoadContext.)

---

Some comments on the implementation:

For general-purpose deserialization, there is no obvious way to specify the load context, because a composite object may contain objects from multiple assembly load contexts.

But in the special case of cloning (serialize-then-deserialize), there is an easy solution: cache the assemblies seen during serialization, and look them up in the cache during deserialization.

Resolving generic types can be complicated because the type arguments need to be resolved recursively, but fortunately the [overload of Type.GetType()](https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype#system-type-gettype(system-string-system-func((system-reflection-assemblyname-system-reflection-assembly))-system-func((system-reflection-assembly-system-string-system-boolean-system-type)))) does the heavy lifting.

### DIFF
--- a/APSIM.Shared/Utilities/ReflectionUtilities.cs
+++ b/APSIM.Shared/Utilities/ReflectionUtilities.cs
@@ -10,6 +10,7 @@
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using System.Runtime.Loader;
     using System.Runtime.Serialization;
     using System.Runtime.Serialization.Formatters.Binary;
 
@@ -318,7 +319,7 @@
         /// <summary>
         /// Binary serialise the object and return the resulting stream.
         /// </summary>
-        public static Stream BinarySerialise(object source)
+        public static Stream BinarySerialise(object source, SerializationBinder binder = null)
         {
             if (source == null)
                 return null;
@@ -327,6 +328,7 @@
                 throw new ArgumentException("The type must be serializable.", "source");
 
             IFormatter formatter = new BinaryFormatter();
+            formatter.Binder = binder;
             Stream stream = new MemoryStream();
             formatter.Serialize(stream, source);
             return stream;
@@ -335,12 +337,13 @@
         /// <summary>
         /// Binary deserialise the specified stream and return the resulting object
         /// </summary>
-        public static object BinaryDeserialise(Stream stream)
+        public static object BinaryDeserialise(Stream stream, SerializationBinder binder = null)
         {
             if (stream == null)
                 return null;
 
             IFormatter formatter = new BinaryFormatter();
+            formatter.Binder = binder;
             return formatter.Deserialize(stream);
         }
 #pragma warning restore SYSLIB0011
@@ -546,9 +549,56 @@
         /// </summary>
         public static object Clone(object sourceObj)
         {
-            Stream stream = BinarySerialise(sourceObj);
+            CachingSerializationBinder binder = new CachingSerializationBinder();
+            Stream stream = BinarySerialise(sourceObj, binder);
             stream.Seek(0, SeekOrigin.Begin);
-            return BinaryDeserialise(stream);
+            return BinaryDeserialise(stream, binder);
+        }
+
+        /// <summary>
+        /// Custom SerializationBinder that records the assemblies seen during serialisation
+        /// and reuses them during deserialisation.
+        /// This is useful when working with assemblies from a non-default AssemblyLoadContext,
+        /// because BinaryFormatter cannot deserialise them otherwise.
+        /// </summary>
+        class CachingSerializationBinder : SerializationBinder
+        {
+            private Dictionary<string, Assembly> assemblyCache = new Dictionary<string, Assembly>();
+
+            public override void BindToName(Type serializedType, out string assemblyName, out string typeName)
+            {
+                assemblyName = serializedType.Assembly.FullName;
+                typeName = serializedType.FullName;
+                if (!assemblyCache.ContainsKey(assemblyName))
+                {
+                    assemblyCache[assemblyName] = serializedType.Assembly;
+                }
+                else
+                {
+                    if (assemblyCache[assemblyName] != serializedType.Assembly)
+                    {
+                        throw new FileLoadException(String.Format("Assemblies with the same name from different load contexts are not supported: '{0}'.", assemblyName));
+                    }
+                }
+            }
+
+            public override Type BindToType(string assemblyName, string typeName)
+            {
+                string qualifiedTypeName = String.Format("{0}, {1}", typeName, assemblyName);
+                return Type.GetType(qualifiedTypeName, assemblyResolver: ResolveAssembly, typeResolver: null);
+            }
+
+            private Assembly ResolveAssembly(AssemblyName assemblyName)
+            {
+                if (assemblyCache.ContainsKey(assemblyName.FullName))
+                {
+                    return assemblyCache[assemblyName.FullName];
+                }
+                else
+                {
+                    return AssemblyLoadContext.Default.LoadFromAssemblyName(assemblyName);
+                }
+            }
         }
 
         /// <summary>


hol353:
add to whitelist Jenkins

tom93:
Thanks for merging!

Issue #4631, opened 2020-01-22
New Cotton Model


hut104:
A new cotton model is under development for APSIM Next Generation.

sarahcleary:
@hut104 - - just following up the concept note.  For background, please refer to https://github.com/APSIMInitiative/ReferencePanel/issues/57#issue-595681585

Amitava-Soil:
I am trying to simulate cotton production in Mississippi, USA. I tried to modify the cultivar coefficient using S71BR cultivar but the LAI is way below the measured but the yield and evapotranspiration estimation is relatively close. Here is the cultivar parameters, I used. I appreciate any help.
- Reading Cultivar Parameters
         ---------------------------------------
         Cultivar   =  _DEFAULT_S71BR
         percent_L  =  45.00
         scboll     =  4.0
         respcon    =  0.016
         sqcon      =  0.022
         fcutout    =  0.5411
         flai       =  0.87
         ddisq      =  400.0
         dlds_max   =  0.12
         popcon     =  0.03633
         acotyl     =  525.0
         rlai       =  0.010
         frudd      =  40  130  200  275  375  500  675  800  
         bltme      =  0.00  0.00  0.00  0.07  0.21  0.33  0.55  1.00  
         wt         =  0.0104  0.0272  0.1441  0.0988  0.5042  0.9617  1.0000  0.5785  
         fburr      =  1.23
         rate_emergence       =  1.0
         background_retention =  0.80

[Continuous Irrigated Cotton - deficit rule.txt](https://github.com/APSIMInitiative/ApsimX/files/9562448/Continuous.Irrigated.Cotton.-.deficit.rule.txt)

edvlmaas:
Hi, is there an estimate when the cotton model will be available? I could use it for a project I'm currently working on. Please consider this a request to prioritize it!  :-)

BrianCollinss:
Hi @hut104. May I please know if the cotton model has reached a state that can be used by a student in a PhD project?

par456:
Todo:
- Investigate why the stem wt is overpredicting in narribri and reducing boll wt
- Cotton not moving to maturity in cold environments when it stops growing.

PR #7489, opened 2022-09-05
Updated Cotton Starting Point

hut104:
working on #4631


hut104:
@hol353 This can be merged as a starting point for Thiemen.

Issue #7487, opened 2022-09-02
MungBean validation set needs some cleanup


hut104:
MungBean validation set needs some cleanup. Will reformat graphs and add more stats to test set

PR #7491, opened 2022-09-06
Heathers Obs Data

hrpasl:
working on #7487


hol353:
add to whitelist jenkins

PR #7497, opened 2022-09-08
Don't create a Soil Nitrogen component but create a Nutrient component…

### LINKED ISSUE #7471: Error in importing old apsim file
Following the similar bug in #7389. After importing old *. apsim file, the nitrogen module is located under `SoilNitrogen` (red circle), instead of `Soil` (blue circle).

image

The error code can be found from here: https://github.com/APSIMInitiative/ApsimX/blob/5ed3f61a0b326d7d107f6fb8d548e14e42e8365b/Models/Soils/Soil.cs#L298

```
[Soil].NO3
[Soil].NH4
[Soil].Urea
```

her123:
resolves #7471
… along with NO3, NH4, UREA and initialise them.

### DIFF
--- a/Models/Core/Apsim710File/Importer.cs
+++ b/Models/Core/Apsim710File/Importer.cs
@@ -232,6 +232,10 @@
                     XmlNode newFolder = this.AddCompNode(destParent, "Folder", XmlUtilities.NameAttr(child));
                     this.AddFoldersAndSimulations(child, newFolder);
                 }
+                else
+                {
+                    this.AddComponent(child, ref destParent);
+                }
                 child = child.NextSibling;
             }
         }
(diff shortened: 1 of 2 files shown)


her123:
resolves #7471

hol353:
@her123 This PR doesn't resolve the problem of importing DroughtSoilsOriginal.zip in the referenced issue.

BTW: the resolves keyword in the second comment box needs to be in the first comment box. Jenkins only looks in the first comment box. You can edit the first comment box by clicking the three dots menu.

her123:
The importer process was designed to have either folders or simulations at the top level. Folders can contain simulations but components must live under a simulation. The DroughtSoilsOriginal is only some folders with soil components in them. Should the importer be changed to support importing this type of input?

hol353:
Yes we probably need to support this. The DroughtSoilsOriginal file is very similar to an APSoil collection of soils.

her123:
@hol353 looks like jenkins failed at the documentation process. I don't think this was my changes.

hol353:
retest this please jenkins.

There is something strange in the canola doc creation step that fails occasionally.

hol353:
retest this please jenkins

PR #7535, opened 2022-10-06
Make Predicted / Observed more flexible

### LINKED ISSUE #7534: More flexibility for Predicted / Observed models
Currently predicted / observed models always use SimulationName as one of the fields to match on when comparing predicted and observed data. There are use cases (e.g. when comparing 2 water balance models - via an experiment - same observed data applies to both simulations) where this field needs to be removed from the fields to match on.

hol353:
Resolves #7534


jbrider (commented):
The documentation for sorghum has added strange characters again.
It looks like it is replacing hyphens with something unknown.

hol353:
Where exactly are the strange characters?

jbrider:
image

There are more instances of the hyphen issue further down.

hol353:
Hmm ok. Add it as a GitHub issue.

PR #7544, opened 2022-10-13
Changed calculation for x

### LINKED ISSUE #7543: SWIM Node Structure
Nodes should be in the middle of each layer or one the upper and lower boundary condition.

hut104:
resolves #7543


hut104:
@hol353 I think we can merge this.  The changes have significantly improved the stats for outputs in water table situations.

The changes to Hudson will be due to changes in space weighting of conductivites etc.  We need to revisit Huson some time to better capture the parameters used here for zero till stubble management.

PR #7552, opened 2022-10-18
Add an IsInitialising field as a semaphore in Simulation to prevent e…

### LINKED ISSUE #7288: Problem with incomplete initialisation not being detected
I've recently noticed problems with the data store, but haven't worked out exactly what is going wrong. As an example, I just updated my installed version of ApsimX and ran an old wheat simulation. The simulation appeared to run normally, but at the conclusion there was nothing in the data store other than the initial conditions. Other outputs were apparently either not stored, or were wiped away at the end of the run. Running the same simulation a second time provided normal results. I've sometimes noticed a similar problem when moving between the installed version and my locallly compiled version - the run appears to execute normally, but the data store holds nothing beyond initialisation values.

zur003:
…xecution of Simulations that have been only partially initialised.

Resolves #7288, but is rather inelegant. Is there a better approach (with mutexes or whatever)?


zur003:
Retest this please jenkins

Models/Core/Run/SimulationDescription.cs
```diff
@@ -140,6 +140,11 @@ public Simulation ToSimulation()
         {
             try
             {
+                // It is possible that the base simulation is still in the process of being
+                // initialised in another thread. If so, wait to let it finish.
+                while (baseSimulation.IsInitialising)
+                    Thread.Sleep(10);
+
```
hol353, line 147: This makes me nervous. Potentially, this could hang APSIM while it waits indefinitely.  Need to put a timeout in here and then throw an Exception?

zur003:
@hol353 - Is the Jenkins build failing due to [this](https://github.com/APSIMInitiative/Jenkins/commit/1ee49cc6f0ced5d51553cdf2b0566a5a1ba67605) change?

hol353:
@zur003 Thanks for continuing to contribute to APSIM. Your 'flag' approach is a bit awkward but I can't think of a simple, more elegant approach. 

Your error on Jenkins should be resolved if you do a pull before pushing back to your branch. The error was caused by a change I had to make to Jenkins (remove the password protected NPI .7z file) to get the CAMP pull request to run.

Models/Core/ApsimFile/FileFormat.cs
```diff
@@ -132,17 +132,28 @@ public static string WriteToString(IModel model)
         /// <param name="errorHandler"></param>
         public static void InitialiseModel(IModel newModel, Action<Exception> errorHandler)
         {
-            foreach (var model in newModel.FindAllDescendants().ToList())
+            IEnumerable<Simulation> simulationList = newModel.FindAllDescendants<Simulation>();
```
hol430, line 135: Probably a good idea to force greedy evaluation here, as we do a few lines later (e.g. by adding `.ToList()`). Otherwise you can run into problems if models get added to or removed from the simulation tree during the initialisation process.

There may be other gotchas which don't spring to (my) mind immediately as well. I'm not sure what the alternative is though; perhaps it's possible for the GUI to wait for the initialisation thread to finish before calling `Run()`, but whether that makes a practical difference, I don't know.

Issue #7550, opened 2022-10-17
Fix Stock conception rate calculations


her123:

her123:
Changes working correctly

PR #7553, opened 2022-10-19
Fixed the lactation response parameter

her123:
working on #7550
Added the missing c-i-21 parameter (Potential intake - curvature of lactation response when number of sucking = 0) for small ruminants.


her123:
Changes for small ruminants complete

Issue #7536, opened 2022-10-09
Characters being replaced inside Memo


jbrider:
Hyphens and degree symbols are being changed within a Memo component by an external process?
It's being changed without the user editing the component.

Mentioned:
https://github.com/APSIMInitiative/ApsimX/pull/7535#issuecomment-1270844768
https://github.com/APSIMInitiative/ApsimX/pull/7462#pullrequestreview-1089568467

Possibly related:
https://github.com/APSIMInitiative/ApsimX/issues/7442

zur003:
@jbrider - Can you indicate the steps needed to reproduce this problem? I haven't been able to reproduce it myself. It appears to be an issue associated with non-ANSI characters, but in theory we should generally be able to handle that.

jbrider:
No. I don't know what's causing it - only that it appears to happen when updates are made to json files.
That's when the memo comments seem to be changed without anyone intending to do so.
I don't know if the original was put back as it was merged before I saw that it had happened again.

zur003:
A search revealed that in several files (most notably APSIM.bib, Soybean.json, and Sorghum.json), non-ASCII characters like en-dash, the degree symbol, and opening/closing quotes had been replaced with the Unicode "Replacement Character" (&#xFFFD;, which is 0xFFFD, or 0xEF 0xBF 0xBD in UTF-8 encoding). Exactly how those replacements happened is not at all clear. This sort of thing usually happens when text is edited using an editor which assumes the wrong encoding - perhaps Windows 1252 instead of UTF-8 - when the file is read. Inside Apsim, I think we are always assuming UTF-8 encoding. My suspicion is that these errors were introduced by some external editor, and that this is not, strictly speaking, an Apsim issue.

In PR #7558, I have attempted to restore all these occurrences back to their original intended values. The question remains whether that is all that needs to be done, or is there still some flaw within Apsim that allows these errors to develop? The most recent such introduction was in a commit a couple of weeks ago by @hol353 in Sorghum.json. @hol353 - can you think of anywhere in the editing cycle that this problem may have occurred? Was Sorghum.json ever edited outside of Apsim (by you or anyone else)?

zur003:
Oh, I should also note that although PR #7558 corrects errors in APSIM.bib, it does NOT make the same corrections to Tests/Validation/Mungbean/APSIM.bib, because I assumed the latter file should not really be there at all. Can it be deleted, or is there a reason for its continued existence?

par456:
[...]

PR #7558, opened 2022-10-21
Replace Unicode "Replacement characters" with their original values.

zur003:
Working on #7536.

This restores characters that were garbled in Sorghum.json and Soybean.json to their original values, but doesn't address how they got garbled in the first place.


zur003:
Retest this please, Jenkins.

PR #7559, opened 2022-10-24
add github action

chokputtaraksa:
This action will process following these steps
1. test unittest
2. build docker file
3. publish to repository

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,71 @@
+name: unittest
+
+on:
+  push:
+    branches:
+      - wgp
+      - master
+  pull_request:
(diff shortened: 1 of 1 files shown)


hol353:
Thanks for raising a pull request. What is this pull request trying to do? It looks like it creates a GitHub action for a unit test. We use Jenkins to do this instead. Also, you will need to create a GitHub issue that describes the problem you're trying to fix and then reference that issue from the top comment box like this:

working on #????
or
resolves #????

where ???? is the GitHub number of the issue you create.

hol353:
Actually, I just noticed the PR is closed so ignore my comment above.

Issue #7560, opened 2022-10-25
Default search radius for searching APSOIL data base is too large


LindsayWBell:
Please replace the default of 100 km radius for searching soils from the database to 10 km. @her123

ric394:
Resolved by #7591

PR #7591, opened 2022-10-28
Soil downloads work better

### LINKED ISSUE #7561: Soil downloads should be able to be cancelled
Using the soil download can be slow. It needs to allowed to be cancelled.

her123:
resolves #7561


her123:
resolves #7560

her123:
@hol353 this is now ready to merge

PR #7595, opened 2022-10-31
Refactor Locator and Locater, merging into a single unit, with details of behaviour controlled by flags.

### LINKED ISSUE #7566: Hard crash of apsim
Output variable "no3()" causes a hard crash of apsim.
[RizwanCalibration.zip](https://github.com/APSIMInitiative/ApsimX/files/9859728/RizwanCalibration.zip)

zur003:
Resolves #7566

and also does a lot of other stuff. For example, it attempts to detect and handle the recursion resulting when someone is silly enough to define a Report column like "n + 1 as n". 

Every Model now has its own Locator (if needed), rather than relying on the Simulation to provide one.

One possible risk: the caching mechanism doesn't consider the flags used in a search, which could result in a cached value being returned when it shouldn't. However, I think this is unlikely to be a problem with current usage.


Models/Core/Model.cs
```diff
@@ -570,20 +569,39 @@ public void ParentAllDescendants()
...
+//        protected virtual Locater Locator()
+//       {
+//            Simulation sim = FindAncestor<Simulation>();
+//            if (sim != null)
+//                return sim.Locater;
+//
+//            // Simulation can be null if this model is not under a simulation e.g. DataStore.
+//            return new Locater();
+//        }
+
```
hol353, line 587: Can you please remove the commented code.

Tests/UnitTests/Core/LinksTests.cs
```diff
@@ -391,7 +391,7 @@ public void EnsureServicesResolve()
 
             var modelWithServices = simulations.Children[1].Children[2] as ModelWithServices;
             Assert.IsNotNull(modelWithServices.storage);
-            Assert.IsNotNull(modelWithServices.locator);
+//            Assert.IsNotNull(modelWithServices.locator);
             Assert.IsNotNull(modelWithServices.events);
```
hol353, line 395: Did you mean to remove the assert? If so, can you remove the commented code.

Tests/UnitTests/Report/ReportTests.cs
```diff
@@ -395,7 +395,7 @@ public void FactorsTableIsWritten()
             };
             var storage = new MockStorage();
             Utilities.InjectLink(report, "simulation", sim);
-            Utilities.InjectLink(report, "locator", new MockLocator());
+//            Utilities.InjectLink(report, "locator", new MockLocator());
             Utilities.InjectLink(report, "storage", storage);
```
hol353, line 399: Likewise, remove commented code.

hol353 (changes requested):
Good job Eric. Really well done. This removes a lot of cruft from APSIM. Thank you. Just remove some of the commented code.

hol353 (approved):

Issue #7454, opened 2022-08-21
Soils Interface Issues


jbrider:
I found these issues while looking for a simple example of a variety of locations for the sorghum release.
I found a couple in Chickpea for Katherine and Launceston which seemed like a good place to start.
Both locations use the same derived sandy soil that appears worse than the sandy soils in WA.

**Physical Node within a soil**

- Poor error message when you are missing a crop node underneath the Physical node.
- Needs a specific syntax that is not mentioned in the error or the help file
- Where is the help for Soils?
- Copying from another soil with a different layer structure caused corruption and errors that were difficult to recover from.
- Classic had a button called _Manage crops_ the functionality of which does not appear to be replicated in Nextgen.
- The graph is not resizable - no matter how big your monitor is.
- Is there a way to show the Crop specific lines on the graph? No graphs are shown on the Crop specific nodes below Physical.
- The table shows columns for Sand, Silt, Clay and Rocks (90% of the time I imagine these are empty?) in the left of the table - the mot prominent position for viewing shows nothing most of the time.
- Also 90% of the time, the simulation will contain a single crop and soil - would it be difficult to show the details for the current crop type on the left instead of empty columns? It would be a good place for an error message before trying to run the model as well.

**Download a soil**

- Assuming you know there is somewhere you can download a soil file, it takes a while to find that you have to right click the Soil node and click Download Soil.
- Like most Select boxes (combo box) in Nextgen, the drop down list takes up most of the screen height if the list has a long list of items in it.
- If you scroll towards the top - it will scroll well past the first item to show empty space scrolling down the screen.
- Searching for a soil has resulted in the following error both times I searched:

```
Error resolving type specified in JSON 'Models.Soils.InitialWater, Models'. Path 'Children[0].$type', line 25, position 50. --> Could not find type 'Models.Soils.InitialWater' in assembly 'Models, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

- Clicking the Add selected soil to simulation button will add a new soil **inside** of the soil you had just right clicked.
- Copying the new child soil and attempting to paste it in the Field above will result in the following error:

```
Duplicate models found: ~paddock.Synthetic Soil Derived from ISRIC SoilGrids has 2 children named Water
```
- It didn't create a newly pasted soil. (Apparently it did create it, but you have to go to another tab and come back again to see it. When you try to copy that node and paste it - it will now say there are 3 Water nodes).
[...]

PR #7604, opened 2022-11-05
Help Water.cs find the "right" Physical and SoilWater nodes.

### LINKED ISSUE #7562: Problem entering new initial soil water %
When entering numbers into the initial soil water % field random numbers appear. This behaviour is only on some soils but is repeatable.
See the Site permutation = Red Kandosol (Berrigan No537-YP) in the attached example.
[ExperimentBased_Scenarios.zip](https://github.com/APSIMInitiative/ApsimX/files/9857349/ExperimentBased_Scenarios.zip)

zur003:
Resolves #7562

### DIFF
--- a/Models/Soils/Water.cs
+++ b/Models/Soils/Water.cs
@@ -203,11 +203,11 @@
             }
         }
 
-        /// <summary>Finds the 'physical' node.</summary>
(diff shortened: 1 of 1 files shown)


zur003:
I think this will also address part of #7454.

Issue #3178, opened 2018-10-18
Name of "Plantain" model


hut104:
Plantain is one of the most important food crops across the world.  Some lists have it within the top ten food crops, after crops such as maize, wheat, soybean,  cassava, taro etc.  Annual production is over 34 million tonnes worldwide.  It is mostly eaten in Asia and West Africa. It is a plant very similar to common banana crops.
The Plantain model within APSIM is not this crop.  Plantain is the local name for a fodder crop in new zealand.  I am not sure how widely this name is used outside of new zealand.
In any case, I do wonder if the name of the model should be reconsidered at some point.  I know there is work from time to time on a banana model in APSIM.  I can see confusion from an international audience when banana means banana but plantain means a fodder crop.

HamishBrownPFR:
Should we be using Latin names for crop models to avoid confusion from colloquial naming.

rcichota:
@hol353 @hol430 
As I am going to be working on the Plantain model, I think we can tackle this issue. I am thinking in  renaming the model to 'PlantainForage', so we avoid much confusion with the Banana sibling as Neil pointed out.
The question is what it takes to do that? Where should I change the references to the plantain model?
I have just tried to rename all instances I could find in the json file in Resources and the test simulation, but when I run it, Apsim spits an error message saying it cannot find the model (or rather that PlantainForage doesn't have such and such variable). If I open the simulation and type PlantainForage on the report, for example, and then 'dot' I get access to all the variables... 
I search all APSIM directory for the word Plantain, but none other than simulations come with any relevant reference to Plantain. Nothing urgent, here, but I would like to do this before I, and colleagues, start using the model and then the name changes... maybe one of those conversion/updater thingys will have to be written?

rcichota:
Correction. when using the Report to type PlantainForage and then 'dot', I get access to a number of variables, but not all, missing the organs... And I cannot visualise the model structure...
If I paste the json file in the replacements I can see all the structure, and the simulation works...
Must be something little I am missing

hol353:
What you've done sounds right to me. Can you attach a simulation to this issue so I can have a look?

rcichota:
Thanks @hol353 
I am working in the Tests\Validation\Plantain simulation. I have pushed into the save working branch you merged early on (https://github.com/rcichota/ApsimX/tree/Issue/FixPlantainSim%237207). Hopefully is easier for you to get it from there. 
[...]

PR #7605, opened 2022-11-07
Renaming plantain 

rcichota:
Working on #3178 
I have renamed the model to PlantainForage to avoid or at least reduce confusion with other usages of the word ' plantain' (particularly bananas).
I've decided on PlantainForage to minimise changes in the name and leave it in the same alphabetical order. This word order seems to be equally valid or have the same usage as 'ForagePlantain', which has a different 'first name' and so may make things slightly more difficult...


rcichota:
@hol353 
This should be done now, the stats are 'red' because the system counts this as a new model (as far as I can see). So, if you want I think you can accept and merge this.
resolves #7605

Issue #6726, opened 2021-08-18
Maize leaf model improvements


HamishBrownPFR:
ISU testing is showing up some problems with the leaf model in the way it represents and responds to N stress.  Investigate problems and implement fixes

PR #7620, opened 2022-11-14
Maize leaf

hut104:
working on #6726


hol353:
retest this please jenkins

jbrider:
@APSIMInitiative/reference-panel  I bought this to Graeme's attention and he would like to have a discussion about the maize model given we have an established working Nitrogen model around leaf dynamics.

hut104:
@jbrider Thanks for that.
Currently I am just converting maize to use simple leaf rather than complex leaf as part of our migration to a more maintainable and usable code base.  I am not looking to make any major change to the science at this point as I'm just in a 3-day sprint for code maintenance and was looking to sort out some issues raised by users about stress outputs etc.
However, it is worth checking.  Are you able to send a link to the code you are suggesting? Would it be a simple change (e.g. adding a different class)?

jbrider:
https://github.com/APSIMInitiative/ApsimX/tree/master/Prototypes/C4Maize
This is some of the main work Drew did for us.
It uses the same leaf structure to Sorghum with some changes in parameterisation - so I doubt it is an easy change for the current Maize model.

par456:
Closing due to age.

PR #7626, opened 2022-11-14
Replacements change

### LINKED ISSUE #7625: Replacements Folder Issue
It looks like the new replacements folder behavior will replace any matching node, regardless of type. A Maize entry in the replacements folder will replace any simulation named "Maize" for instance, which leads to an object reference error when the replacement is applied (simulation objects don't necessarily have a parent).

ilhuber:
Resolves #7625 

This prevents replacements from trying to replace the top level Model, which always results in a null reference error.


hol430:
I think the old behaviour may have been intentional here. I think there was a requirement to be able to replace all instances of a model with another implementation of that model. Whether that's still a relevant use case, I don't know. Jenkins did run green, so perhaps it doesn't matter.

ilhuber:
Ahh, I didn't know that. It seems like the changes to how replacements are handled broke a few of our workshop exercises and I thought this might be the cause. I'll look deeper into the issue.

ilhuber:
retest this please jenkins

PR #7637, opened 2022-11-16
Fix to Generic Phase

### LINKED ISSUE #7636: Generic Phase fails if target drops below progress
If something suddenly drops the target for a generic phase, the code results in undesirable behaviour.
When entering the timestep, if target is already exceeded, the timestep should be provided to the next phase.

hut104:
resolves #7636


hut104:
@hol353 These changes are only for chickpea and are mostly positive with improved performance.  Only maturity date was slightly affected and are probably due to removal of errors.
I suggest merging of this bug fix.

PR #7667, opened 2022-11-20
Fixed a number of bugs around initial water UI

### LINKED ISSUE #7409: FractionFull Error
When I manually adjust the fractionfull for in the water node under soil within the "Site" factor "Breeza" under the "Sim" factorial, I get an error that the water in the bottom 2 layers is below the lower limit. When I enable the IW factor to adjust the fractionfull, I do not get an error.

See attached simulation
[Simulation.zip](https://github.com/APSIMInitiative/ApsimX/files/9231187/Simulation.zip)

hol430:
Resolves #7409.

I'm marking #7409 as resolved, but I don't actually remember the root cause of the reported problem. I do remember reproducing the original problem, and it no longer shows up after all of these changes, so I guess it's fixed. Otherwise, feel free to reopen the issue.

The reason I've touched so many files is that I changed a couple of utility functions used for converting between strings and doubles such that they now take a culture argument. The culture used for conversions depends on the data source.

Someone really ought to double check my changes to SoilsDataTable.cs. I can't find where this is being used, so I have no idea where the data table is coming from and therefore no idea which culture should be used.


hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
retest this please jenkins

hol430:
I changed some utility functions so that they now require an extra CultureInfo argument. I don't think there's really a sensible default value we could use; it really depends on the context, so the caller needs to decide what to pass in.

The changed .apsimx files contain manager scripts which call this api, so I needed to fix them manually in order to get them running. If I revert the changes to the .apsimx files they will give an error. If I revert the change to the api I'd be reintroducing one of the bugs I've fixed here.

I suppose we could set it to use the invariant culture by default, and that would avoid a potentially breaking change for people's .apsimx files. But I feel that approach encourages callers of the api to bury their heads in the sand rather than take the time to think about how these conversions should occur, and ultimately lead to more bugs like the ones (plural!!) that have been fixed here.

PR #7704, opened 2022-12-05
New ActivityTimerRuminantLevel

### LINKED ISSUE #7703: Add timer based on filtered herd state
Create a timer to determine when activities are performed based on the number or summarised properties of a specified filter of the ruminant herd

lie112:
Resolves #7703


hol353:
Unsure why this failed. I reran it on jenkins and it seems to have worked.

PR #7715, opened 2022-12-19
Allow overloaded methods to be included in Report variables

### LINKED ISSUE #7714: Report cannot handle overloaded methods
Report throws "Ambiguous" error when trying to report an overloaded method.

lie112:
Resolves #7714 

Added code to identify argument type by the passed values in a method included as variable in reports
This allows us to link the correct overloaded version of the method to the report.


lie112:
What's the apsim convention for adding strings to the variable entry for reports?
Is it to use single quotes? I was using escaped double quotes but it's messy as actually needs the escape escaped! \\\"
I will change this to single quotes for consistency and ease for users if that's the case.

hol353:
>What's the apsim convention for adding strings to the variable entry for reports?
I don't understand what you're asking.

lie112:
What I meant was if we need to pass a string to a method in report do we ignore text delimiters or use quotes or single quotes.
Model.WorkWithModel("Modelname", 2) for WorkWithModel(string, int32)
We were working to get a method call working and I was struggling to get a string passed or it had the quotes included with the string. I think I worked it out, but now can't remember the solution.

hol353:
Ahh ok. I'm not sure if you need quotes or not. I can't remember.

PR #7719, opened 2022-12-30
Replace Bitmap and Image with Gdk.Pixbuf; replace SharpMap with Mapsui.

### LINKED ISSUE #7469: Replacement for System.Drawing.Common
While looking into issue #7139, I became aware that Microsoft is encouraging developers to move away from the use of System.Drawing.Common because of issues with cross-platform support. We are currently using version 5, although version 6 is available. However, version 6 emits compiler warnings if non-Windows targets are supported. Microsoft describes the situation [here](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only). They suggest three alternative for cross-platform graphics support: ImageSharp, SkiaShap, or Microsoft.Maui.Graphics. Which of these options will be best for us? We are already pulling SkiaSharp into APSIM.Interop, so there may be less overhead there. On the other hand, Microsoft.Maui.Graphics may offer a simpler upgrade pathway (or maybe not). Interestingly, I think it actually wraps bits of SkiaSharp for its Linux support (though I might be wrong about that).

So what do people think is the best alternative? You may think we're not doing much with System.Drawing.Common, but that's not the case. The classes Color, Font, Point, and Rectangle all live there, and they are used in many places. For those specifically, we could probably use roughly equivalent Gdk classes, so that's another alternative to consider.

The problem is not immediate, but will need to be addressed in the longer term if we are to maintain cross-platform support.

zur003:
Resolves #7469.

Eliminates use of Windows-based elements of System.Drawing.Common, by using Gdk.Pixbuf instead of Image or Bitmap. Also converts mapping from SharpMap to Mapsui. [Mapsui](https://mapsui.com) was developed from SharpMap, but uses SkiaSharp for its rendering, and is more actively maintained and updated. Map retrieval is done asynchronously, and uses the same Brutile cachinbg mechanism as did Sharpmap.

### DIFF
--- a/APSIM.Documentation/Program.cs
+++ b/APSIM.Documentation/Program.cs
@@ -32,6 +32,7 @@ namespace APSIM.Documentation
         {
             try
             {
+                Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
                 // Get autodocs config - ie which models to document.
                 IEnumerable<IDocumentationTable> tables = new[]
                 {
--- a/APSIM.Interop/APSIM.Interop.csproj
+++ b/APSIM.Interop/APSIM.Interop.csproj
@@ -25,7 +25,7 @@
     <PackageReference Include="PdfSharpCore" Version="1.2.11" />
     <PackageReference Include="OxyPlot.Core" Version="2.1.0-Preview1" />
     <PackageReference Include="OxyPlot.SkiaSharp" Version="2.1.0-Preview1" />
-    <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="2.88.0-preview.127" />
+    <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="2.88.3" />
     <PackageReference Include="HarfBuzzSharp.NativeAssets.Linux" Version="2.8.2-preview.127" />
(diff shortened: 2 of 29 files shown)


zur003:
@hol430 - It appears from the log that libgdk_pixbuf-2.0-0.dll is not being loaded in the Jenkins tests. Do you know why this might be?

hol430:
Yeah, the jenkins tests run in a docker container which doesn't have gtk (or any of the related libs) installed. Even if it were installed, calls to `gtk_init()` would fail anyway (something related to lack of a display as I recall). The error is occurring because some of the APSIM.Interop unit tests exercise the markdown renderers with a mocked-out pdf builder.

This might(?) be why I used the System.Drawing types in the api in the first place here, to avoid calling into the gdk api. Does skia offer an `Image`-like type? If so, using that in the api might solve the problem, if the skia libraries can be loaded in the test environment.

zur003:
@hol430 - Thanks for the explanation - I thought it might be something like that. I'll see what I can do with SKImage, at least for the bits in APSIM.Interop.

zur003:
Retest this please, Jenkins

PR #7720, opened 2023-01-05
Use IClock instead of Clock consistently

mpastell:
This pull request replaces the use of Clock with IClock in all of the models where Clock was used. This enables implementation of custom clocks, which I assume (hopefully correctly) is the reason that IClock exists.

The motivation for using a custom Clock for me is to be able to run simulation with daily time step trough Python.

I'm happy to make revisions based on feedback. After making the changes  874 tests pass and 13 fail.

### DIFF
--- a/Models/AgPasture/PastureSpecies.cs
+++ b/Models/AgPasture/PastureSpecies.cs
@@ -29,7 +29,7 @@ namespace Models.AgPasture
 
         /// <summary>Link to APSIM's Clock (provides time information).</summary>
         [Link]
-        private Clock myClock = null;
+        private IClock myClock = null;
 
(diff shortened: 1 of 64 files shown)


hol353:
add to whitelist jenkins

hol353:
Thanks for contributing to APSIM. 

This PR needs to have a 'working on' or 'resolves' message in the top comment box that refers to a related GitHub Issue. See https://apsimnextgeneration.netlify.app/contribute/sourcetree/pushandpullrequest/ - look under 'Open Pull Request' section. You can edit the top comment box by clicking ... at the right of the comment box.

mpastell:
Replaced with #8016

Issue #7734, opened 2023-01-23
Cultivar changes sticking around


ilhuber:
It seems like cultivar parameters will remain changed after the crop has ended. If the next cultivar of the same crop to be sown does not change a parameter a previous cultivar changed, then this new cultivar will inherit the older parameter changes.

ilhuber:
With #7735 merged, this issue has been resolved.

PR #7735, opened 2023-01-23
Undo cultivar changes on plant ending.

ilhuber:
Working on #7734 

OilPalm also uses Cultivar objects, but it doesn't look like that plant is ever meant to be ended so I didn't make any changes.


ilhuber:
retest this please jenkins

PR #7741, opened 2023-01-25
Fixes to cell editing in SheetView

### LINKED ISSUE #7738: Grid data entry - key different to keyboard numbers
Probably can be merged with another issue but I can't find one. 

There is a problem where it can be hard to get into and then stay in edit mode to enter values in a grid. Today I found that if I use the numbers keys above the main keyboard I get data entry but if I use the keypad then I have to double-click on the cell first. Can this be addressed?

hol430:
Mainly related to the use of the numpad, although I've also added support for some other, more obscure keyboards. To be honest there is probably more to be done here: e.g. using tab/enter for grid cell navigation when *not* in edit mode (currently this does nothing).

Resolves #7738.


zur003:
Retest this please Jenkins

PR #7754, opened 2023-01-31
Interact with the GUI on the main thread.

### LINKED ISSUE #7752: The GUI freezes under Ubuntu Linux when MapView is activated.
I though this had been working, but with a Net 6 build on Ubuntu Linux, I'm finding that the GUI stops responding after I select a "Map". If I try any operations, I get a '"ApsimNG" is not responding' dialog with the options to "force quit" or "wait". Waiting doesn't help. 

This is rather odd because I can generate Documentation for the simulation that includes the map in the resulting PDF as intended, and the code used in MapView is mostly similar. Other components for the view (e.g., for entering zoom level) appear correctly, and the map itself appears to render briefly before the map image region goes blank and everything locks up.

I've been frustrated in my efforts to debug this, as I've been unable to debug ApsimNG under MS Code. Code does the build properly, but ApsimNG starts then quickly fails with a GException indicating a failure to load libpixbufloader-svg.so and some other stuff suggesting it might be a problem with __libc_vfork and glibc. However, the executable assembly that Code has built loads and runs correctly outside the GUI. I suspect a possible conflict between different versions of glibc, but haven't been able to sort it out.

zur003:
Resolves #7752

### DIFF
--- a/ApsimNG/Views/MapView.cs
+++ b/ApsimNG/Views/MapView.cs
@@ -253,12 +253,12 @@
                         while (countryLayer.Busy && nSleeps++ < 700)
                             System.Threading.Thread.Sleep(10);
                     }
-                } 
+                }
+                osmLayer.DataChanged -= OsmLayer_DataChanged;
                 Mapsui.Rendering.Skia.MapRenderer renderer = new Mapsui.Rendering.Skia.MapRenderer();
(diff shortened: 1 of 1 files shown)


zur003:
Retest this please Jenkins

zur003:
Retest this please Jenkins

PR #7756, opened 2023-02-01
Update from .Net 3.1 to .Net 6.0

zur003:
Working on #7139.

Do not merge (yet). This is for testing purposes


hol430:
This looks great Eric, and I'm able to run this without any obvious problems. A couple of thoughts:

- Probably safe to delete the old .net core 3.1 installer scripts. If anyone needs these they can always dredge them up from version control
- Do we still need the gtk-sharp3 dependency in the control file for the debian installer? I'm not sure what this provides -- maybe the glue libraries for the old gtk-sharp (ie not GtkSharp) dependency? Either way, I don't think the new GtkSharp from nuget has any dependencies other than the native libraries, meaning we should be able to replace the dependency on `gtk-sharp3` with a dependency on `gtk3` (or whatever this package is called in Debian-land)

I'll also mention that I had to set the environment variable `CLR_ICU_VERSION_OVERRIDE=72.1` in order to get this to run (on Linux), but this is a problem with my .net setup, not an apsim problem. I may have a more recent version of libicu installed than is supported by my .net runtime installation. Just mentioning this here in case someone else hits the same snag.

hol353:
retest this please jenkins

hol353:
I've moved to .NET 6.0 in our apsimng-build docker container and rerun this PR. The problem now is that various scripts in the 2 repos Jenkins and APSIM.Docker refer to netcoreapp3.1.

e.g. Jenkins\NextGen\container-scripts\run-validation-tests:

```
find . -not \( -path ./Tests/UnitTests -prune \) -not \( -path ./ApsimNG -prune \)  -name '*.apsimx' -print0 | xargs -0 dotnet bin/Release/netcoreapp3.1/Models.dll --verbose --run-tests
```

If I change these files and commit and push the changes to the 2 repos then this will break the build system until this PR is merged. I think one of us (probably me) will have to get the build system running on my local PC to test all this and then commit, push and merge everything all at once.

Take home message: all files for the build system need to be in the same repo so that a PR can make changes in APSIM as well as the build scripts.

zur003:
@hol430 - Unfortunately, building for both .Net 3.1 and 6 isn't quite that simple. Some of the packages (e.g., Kubernetes) don't support both simultaneously, or at least can do so only with some difficulty.

jbrider:
We're not using kubernetes from within ApsimServer any longer - so it can be removed if that is an obstruction.
We're not using ApsimServer either - but I think others might be.

zur003:
Following the suggestion of @hol430, this now builds for both .Net 3.1 and 6.0, so it can probably be merged now, with changes made to the build scripts later.

zur003:
@hol353 - I think you're correct in that we should be able to change the Debian dependency from gtk-sharp3 to libgtk-3-0. I haven't actually tested this, though.

PR #7758, opened 2023-02-03
Use water concentrations, not amounts, for layer-mapping when getting…

### LINKED ISSUE #7739: What on earth is going on with the initial water content?!?
This is what it looks like to start with:
image

Then I enter a "7" in the "Percent full" box and get - of all things!!!
image

The I use the backspace key to remove the final "2" of the 112 that I did not enter and get (note where my cursor ends up - might assist with the bug fixing?)
image

zur003:
… water FractionFull. Resolves #7739.

### DIFF
--- a/Models/Soils/Water.cs
+++ b/Models/Soils/Water.cs
@@ -172,9 +172,9 @@
         {
             get
             {
-                double[] dulMM = SoilUtilities.MapConcentration(Physical.DULmm, Physical.Thickness, Thickness, Physical.DULmm.Last());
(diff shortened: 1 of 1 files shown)


hol430 (approved):
Makes sense to me, though not really my area of expertise. Would this be equivalent to calling MapMass() with the dulmm values?

zur003:
@hol430 - Good point. They normally give the same overall result, but MapConcentration has an additional argument which provides a default value to use if the bottoms of the profiles don't align. There may be cases where that would be useful.

hut104:
Should we be adding tests for these sorts of changes to ensure that they do not get broken again?

PR #7768, opened 2023-02-08
Models now displaying all resources in add model context menu

### LINKED ISSUE #7467: Adding released crop models into the standard toolbox
I noticed that some of the released crop models are not updated in the standard toolbox while working on some APSIMX tutorials. Might be worth adding them in please.

ric394:
Resolves #7467

- Crops were already displaying in the add model context menu.
- Removed standard toolbox from GUI
- Removed associated files.


rowaneisner:
So is irrigation no longer supported? It doesn't work without the module from the standard toolbox. And I'm teaching it the day after tomorrow!

rowaneisner:
Ah sorry, just found it from your previous answer.

PR #7776, opened 2023-02-10
MacOS installer: always build in self-contained mode

### LINKED ISSUE #7765: Mac installs broken?
Can anybody verify that the mac installer is working? A current install (7137) won't start, but from last year (7130) does.

It appears the dotnet runtime isnt being installed: image

hol430:
Resolves #7765

This might fix #7765 but I don't have a mac to test on so don't really know for sure. I had thought `--self-contained` was true by default when publishing, but maybe that has changed in recent sdk versions.


hol353:
retest this please jenkins

Issue #7086, opened 2022-02-09
Accessing messages from Summary


lie112:
Has the previous Summary.Messages() been deprecated and replaced by Summary.GetMessages()?
It seems the private DataTable messages only holds the empty table (schema) for the messages table (created during Initialise()) and no longer holds reported messages. It doesn't appear this is used anywhere. The private messages that it returns is Cloned() in WriteMessages to create the table to be written.
Previously I used Summary.Messages to access the current messages, but it now returns only the empty table.
I found the GetMessages() method returning a much better list of Messages.
Would it be worth removing the old Summary.Messages() method to avoid confusion as it doesn't appear to be used anywhere?

hol430:
Yeah, `Messages()` should be removed.

hol353:
@ric394 : Just remove Messages() from Summary.cs

PR #7807, opened 2023-02-20
Issue #7086 fix Remove messages() method from Summary.cs

ric394:


ric394:
Additionally commits included. I did not want these included.

PR #7809, opened 2023-02-20
Issue #5528 fix

### LINKED ISSUE #5528: Number should be zeroed in clear method in ReproductiveOrgan
Number is grain number.  The masses are zeroed in Clear but not the number.

ric394:
Resolves #5528

@hut104 @hol353 Does it matter if the properties are cleared instead of reinitialised?


hol353:
Clearing the properties is fine.

PR #7816, opened 2023-02-21
Prevent dry soils from reporting a water table depth of 0.

### LINKED ISSUE #7815: Water table depth reported as 0 in dry soils.
It looks like in the default WaterBalance model, water table depth will be reported as 0 when no layers are saturated, rather than the bottom depth of the profile.

ilhuber:
Resolves #7815


ilhuber:
The issue with the FACTS watertable simulation is expected, but I am not sure what is causing the difference in the Oats simulation.

PR #7825, opened 2023-02-23
Small changes to get YP running Next Gen

### LINKED ISSUE #7824: The accumulator class needs to be more flexible
Currently the accumulator class only has a Sum method. Expose the underlying values via a public property so that other stats can be calculated.

hol353:
Resolves #7824


hol353:
retest this please jenkins

Issue #7835, opened 2023-03-01
Sum("non-scalar type") report variable not raising error on simulation run


ric394:
Tried to have a report variable as:

"sum([Soil].Nutrient.Denitrification) as Denitrification"

This did not report an error on run when it should have, as 'summing' all the members of that class would not make sense.

peter-devoil:
In the same area, adding a second report variable with the same "as" name as another variable (specifically, a factor name under the simulation's experiment node) is silently ignored

ilhuber:
Looks like IFunction types will also report 0 when used in an expression, while reporting their .Value() (without any index argument) when not. The code to handle these cases appears to be in:
https://github.com/APSIMInitiative/ApsimX/blob/71491ae2064212ec17d2f94f451f03b303f3be90/Models/Core/VariableExpression.cs#L133

par456:
This issue has been fixed.

PR #7837, opened 2023-03-03
Throw an error for unrecognized types in VariableExpression.

ilhuber:
Also, allows for IFunctions to have their .Values() used in expressions. Previously, unknown types would silently substitute in scalar 0 when used in expressions. Doesn't address @peter-devoil's concerns about multiple 'as' variables with the same name being silently ignored. 

Working on #7835


ilhuber:
retest this please jenkins

ilhuber:
I am a little confused. If a simulation in the test set was encountering an error due to an unrecognized type, shouldn't we have seen that in the Jenkins error message?

Issue #7869, opened 2023-03-21
SCRUM upgrade to be more general


HamishBrownPFR:
Work approved by Steering comittee and funded by AI.  See attached for project description
[Proposal for APSIM initiative SCRUM.docx](https://github.com/APSIMInitiative/ApsimX/files/11034756/Proposal.for.APSIM.initiative.SCRUM.docx)

PR #7870, opened 2023-03-21
Alter sow method so cultivar overwrite object can be passed in rather…

HamishBrownPFR:
… than needing to find it from the crop model.  This will be needed for the generalised verson of scrum that will calculate parameters externally and pass them into the model

Working on #7869


hol353:
Why do we need a 'cultivarOverwrites' argument added to the sow method call? What is the use case? 

We already have 'Cultivar' class which overrides things in the plant model.
We also have a 'ModelOverrides' class that can be dropped on a parent plant to allow for a generic plant with overrides for specific plant species. 

Both of these have a public string[] property that specifies what needs to be overridden. These properties could be modified by a manager script or some other model if you are wanting to calculate parameters at runtime.

HamishBrownPFR:
@hol353.  There will be some parameters in the generalized version of SCRUM that are calculated from user specified information each time the crop is sown so the Cultivar and ModelOverwrites approaches are not suitable.  I was thinking passing a cultivar object in at sowing would be a tidy way of doing this but can look at writing something that using the set method instead.

hol353:
Presumably you must already have a manager script that creates the cultivar to pass to SCRUM. Rather than changing PMF, can you use set lines in the manager script instead? I don't really want more complexity and more ways to do things in PMF if there is an alternative.

HamishBrownPFR:
I will see what I can come up with.  Design for the Scrum is still pretty fluid at this stage.  Would you be keen for me to run you through where I am at to get your feed back?

hol353:
Yes, might be good to get a quick update on where you're going with SCRUM.

HamishBrownPFR:
I have changed the approach so cultivar does not need to be passed in.  Closing this pull request as redundant now

PR #7873, opened 2023-03-22
refactor HourlyInterpolation.cs + inner classes

### LINKED ISSUE #7856: HourlyInterpolation.cs needs refactoring
1. There are multiple classes within the one cs file
2. The hourlyinterpolationclass does not necessarily do hourly interpolation - the name is misleading.  This class is used most regularly with 3-hourly interpolation.  It should be renamed to SubDailyInterpolation or something
3. The ThreeHourSin class does not use a sine curve.  The name is misleading
4. The file should be refactored to be much clearer for users reading the code and viewing models in the UI.

ric394:
Resolves #7856


ric394:
retest this please jenkins

PR #7874, opened 2023-03-22
Updated label for soil download glade

### LINKED ISSUE #7822: Download soil placename means postcode
Using the Eucalyptus example, for example, downloading an Australian soil in a simualtion using right-click on a Zone and providing place name 'Mount Gambier' returns no soils, but justing using it's postcode '5290' does.  A quick fix would be to change the input required to Postcode, but providing place name and postcode options would be more useful.

ric394:
Resolves #7822 
- Improves clarity on search terms.


ric394:
retest this please jenkins

PR #7875, opened 2023-03-23
#7304 report conversion and clear status msg

### LINKED ISSUE #7304: Report simulation file conversion and clear errors
It would be useful if the status window reported that a simulation file had been converted to the latest version for full transparency that the file may have been modified.
Also if you get a json error deserialising the simulation file it would be good if this was deleted when opening another simulation to avoid confusion.
On file open, clear the status window. Is there any reason old messages would be wanted upon opening new simulation, even with multiple tabs open?

ric394:
Resolves #7304

Clears status messages and now reports if simulation has been converted.


hol353:
Please remove wheat.apsimx from this PR.

hol353:
wheat.apsimx is still in this PR.

PR #7880, opened 2023-03-24
Changed Enabled property of models to enable children as well. 

### LINKED ISSUE #7860: Running SWIM and SoilWat together using disable/enable factors
Attached run was intended to work as follows:
- have all SWIM and SoilWat specific nodes (6 of them) disabled
- use factors to enable relevant nodes

1. This intermittently results in an error "Cannot find a match for link lateralFlowModel in model.Yuluma2021HeaderStripperModelSoilWAT.Field.Soil.SoilWater
The error does not always show up.
If opened from scratch with SoilWater disabled it runs fine.
But error occurs when it has first been run with SoilWater node (erroneously) enabled and then rerun with it (correctly) disabled. 
It must keep something in its memory, even if data store is emptied before rerunning.

2. If erroneously having SoilWater node enabled, then the SWIM runs will find the relevant SWIM related nodes, but also the SoilWater node. This seems to override the simulated soil water contents and Es (and possibly more, but those were the ones I plotted) - resulting in no change in these variables. 
The simulated soil water potentials did show the correct behaviour, which is how we picked this up.
Would be good to include a warning that SWIM can't be run with SoilWater enabled.
However, please be careful with the implementation of that as our new way of doing these runs is to have them all enabled and using the factor to disable the relevant nodes. So you wouldn't want the check for "wrongly enabled nodes" come too early.

[Yuluma.zip](https://github.com/APSIMInitiative/ApsimX/files/11012594/Yuluma.zip)

hol353:
Also added check for multiple soil water models.

Resolves #7860


hol353:
retest this please jenkins

par456:
retest this please jenkins

PR #7889, opened 2023-03-31
fix for invalid leaf sizes

### LINKED ISSUE #7888: Leaf calculation allowing negative leaf area
certain leaf configurations are causing leaf area calcs to return leaf area's less than 0.

jbrider:
Resolves #7888


jbrider:
@hol353 this can be merged.

PR #7896, opened 2023-04-03
changed model override to only be child of plant

### LINKED ISSUE #7642: One crop exists, but there are two crops selectable in a manager script
I have Wheat in the simulation, and another Wheat under "ModelOverrids". 

image

ric394:
Resolves #7642


ric394:
retest this please jenkins

PR #7897, opened 2023-04-03
Modified Plant.cs to have same value as IPlant for rowConfig value

### LINKED ISSUE #7206: Row Config default value when sowing plant
The IPlant definition of the Sow() method has a different default value for rowConfig than the implementation in Plant.cs. Therefore, calling this method via the interface can give different behaviour than calling the method via a concrete reference to a Plant instance.

```csharp
[Link] IPlant iplant = null;
[Link] Plant plant = null;

private void SowCrop()
{
    iplant.Sow(cultivar, popn, depth, rowSpacing); // passes in rowConfig of 1
    plant.Sow(cultivar, popn, depth, rowSpacing); // passes in rowConfig of 0
}
```

ric394:
Resolves #7206

### DIFF
--- a/Models/Core/IPlant.cs
+++ b/Models/Core/IPlant.cs
@@ -40,7 +40,7 @@ namespace Models.Core
         /// <param name="seeds">The number of seeds sown (/m2).</param>
         /// <param name="tillering">tillering method (-1, 0, 1).</param>
         /// <param name="ftn">Fertile Tiller Number.</param>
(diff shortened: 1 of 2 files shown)


hol353:
RowConfig needs to have a default of 0, not 1. A value of 1 means every second row isn't planted. This isn't what users would expect.

ric394:
retest this please jenkins

PR #7906, opened 2023-04-04
Fix problem with code completion service.

### LINKED ISSUE #7893: At the tip revision (29 March 10:18) intellisense in Managers is locking up
It is working in Report
Reverting back to to the revision on 23 March 17:09 fixed it
UI completely locks up. If in Debug mode then can stop from VS otherwise need task manager

zur003:
Resolves #7893.

Well, it works, although I don't really understand synchronization contexts enough to explain _why_ it works. Perhaps @hol430 could review this. How did you work out how to use code completion? It doesn't seem to be documented by Microsoft.

### DIFF
--- a/ApsimNG/Intellisense/CodeCompletionService.cs
+++ b/ApsimNG/Intellisense/CodeCompletionService.cs
@@ -127,9 +127,8 @@ namespace UserInterface.Intellisense
 
             foreach (Document document in documents)
             {
-                SourceText source = await document.GetTextAsync();
                 CompletionService service = CompletionService.GetService(document);
-                CompletionList completionList = await service.GetCompletionsAsync(document, offset);
+                CompletionList completionList = await service.GetCompletionsAsync(document, offset).ConfigureAwait(false);
 
                 if (completionList != null)
                 {


zur003:
Retest this please Jenkins

hol430:
I believe that `.ConfigureAwait(true)` (the default) can deadlock in some situations when synchronously waited upon. If this fixes the problem, then I wonder how this ever worked in the first place. It may be worth using `.ConfigureAwait(false)` for the other awaited expressions in this function as well, though I haven't tested this.

As for how I implemented this in the first place...good question. I think I found an open source C# repl which used roslyn internally, and had a look at its code to get a feel for the api. I can't remember the name of the project though. There was definitely a frustrating lack of documentation at the time.

ilhuber:
The following diff also appears to resolves the problem without needing to ConfigureAwait:
```diff
--- a/ApsimNG/Classes/ScriptCompletionProvider.cs
+++ b/ApsimNG/Classes/ScriptCompletionProvider.cs
@@ -275,19 +275,15 @@ public bool Match(CompletionContext context)
         /// gtk_source_completion_context_add_proposals() function.
         /// </summary>
         /// <param name="context"></param>
-        public void Populate(CompletionContext context)
+        public async void Populate(CompletionContext context)
         {
             try
             {
                 string code = context.Iter.Buffer.Text;
                 int offset = context.Iter.Buffer.CursorPosition;

-                Task<IEnumerable<NeedContextItemsArgs.ContextItem>> task = service.GetCompletionItemsAsync(code, offset);
-                task.Wait();
-                if (task.Result == null)
-                    return;
+                var contextItems = (await service.GetCompletionItemsAsync(code, offset));

-                IEnumerable<NeedContextItemsArgs.ContextItem> contextItems = task.Result.OrderBy(c => c.Name);
                 // Iff owned is true, the list will be freed by calls to Dispose().
                 bool owned = true;
                 // Iff elementsOwned is true, the list elements will be freed
```
Looks like Task.Wait on async code is not recommended: https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html. The author also seems to not be a fan of ConfigureAwait as a catchall solution, but I think the usage here is probably fine?

PR #7923, opened 2023-04-13
7499 Manager Quirks

### LINKED ISSUE #7499: Manager script quirks
A nitpick over the manager presenter:

1. Open any script, 2. type in some garbage that wont compile, 3. press run -> No errors are displayed. I think a red message flashes for a microsecond, but cant be sure. The user eventually finds the error when clicking on another component, but not when running the simulation

par456:
Resolves #7499

The run button will now recompile all the scripts in the model before running the simulation. Will not run the simulation if a script has an error and will show script error.

A context menu option has been added for manager nodes, this allows you to compile the script without the page refreshing like when clicking a node or the run button. This should fix the undo problem and keeps the script tab open.

### DIFF
--- a/ApsimNG/Menus/MainMenu.cs
+++ b/ApsimNG/Menus/MainMenu.cs
@@ -1,19 +1,17 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Reflection;
-using APSIM.Shared.JobRunning;
-using APSIM.Shared.Utilities;
-using global::UserInterface.Commands;
-using global::UserInterface.Hotkeys;
-using Models;
-using Models.Core;
-using Models.Core.Run;
-using Utility;
-
-namespace UserInterface.Presenters
+﻿namespace UserInterface.Presenters
 {
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Reflection;
+    using APSIM.Shared.JobRunning;
+    using APSIM.Shared.Utilities;
+    using global::UserInterface.Commands;
+    using global::UserInterface.Hotkeys;
+    using Models.Core;
+    using Models.Core.Run;
+    using Utility;
 
     /// <summary>
     /// This class contains methods for all main menu items that the ExplorerView exposes to the user.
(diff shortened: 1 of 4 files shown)


ApsimNG/Presenters/ManagerPresenter.cs
```diff
@@ -170,12 +170,16 @@ private void BuildScript()
 
                 string code = managerView.Editor.Text;
 
+                //set this to false prior to compiling in case an error is thrown
+                manager.SuccessfullyCompiledLast = false;
+
```
hol353, line 175: Why doesn't manager set this flag when it tries to do a compile?

Models/Management/Manager.cs
```diff
@@ -99,6 +99,14 @@ public string Code
...
 
+        /// <summary>
+        /// Stores the success of the last compile
+        /// Used to check if the binary is up to date before running simulations
+        /// Prevents an old binary brom being used if the last compile had errors
+        /// </summary>
+        [JsonIgnore]
+        public bool SuccessfullyCompiledLast { get; set; } = false;
+
         /// <summary>
```
hol353, line 110: If manager sets this property then it can be made private set;

ApsimNG/Menus/MainMenu.cs
```diff
@@ -199,6 +201,17 @@ public void OnRun(object sender, EventArgs e)
...
+                foreach (Manager script in scripts)
+                {
+                    if (script.SuccessfullyCompiledLast == false) {
+                        //one of our scripts had an error
+                        //throw an expection to stop simulations from running with old binaries
+                        throw new Exception("Errors found in manager model " + script.Name);
+                    }
+                }
+
                 Runner runner = new Runner(model, runType: Runner.RunTypeEnum.MultiThreaded, wait: false);
```
hol353, line 215: Instead of doing the above code in MainMenu.cs, the manager could throw the exception in it's 'OnStartOfSimulation' method. This method will be called whenever a simulation starts running.

hol353 (changes requested):

ApsimNG/Presenters/ManagerPresenter.cs
par456: That is a good point, I've moved the flag setting to the Manager.RebuildScriptModel() function where it should have been the first time. I've then reverted all changes to the ManagerPresenter.

Models/Management/Manager.cs
par456: Made private now that manager handles the flags

ApsimNG/Menus/MainMenu.cs
par456: The exception is now thrown from the OnStartOfSimulation function.

hol353 (approved):

PR #7933, opened 2023-04-18
7524 X2Y2 field now controlled in DisableUnusedControls

### LINKED ISSUE #7524: Error after switching off graph series "on right?"
If the "On right" switch has been set previously and then switched off the values of X2 and Y2 remain set in the apsimx json file.
This is fine while these fields remain in the database but if they are changed or removed a warning is thrown when selecting the series 

WARNING: .Simulations.Graphs_Summary.Females numbers by class.FBreeders: Selected X2 field name 'TotalAE' does not exist in the datastore table 'ReportResourceBalances'. Have the simulations been run?

There is no way to fix this other than edit the json file to set X2 and Y2 to null or delete and recreate the series.

When "on right" is turned from on to off, the values need to be set to null.
I cannot determine a way of achieving this through the UI.

par456:
Resolves #7524

There were a few bugs happening here.

First, the X2 and Y2 field would only display if the Region was selected before the node was clicked in the tree, changing the graph type would not refresh the field's visibility. That was fixed by moving the ShowX2Y2 function call into DisableUnusedControls which is called both when the page is first made, and when the graphtype field is changed

To solve the actual reported bug, the X2 and Y2 field names are now null'ed if the fields are disabled, such as if the graph type is changed from region.

Lastly the X2 and Y2 field were appearing in the wrong order in menu (x, y, y2, x2), so I switched their positions around.

### DIFF
--- a/ApsimNG/Presenters/SeriesPresenter.cs
+++ b/ApsimNG/Presenters/SeriesPresenter.cs
@@ -218,6 +218,7 @@
 
             // This doesn't quite work yet. If the previous series was a scatter plot, there is no x2, y2 to work with
             // and things go a bit awry.
+            // this function is now called in disableUnusedControls if this needs to be added back later
             // this.seriesView.ShowX2Y2(series.Type == SeriesType.Area);
 
             // If the series is a box plot, then we want to disable certain unused controls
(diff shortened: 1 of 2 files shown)


par456:
retest this please jenkins

par456:
retest this please jenkins

PR #7936, opened 2023-04-21
7847 Incorporated/Tilled Event

### LINKED ISSUE #7847: Add an Incorporated or Tilled event into SurfaceOrganicMatter
Use case: A manager script needs to trap (subscribe to) the event to perform calculations. The event data should contain fraction incorporated and depth of incorporation. These will be useful in the manager script.

par456:
Resolves #7847

Surface Organic Matter will now invoke a Tilled event when Incorporate is called from a manager script. 

Unit test has also been created that will run a small simulation to check the event is being called, checking that it invokes four times and reduces LyingWt as expected with the Incorporate function.


par456:
retest this please jenkins

PR #7938, opened 2023-04-21
7935 Column Captions set to Column Name after read

### LINKED ISSUE #7935: Failure to successfully upload the wather data
I'm new to using APSIM, and therefore I have been using the manuals to learn. I formatted my Excel file containing weather data following the format recommended but am not getting the correct data when I import it into APSIM. 

This is from the manual
Apsim Issue 0

Data view after importing

Apsim Issue 1

Summary of my imported data
Apsim Issue 1 1

Apsim Issue 1 2

Apsim Issue 1 3

par456:
Resolves #7935

A weird little bug where the excel to table converter used by the weather class would only set the column name and not the column caption when reading from an Excel file. Ended up adding a loop right after it calls the system code to copy the name into the caption for each column.

### DIFF
--- a/APSIM.Shared/Utilities/ApsimTextFile.cs
+++ b/APSIM.Shared/Utilities/ApsimTextFile.cs
@@ -389,7 +389,19 @@ namespace APSIM.Shared.Utilities
             }
             //will I ever hit this without having any data???
 
-            return data;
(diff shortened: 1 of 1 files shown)


par456:
retest this please jenkins

PR #7942, opened 2023-04-24
7495 Cultivars under a resource can now be edited

### LINKED ISSUE #7495: Unable to rename pasted nodes (for example new cultivars)
Hi @hol353
So the problem of being unable to rename the cultivars is still happening.
It's keeps on saying that the "it's read-only".
The error information is:
Models.Core.ApsimXException: Unable to rename Kerrin1 - it is read-only.
   at UserInterface.Commands.RenameModelCommand..ctor(Model modelToRename, String newName) in C:\Users\andreucm\ApsimX\ApsimNG\Commands\RenameModelCommand.cs:line 33
   at UserInterface.Presenters.ExplorerPresenter.OnRename(Object sender, NodeRenameArgs e) in C:\Users\andreucm\ApsimX\ApsimNG\Presenters\ExplorerPresenter.cs:line 1007

Thanks for that Dean :)

 
RenameIssue

par456:
Resolves #7495

Adds a simple check into the "Show Model Structure" and "Read Only" context menu options so that Cultivar nodes under a resource can be changed from read only and be edited like they are supposed to.

### DIFF
--- a/ApsimNG/Menus/ContextMenu.cs
+++ b/ApsimNG/Menus/ContextMenu.cs
@@ -816,7 +816,7 @@ namespace UserInterface.Presenters
                     foreach (IModel child in model.FindAllDescendants())
                     {
                         child.IsHidden = hidden;
-                        if (isResource && String.Compare(child.ToString(), "Models.PMF.Cultivar") != 0)
+                        if (isResource && !(child is Models.PMF.Cultivar))
                             child.ReadOnly = true;
                     }
                     foreach (IModel child in model.Children)
@@ -898,7 +898,7 @@ namespace UserInterface.Presenters
                     return;
 
                 // Don't allow users to change read-only status resource children, except for Cultivars
-                if ((model.FindAllAncestors().Any(a => !string.IsNullOrEmpty(a.ResourceName))) && String.Compare(model.ToString(), "Models.PMF.Cultivar") != 0)
+                if ((model.FindAllAncestors().Any(a => !string.IsNullOrEmpty(a.ResourceName))) && !(model is Models.PMF.Cultivar))
                     return;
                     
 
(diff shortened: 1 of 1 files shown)


ApsimNG/Menus/ContextMenu.cs
```diff
@@ -816,7 +816,7 @@ public void ShowModelStructure(object sender, EventArgs e)
                     foreach (IModel child in model.FindAllDescendants())
                     {
                         child.IsHidden = hidden;
-                        if (isResource)
+                        if (isResource && String.Compare(child.ToString(), "Models.PMF.Cultivar") != 0)
```
hol353, line 819: How about saying:

```c#
if (isResource && child is Models.PMF.Cultivar)
```

ApsimNG/Menus/ContextMenu.cs
```diff
@@ -892,10 +892,15 @@ public void ReadOnly(object sender, EventArgs e)
...
-                
+
                 // Don't allow users to change read-only status of released models.
-                if (!string.IsNullOrEmpty(model.ResourceName) || model.FindAllAncestors().Any(a => !string.IsNullOrEmpty(a.ResourceName)))
+                if (!string.IsNullOrEmpty(model.ResourceName))
+                    return;
+
+                // Don't allow users to change read-only status resource children, except for Cultivars
+                if ((model.FindAllAncestors().Any(a => !string.IsNullOrEmpty(a.ResourceName))) && String.Compare(model.ToString(), "Models.PMF.Cultivar") != 0)
                     return;
```
hol353, line 902: Same as above. Use ```model is Models.PMF.Cultivar```

ApsimNG/Menus/ContextMenu.cs
par456: Agreed, that's a much simpler way.

ApsimNG/Menus/ContextMenu.cs
par456: And will do the same here.

PR #7943, opened 2023-04-26
7859 Check m to avoid dividing by 0

### LINKED ISSUE #7859: Spikes in predicted soil water potential with SoilWat
Simulation where we simulate soil water potential. On two occasions where the conditions get wet, we get a wild negative spike in water potential. See 6 cm water potential and also the water retention curve figure.
[Temora.zip](https://github.com/APSIMInitiative/ApsimX/files/11012532/Temora.zip)

par456:
Resolves #7859 

When doing suction calculations for water potential, it was possible to divide by 0 and end up generating very strange numbers. I've added a check to not do the calculation if m is 0, though this might fail unit tests if any had accidentally been triggering this issue.

**Is stopping the calculation is the correct action? or if it should it just use a minimum m instead.**

### DIFF
--- a/Models/Soils/SWIM/HyProps.cs
+++ b/Models/Soils/SWIM/HyProps.cs
@@ -229,7 +229,7 @@ namespace Models.Soils
 
                     double m = (est2 - est) / dpF;
 
-                    if ((Math.Abs(est - theta) < tolerance) || (m == 0))
+                    if ((Math.Abs(est - theta) < tolerance) || Math.Abs(m) < tolerance)
                         break;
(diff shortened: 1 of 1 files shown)


par456:
retest this please jenkins

hol353:
Was this merged?

par456:
Nope. I had a pretty long chat with Neil trying to work out why it was becoming 0, and he worked out there is a flaw in the calculations that should be addressed instead. So instead of patching the problem, he's going to investigate a bit further.

PR #7949, opened 2023-04-27
7454 Soil Interface

### LINKED ISSUE #7454: Soils Interface Issues

### LINKED ISSUE #7574: Sorghum Tillering method to use
Is setting tillering method to "rule of thumb" using the same rules that we used when we let the sorghum module in Classic set fertile tiller number using inbuilt rules based on population, row configuration, latitude and sowing date?

par456:
Resolves #7454 

This was a big old issue that had lots of little things that needed updating. I split each issue that was fixed across a different commit.

This pull request, combined with the other two that were done previous on this issue, should now resolve all the problems that were identified. **Main thing to check is how I implemented the graph on the SoilCrop node and if that is correct.**

### DIFF
--- a/ApsimNG/Presenters/ProfilePresenter.cs
+++ b/ApsimNG/Presenters/ProfilePresenter.cs
@@ -64,7 +64,11 @@
             propertyPresenter.Attach(model, propertyView, explorerPresenter);
 
             graph = view.GetControl<GraphView>("graph");
-            graph.SetPreferredWidth(0.3);
+
+            //get the paned object that holds the graph and grid
+            Gtk.Paned bottomPane = view.GetGladeObject<Gtk.Paned>("bottom");
+            int paneWidth = view.MainWidget.ParentWindow.Width; //this shoudl get the width of this view
+            bottomPane.Position = (int)Math.Round(paneWidth * 0.75); //set the slider for the pane at about 75% across
 
             numLayersLabel = view.GetControl<LabelView>("numLayersLabel");
 
--- a/ApsimNG/Presenters/ProfilePresenter.cs
+++ b/ApsimNG/Presenters/ProfilePresenter.cs
@@ -7,6 +7,7 @@
     using Models.Soils;
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using Views;
 
     /// <summary>A presenter for the soil profile models.</summary>
@@ -56,8 +57,9 @@
             gridPresenter = new NewGridPresenter();
             gridPresenter.Attach(model, v, explorerPresenter);
 
-            physical = this.model.FindInScope<Physical>();
-            water = this.model.FindInScope<Water>();
+            Soil soilNode = this.model.FindAncestor<Soil>();
+            physical = soilNode.FindChild<Physical>();
+            water = soilNode.FindChild<Water>();
 
             var propertyView = view.GetControl<PropertyView>("properties");
             propertyPresenter = new PropertyPresenter();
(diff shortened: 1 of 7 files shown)


par456:
Yep, I totally put down the wrong issue number when I made this branch. Thanks for that, fixed it.

hol353:
> Main thing to check is how I implemented the graph on the SoilCrop node and if that is correct.

Just pulled your branch and ran it. Looking at the SoilCrop graph:
* Need to change first series data and legend item to 'PAW relative to Wheat LL', not LL15.
* The CLL line looks like it is actually LL15 - change the wording in the legend to LL15.
* Change the bottom graph series legend item (e.g. WheatSoil LL) to just Wheat LL.

Everything else looks good.

par456:
Ignore the branch management messages here, it got very upset when the branch name got changed.

Issue #3871, opened 2019-05-21
The HoldFunction doesn't work if ValueToHold points to Plant.SowingData


hol353:
This is due to Plant.SowingData being null before sowing. The HoldFunction tries to get a value every day from start of simulation. This means an exception is thrown when the crop isn't yet sown.

PR #7950, opened 2023-04-28
7350 Exception is now thrown up the stack

### LINKED ISSUE #7350: Exceptions about PMF model structure are being hidden
When exceptions are thrown by a model in PMF not able to find another model, the exceptions are returned to the GUI or summary file. The user has no indication that something is wrong. The attached demonstrates this problem. The attached simulation runs very slowly because of the exceptions. The rice model in the attached isn't even being sown.

[test.zip](https://github.com/APSIMInitiative/ApsimX/files/8954239/test.zip)

par456:
Working on #7350

The error exception was being discarded in the worker threads with an empty catch, now that will pass the exception up the line and it now gets to the GUI, halts the simulation and displays the error.

This might break a unit test if any had errors that were being hidden by this empty exception handler. Though that's a good thing, since these mistakes slow down the simulation by a huge amount and should be fixed.

### DIFF
--- a/Models/Functions/HoldFunction.cs
+++ b/Models/Functions/HoldFunction.cs
@@ -81,9 +81,9 @@ namespace Models.Functions
             {
                 _Value = ValueToHold.Value();
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-
+                throw new Exception(ex.Message);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


par456:
This exception hiding seems to have been a patch for issue #3871. Checking that problem again now with this branch, trying to get SowingData in a report before any sowing has happened, no longer causes an exception.

So I'm removing the try-catch around in the Hold function and rewriting the unit test to actually check that an exception is thrown.

par456:
retest this please jenkins

PR #7953, opened 2023-05-02
7180 2nd check for events with space in name

### LINKED ISSUE #7180: Models with space in name cannot subscribe to events in reports
Since #7156 models with a space in their name cannot subscribe their events as reporting frequency

For example, I have a reporting frequency
[Resources].Ruminants.Bos indicus.ConceptionStatusChanged
and when it is passed through EventReportFrequency.TryParse, the tokens are created based on splitting by space and it's not a date or expression report frequency.

par456:
Resolves #7180 

This should fix the follow up problems that Adam reported. When it tries to make an event, if fails with the single token, it will then also try with the full line. If that also fails, then it throws an exception same as before.

### DIFF
--- a/Models/Report/EventReportFrequency.cs
+++ b/Models/Report/EventReportFrequency.cs
@@ -24,7 +24,6 @@ namespace Models
             if (tokens.Length == 1)
             {
                 new EventReportFrequency(report, events, tokens[0]);
-                return true;
             } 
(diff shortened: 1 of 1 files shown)


par456:
retest this please jenkins

par456:
retest this please jenkins

PR #7954, opened 2023-05-02
Orchard model

HamishBrownPFR:
working on #3801


HamishBrownPFR:
@hol353 for some reason this pull request is throwing a error in trying to build the nutrient documentation.  However, I don't think I have touched anything that relates to that.  Any suggestions on what is going on here appreciated.

Hamish

hol353:
This is an intermittent problem that we need to fix.

retest this please Jenkins.

hol430:
> This is an intermittent problem that we need to fix.

This is probably a result of the way the mapsui -> skia image conversion has been implemented. I suspect the download is timing out (see `MapRenderer.cs:107`); I don't really have time to look into this now, but a quick and easy fix might be to increase the timeout duration a bit, and to explicitly throw an exception with a helpful error message if `result` is still null after the loop there. But maybe there's a more elegant way to handle this in the mapsui API, I can't say I'm really familiar with it.

HamishBrownPFR:
@hol353 could you merge this one please.

PR #7970, opened 2023-05-09
Four issue templates to remind people to add info

### LINKED ISSUE #7952: Add a bug report template
Look into writing up a template for people when reporting bugs. Need to remind them to include apsimx files with bug reports and generally try and pre-empt the most common mistakes people make when writing an issue.

par456:
Resolves #7952 

We'll start with these and update as needed in the future.


lie112:
The Question template could be more generic than "crop model" as it may also relate to Stock, Pasture or CLEM.
It's difficult to use the term model with the range of users as it covers everything from APSIM itself, to each model developed in APSIM, to the development end of things where each component used is of type IModel.

PR #7973, opened 2023-05-09
Add report column and row filter doc

### LINKED ISSUE #7972: Documentation for report row and column filters
It would be good to have report column and row filter documentation on how to use them.

ric394:
resolves #7972

details how to use column and row filters.


Docs/content/Usage/Report.md
```diff
@@ -0,0 +1,48 @@
...
+```
+[Clock.Today] > "1905-**-**" and [Wheat.AboveGround.Wt] > 1000
+```
+
+## Using report column filters
+
+To filter the column data in you report node, you can use a column filter.
+
+To do so you type the name of the column header. Some examples are:
+
```
hol353, line 39: Would be good to mention that you can also specify a partial column name e.g. 'Wheat' to get all the wheat columns.

ric394: That's added now.

PR #7974, opened 2023-05-10
Allow Max and Min to take two doubles as well

### LINKED ISSUE #7963: Expression parser not fully functional
I am trying to put the following function into the Expression function 
0.22 * exp(-0.001 * max(0,([Physical].DepthMidPoints - 200)))

This doesn't work for two reasons.
1) so far as I can see the code is incomplete and it throws and error when ever more than one argument is given in a max function (infact it appears there are quite a few functions that are not complete but the code is a bit hard to follow)
2) [Physical].DepthMidPoints is an array.  Some parts of the expression evaluator will deal with arrays but I does not appear it will work for in a min function and other parts as well. 

image

par456:
Resolves #7963

Max and Min functions in an expression now work similar to C and will take either an array of number, or two single numbers.

### DIFF
--- a/APSIM.Shared/Utilities/ExpressionEvaluator.cs
+++ b/APSIM.Shared/Utilities/ExpressionEvaluator.cs
@@ -1031,9 +1031,39 @@ namespace APSIM.Shared.Utilities
                     }
                     else if (args.Length == 2)
                     {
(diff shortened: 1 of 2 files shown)


par456:
retest this please jenkins

par456:
Unsurprisingly, allowing bad function expressions to actually throw their errors has caused a bunch of problems to appear. So instead of tackling those here, I'm going to make it a new issue.

PR #7976, opened 2023-05-11
Document XY will use up to 6 decimal places

### LINKED ISSUE #7967: Are MinimumNConc and MaximumNConc in Potato equal to zero?
It seems there is a rounding issue in these two tables, as the values are non-zero in the graphs.

Page 10:
image

Page 11:
image

par456:
Resolves #7967

Previously all XY pairs in a auto-doc would be rounded to 1 decimal place, which made graphs and tables not agree on the numbers. Now it will use up to 6 decimal places depending on what is needed.


par456:
restest this please jenkins

par456:
restest this please jenkins

PR #8008, opened 2023-05-23
Removed unnecessary function from root class

### LINKED ISSUE #5521: Optional function DltThermalTime can be removed from Root class
There is no reference to this function in the source code.  It is referenced in PMF functions in JSON but this can be achieved without the requirement for a function in the class.  

        /// <summary>Used to calc maximim diffusion rate</summary>
        [Link(Type = LinkType.Child, ByName = true, IsOptional = true)]
        public IFunction DltThermalTime = null;

ric394:
resolves #5521


ric394:
retest this please jenkins

PR #8012, opened 2023-05-23
4865 Change CriticalNConc to reference correct MinNConc

### LINKED ISSUE #4865: Eucalyptus Coarse Root Crit N conc points to stem
Eucalyptus Coarse Root Crit N conc points to stem rather than using the local Crit N conc

par456:
Resolves #4865

Followed the instructions from the issue to update the Eucalyptus resource so CourseRoot.CriticalNConc points to the correct MinimumNConc


hol353:
I'm surprised this didn't cause a change in the stats for the Euc validation.

PR #8015, opened 2023-05-23
8014 Changed that property to just net 6.0

### LINKED ISSUE #8014: APSIM.Shared still has reference to .net 3.1
### What happened?

Found this when looking at build settings:

image

Going to search for any others while I'm at it.

Windows

par456:
Resolves #8014

Did a search of the project for any other references to netcore or net 3.1, nothing else important came up.


par456:
On second thought, I check to see whether the NETSTANDARD property was actually being used anywhere in the project, and it is not. We should be able to completely remove that now that 3.1 isn't being targeted.

Issue #7902, opened 2023-04-04
Update files to conform to new style rules


ric394:
Files require restyling to conform to new style rules.

lie112:
Were are the full style rules outlined?

ric394:
@lie112  they are outlined in a file called ".editorconfig".

hol353:
@lie112: We are following the Microsoft C# guidelines here: 

https://apsimnextgeneration.netlify.app/development/software/codingstyle/
https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions

ric394:
working on #7902

My original approach of fixing all errors caused by the turning on of a rule all at once, was too ambitious. 

New approach:
- A new rule will be switched on as a warning.
- Create new branch for fixing specific directory in the project.
- Fix all files in a specific project's directory until it is completed. (This allows others to assist as well if interested)
- Create pull request with "working on" comment.
- Then do the same for each project in the repo.
- At this point the rule will be switched to error, enforcing the standard for all subsequent pull requests
- Then repeat until all desired rules are active.

The location of the rules: Solution Item > .editorconfig
![image](https://github.com/APSIMInitiative/ApsimX/assets/123442863/063fa9ac-c067-4e55-b59f-503c3ceda9ac)

lie112:
Have we considered changing to file scoped namespace (C#10) to remove an unnecessary level of indentation from all our files and manager scripts?

PR #8018, opened 2023-05-24
agpasture directory corrected for using directive placement rule

ric394:
working on #7902


ric394:
retest this please jenkins

PR #8027, opened 2023-05-26
Improved emergence detection in Phenology

### LINKED ISSUE #3558: Remove dependancy between Phenology class and EmergingPhase class
The phenology model requires the emergence phase to be modelled using an emerging phase class.  however, this class could be replaced with a more flexible generic phase to allow much more user input on drivers of the emergence process.

However, the phenology class has the following
    if ((CurrentPhase is EmergingPhase) | (CurrentPhase.End == structure?.LeafInitialisationStage))
                    {
                         Emerged = true;
                    }

The model should be changed to allow emergence to be defined by something other than a class type.

hol353:
Resolves #3558

### DIFF
--- a/Models/PMF/Phenology/Phases/DAWSPhase.cs
+++ b/Models/PMF/Phenology/Phases/DAWSPhase.cs
@@ -1,10 +1,6 @@
 using System;
 using Models.Core;
-using Models.PMF.Organs;
 using Newtonsoft.Json;
-using Models.PMF.Struct;
-using System.IO;
-using Models.Functions;
 using Models.Climate;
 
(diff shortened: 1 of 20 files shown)


hol353:
A better approach is to do what the Chemical class does. It has the same problem of solutes potentially being in different layer structures. It has a method called GetStandardisedSolutes that calls SoilUtilities.Map to map values in one layer structure to another layer structure. I suggest you use the same approach for SoilCrops.

hol353:
Accepted stats because SLURP Phenology.Enabled never used to be true and with this PR it is.

PR #8058, opened 2023-06-07
7568 clock docs

### LINKED ISSUE #7568: Clock Event Documentation
Knowledge of the order of clock events is beneficial for modellers writing manager scripts or sometimes for reporting.

1) The list of events in report (for reporting frequency) includes the CLEM events (which are many).  It is good that they have CLEM in the name to help.  Is it possible for CLEM to use a different list to APSIM?

2) It would be good to have documentation that tells APSIM users the order of clock events during an APSIM simulation. It would be difficult to currently autogenerate this (and therefore it may become incorrect when clock is edited).

par456:
Resolves #7568

A generic documentation function called GetModelEventsInvoked has been added that can be used to read a function in a model and get a list of all the event evoked in that function in the order they appear. This has then been added to clock to print out all the events it can trigger.


hol353:
Why is SpeciesTable.apsimx in this PR?

par456:
> Why is SpeciesTable.apsimx in this PR?

It had a debugging breakpoint in one of it's scripts that was blocking the documentation application from running in debug mode. I can remove that file change if needed, though I think it was just leftover from a past job.

Issue #8066, opened 2023-06-08
Inconsistent Crash in CLEM_Sensibility_IndividualAnimal


par456:
### What happened?

Jenkins has been occasionally reporting an exception and crash in the CLEM_Sensibility_IndividualAnimal validation simulation. Sometimes it runs through fine, other times it crashes.

![image](https://github.com/APSIMInitiative/ApsimX/assets/130029141/a83f153b-def4-40ff-a9d5-adf8dd06d20f)

Running the sim locally shows the same behaviour, with the crash occurring in the function OnCLEMAnimalDeath and is here:
![image](https://github.com/APSIMInitiative/ApsimX/assets/130029141/3224cdaf-4f0a-468f-9c15-a4a09570b724)

The crash happens when the herd has 0 members in it, as the code tries to select from an empty array without checking. This could be checked here, but this could be a symptom of a larger problem if the herd is not supposed to fall to 0.

Looking at the sim, the generator seed is set to 0 (which means it uses the clock as the seed), so alongside addressing this, we'd probably want a set seed here for validation purposes so that Jenkins isn't doing randomised validation.

![image](https://github.com/APSIMInitiative/ApsimX/assets/130029141/1cf99bf2-2c0d-4119-9e8e-0d45138cb408)

### What Operating System are you on?

Windows

par456:
@lie112 This might be relevant to have a look at.

lie112:
Happy to have this deleted for now. I'll look into this issue as it will happen in simulations where the herd crashes to zero too. 
This is some recently changed code, but I have never seen the individual die and the previous code must have handled it
I'll fix this and set the seed so it's not randomised. I am doing some animal production code updates soon so it all needed to change anyway,

PR #8068, opened 2023-06-08
8066 Removing CLEM_Sensibility_IndividualAnimal until fixed

par456:
Working on #8066 

Just removing this from the validation tests until it can be fixed so that other pull requests can be completed, leaving the issue up so it is tracked.


hol353:
retest this please jenkins

PR #8072, opened 2023-06-11
Add error message  if a biomass removal is called when plant ~IsAlive

### LINKED ISSUE #8071: It should not be possible to harvest a crop if it is not alive
### What happened?

Some examples have a harvest() event following and EndCrop() event.  This harvest event calls through some crop code and changes settings even though the crop is not in the ground and can cause problems.  The user is probably also unaware that their harvest event is not at the right time.

Windows

HamishBrownPFR:
resolves #8071

This will throw some errors in the test and redundant Harvest operations will need to be removed to resolve


HamishBrownPFR:
@hol353 Can you merge this one if happy with it

Issue #8023, opened 2023-05-25
Phenology skip not working


HamishBrownPFR:
### What happened?

Phenology.SetToStage(stage) is not working if setting to an advanced stage (skipping forward)  works if rewinding.  I suspect the code for skipping has never been exercised

### What Operating System are you on?

Windows

PR #8073, opened 2023-06-11
Scrum upgrade

HamishBrownPFR:
Working on #7869

This still requires some work and issues #8071 and  #8023 to be resolved.  

However @hol353 I am keen for some review on how I have dealt with weather data.  The Upgraded Scrum model either calculates its Tt parameters from user provided establishment and harvest dates, or calculates its harvest date from user provided establishment date and Tt duration.  It is a fundamental requirement for the new version of scrum to do this.

To facilitate this I have added [Models/Climate/GetTempSum.cs](https://github.com/APSIMInitiative/ApsimX/compare/master...PlantandFoodResearch:ApsimX:SCRUMupgrade?expand=1#diff-085766bb605899dab0d67f5c9e0041a799b4230442afc4a35169317fe7d37ef2) which is called by [Models/PMF/SCRUM/ScrumCrop.cs](https://github.com/APSIMInitiative/ApsimX/compare/master...PlantandFoodResearch:ApsimX:SCRUMupgrade?expand=1#diff-7938896a719b97bf76085ba3e6dab9f329b0895dcf11fa62a1fed00d4854317f)
I am keen for your review on this as I suspect there is a better way.  I have had to make another class as calling direct into the Weather.cs with future dates causes problems.  

I wonder if we could make weather data access more flexible in simulations by reading everything between Start and End into objects at the start of the simulation.  
for example we could change Weather propeties from doubles (eg. public double MaxT {get; set;}) to dicts (Dictionary<DateTime, double> MaxT) which then need to be referenced with a relevent date (MaxT[Clock.Today])
I have set up dates like this in another model I have written and it works OK.  Not sure what the run time consequences would be though.  May be a saving for SCRUM as don't have to read the weather twice.  Are there any other applicaitons that could benefit from being able to access future weather data?


HamishBrownPFR:
@hol353 There are some changes in here I will put through in separate pull requests so don't merge this please.
To do list for me
[Models/Functions/DeltaFunction.cs](https://github.com/APSIMInitiative/ApsimX/pull/8073/files#diff-9dc92eb1872627a022ccce96750fbd976df5b54e80104703f552d55edf78aec7)
[Models/Functions/DemandFunctions/PartitionFractionDemandFunction.cs](https://github.com/APSIMInitiative/ApsimX/pull/8073/files#diff-6cf3bf77cc26c262b605c48034f02ac8add4bd919519393e01e23fbc5b03667d)
[Models/PMF/PlantTypes.cs](https://github.com/APSIMInitiative/ApsimX/pull/8073/files#diff-294af5d41ed412f17c840394d6c63645e129364de815ede824f43c0ca39d3237)

hol353:
It looks like this method in Weather almost does what you want:

```c#
private DailyMetDataFromFile GetMetData(DateTime date)
```

It even has code to seek to to the correct position in the file, however it only does it once at initialisation to find the start date of simulation. I wonder if this seeking mechanism could be tweaked to work at any time during the simulation to find the correct date.

Regardless of whether my suggestion works (or is easy), I think we need to change the Weather class to support getting future weather data. This will mean you won't need your GetTempSum class which seems to duplicate a lot of code from Weather. We can probably do this for you next week (we're busy this week on non APSIM stuff). In the meantime, you can keep doing it the way you are for now. I suspect you can also remove the other bits of GetTempSum e.g. GetTtSum and GetHarvestDate back into SCRUM crop as they seem to be SCRUM things rather than weather things.

I won't merge this PR.

HamishBrownPFR:
Thanks @hol353.  I spent quite alot of time trying to use GetMetData() to do this job.  I forget the exact problem now but I has some property that is incremented each time it is called so if it is called more than once per day it throws errors.  This could probably be over come but I needed to keep moving so duplicated code into  getTTSum.cs as a interim solution.  If future weather data is available in an object I will be easy to bring this functionality back into SCRUM where it belongs.
Happy to help out with implementing/testing future weather functionality.

Hamish

HamishBrownPFR:
@hol353 I think it will be OK to merge this one now if you are happy with it.  Noting that some of the original content of this pull request was closed off in seperate issues, #8078, #8077, #8074, #8072 and #8023 and that I have opened a separate issue for dealing with future climate #8085 which will enable the deprocation of GetTempSum.cs once that work is completed.
I am keen to review this further with you  and @hol353 as well as my team at PFR but keen to get this merged in as a progress toward a more generic simple crop model.
We also need to have a discussion about how this is deployed: as a replacement to the existing SCRUM or as an alternative model.  I haven't gone through what would be involved in converting exisint SCRUM simulations but suspect it would be tricky.

PR #8075, opened 2023-06-11
Add error message to residue so it throws if attempting to set invalid fractions

### LINKED ISSUE #8074: Add error if invalid remove or return fractions set in residue
### What happened?

RemovalFractions is passed into BiomassRemove() in plant to specify the proporitons of biomass removed (from the system) and returned (to SOM).  They should only ever be between 0 and 1 but currently can be set at any value.  Need error message

Windows

HamishBrownPFR:
resolves #8074.

Will need to fix error in test set if any throw

### DIFF
--- a/Models/PMF/PlantTypes.cs
+++ b/Models/PMF/PlantTypes.cs
@@ -169,23 +169,30 @@ namespace Models.PMF
         ///</summary>
         public void SetFractionToRemove(string organName, double fraction, string biomassType = "live")
         {
-            if (removalValues.ContainsKey(organName))
+            if ((fraction > 1.0) || (fraction < 0.0))
             {
-                if (biomassType.ToLower() == "live")
-                    removalValues[organName].FractionLiveToRemove = fraction;
-                else if (biomassType.ToLower() == "dead")
-                    removalValues[organName].FractionDeadToRemove = fraction;
-                else
-                    throw new Exception("Type of biomass to remove should be either \"live\" or \"dead\"");
+                throw new Exception("Fraction removed from " +organName+" must be between zero and one");
             }
             else
             {
-                if (biomassType.ToLower() == "live")
-                    removalValues.Add(organName, new OrganBiomassRemovalType() { FractionLiveToRemove = fraction });
-                else if (biomassType.ToLower() == "dead")
-                    removalValues.Add(organName, new OrganBiomassRemovalType() { FractionDeadToRemove = fraction });
+                if (removalValues.ContainsKey(organName))
+                {
+                    if (biomassType.ToLower() == "live")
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@sme016.  I have added an error check into the plant biomass removal (see https://github.com/APSIMInitiative/ApsimX/issues/8074.) to stop inadvertently passing in removal fractions greater than 1 or less than zero.  
This causes the Eucalyptus test set to throw a number of error because there are some simulations that have negative proportions passed into thinning opperations (see screen shot below for eg.).  I 

image

The code currently treats negative values as zeros but so I will change the negatives in the test set to zeros to ensure the current behaviour is maintainted but the new error is not triggered.  
Its not clear why there would be negative thinning fractions in these tests but if the change I have made here is not acceptable please let me know

HamishBrownPFR:
@sme016 there was also one in the Pinus test which I have also set to zero

image

HamishBrownPFR:
@hol353 minor stats change because negative thinning props had been fed to biomasRemove.  This PR adds an error check to stop passing unfeasible numbers in for removal fractions but I have had to make some negatives zero to stop the error throwing for eucalypt and pine.  I am happy for this to be merged and have raised it with @sme016 to get his input first.

HamishBrownPFR:
@hut104 I am still waiting on comment from @sme016.  Do you have any suggestions on this one.  There were some thinning events that were being passed negative values.  The code sets negatives for biomass removal but the negatives were increasing stem population.  The net effects on the stats is very small but I am reluctant to suggest merging this without comment from someone who is clearer about what the intent was with the negative removal fraction in the thinning events

HamishBrownPFR:
Thanks @sme016  and apologies.  I was looking for your reply in the wrong place.  
Your explanation of the use of negative thinning proportions makes sense.  I the case of the model it would have been adjusting stem number up to account for multistemming (i.e plant population set allows for increases) but not adjusting biomass components (Biomass removals only allows positive removals).
As a way forward.  I will make a more sophisticated change to your TreeThinningScripts so negatives are not passed into the biomass removals but are still passed into the SetThinningProporiton and population set.  This should give the same results as currently.  I will also raise and Issue and assign to you to deal with the fact that thinning is not currently affecting biomass where population increase occurs and is specified by a negative thinning.
Sound OK?

sme016:
That sounds fine. So please go ahead. 
Is it also possible then to add the TreeThinning tool to the Management toolbox?
I'll then consider my part where and if it might be necessary. 
Thanks

HamishBrownPFR:
Thanks @sme016.  I will set up an issued to create a thinning script in the manager toolbox.

HamishBrownPFR:
@hol353 this one should be good to merge now

PR #8097, opened 2023-06-21
8085 Getting Future Weather

### LINKED ISSUE #8085: Future weather
### Describe the new feature

#8073 needs to access future weather to derive coefficients for each sowing of SCRUM.  It has implemented some bad code to do this but there will be a need for other models to access future weather so would be useful to implement a tidy and efficient way of doing this.

Currently weather properties (e.g MinT) are read in each day of the simulation time step and trying to call GetMetData(Date) causes errors if called more than once per day.

Maybe change each property to a DateTime indexed dictionary: 
Dictionary<DateTime, double> MinT
Which are established on start of simulation:
Current references to MinT would be replaced with MinT[Today] and future (or past) references would a be MinT[DateOfInterest]

par456:
Resolves #8085

There are now two caches when reading weather data. The first is on the weather file which marks the date and file position of lines it reads. If you need to seek to a new position it now tries to find the closest position it has been before and works from there.

The weather also keeps a cache of every met data record it has already read. This means it doesn't need to re-read the file when looking at previous data, or if it's fetched future data and has then gotten to that future date during the same simulation.

Yesterday and Tomorrow's data is now grabbed when they are needed instead of being properties that are updated every day that is simulated. I may need to change that back for performance.

Based on the tests, it shows some values have changed. This is most likely because previously yesterday's weather was always set to today's weather for the first day of the simulation. With this new way, it will now actually get yesterday's weather if it exists. I think that is what is causing the minor variance, but will need to double check that.

Based on the speed it ran on the previous test, I may need to continue to work on this to improve the speed of this solution.

### DIFF
--- a/APSIM.Shared/Utilities/ApsimTextFile.cs
+++ b/APSIM.Shared/Utilities/ApsimTextFile.cs
@@ -48,6 +48,34 @@ namespace APSIM.Shared.Utilities
     }
 
     /// <summary>
+    /// A simple type for encapsulating a text entry
+    /// </summary>
+    public class ApsimTextCache
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApsimConstant"/> class.
+        /// </summary>
+        /// <param name="date">The datetime of this entry.</param>
+        /// <param name="val">The value.</param>
+        /// <param name="pos">The file position of this entry.</param>
+        public ApsimTextCache(DateTime date, string val, long pos)
+        {
+            Date = date;
+            Value = val;
+            Position = pos;
+        }
+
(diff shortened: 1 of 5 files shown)


par456:
Speed problems have been solved and the new solution now runs just as fast the original code. However now Weather.GetMetData() can be used to fetch weather from any date without causing consecutive date errors.

This will still run red due to the Stats, as previously yesterday's weather was always set to today's weather at the start of the simulation. It is now correctly set to yesterday's weather, unless the start date is the same as the first date in the weather, in that case it will use today's weather like before.

par456:
retest this please jenkins

PR #8102, opened 2023-06-23
Date utilities refactor

### LINKED ISSUE #8101: Date Utilities keeps giving errors on edge cases
### Describe the new feature

This plan is to refactor the Date Utilities class to make it easier to use and to do way more error checking.
----
- [x] Put private functions at the bottom and structure functions together
- [ ] GetDate variations: (rename parseDate and deprecate old?)
    - [x] string - top level that can take in any recognised format
    - [x] string, string - any format with a formatting string
    - [x] string, int - day-mon or mon-day string with year as number
    - [x] int, int - day of year and year (met data values)
    - [x] int, int, int - year, month, day
- ParseDate is public, then each format is a private function
----
- [x] Deprecate Unused Functions
    - [ ] Julian dates (why do we need them?)
    - [ ] DMYtoDate
    - [ ] validateDateString (and replaced)
----
- [ ] Rewrite DMYtoISO to use the rest of the class
- [ ] rewrite DatesAreEqual to use new functions
- [ ] Unit tests on everything new (see whiteboard)

Document places that dates are being stored as string to convert them to storing as datetime if possible

ric394:
Resolves #8101

Big refactor of Date Utility class. Removing duplicate code and deprecating functions that should be replaced. Will be a longer branch as many changes still need to be made.

Current Plan:
- [x] Write new parsing functions
- [x] Link old functions to new parsers
- [x] Edit apsim classes to stop using the old functions
- [x] Write large unit test to prove new parser is working.

Using a deprecation system on the old functions so that manager scripts or such that use them will still work, but aim is to completely replace the old system.

### DIFF
--- a/APSIM.Shared/Utilities/DateUtilities.cs
+++ b/APSIM.Shared/Utilities/DateUtilities.cs
@@ -24,7 +24,7 @@ namespace APSIM.Shared.Utilities
             rxDMY = new Regex(@"^(\d{1,2})[-/](\d{1,2})[-/](\d{4})$"),
             rxddMMM = new Regex(@"^(([0-9])|([0-2][0-9])|([3][0-1]))\-(Jan|jan|Feb|feb|Mar|mar|Apr|apr|May|may|Jun|jun|Jul|jul|Aug|aug|Sep|sep|Oct|oct|Nov|nov|Dec|dec)");
 
-        private const int DEFAULT_YEAR = 1900;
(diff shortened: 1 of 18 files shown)


par456:
Refactor is now building and validation, but it is both running slower than expected and changing stats, so that's the next job here.

par456:
Completed and ready to merge.

PR #8117, opened 2023-07-05
Remove biomass in perennial leaf wasn't updating live and dead pools

### LINKED ISSUE #8051: CourseRoot organ in Eucalyptus and Pinus shouldn't be aboveground
### What is your question?

The CourseRoot organ in the Pinus and Eucalyptus models is currently aboveground. This was probably not an issue for the validation work orginally, as those datasets were all or mostly only for one rotation. (I need to check if any went for more than one rotation.)  

Now I am doing multi-rotation simulations, courseroot biomass, which can be considerable, is being added to the surface at harvest, which might upset to some degree soil evaporation, erosion, energy balance, decomposition and N availability. Overall, I suspect it will only be having a minor effect on tree growth and wood yield, but we should change it to better reflect reality anyway.

Overall allocation to courseroots should not change in the models, I suspect, but it needs to be distributed amongst soil layers. I suspect copying fineroot proportional allocations to depths would be a good enough approximation. As courseroots have minimal or no sensence, they should accumulate in sizable biomass mostly in surface layers, which could be in contrast to fine roots that have a short lifespan. There are few data in the validation datasets on this, but some.

In changing this organ to below-ground, should I detick the 'aboveground box' of the organ? Then what? Do I need to add the RootShape model to that organ?

If you want me to have a go at this, please set me up with an editable version of the Pinus and Eucalyptus models, which so far I have not had success in doing by working with them in Replacements.

hol353:
Working on #8051

### DIFF
--- a/Models/PMF/Organs/Cohorts.cs
+++ b/Models/PMF/Organs/Cohorts.cs
@@ -125,23 +125,6 @@ namespace Models.PMF.Organs
         }
 
         /// <summary>
-        /// Set the total live biomass of the cohort.
-        /// </summary>
-        public void SetLive(Biomass b)
-        {
-            live = b;
-        }
-
-        /// <summary>
-        /// Set the total dead biomass of the cohort.
-        /// </summary>
-        public void SetDead(Biomass b)
-        {
-            dead = b;
-        }
-
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


hol353:
retest this please jenkins

Issue #6772, opened 2021-09-06
Improve job runner performance


hol430:
The job runner used by the storage writer thread currently spends a non-trivial amount of time running empty jobs and sleep jobs. This should be refactored so that this job runner ends/aborts when there is no more work to be done, rather than doing busy waits until work is available.

PR #8119, opened 2023-07-05
8104 StorageWriter Busy Waiting Fix

### LINKED ISSUE #8104: 100% CPU usage related to EmptyJob
### What happened?

ApsimX uses 100% CPU after running a simulation. #6772 (from Sept 2021) already mentions that "The job runner ... spends a non-trivial amount of time running empty jobs and sleep jobs. This should be refactored ...". However, I think 100% CPU usage when the program is supposed to be idle deserves more attention.

It happens on master @ 433d5904f7632c818abe49aa045caaa1417c6343, and on older commits such as 6976310bc69a40078a970388bba4dd88bfc8625e (from right before [#6754](https://www.github.com/APSIMInitiative/ApsimX/pull/6754)) and 675c950e074003bfbedd5e240d5ec809c36c1593 (from Aug 2021).

### Workaround

Insert `Thread.Sleep(20);` in `EmptyJob.Run()`. This reduces the CPU usage to ~0%. I'm not sure whether it has other undesirable effects. \[*Edited.* I originally wrote `Thread.Sleep(1000)`, i.e. 1 second, but that adds a delay at the end of the simulation, which can become noticeable when running several simulations. I reduced the sleep to 20ms.\]

### Steps to reproduce

Commands (`sh` syntax):

```sh
# Add logging to EmptyJob.Run().
git apply <<"EOF"
diff --git a/Models/Core/Run/EmptyJob.cs b/Models/Core/Run/EmptyJob.cs
index 9b731d7..e62ae21 100644
--- a/Models/Core/Run/EmptyJob.cs
+++ b/Models/Core/Run/EmptyJob.cs
@@ -15,9 +15,12 @@ public void Prepare()
         }

+        private static int count = 0;
         /// <summary>Called to start the job. Can throw on error.</summary>
         /// <param name="cancelToken">Is cancellation pending?</param>
         public void Run(System.Threading.CancellationTokenSource cancelToken)
         {
-            //do nothing
+            count++;
+            if ((count & -count) == count) // Only print powers of two to reduce noise.
+                Console.WriteLine("EmptyJob.Run(): count=" + count);
         }

EOF

# Overwrite APSIM.Cli/Program.cs with a demo program.
mv -n APSIM.Cli/Program.cs APSIM.Cli/Program.cs.orig
cat > APSIM.Cli/Program.cs <<"EOF"
using Models;
using Models.Core;
using Models.Core.Run;
using Models.Storage;
using System;
using System.Threading;

public class Program
{
    public static void Main()
    {
        Simulation sim = new Simulation
        {
            Children =
            {
                new Clock
                {
                    StartDate = new DateTime(1980, 1, 1),
                    EndDate = new DateTime(1980, 1, 2)
                },
                new Summary(),
                new DataStore(),
            }
        };
        Simulations sims = Simulations.Create(new Simulation[] { sim });
        Runner runner = new Runner(sims);
        Console.WriteLine("Running the simulation...");
        List<Exception> errors = runner.Run();
        if (errors != null && errors.Count > 0)
        {
            throw new AggregateException("Simulation ran with errors", errors);
        }
        Console.WriteLine("Finished running the simulation");

        Console.WriteLine("Sleeping...");
        Thread.Sleep(10*1000);
        Console.WriteLine("Finished sleeping");
    }
}
EOF

# Build.
dotnet build

# Run.
dotnet run --project=APSIM.Cli --framework=net6.0
```

Output:

```
Running the simulation...
EmptyJob.Run(): count=1
EmptyJob.Run(): count=2
Finished running the simulation
Sleeping...
EmptyJob.Run(): count=4
EmptyJob.Run(): count=8
EmptyJob.Run(): count=16
EmptyJob.Run(): count=32
EmptyJob.Run(): count=64
EmptyJob.Run(): count=128
EmptyJob.Run(): count=256
EmptyJob.Run(): count=512
EmptyJob.Run(): count=1024
EmptyJob.Run(): count=2048
EmptyJob.Run(): count=4096
EmptyJob.Run(): count=8192
EmptyJob.Run(): count=16384
EmptyJob.Run(): count=32768
EmptyJob.Run(): count=65536
EmptyJob.Run(): count=131072
EmptyJob.Run(): count=262144
EmptyJob.Run(): count=524288
EmptyJob.Run(): count=1048576
EmptyJob.Run(): count=2097152
EmptyJob.Run(): count=4194304
EmptyJob.Run(): count=8388608
Finished sleeping
```

CPU usage is 100% during the sleep. The output shows that `EmptyJob.Run()` is called millions of times.

Dockerfile for convenience:

```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:6.0
RUN apt-get update && apt-get install -y libsqlite3-dev nano procps
# Shallow clone of ApsimX. The commit is the latest on master at the time of writing.
RUN git init ApsimX && cd ApsimX && git remote add origin https://github.com/APSIMInitiative/ApsimX.git && git fetch origin --depth=1 433d5904f7632c818abe49aa045caaa1417c6343 && git checkout FETCH_HEAD
WORKDIR ApsimX
RUN dotnet restore
```

par456:
Resolves #8104 

StorageWriter previously used an EmptyJob to busy wait when it didn't have anything to write. An existing SleepJob class already existed, so that was added instead with a 10ms sleep delay.

No hit to performance that I could see, however as noted by Drew in #6772 this could be refactored in the future to actually solve the problem.

### DIFF
--- a/Models/Storage/DataStoreWriter.cs
+++ b/Models/Storage/DataStoreWriter.cs
@@ -24,7 +24,7 @@ namespace Models.Storage
         private List<IRunnable> commands = new List<IRunnable>();
 
         /// <summary>A sleep job to stop the job runner from exiting.</summary>
-        private IRunnable sleepJob = new EmptyJob();
+        private IRunnable sleepJob = new JobRunnerSleepJob(10);
 
         /// <summary>The runner used to run commands on a worker thread.</summary>
         private JobRunner commandRunner;


tom93:
Thanks!

PR #8121, opened 2023-07-05
In soil evap, added code to reset sumes1 and 2 when U changes 

### LINKED ISSUE #8112: SoilWat resetting U/Cona when switching summer/winter U-Cona
### What happened?

In absence of rainfall, APSIMX SoilWat resets U/Cona when it switches to summer U/Cona values. (The same may happen when it goes to winter U/Cona, but the example simulation had rainfall at that time, so not as noticeable.)

In the example simulation we see a sudden drop in soil water content in surface layers and accompanying spike in Es compared with parallel SWIM3 run (as part of an experiment with Model as treatment).
image
[ErrorUConaReset.zip](https://github.com/APSIMInitiative/ApsimX/files/11929057/ErrorUConaReset.zip)

Windows

hol353:
(e.g. summer / winter turnover). 

Resolves #8112

### DIFF
--- a/Models/Soils/WaterModel/EvaporationModel.cs
+++ b/Models/Soils/WaterModel/EvaporationModel.cs
@@ -75,6 +75,9 @@ namespace Models.WaterModel
         /// <summary>Is simulation in summer?</summary>
         private bool isInSummer;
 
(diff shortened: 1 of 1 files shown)


hol353:
The change in stats is expected due to small change in ES calculation

PR #8122, opened 2023-07-05
Fixed goto phase so that it doesn' throw when phenology is missing.

### LINKED ISSUE #8092: PhaseLookUpValue function throws an error in the UI because phenology is null
### What happened?

Building a PMF model in the UI, whenever you click on a PhaseLookUpValue Function it throws a null reference error because phenology is null.  This is disruptive because it means you can't set the start and end properties on the PhaseLookUpValue through the UI.

Example open Prototypes/Orchard/Orchard.apsimx.  Any PhaseLookUpValue Function will do it but here is a specific example

image

Windows

hol353:
Resolves #8092

### DIFF
--- a/Models/PMF/Phenology/Phases/GotoPhase.cs
+++ b/Models/PMF/Phenology/Phases/GotoPhase.cs
@@ -34,6 +34,8 @@ namespace Models.PMF.Phen
         {
             get
             {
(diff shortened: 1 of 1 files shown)


hol353:
retest this please jenkins

Issue #7726, opened 2023-01-16
Tidy up Camp mechanisums


HamishBrownPFR:
We have found a tidier way to formulate the CAMP code that makes to model easier to explain.  Also allows better fits with parameters.

PR #8145, opened 2023-07-16
Camp tidyup

HamishBrownPFR:
working on #7726


hut104:
@hol353 Please merge this.  The results are very similar for most simulation test sets.  In some cases, we have taken the opportunity to clean up some results to improve fit.  The NZ simulations are not currently as good.  @HamishBrownPFR will be spending time addressing these in the near future.

PR #8150, opened 2023-07-17
8133 Add an offset to the ResetOnFixedDate Script

### LINKED ISSUE #8133: Reset water
### What is your question?

I have 10 planting dates, is there a function to reset the water to one month before each sowing date automatically?

I know there is an option to _reset water to the sowing date_, and also to _reset water to a specific date_, but if the experiment is to test many sowing dates is it possible to reset to one month before the sowing date?

par456:
Resolves #8133

This will allow you to offset the reset date by an amount of days. Negative days will make the reset happen before the given days, positive days will make the reset happen that long after the given date.


hol353:
Can you add the same logic to the *Reset on Sowing* rule?

Issue #8171, opened 2023-07-19
Remove SoilNitrogen model - use Nutrient instead


hol353:
### Describe the new feature

- Need to remove all occurrences of SoilNitrogen from all .apsimx files.
- Remove commented converter (which never really worked).
- Delete source code.

PR #8172, opened 2023-07-20
Removed SoilNitrogen from .apsimx files in Prototypes and Simulation

hol353:
Working on #8171


hol353:
Changing SoilNitrogen to Nutrient caused small diffs in stats.
Removing 'SetSoilParams' from Barley caused bigger changes in stats. Barley is currently being recalibrated.

Issue #8189, opened 2023-07-27
Add class to set biomass removal options so this doesn't have to be done in manager script


HamishBrownPFR:
### Describe the new feature

The manager scripts to do biomass removal are messy and not intuitive.  Add a UI component for specifying biomass removal settings that is then called from a manager to make biomass removal happen

HamishBrownPFR:
Issue resolved with https://github.com/APSIMInitiative/ApsimX/pull/8190

PR #8190, opened 2023-07-27
Defoliator

HamishBrownPFR:
working on #8189


HamishBrownPFR:
@hol353 I have prototyped something to get Xiumei moving with the Lucerne model but keen for some review to see if this is the right way to go and if it can be made more general.  I have set up a test in SCRUM which throws a null reference error that I can not solve and will need some pointers on that.  It works OK in lucerne.

HamishBrownPFR:
@hol353 @par456 @ric394 @hut104 
I have done a review of this with Dean and tidied up some code implementation.  There is a working example in Tests/Validation/SCRUM "Simple test with Biomass Cut".
I have added two classes to this simulation feed biomass removal settings into the simulation.
1. BiomassRemovalSetting.cs which specifies the crop that the settings apply to and the type of biomass removal event and has children that specify the removal fractions for each organ and the (optional) phenology rewind stage
2. OrganBiomassRemoval.cs which has children that specify the fractions of biomass to remove.

We don't think this is the best way to set this up in the user interface.  Instead it would be better with a grid view with a row for each organ on the crop and a column for each of the biomass removal fractions (LiveToRemove, DeadToRemove, LiveToResidue, DeadToResidue).  I have attached a mock up Dean made below.  Implementing this exceeds the scope of my programming skills.  The other column that could be included in the table is the type of biomass removal event (Cutting, Grazing, Pruning, Harvesting).  At some stage we should probably have a think if we need all of these event types.  I suspect at least Pruning could be removed and cutting used instead but that is a job for later.  We need to think about how much we want to break things up.  We could have one large table that is populated with all Crop x Organ x RemovalType combinations in scope.  Or we could have a seperate table for each Crop x RemovalType or a seperate table for each Crop or a separate table for each removal type.  My feeling is a table for each crop with Organ x removal type would be best at this stage.

image

par456:
I've reworked the class a bit with input from Dean, just so it matches the same pattern as existing classes. It's also been renamed to BiomassRemovalFractions so that it's consistent with the existing BiomassRemoval class. Each row of the table is stored as a BiomassRemovalOfPlantOrganType object, which is basically just a class to hold all the variables it needs to store.

At the moment I've left the SetStage and event calls out of the class, we'll need to add those back in once we have a plan on how events could be hooked up to this. It's also lacking unit tests, so I'll probably go in and add those down the road once we have settled on the functionality of the class and interface a bit more.

Let me know if its working how you expect.

HamishBrownPFR:
Thanks @par456.  I like what you have done with the grid.  There is a problem becasuse currently when we call the BiomassRemovalFractions it steps through all biomass removal types but for a single biomass removal event we would only want to apply one removal type.  I think the solution here will be to remove Type from the table so that the table represents a specific type.  For example we would have a table that has the fractions to remove for cutting and if we wanted to do a grazing biomass removal in the same simulation we would include a second table with the removal fractions for that type. 
I will make some changes along these lines and push up for your review.

HamishBrownPFR:
@par456 @hol353 I have changed BiomassRemovalFraction so each instance is specific to a BiomassRemovalType and have added a property so the user can select which type.  The deals to the problem outlined above but the selector is not showing up in the view.  Doing it this way makes it easy to reactivate the events.  
I have moved the phenology stage set out of this class as it does not fit that well there.  I have put it back in the manager for now. As I write this I am wondering if we should write a class that that is specifically for triggering defoliation events (what the manager is currently doing).  It would do the call the BiomassRemovalFraction and phenology sets.  We could also move the BiomassRemovalType property to this class and pass it to BiomassRemovalFraction.  That we we could expand the table to include all types again but only remove for the type specified on a given call.  I can rough something up for critique if you think this is worth investigating

HamishBrownPFR:
@par456 @hol353.  I have added a class specifically for triggering biomass removal events.  This class holds settings for what type of removal will happen (cutting, grazing etc) and if phenology is to be set.  It then calls BiomassRemovalFractions passing the removal type so it can trigger the removal from organs. It also calls the phenology rewind.
This is now fully functional in terms of biomass removal from organs, event calls and phenology resets.  let me know if you see any issues

par456:
I think that is a better solution for handling the phenology and remove types, though I wonder if they should be a child of the BiomassRemovalFractions since they are so directly related to each other.

The date field is going to need a rework on the Events view, as we would want to allow dates without a year and support all the the same formatting as the rest of apsim. Not totally sure how to change it yet though, dates inputs are always hard.

Otherwise I think that's a good way to do this, we'll see what Dean thinks.

HamishBrownPFR:
@par456 @hol353 .  The date thing is not critical to this so no urgency.  Will be nice to fix one day.  Can any one tell me why this pull request if failing.  I can't work out what is causing the error message

par456:
Looks like Lucerne was set up with SoilNitrogen which got patched the other week into Nutrient. In the DiagnosticsVariable manager scripts you have this variable:

```c#
[Link(ByName = true)]
private Nutrient SoilNitrogen;
```

Which will need to be removed as SoilNitrogen doesn't exist in those sims anymore. There are about 30 of them, I just went through and did it myself to check, so I can push that change up if you'd like.

HamishBrownPFR:
@par456 Thanks.  I thought I had reverted that Lucerne test to master.  Will have another go

par456:
No worries. I just pushed up the fix for the date issue you were having as well, and it'll also handle dates without years now too, which matches other existing classes that we have. 

Figured we should deal with that sooner rather than later since those values are getting saved to file.

HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
@hol353 I fixed a bug in the date parser without realizing @par456 has also fixed it in #8206.  Please merge #8206 first then I will tidy up the conflicts here and request final review.

HamishBrownPFR:
@hol353 @par456 Can you guys please review and let me know if any further changes needed.  Please merge if not.

PR #8199, opened 2023-08-01
6706 Models using IFunction can have valid parent restrictions

par456:
Working on #6706

Looking through an old issue and I noticed that models using the IFunction interface would always appear in the Add Model menu, even if they had been given a ValidParent tag to restrict them. Rewrote the function that handles valid parent checking so that it respects that and it cleans up most of the random classes that always show up in the add model menu.

After that only 3 models still appeared that shouldn't: Soils.WaterModel.CNReductionForTillage, Soils.WaterModel.CNReductionForCover and PMF.Structure.HeightFunction.

CNReductionForTillage and CNReductionForCover were only used under a WaterBalance node, so I gave them that parent.

Not sure what to do with HeightFunction however, as it does not seem to be used in any of our examples or validation simulations. The last big edit to it seems to be in 2017, so is this an old leftover class?

### DIFF
--- a/ApsimNG/Presenters/AddModelPresenter.cs
+++ b/ApsimNG/Presenters/AddModelPresenter.cs
@@ -73,8 +73,21 @@
                 AddTreeNodeIfDoesntExist(modelThatCanBeAdded, rootNode);
 
             tree.Populate(rootNode);
-            if (models.Count() < 10)
+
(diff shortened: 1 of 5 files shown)


lie112:
One of those appreciated improvements to the UI functionality awaiting a fix, thanks.

PR #8215, opened 2023-08-08
Change SurfaceOM to decompose residues to a depth of soil

### LINKED ISSUE #8107: Residue breakdown when using SWIM
### What happened?

In a system that doesn't have a lot of surface mineral N, there is now a large difference in breakdown of residue under SWIM compared with SoilWat. This appears to be related to SWIM now only checking the first layer for mineral N available for residue decomposition (immobilisation N demand). While attached simulation still shows some breakdown, others don't show barely any at all.

image
[Gatton.zip](https://github.com/APSIMInitiative/ApsimX/files/11864150/Gatton.zip)

Windows

hol353:
rather than top layer.

Resolves #8107


hol353:
There are diffs in the stats because simulations have a layer thickness in the top layer < 100mm. In these situations, this PR will decompose residues in the 2nd layer as well (down to 100mm from the surface).

PR #8231, opened 2023-08-14
Surface om tests

### LINKED ISSUE #8225: Addition of standing crop residues
### Describe the new feature

SurfaceOrganicMatter provides the ability to simulate residues that include those lying flat on the ground surface, and those standing after crop harvest.  There is currently the ability to specify the faction of standing residues when initialising the SurfaceOrganicMatter model.

We need the ability to specify residue orientation when adding further residues.  Whilst the crop models will need to be enhanced to allow this, adding the capability to SurfaceOrganicMatter would allow manager scripts to be developed to further test this capability by researchers working in this space currently.

The addition of extra functionality in our crop models could be added at a later time given sufficient time for testing and dealing with possible changes to model behaviour if this change is broadly applied within the model.  It is likely that broad application of this functionality will change results significantly (based on work on the importance of residue orientation).

hut104:
resolves #8225


hut104:
@hol353 I've added an initial test set for SurfaceOM.  Datasets from the stubble project could ultimately be added here.  Three sensibility tests have been added.  The third one tests the Add method implemented recently to resolve this issue.

PR #8233, opened 2023-08-16
Defoliator

### LINKED ISSUE #8209: Some issues with BiomassRemovalFractions when crops are removed from simulation
### What happened?

I have been setting up some examples for the BiomassRemoval approach we have developed.  Needed to make a few changes so it can deal with multiple crops in the same simulation.  

The problem I encountered was copying a simulation with maize and wheat (which had maize and wheat specified in BiomassRemovalFractions) and then removing maize (I only needed wheat in the second example).  The view for the grid only shows up wheat which is expected but the BiomassRemovals list in BiomassRemovalFractions.cs has not been cleaned out so still contains maize.  An object ref error is then thrown when BiomassRemovalFractions steps through each member of BiomassRemovals and can not find the maize model (which is still in the list by not in the simulation).

To reproduce the error pull branch https://github.com/HamishBrownPFR/ApsimX/tree/Defoliator build and then try to run the "SendingDatesFromOpperations" simulation in C:\GitHubRepos\ApsimX\Examples\BiomassRemovalsFromPlant.apsimx you will see what I mean

Windows

### LINKED ISSUE #8210: Review BiomassRemovalTypes
### Describe the new feature

The BiomassRemovalFractions has a row for each BiomassRemovalType (cutting, grazing, harvesting, pruning) by each organ by each crop.  When there are multiple crops in a simulation the table becomes quite large.

Organs deal with biomass removal in a generic manner (they don't know what type of removal it is) but the types are used for broadcasting different events that other classes respond to.  We need to review what is using these events and how to see if this functionality can be generalized into a single event or put into an ICareAboutBiomassRemoval interface and actions achieved with methods calls via the interface instead.

HamishBrownPFR:
Resolves #8209
Resolves #8210


HamishBrownPFR:
@hol353, Please review and merge if you are happy.  @par456 thanks for your help with this.

Models/Management/BiomassRemovalEvents.cs
```diff
@@ -23,15 +20,111 @@ namespace Models.Management
...
+            /// <summary>Biomass is harvested</summary>
+            Harvesting,
+            /// <summary>Biomass is pruned</summary>
+            Pruning
+        }
+
+        /// <summary>Stores a row of Biomass Removal Fractions</summary>
+        [Serializable]
+        public class BiomassRemovalOfPlantOrganType
+        {
```
hol353, line 45: I don't really like classes within classes (except for internal, private classes). Separate this public class into a separate .cs file.

Models/Management/BiomassRemovalEvents.cs
```diff
@@ -23,15 +20,111 @@ namespace Models.Management
...
+            public double LiveToResidue { get { return Convert.ToDouble(LiveToResidueString); } }
+
+            /// <summary>Fraction of live biomass to remove from organ and send to residue pool (0-1) </summary>
+            [Units("g/m2")]
+            [JsonIgnore]
+            public double DeadToResidue { get { return Convert.ToDouble(DeadToResidueString); } }
+
+            /// <summary></summary>
+            public string LiveToRemoveString { get; set; }
+
```
hol353, line 82: Why do we have a string and double properties for the same thing?

Models/Management/BiomassRemovalEvents.cs
```diff
@@ -66,18 +176,219 @@ public DateTime[] RemovalDates
...
+                        try
+                        {
+                            string input = Convert.ToDouble(row["Live To Remove"]).ToString();
+                            liveToRemove = input;
+                        }
+                        catch
+                        {
+                            liveToRemove = row["Live To Remove"].ToString();
+                        }
+                    }
```
hol353, line 265: Lots of duplicate code that could be removed. e.g. create a helper method that converts a double to a string. Also don't rely on exceptions to do conversion. Use double.TryParse instead. Also, why not store doubles in the DataTable as well as this class so that no conversion is necessary?

Models/Management/BiomassRemovalEvents.cs
```diff
@@ -66,18 +176,219 @@ public DateTime[] RemovalDates
...
+                    try
+                    {
+                        if (!String.IsNullOrEmpty(removal.LiveToRemoveString))
+                            liveToRemoved = removal.LiveToRemove;
+                    }
+                    catch (Exception e)
+                    {
+                        throw new Exception($"Error parsing Biomass Removal Variable 'LiveToRemove'. Value {removal.LiveToRemoveString} could not be parsed to number.", e);
+                    }
+
```
hol353, line 338: When would the event be triggered. The exception message says that parsing is happening but that's not the case. remove try/catch? Could also simplify by having a helper method to check for null and return 0 if so. Then this code becomes something like this:
```c#
double liveToRemove = GetRemovalFraction(removal.LiveToRemoveString);
double deadToRemove = GetRemovalFraction(removal.DeadToRemoveString);
```

Models/Management/BiomassRemovalEvents.cs
```diff
@@ -66,18 +176,219 @@ public DateTime[] RemovalDates
...
+            if (PlantToRemoveFrom.ToString() == BiomassRemovalType.Pruning.ToString())
+                Pruning?.Invoke(this, new EventArgs());
+            if (PlantToRemoveFrom.ToString() == BiomassRemovalType.Harvesting.ToString())
+                Harvesting?.Invoke(this, new EventArgs());
+
             if (StageToSet >= 1.0)
             {
+                Phenology phenology = PlantToRemoveFrom.FindChild<Phenology>();
                 phenology?.SetToStage(StageToSet);
             }
```
hol353, line 393: Probably want an exception thrown if there is not phenology model but a StageToSet was supplied.

Models/Management/BiomassRemovalEvents.cs
par456: Move to own file

Models/Management/BiomassRemovalEvents.cs
par456: Strings removed, double are set up correctly now.

Models/Management/BiomassRemovalEvents.cs
par456: Parsing made simpler with TryParse

Models/Management/BiomassRemovalEvents.cs
par456: Useless exception handling removed.

Models/Management/BiomassRemovalEvents.cs
par456: Check added. StageToSet changed to a string and parsed here so that it can be left empty in the GUI without it showing up as NaN.

hol353:
Much improved. Well done.

PR #8251, opened 2023-08-27
MicroClimate can only be a child of Simulation

### LINKED ISSUE #8250: Ensure MicroClimate only runs at the top level of a simulation
### Describe the new feature

* Throw exception if not running at the top level of a simulation
* Ensure that MicroClimate can only be added, dropped or pasted at the top level of a simulation.

ric394:
Resolves #8250 
- [x] Remove zone ValidParent annotation
- [x] Add exception to make sure existing MicroClimate nodes are now known to be in incorrect position.


BrianCollinss:
This is not linked to the corresponding issue: https://github.com/APSIMInitiative/ApsimX/issues/8250

ric394:
retest this please jenkins

ric394:
This has been put on hold until more advice is gained for dealing with AGPRyegrass issues.

hol353:
The specific problem: in the Stock example (UsingTagsExample simulation) there is a divide by zero at PastureSpecies.cs line 177.
 microClimate.RadiationInterceptionOnGreen = 0 because it is returning the value from the feedlot rather than the value from paddock 1. This is a potential bug in MicroClimate. Need to talk to @hut104.

par456:
Considering how old this is, and the number of files it would have to change, I think we'd restart this PR on a new branch when we try to address this change.

PR #8257, opened 2023-08-29
Disable stylecop directive and unused parameter warnings

### LINKED ISSUE #8253: Remove Stylecop error/warnings until addressed
### What happened?

Just adding on from the issue we raised this morning in the meeting. Can we turn off the stylecop errors for unused parameters and using directives inside of a namespace, so they stop cluttering up the error list when building.

They should be added back in again later when that is worked on again, but while it's not, it would be good to exclude them so our error lists are useful again.

Windows

ric394:
resolves #8253
working on #1281

Warnings were cluttering up error window in VS. Will turn on warnings once StyleCop rules are being worked on in the future.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -170,7 +170,7 @@
 ; csharp_style_expression_bodied_local_functions = false:silent
 ; csharp_indent_labels = one_less_than_current
 csharp_indent_labels = one_less_than_current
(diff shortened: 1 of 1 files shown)


ric394:
retest this please jenkins

PR #8262, opened 2023-08-31
Mulitple changes to allow Firebird to continue to be used as a data s…

### LINKED ISSUE #2821: Improve Datastore database performance

zur003:
…tore.

Resolves #2821

I'm flagging this as "resolved" even though the original intent - to improve database performance - was never achieved (neither with this nor the earlier Firebird code). In fact, it remains substantially slower than SQLite. However, it does allow for the possibility of the datastore residing on a remote server rather than a local file. Note, however, that this capability has not been fully implemented, as it would require introducing a mechanism in the GUI and in Report for entering and storing the name of the server, path to the database on that server, and authentication information. This should be easy to do, if needed.

One might argue that Firebird support should be dropped entirely, as it is slower and no-one has been using it. However, retaining it means that (a) we can set up to use a client-server database model very quickly and (b) it helps keep us aware of dependencies and quirks in our use of SQLite, making it easier to add support for some other database engine should we so desire.

My goal with this set of changes was to reach the point where the Wheat validation suite could be run reliably when using Firebird as the datastore. That has been achieved (although it's disappointingly slow).

There are a few "quirks" I should mention:

1. The restriction in older version of Firebird of having column names (and other identifiers) limited to 31 characters meant that variables with names like Wheat.Phenology.CurrentStageName could not be used without truncation. I had previously implemented a work-around that used an additional table to map between "long" column names and "short" column names (which were effectively just column numbers). This was awkward and hard to maintain. The current version of Firebird (4.0) increases the limit on identifiers to 63 characters, which we are much less likely to exceed, so I've removed the old work-around. 
2. Some constraints arise in the ADO.NET provider for Firebird rather than from the Firebird engine itself. As an example, values returned from SELECT queries still suffer from a 31 character limitation, but only when Firebird is used in embedded (rather than client/server) mode (I've included a work-around for this problem). This is because the provider is not yet using the current updated API when in embedded mode. This should be corrected in the future.
3. The provider can handle multiple threads, but each thread must use a separate Connection. I've implemented this, and tried implement the use of multiple threads in DataStoreWriter. Curiously, this seemed to give performance advantages when in client/server mode (with _localhost_ providing the server), but poorer performance in embedded mode. 
4. When there are multiple simulations, there is often quite a long "Waiting for datastore to finish writing" delay. I've added a "counter" to the status message in the GUI to give the user some feedback on the progress. I also did some tinkering to be sure that if the user presses "Stop" during this period that it actually stops writing to the database.


hol353:
Thanks @zur003 for this PR. Is there any reason for the .xlsx file? If not, can you remove it?

zur003:
@hol353 - The .xslx file is there deliberately, for a rather odd reason. Some of the cells in the original version of the spreadsheet were empty, but not being treated as empty. This arose from an oddity in Excel; see, for example, [https://answers.microsoft.com/en-us/msoffice/forum/all/strange-empty-cell-in-excel/4ec25afd-bd50-43df-9bbb-d8ed11631092](https://answers.microsoft.com/en-us/msoffice/forum/all/strange-empty-cell-in-excel/4ec25afd-bd50-43df-9bbb-d8ed11631092) which describes the situation (but doesn't really explain it). These cells were causing difficulty when ExcelReader attempted to write the data table to a Firebird database, when trying to convert values in the problematic cells to doubles. I don't recall the exact details, but it was something along the lines of the value being an empty string instead of null. Nulls were handled correctly, but the empty string gives a conversion error.
Whatever was going on, the new version of the .xlsx file is identical to the original, except that the problematic "empty" cells of the original are now truly empty cells.

PR #8266, opened 2023-09-01
8198 Removal of GridView

### LINKED ISSUE #6904: Truncated column results in data display from Report
I have a column named "Resource" generated by one of my reports. The column contains string entries containing a full stop (e.g. Finance.BankAccount and these values show up when I inspect the database in Browser DB. When this table displays in the PagedDataPresenter the leading characters before the full stop are removed. The columns do not fill the screen so it doesn't seem to be a column size issue.  I have looked through the code but cannot find where any truncation happens as when I inspect the column data at various places I can see the full string. This does not happen for the Category column in the same table

image

### LINKED ISSUE #8198: DualGridView and IModelAsTable should be refactored
Related to pull request #8190 where I ran into this problem.

This view and helper interface are the best for displaying a table GUI for a class, however they are poorly named, as it can do both single and dual grid views. It is also only used with a TablePresenter, which is only used with DualGridView, which should be more obvious.

There is also a GridView and NewGridPresenter that are tied heavily into the soil classes, as they use TabularData not a DataTable to store the grid. This refactor should address this, either by renaming this view and presenter, or by reworking them to use the more generic DualGridView that we have.

My proposal is to do some refactoring on this view and presenter to add additional properties to the ModelAsTable interface and rename each to be consistent.

In the related pull request, I added an additional panel to the top of this view to show a description and instructions on what the class is and how it can be used in the GUI. This should be incorporated into the other classes that use this view and presenter.

Summary:

ModelAsTable:
- Rename this interface to be consistent with the view and presenter
- Add Description/Instructions property that classes can provide to be shown in the GUI
- Are there other properties about the table that classes should be able to set individually without breaking our model-view pattern?

DualGridView:
- Adding a section where a description can be shown at the top of the window (should be in related pull request)
- Automatically setting the height to match the amount of text.
- Renaming this back to our default grid view, will need to deal with the soil GridView to do this.
- This should use SheetWidget instead of directly using a grid

TablePresenter:
- Give this presenter the same name as the view so that it's clear they are used together
- Find out why Grid2 has intellisense hooked up but Grid1 does not

Work out if GridView and NewGridPresenter can be adapted to work with DualGridView and TablePresenter since they are both doing similar things, but NewGridPresenter is more specialised.

Any other comments on this that should be addressed?

par456:
Resolves #8198 
Resolves #6904

This is a refactors of apsim to finally remove GridView and all its parts and finish converting over to using Sheet Widget.

Refactor is complete. Will need some people to pull the branch and find problems however.


hol353:
I've pulled this branch and in general it works really well. Just found a few problems.

- [ ] click on datastore. In grid on right, select multiple cells, right click and select copy - get an exception and APSIM hard shutdown. Choosing 'select all' also crashes.
- [ ] in datastore view, crtl right key press moves the view but the selected cell is hidden. ctrl down is the same
- [ ] in physical, if I type '.4' (no leading 0) into the BD column it ignores the leading '.' character.

par456:
When I typed .4 into the BD column it entered as 0.400, same behaviour on Jullian's system. Might need to check that again and see if it's working now.

Issue #8227, opened 2023-08-11
Add report feature to find useful outputs quickly


hol353:
### Describe the new feature

# 1. Quick list for selecting variables
Currently we rely on properties in classes to sum values across layers or instances of other objects. An example is the MineralN output in nutrient (Nutrient has lots of other examples as well). MineralN is the sum of all solute values in kg/ha. It looks like this:

```c#
        public IReadOnlyList<double> MineralN
        {
            get
            {
                double[] values = new double[numLayers];
                double[] nh4 = NH4.kgha;
                double[] no3 = NO3.kgha;
                values = MathUtilities.Add(values, nh4);
                values = MathUtilities.Add(values, no3);
                if (Urea != null)
                {
                    double[] urea = Urea.kgha;
                    values = MathUtilities.Add(values, urea);
                }
                return values;
            }
        }
```

This only exists to make it easier to report mineralN. Nutrient shouldn't be doing this.

Report can do it like this
```
NO3.kgha + NH4.kgha + Urea.kgha as MineralN
```

This is difficult for users so can we have a way for Report to show useful equations (from a quick list) that you can drop onto a report file? Perhaps users should be able to add to their own quick list and even submit useful calculations back to a cloud based list.

Another example of a variable could be in crop rain or wheat yield in kg/ha.

# 2 More flexibility
There are some things that report can't do. For something like total soil organic carbon you need to iterate all OrganicPools and sum up the carbon (C) arrays. In C#, LINQ can do this on one line:

```c#
Nutrient.OrganicPools.Select(pool => pool.C).Sum();
```

We can't do this in report. I had a go at this in #7675 but ultimately gave up. Perhaps we need to rethink it.

par456:
To Do List:

- [ ] Panel on the right side of the variables and frequency Panels
- [ ] Show list of options in a list or table view in new panels (use MineralN at first)
- [ ] Drag/double click an option from the list to report and show code that makes that variable
- [ ] Store premade list of options as a resource in Apsim directory
- [ ] Right click on a line to add it to list (use regex to parse, line must have "as Name" on it)
- [ ] Store user created items in configuration file
- [ ] Add filter to panel to reduce number of options

Stretch Goals
- [ ] Hide code that's dragged in until shown
- [ ] Common place to share variables (cloud)
- [ ] Give options for crop names instead of using generic Plant

PR #8292, opened 2023-09-15
Add new report common event/variable issue template

ric394:
working on #8227 

Need template submitted before new report changes submitted as the final URL for the template is needed for a link.


hol353 (commented):
What about if they want to add several / lots of variables. Can they do that all in one Issue? They probably won't like creating separate issues if they want to add say 10 variables.

ric394:
@hol353 very good point, I'll modify it

PR #8293, opened 2023-09-17
New Report Functionality

### LINKED ISSUE #8227: Add report feature to find useful outputs quickly

ric394:
resolves #8283
resolves #8227

# Changes

 - Report has new changes to make locating report variables and events easier.
- Additional windows have been added to the report node to assist in this.
- Filtering of events and variables is now active while typing on a new line.
- Documentation has been added to explain and describe the reports features and usage.


hol353:
I've had a play with your branch. Some things I noticed:
* Dragging a variable causes an Exception and hard crash of GUI.
* The width of the 'common report variables' window isn't remembered
* Clicking on a variable overwrites whatever is in the current row. I think it should always append the variable to the bottom of the report variables window. That way you can quickly double click on multiple variable and they all get appended.
* After adding a variable (e.g. in crop rain), the number of items shown in the common items windows reduces.

ric394:
Suggestions have now been implemented.

* In response to the last comment:
This is intended behaviour. The common variable/event lists should update when a variable/event is added to the text editors. Because the selected lines text has changed the common report events/variables list view contents will change to reflect the new lines contents and filter on the text in that line.

PR #8295, opened 2023-09-20
Remove exception when less biomass grazed than requested - not needed

### LINKED ISSUE #8294: SimpleGrazing was generating an exception when there was no error
### What happened?

There was a condition in which an exception was generated when the user requested more biomass than was available. The code already applies minimum harvestable amounts so this exception is not needed.

Windows

sno036:
resolves #8294


sno036:
resolves #8295

PR #8305, opened 2023-09-25
Change SimpleTheta to avoid cubic approximations where not required.

### LINKED ISSUE #8304: Monotonicity in SWIM3 moisture characteristics
### What happened?

In sandy soils, if LL15 is very low (e.g. < 0.01) the spline interpolation can become non-monotonic.  Given that this part of the characteristic is approximating a linear relationship (as attempted in the way that the cubics are specified) perhaps use a linear calculation directly.

Windows

hut104:
resolves #8304


hut104:
@hol353 The results are different, but all differences are improvements in the statistics.
I think that we can accept the improvements and merge this in.

PR #8313, opened 2023-09-26
Report's vertical splitter position issue fix

### LINKED ISSUE #8309: New report VerticalSplitterPosition has odd interactions with tree view/presenter divider.
### What happened?

It seems like when the value of ReportView's VerticalSplitterPosition is low (far to the left of the presenter) for the new report, it is very difficult to resize because the divider between the simulation tree view and the right-hand presenter has priority.

Additionally, when resizing the simulation tree/right-hand presenter with a Report selected the reportVariableBox (top pane) and reportFrequencyBox (bottom pane) don't have their vertical splitters change position evenly. Clicking away and back re-aligns the vertical splitters to the reportFrequencyBox.

I encountered both of these behaviors on d76ee20.

Windows

ric394:
resolves #8309

# Issue

The vertical position starts off not set making the common report and events windows take up the entire report view.

# Fix

Added check if the position is not set yet, to make it a sufficient number to make all report views seen.


hol353 (approved):
This seems to fix the problem on my computer

PR #8320, opened 2023-09-28
Empty drop down box options for PredictedObserved Node

### LINKED ISSUE #8252: No way to deselect a variable in dropdown list
### What happened?

In the PredictedObserved post simulation component, there is no way to deselect an already selected variable from the dropdowns  - it needs a special entry "<none>" or somesuch. Code is [here](https://github.com/APSIMInitiative/ApsimX/blob/b65712a1746b80cdf40c94a51036cbe1a84110d0/Models/PostSimulationTools/PredictedObserved.cs#L52) - the Values= ... argument.

As well, a scrollbar for the dropdown list is not always shown (appears to be at random) in a longish list, eg wheat validation.

Linux

ric394:
resolves #8252 

predictedobs-dropdown-img

### DIFF
--- a/ApsimNG/Views/PropertyView.cs
+++ b/ApsimNG/Views/PropertyView.cs
@@ -1,14 +1,14 @@
 namespace UserInterface.Views
 {
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
     using APSIM.Shared.Utilities;
     using Classes;
     using EventArguments;
     using Gtk;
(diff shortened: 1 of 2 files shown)


lie112:
Will this empty space appear at the top of every dropdown?
Could we have a property attribute that allows us to specify if an empty (null) entry is available.
For example an enum dropdown should not have a blank in the list, or is that the developer's responsibility to check in validation, or will this only be available for dropdowns of nullable properties?

PR #8326, opened 2023-10-03
8324 Allow Manager scripts to link to models when used under a factor

### LINKED ISSUE #8324: TreeManagement at a factor level
### What happened?

image
The TreeManagement tool is not allowing choice of 'tree to manage' or 'cultivar' when placed a factor level in an experiment. Please  repair it.
See screenshot above
Simulation attached
[APSIM2.zip](https://github.com/APSIMInitiative/ApsimX/files/12779575/APSIM2.zip)

Windows

par456:
Resolves #8324 

A bit of a weird bug where the FindAllInScope method didn't work when the calling manager script was located under a factor in an experiment. 

To fix this, if the script didn't have a simulation as an ancestor, it would then look for an experiment instead and use that to find the simulation model. Then when returning all models in scope, it would return the models in scope under the factor first, before then also returning the models under the simulation so the script could link to models back in the simulation. This is a bit slower, but also fairly rare to do, and better than modifying the existing FindAll function in the Scope class.

### DIFF
--- a/Models/Core/Model.cs
+++ b/Models/Core/Model.cs
@@ -451,28 +451,9 @@ namespace Models.Core
         public IEnumerable<IModel> FindAllInScope()
         {
             Simulation sim = FindAncestor<Simulation>();
-            Experiment exp = null;
-            if (sim == null) //if model is under an experiment factor, look for the simulation under the experiment
-            {
-                exp = FindAncestor<Experiment>();
-                if (exp != null)
-                    sim = exp.FindChild<Simulation>();
-            }
-
             ScopingRules scope = sim?.Scope ?? new ScopingRules();
             foreach (IModel result in scope.FindAll(this))
                 yield return result;
-
-            //scope may not work for models under experiment (that need to link back to the actual sim)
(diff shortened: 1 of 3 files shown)


Models/Core/Model.cs
```diff
@@ -451,9 +451,28 @@ public IEnumerable<IModel> FindAllChildren()
...
             Simulation sim = FindAncestor<Simulation>();
+            Experiment exp = null;
+            if (sim == null) //if model is under an experiment factor, look for the simulation under the experiment
+            {
+                exp = FindAncestor<Experiment>();
+                if (exp != null)
+                    sim = exp.FindChild<Simulation>();
+            }
+
             ScopingRules scope = sim?.Scope ?? new ScopingRules();
```
hol353, line 462: I'd rather change the scope.FindAll method than Model.cs. The scope class is probably used elsewhere in APSIM. This edge-case fix should be in scope class so that all uses of the class get the fix.

par456: Yep, changes made.

PR #8343, opened 2023-10-11
Changed divide calls

### LINKED ISSUE #8342: MicroClimate does not do something useful when avoiding divide by zero
### What happened?

When canopy conductance goes to zero, or very near zero, a divide function tries to avoid overflow.

However, it returns a zero, not an infinite (ie high) number.  As a result, when the divide triggers, GC jumps unexpectedly.

https://github.com/APSIMInitiative/ApsimX/blob/36798c99979d5819ebc34accda0d055fc0b3fb82/Models/MicroClimate/MicroClimateZone.cs#L851C4-L851C4

Windows

hut104:
resolves #8342


par456:
Just checking this change, did you actually mean to use 1e6 (1,000,000) , or was that supposed to be a 1e-6 (0.000001) ?

PR #8344, opened 2023-10-12
Database column types now update correctly.

### LINKED ISSUE #6807: Database column types are not always updated correctly
When refreshing the database, any changes to table column types in the inputs (such as excel inputs) are not necessarily applied to the tables stored in the database. More precisely, the column type in the apsim database is only updated if the column does not already exist in that table in the database. So if the user changes the input column data type (e.g. in excel), then that change will only affect the apsim database if the user drops that column from the table in the apsim database before running the excel input.

The typical workaround for this is to delete the .db file.

ric394:
resolves #6807


ric394:
retest this please jenkins

ric394:
# Blocking issue
- There are ExcelInput nodes referencing sheets that are not named correctly or are missing
-  Or there are entire excel files missing causing a problem.
- These need to be corrected before this can be merged in, as this fix throws exceptions when these two conditions occur.

par456:
Considering this is related to the work I am doing with ExcelInput, I may fold these changes into my branch and then we can deal with updating observed files when that comes up to being merged.

ric394:
Some of the fixes will be incorporated into other PRs. Closing PR.

PR #8350, opened 2023-10-13
8203 bubble chart improvements

### LINKED ISSUE #8203: BubbleChartPresenter has unfinished parts
There are a lot of comments left in BubbleChartPresenter for work that should be done and known bugs. Some examples:

image
...
image
...
image

Just needs a general cleanup along with the bubblechartview so they provide the intended functionality. 

- Address all the todo and fixme comments
- Determine if any functionality from the view should be in the presenter
- Cleanup commented out code if it's no longer needed.

par456:
Resolves #8203

This view was only half implemented when it was created with lots of comments through it with things missing. This branch will address those problems and generally improve the interface so that people can actually use it.

To do list:
- [x] Rewrite View to use a glade file for structure, and functions to build the separate boxes that are displayed.
- [x] Place node at cursor when created
- [x] Make default color randomized
- [x] Resizeable script panes on arc selections
- [x] ~~Show script errors when they happen~~
- [x] Make Arcs easier to click on
- [x] Fix arcs position/curve when node is moved
- [x] Fix Intellisense positioning
- [x] Use Node Ids instead of names
- [x] Generate ids for nodes
- [x] Use property presenter for variables that can be changed
- [x] Allow right clicking on node to both select it and then open a context menu
- [x] Click and Drag to select multiple nodes/arcs at the same time
- [x] Fix the GetLine function
- [x] Hover Effects


par456:
Currently waiting on functionality added in #8529 to be merged in so I can use it here. I need to be able to store the cursor position in the code editors between the view refreshing.

Otherwise this branch is very close to being completed.

par456:
This is probably at a point now where it can use some testing by other folks. See what breaks in the interface and suggest some sections that we could put unit tests on to make sure it continues into the future.

ric394:
# Notes/Improvements

- [ ] The context menu will appear at any location when a node is selected. This should only appear when the user right-clicks that specific node. Also occurs for arcs.
- [ ] Dragging a node outside of the visible 'canvas' makes the node no longer viewable/selectable. Changing to another node in the ExplorerView and then reselecting RotationManager does not reset the nodes in the 'canvas'.
- [ ] The arc name field does not behaviour as expected. Changing the name in the field does not update the arc name in the canvas instantly, and does not always update the name. This also occurs for node names.
- [ ] Changing an arc's name and clicking away makes the 'name' field lose the entered text and does not update the arc's name. This also occurs when changing node names.
- [ ] Changing a node's description field and clicking away makes the field lose the entered text.
- [ ] Entering text into a node's description field and then pressing enter does not update the description. 
- [ ] Entering text into a node's description field without any other action does not retain the new description text
- [ ] Selecting a new node colour from the drop down does not take effect until the user moves the mouse onto the 'canvas'.

par456:
Presenter and GUI are in a good state now. Need to add unit tests to the parts that can use them.

Issue #8388, opened 2023-10-23
EvaporationModel has some trouble with the northern hemisphere.


ilhuber:
### What happened?

The Initialize method of EvaporationModel can sometimes be called before Clock.Today has been initialized, leading to a DateUtilities method trying to subtract a year from the earliest representable DateTime for cropping systems with a winter date later in the year than their summer date.

### What Operating System are you on?

Windows

ilhuber:
Looks like I may have broken reset with that pull request. Simulations that have a reset day not equal to their starting day could have incorrect values for a time.

PR #8390, opened 2023-10-23
Evaporation initialize date

ilhuber:
Working on #8388 

The previous PR relating to this issue (#8389) may have introduced an error where simulations using the reset feature could run with incorrect values for a time.


ilhuber:
I have a better idea for this.

PR #8392, opened 2023-10-24
Attempt to fix reset error with evaporation model.

ilhuber:
Working on #8388 

My previous fix attempt (#8389) introduced a bug where EvaporationModel would run with wrong values after a reset under certain conditions. The other PR trying to fix my error (#8390) revealed a few tests are running without initialized Clocks, but that is another issue.


ilhuber:
Should I open a fresh issue relating to this? Sorry if I am not doing things quite right.

hol353:
I've merged this one. Normally one PR fixes one issue so you've done the right thing.

PR #8426, opened 2023-11-07
Change formulas to values in observed

### LINKED ISSUE #8425: Formulas in observed data aren't converted to numeric
### What happened?

The chickpea observed harvests contain a few formulas in the MaturityDAS column that aren't converted to numerics when graphing

Windows

peter-devoil:
resolves #8425


hol353:
This PR hides the problem in one observed spreadsheet. I would guess there are lots of formulas in lots of spreadsheets. APSIM is supposed to be able to deal with formulas in spreadsheets so better to fix the underlying defect. @par456  or @ric394 can you have a look at this to see what the problem is?

peter-devoil:
Abandoned via #8425

Issue #8360, opened 2023-10-17
Ability to suppress non-fatal error messages in Summary file


sno036:
### Describe the new feature

It would be quite useful if it were possible to optionally suppress messages to the Summary file. The DB file can get 'rather' bloated with these messages - totally out of proportion to the size of the actual output. This becomes more of an issue as large simulation jobs (grids across the country/world) are run. Would this be possible?

ric394:
@sno036 Are you looking for a more specific option for only fatal errors in the capture rules drop down box? or just a checkbox to ignore the writing of non-fatal errors when using any of the other capture rule options?
<img width="914" alt="image" src="https://github.com/APSIMInitiative/ApsimX/assets/123442863/6aee088a-2961-4a80-9cb4-70f0da674bb7">

PR #8428, opened 2023-11-07
Remove message and initial conditions filters from summary node

### LINKED ISSUE #8378: Summary node refactor

Summary.cs
SummaryPresenter.cs
SummaryView.cs

### What needs to be changed?

# Tasks
- [x] Remove initial conditions check box and treat as information type
- [x] Remove the filter box and make capture dropdown perform both functions.

To simplify the node. The thinking is to remove filtering and only store the 'level' of data chosen from the capture drop down box.

### LINKED ISSUE #8398: Summary Presenter is Difficult to use
### What happened?

1) The upper part of the presenter (ie settings) extends beyond the right hand side of the screen when working on a tablet and so cannot to used.
2) No vertical scroll bar is visible and so it is hard to navigate the summary log.
3) When using page down to navigate, the page scrolls right and so you cannot see the start of each line.

Windows

### LINKED ISSUE #8429: Summary UpdateView function refactor

SummaryPresenter

### What needs to be changed?

The UpdateView() method

It's very large and hard to read. Needs to be made much more modular for maintainability.

ric394:
resolves #8398
resolves #8378
resolves #8429

As a result of doing this, the summary node's view is much nicer on smaller screens and the controls are much tidier.

I was unable to replicate the page up and down issue mentioned in #8398.

image


ric394:
@hol353 This may also resolve #8360. Can you confirm if the capture setting does what @sno036 was asking?

hol353 (approved):

hol353:
Yes this also resolves #8360.

Issue #8175, opened 2023-07-20
Make orchard model more general


HamishBrownPFR:
### Describe the new feature

Need to progress set up of this model so it can be used to represent different tree crops

PR #8441, opened 2023-11-14
Orchard model

HamishBrownPFR:
working on #8175


HamishBrownPFR:
@hol353 There are a few trivial diffs in the performance which I think are OK to proceed with.  Can you please review this pull request and let me know if any changes are needed.

Models/MicroClimate/MicroClimate.cs
```diff
@@ -305,7 +305,7 @@ private void CalculateStripZoneShortWaveRadiation()
             foreach (MicroClimateCanopy c in tallest.Canopies)
             {
 
-                if ((c.Canopy.Height - c.Canopy.Depth) > 0 && c.Canopy.Width > Wt)
+                if (c.Canopy.CanopyType == "STRUM")
                     TallestIsTree = true;
```
hol353, line 309: When is CanopyType ever set to STRUM? I can't see it in this PR.

I also don't like having a flag for 'TallestIsTree' (or TallestIsVine)

```c#
            bool TallestIsTree = false;
            bool TallestIsVine = false;
            double Wt = (tallest.Zone as Zones.RectangularZone).Width * 1000;    // Width of tallest crop zone
            foreach (MicroClimateCanopy c in tallest.Canopies)
            {

                if ((c.Canopy.Height - c.Canopy.Depth) > 0 && c.Canopy.Width > Wt)
                    TallestIsTree = true;

                if ((c.Canopy.Height - c.Canopy.Depth) > 0 && c.Canopy.Width <= Wt)
                    TallestIsVine = true;
            }
            if (TallestIsTree)
            else if (TallestIsVine)
                DoVineStripShortWaveRadiation(ref tallest, ref shortest);
            else
                DoStripCropShortWaveRadiation(ref tallest, ref shortest);
```

Why do we have 3 different methods to do short wave radiation? Can they be combined? Can we detect somehow whether tree or vine is tallest? Microclimate needs a refactor at some point soon.

HamishBrownPFR: The code to set Canopy type is already in the source code so not in this PR.
Leaf.cs, line 101 
public string CanopyType { get { return parentPlant.PlantType; } }

There is scope for a tidy up of the way we do multiple canopy interception in microclimate.  The three methods were quick attempts to make progress.  Agree they need to be harmonized and tidied up.

HamishBrownPFR: @hol353, any further questions on this

Issue #7075, opened 2022-02-01
Simple pasture model


HamishBrownPFR:
Most people probably think this is not needed but I build it because it was useful for a job I was doing and I think it will be useful for others.

Justification:  Sometimes we just want an approximate pasture model to produce at specified levels and use water and nitrogen with a sensible seasonal pattern without having to go to the effort of getting a full blown pasture model and associated management set up.
The Simple Crop Resource Uptake Model (SCRUM) was written to fill this role for crops and now the Simple Pasture Resource Model (SPRUM) has been written to fill this role for pasture:

Key features:
Follows similar philosophy to SCRUM.  Daily growth rates (g DM /oCd) are specified.
Growth rates driven by temperature
Two phenological stages (Canopy expanding and Full Canopy) which have set thermal time durations.  When the end of the Full canopy phase is reached a "GrazeAndRewind" phase is entered which returns the phenogy to the start of the canopy expanding phase.  This give a pasture that will grow up to full cover and be grazed without any manager scripts required to do so.
N fixation varies between 0 - 0.03 g N / g photosynthesis based on a LegumePropn constant.

More details to come but I have to catch a bus now

lie112:
I'd be really interested in learning more about using this pasture (and crop) model as to date I have developed CLEM but not really used APSIM. I'd like to add grazing of SPRUM from the CLEM herd, and discuss the model limits of this pasture (e.g. suitability for tropical or temperate pastures, ability to allow basal area to change with grazing pressure, how to include species composition and handle annual vs perennial grasses) and whether it is possible to modify the soil profile by loss/gain of a macroinvertebrate community and associated macropores that increase infiltration 10+ fold. SPRUM could benefit from some of the functionality of GRASP to be quite a nice little pasture model.

HamishBrownPFR:
Hi @lie112.  I would be happy to discuss this with you at some stage to see what the potential is here.  SPRUM is an early prototype so good time to look at how else it might be used and what changes would be needed to achieve this.

PR #8450, opened 2023-11-20
Move SPRUM to similar implementation as STRUM

HamishBrownPFR:
# working on #7075

Removed Cultivars from SPRUM and added UI component to set pasture parameters specific for each simulation.  Renamed Product -> Residue and Stover -> Leaf and switched biomass parameters so the naming is more logical.  Implemented biomass removal with phenology generated grazings.


HamishBrownPFR:
@hol353 can you review please

PR #8458, opened 2023-11-21
8454 Moved editting check code to drag end and reset variables

### LINKED ISSUE #8454: Wrong Node editted in navigation tree
### What happened?

When making a copy of a node by clicking and dragging, then clicking on the node to rename it, the original gets renamed instead of the copy, despite the copy being the one selected.

This may be caused by the undo function as well, as it only happens some times.

Windows

par456:
Resolves #8454 

When dragging a node to duplicate it, it was possible for the name editing to start on the previous node. Then when you tried to rename the new node you just made, it would fire the renaming finished event on the original node instead.

This should fix it so that the node renaming is reset when dragging ends.

### DIFF
--- a/ApsimNG/Views/TreeView.cs
+++ b/ApsimNG/Views/TreeView.cs
@@ -827,8 +827,6 @@ namespace UserInterface.Views
         {
             try
             {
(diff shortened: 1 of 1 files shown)


par456:
retest this please jenkins

Issue #8493, opened 2023-12-11
Observed Data checks


par456:
A catch-all issue for the work on detecting and cleaning errors from observed data in the existing validation sets.

PR #8494, opened 2023-12-11
8493 Wheat Data Checks

par456:
Working on #8493

**Do not merge**


par456:
Don't need this for now. Deleting to clean up

PR #8500, opened 2023-12-12
Change default no3/nh4 value to use below sample depth

### LINKED ISSUE #8439: N Initialisation Q
### What is your question?

If I specify N distribution on a profile that is incomplete - eg only defined for surface layers in a soil of 1800mm:
image
 the initial N content (66 kg/ha) has some extra N from somewhere. Where could that be?

[no3.apsimx.zip](https://github.com/APSIMInitiative/ApsimX/files/13329716/no3.apsimx.zip)

hol353:
Resolves #8439

### DIFF
--- a/Models/Soils/Solute.cs
+++ b/Models/Soils/Solute.cs
@@ -237,13 +237,7 @@ namespace Models.Soils
         public void Standardise(double[] targetThickness)
         {
             // Define default ppm value to use below bottom layer of this solute if necessary.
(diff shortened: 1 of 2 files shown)


hol353:
The change in stats is acceptable given I changed the default no3/nh4 to zero. This value is used to infill missing values.

hol353:
I'll fix the failing unit test first before this can be merged.

PR #8501, opened 2023-12-13
8460 Graphs only show one demical place

### LINKED ISSUE #8460: Graph Tooltip has 15 significant figures
### What happened?

From Email.

Clicking on a point on a graph brings up a tooltip displaying the XY value of the point. The values have fully written numbers with up to 15 decimal places.

These should be reduced to just 1 decimal place.

Windows

par456:
Resolves #8460


hol353:
I wouldn't change the xy values that get plotted. There are lots of instances where we need more precision than one decimal place (soil water values, soil nitrogen values, anything SWIM related). Instead, can you just change the tool tip somehow?

par456:
Rewrote the code around the tooltips for box area and scatter graphs. I also abbreviated dates that had a time of 12:00am on them so only the date is shown.

PR #8502, opened 2023-12-14
Changes to stats when Lincoln9192 is removed from Wheat

par456:
Working on #8493

**Do Not Merge**

We've noticed that the data from the Lincoln9192 experiment is often outliers to the rest of the data. Checking what the stats do if we remove it from the validation.


par456:
Don't need this for now. Deleting to clean up

PR #8526, opened 2024-01-08
Maize and soybean models removed from all Replacements folders in FACTS

### LINKED ISSUE #8469: FACTS test set is not using standard release models
### What happened?

The test set in tests\validation\system\facts has both soybean and maize models in the experiments.  It is therefore not using either of the standard release models in the validation tests.
These simulations need to be changed to use the released version of the models.

Windows

ric394:
resolves #8469

Reason:
Release versions will be used instead of local file versions. This makes sure that the most recent version of a plant model is being used.


ric394:
The reports are reporting variables for the model in replacements. How do we approach this one @hut104?

hol353:
Try putting the model that is in Replacements into the resource files (and delete from Replacements) and see how much the stats change. If the stats are ok then we might do a diff to see what is different between the models.

PR #8529, opened 2024-01-09
8476 Manager Script Bugs

### LINKED ISSUE #8476: Manager editing buglets
### What happened?

\<ctrl>-T doesnt test the script compilation (for errors) any more

\<ctrl>-S (save) changes insertion to [0,0] and moves view to origin

Clicking on / off the component resets the view to origin (editing position isnt preserved)

Weird stuff happens when errors are present and the component is copied/duplicated (may be related to https://github.com/APSIMInitiative/ApsimX/pull/8458)

Linux

par456:
Resolves #8476 

Makes the script interface a bit easier to use as now your position is saved if you save the file with it open, and the tab is remembered when switching between nodes. Ctrl-T compile has also been updated so that it works all the time.

Only thing from that issue that was not addressed was having the script open to the same position when switching nodes. I had a go at doing that, but gtk wasn't playing nice with the scroll bars. I think just having it open to the correct tab is enough though.


hol353:
Ctrl-T seems to work for me on this branch. When I click away and come back it goes straight back to 'Script' tab (which is good) but it doesn't go back to the previous editing position (which is bad). Can you have another go at remembering the previous editing position?

par456:
For future reference on anyone looking at this PR.

You can only set the scrollbar position after the content on the page has been loaded, so trying to do it in the attach function won't work because gtk hasn't measured everything out yet. I ended up attaching to the Drawn event and setting the scrollbar position on the 3rd draw of the widget onto the screen, as that seems to be the first frame when everything has been loaded in. 

The OnPageSwitch event of the notebook widget that the ManagerView uses can also be used to set the scrollbar, so if the properties page comes up first, the drawn event handler is disabled and the scrollbar set when the page is changed.

PR #8532, opened 2024-01-09
8515 NaN Error messages on Graphs

### LINKED ISSUE #8515: Matching observed harvest data with data in daily report
### What is your question?

I have set up my observations this way:

image

When I use the PredictedObserved table to match the observations and predictions to generate X:Y scatter plots to display DaysAfterSowing, it works fine. But when I to the same for GrainWt, I keep getting this warning:

```
AboveGroundWt: X Axis values contain NaN values.
AboveGroundWt: X Axis values contain NaN values.
AboveGroundWt: X Axis values contain NaN values.
AboveGroundWt: X Axis values contain NaN values.
AboveGroundWt: 4 points are outside of the provided graph axis. Adjust the minimums and maximums for the axis, or clear them to have them autocalculate and show everything.
```

Should I assume that cells with empty values are the reason?

par456:
Resolves #8515

When a cell is fed through as null or empty, it gets converted to a NaN by the graphing code. This has to stay this way, but not all NaNs are bad so I've redesigned the messages it gives back.

Users are still informed of NaN values, but now they are told it could be caused by empty cells in the datastore, the errors are also broken up into points with only NaNs in the X, only NaNs in the Y or NaNs in both axis, with a count given to each.

That should hopefully make it more clear what is being reported and allow users to check their data to confirm it matches.

### DIFF
--- a/ApsimNG/Presenters/GraphPresenter.cs
+++ b/ApsimNG/Presenters/GraphPresenter.cs
@@ -208,18 +208,14 @@
                         }
                         else
                         {
-                            if (xMin != double.NaN)
-                                if (x < xMin)
-                                    isOutside = true;
-                            if (xMax != double.NaN)
-                                if (x > xMax)
-                                    isOutside = true;
-                            if (yMin != double.NaN)
-                                if (y < yMin)
-                                    isOutside = true;
-                            if (yMax != double.NaN)
-                                if (y > yMax)
-                                    isOutside = true;
+                            if (!double.IsNaN(xMin) && x < xMin)
+                                isOutside = true;
+                            if (!double.IsNaN(xMax) && x > xMax)
+                                isOutside = true;
+                            if (!double.IsNaN(yMin) && y < yMin)
+                                isOutside = true;
+                            if (!double.IsNaN(yMax) && y > yMax)
+                                isOutside = true;
 
                             if (isOutside)
                                 pointsOutsideAxis += 1;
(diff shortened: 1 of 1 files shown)


BrianCollinss:
Would it be possible to allow users to use a predefined value (e.g. None or NaN) to tell the graph the empty cells are intentionally empty?

par456:
I had a think about how that could be done when working on this, and it's a bit hard. An empty cell could be deliberately left blank (aka no data), or accidently blank (forgot to enter data), so that should be noted.

And any input into the cell could hide legitimate input, such as a measurement of "None". It also has the problem that a lot of this graph checking code is looking at the values after they've been converted to their graphed values, so an empty cell, or a NaN or something invalid all just comes out as NaN at that point of the graphing.

We are doing some work on the entire Predicted / Observed system at the moment, so I'll keep this problem in mind as we work on that, as it might be better solved with those changes instead of trying to do it in the graphing code.

ApsimNG/Presenters/GraphPresenter.cs
```diff
@@ -160,67 +160,83 @@ public void DrawGraph(IEnumerable<SeriesDefinition> definitions)
...
+                        } else if (double.IsNaN(x) && double.IsNaN(y))
+                        {
+                            bothNaNCount += 1;
                         }
                         else
                         {
+                            if (xMin != double.NaN)
+                                if (x < xMin)
+                                    isOutside = true;
+                            if (xMax != double.NaN)
```
hol353, line 214: Shouldn't test for NaN's using equality. Can this bit of code be simplified to:
```c#
if (!double.IsNaN(xMin) && x < xMin)
   isOutside = true;
```

par456: Whoops, I did that again, I keep forgetting about the IsNaN function. Fixed.

PR #8534, opened 2024-01-10
Orderby statements removed to allow default behaviour.

### LINKED ISSUE #8519: Row order in DataStore
### What happened?

When there are multiple factorial simulations. data in DataStore is messy. As can be seen in the following example, the data is presented for the 2016 experiment ordered by date, then it is cut on day 220 and data for the 2015 experiment starts. Should we be able to sort the data by clicking on the column names?

image

Windows

ric394:
resolves #8519 

The current orderby statements ordered rows by SimulationId then Clock.Today.
These have been removed. The default behaviour now causes rows to be ordered by simulation name then date.

### DIFF
--- a/ApsimNG/Presenters/DataStorePresenter.cs
+++ b/ApsimNG/Presenters/DataStorePresenter.cs
@@ -1,13 +1,13 @@
-﻿using UserInterface.EventArguments;
-using Models.Core;
-using Models.Factorial;
-using Models.Storage;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
-using UserInterface.Views;
 using APSIM.Shared.Documentation.Extensions;
+using Models.Core;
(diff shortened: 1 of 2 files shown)


par456:
This will need some discussion before any merge. It's unclear why the ORDER BY statements in the sql string are not working as intended, but when it is removed, the data is then returned in alphabetical order.

There is something not quite right, and I'm hesitant to remove code when I'm not sure why it wasn't working.

hol353:
We don't want to remove the OrderBy fields! That will impact the graphing of time series data.

The DataStoreReader has code to order by SimulationID and then Clock.Today. This is what we want. Given the screenshot in the issue shows this not being done when displaying the data in a table means the OrderBy code is being bypassed somehow. This needs further investigation.

par456:
Killing this PR so we can figure out what's going on and develop a better solution.

PR #8536, opened 2024-01-10
Added "Germination Temperature" to "GerminatingPhase"

### LINKED ISSUE #8535: Minimum germination temperature
### Describe the new feature

Minimum soil temperature for germination to occur.

BrianCollinss:
Resolves #8535


Models/PMF/Phenology/Phases/GerminatingPhase.cs
```diff
@@ -39,6 +40,12 @@ public class GerminatingPhase : Model, IPhase
         [Link]
         private IClock clock = null;
 
+        [Link(IsOptional = true)]
+        private ISoilTemperature soilTemperature = null;
```
hol353, line 44: There will always be a ISoilTemperature in every simulation. Doesn't need to be optional.

Models/PMF/Phenology/Phases/GerminatingPhase.cs
```diff
@@ -39,6 +40,12 @@ public class GerminatingPhase : Model, IPhase
...
+
+        [Link(Type = LinkType.Child, ByName = true, IsOptional = true)]
+        private IFunction temperature = null;
```
hol353, line 47: We don't like optional functions because they obscure behaviour. If a developer is looking at GerminatingPhase in a model in the GUI, they won't see that GerminatingPhase can have a child temperature function. If you make it explicit and require all GerminatingPhase functions have a temperature function (even if it is a constant function set to 1) then it is obvious.

Remove the IsOptional and then write a converter that inserts a temperature function (constant function set to 1) into every GerminingPhase in every .apsimx file.

Models/PMF/Phenology/Phases/GerminatingPhase.cs
```diff
@@ -74,6 +81,13 @@ public class GerminatingPhase : Model, IPhase
         [JsonIgnore]
         public string GerminationDate { get; set; }
 
+        /// <summary>
+        /// Date for germination to occur.  null by default so model is used
+        /// </summary>
+        [JsonIgnore]
+        [Units("oC")]
+        public double GerminationTemperature { get { return temperature == null ? -999 : temperature.Value(); } }
```
hol353, line 89: If the temperature function isn't optional then there is no need for this property.

Models/PMF/Phenology/Phases/GerminatingPhase.cs
```diff
@@ -91,7 +106,7 @@ public bool DoTimeStep(ref double propOfDayToUse)
                 }
             }
 
-            else if (!phenology.OnStartDayOf("Sowing") && waterBalance.SWmm[SowLayer] > soilPhysical.LL15mm[SowLayer])
+            else if (!phenology.OnStartDayOf("Sowing") && waterBalance.SWmm[SowLayer] > soilPhysical.LL15mm[SowLayer] && sowLayerTemperature >= GerminationTemperature)
             {
```
hol353, line 110: Refer to the temperature function directly rather than to the GerminationTemperature property.

hol353 (changes requested):

BrianCollinss:
@hol353 would you please advise whether I must upgrade all recourses and commit/push again?

hol353:
Yes you will need to upgrade all resources. There is a button on the home screen of the APSIM GUI that will do it for you. You can then commit the changed files and push to this PR.

hol353 (approved):

Issue #6863, opened 2021-10-15
Met files in Examples folder have got messy and there are errors


sno036:
There are met files in there that have errors and others that are not helpful (e.g. with coordinates for names). Suggestions:
- all files retained should be named as CC_LocationName with CC being the usual country code 
- all files retained should be tested with an end to end run of a simple simulation (SoilWater and no crops?) to test for errors
- met files to demonstrate different ways of doing met files should be in the Tutorial folder (and should be short)
- junk files should be deleted to reduce cruft

PR #8546, opened 2024-01-15
Renamed met files with country code and location name.

ric394:
working on #6863

Provides better clarity for included met files.


par456:
Just have to be careful on this one for other people that are using these files. Would it make sense to include converter if they have apsim files using the renamed example weather files?

hol353:
Yes, write a simple converter that changes the weather file name to match the new name. Should only be a small amount of code.

PR #8561, opened 2024-01-19
Playlist functionality added to command line

### LINKED ISSUE #8550: Models.exe runs all simulations despite a Playlist being present
### What happened?

[Test.zip](https://github.com/APSIMInitiative/ApsimX/files/13957724/Test.zip)

I have a Playlist in which 3 simulations are listed. When I run APSIM via the GUI, only these three are run. However, when I run it via the command line on HPC, it seems it tries to run all 9 simulations existing in the apsimx file. (Note: there is an issue with residue type so the model fails to run, but there are 9 errors instead of 3).

Does it mean Playlist only works in the GUI?

**EIDT: It runs all 9 simulations when I run it on my laptop (which does not have the issue with residue type).**

Windows

ric394:
resolves #8550

# Problem
Users were not able to specify a playlist to run in an apsimx file from the command line, but they could from ApsimNG.

# Fix
A user can now run models.exe with the ```--playlist``` or ```-p``` switch followed by the name of a playlist node they'd like run from an apsimx file.  

# Example CLI commands
```models.exe wheat.apsimx --playlist specialPlaylist```
```models.exe --apply commands.txt --playlist playlistName```


BrianCollinss:
Would it be not easer to run all enabled Playlists instead of needing to specify which one to run? Or if the -p flag is not present, all Playlists are run.

par456:
We discussed that type of functionality when the playlist was first implemented, and decided that it should be explicit instead of automatic. The GUI works in the same way, where to run the playlist, you have to click on the playlist to run it. This allows you to have multiple playlists in a file while being able to still run everything without disabling them.

We had a chat and believe the console flag should work the same, again, so you can have multiple playlists and be able to run the entire file without needing to edit it first.

BrianCollinss:
But currently if there's a Playlist in the gui and you run from the gui, it just runs the Playlists, not all simulations.

par456:
Just opened up a test file to check, and yeah, the playlist choices are only used in the GUI if you have the playlist open when you click run (either the button at the top or right click -> run).

If you've got a file that is running a playlist selection when you've got something else selected (like Simulations which should run everything, or node under a Simulation, which should only run that simulation), let me know because you've found a bug I need to fix.

BrianCollinss:
Is this going to be merged?

ric394:
> Is this going to be merged?

This pull request needs to be reviewed before being merged. @hol353 could you please review?

hol353:
I'm not in the office until Friday so will look at it then.

PR #8572, opened 2024-01-22
Replaced the defoliation manager in redclover example 

### LINKED ISSUE #8353: Defoliation manager in Red clover example is not working
### What happened?

In the red clover example for the latest release (2023.10.7317.0) the defoliation manager does not appear to defoliate/harvest the crop.  The work around is to replace it with a defoliation manager from the Plantain Forage example, and then modify the defoliation manager (as it is hard coded to plantain).  I don't particularly like the solution as I preferred the original defoliation manager in the red clover example as it was not hard coded to a particular crop/pasture.    

Windows

hol353:
I used the one in the redclover validation. They have the same user interface except now the user needs to specify fractions of biomass in each organ to remove. Previously these were defaults in each plant organ. It was decided to expose these to the user in the hope they would know how to set them.

Resolves #8353

### DIFF
--- a/Examples/RedClover.apsimx
+++ b/Examples/RedClover.apsimx
@@ -1,7 +1,6 @@
 {
   "$type": "Models.Core.Simulations, Models",
-  "ExplorerWidth": 273,
-  "Version": 163,
+  "Version": 171,
   "Name": "Simulations",
   "ResourceName": null,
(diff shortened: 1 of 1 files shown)


hol353:
retest this please jenkins

PR #8575, opened 2024-01-23
Phase drop downs

### LINKED ISSUE #8574: Phase Name drop downs for Crop phase properties
### Describe the new feature

A number of PFR classes have a property which is a specific phase name but the view to specify this is free form code.  We have a property selector that lists phases of current crop model as emums so should be using this.

HamishBrownPFR:
resolves #8574


HamishBrownPFR:
@hol353 @ric394 @par456 please review and merge if OK

PR #8579, opened 2024-01-23
weather filename path bug fixed with tests

### LINKED ISSUE #8571: The examples don't run out of the box on Linux or MacOS
### What happened?

On Linux and MacOS, the back slashes in the weather path ("%root%\\Examples\\WeatherFiles\\Dalby.met") need to be converted to forward slashes.

Linux

ric394:
resolves #8571

### DIFF
--- a/Tests/UnitTests/Weather/WeatherTests.cs
+++ b/Tests/UnitTests/Weather/WeatherTests.cs
@@ -191,7 +191,7 @@ namespace UnitTests.Weather
         /// Ensures all example .met files contain %root%
         /// </summary>
         [Test]
-        public void TestAllExampleFilesWeatherModelsHaveRootReferenceInFileName()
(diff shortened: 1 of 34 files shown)


Tests/UnitTests/Weather/WeatherTests.cs
```diff
@@ -186,6 +187,30 @@ public void TestGetAnyDayMetData()
             Assert.AreEqual(weather1900.Rain, weather.TomorrowsMetData.Rain, 0.01);
         }
 
+        /// <summary>
+        /// Ensures all example .met files contain %root%
+        /// </summary>
+        [Test]
+        public void TestAllExampleFilesWeatherModelsHaveRootReferenceInFileName()
```
hol353, line 194: Does this also check that the slashes are forward slashes? I think we need that.

PR #8580, opened 2024-01-23
8569 Script Manager Fixes

### LINKED ISSUE #8569: Script Parameters no longer keep changes.
### What happened?

It seems like with the current version (2024.01.7346.0) the Parameters tab of a manger script will no longer keep changes. Oddly, changes to the Parameters tab will be kept if the Script tab has been modified.

Windows

par456:
Resolves #8569 

Bugs have been fixed, testing what can be done with unit tests to stop this sort of bug happening again.


par456:
Have finished working on these fixes. Along with fixing the loading and scrolling bugs we had, I've also implemented a unit test that that individually runs each method in the Manager.cs class and uses reflection to ensure each method and public property has been tested in that unit test.

This should hopefully mean that if someone comes along and extends Manager.cs again, it'll alert them to add unit tests for new methods that they've added.

Will need a review if this is the best way to write this kind of test, but it means every part of Manager.cs has at least one unit test connected to it, and will help enforce more unit tests in the future.

PR #8582, opened 2024-01-24
Added a StartOfFirstDay event to clock

### LINKED ISSUE #8484: Ability to report after initialisation on the first day of the simulation
### Describe the new feature

Particularly useful for getting changes in pools. There is currently a "FinalInitialise" - unsure what this really is but soil arrays are reported with zeros when it is used.

hol353:
Resolves #8484


par456:
In the issue you mentioned that people shouldn't try to read values on the OnStartOfSimulation event. Could you add a comment about that to the summary on that event in the clock? That way it'll get captured in the documentation and might stop someone else accidently doing that.

hol353:
Good idea - done.

PR #8586, opened 2024-01-25
Fixed leaf detachment

### LINKED ISSUE #3559: Leaf Class Detachment output does not work
The leaf class has an output of type biomass to report daily detachment.  The value in this class is not updated with daily detachment.  It reports only manual detachment by external actions.  The detachment that it computes is never added to this value.

hol353:
Resolves #3559


hol353:
retest this please jenkins

ric394 (approved):

PR #8587, opened 2024-01-25
Change TreeProxy to do N uptake in SetActualNitrogenUptakes method

### LINKED ISSUE #3566: Nuptake in TreeProxy seems to be set incorrectly
Nuptake appears to be set when providing estimates to soil arbitrator rather than when the arbitrator provides the actual uptakes back to the tree proxy

hol353:
Resolves #3566


ric394 (approved):

hol353:
Thanks for approving - please merge when ready.

Issue #8564, opened 2024-01-20
Water node is empty when an XML soil is added to GUI


BrianCollinss:
### What happened?

[test.zip](https://github.com/APSIMInitiative/ApsimX/files/13996541/test.zip)

![image](https://github.com/APSIMInitiative/ApsimX/assets/39573622/944006b5-f208-4c20-9b08-c81a57cbe0c7)

### What Operating System are you on?

Windows

BrianCollinss:
@par456, for some reason, the **UpgradeToVersion152** does not do anything as its first line cannot find a Soil model! This is the converter that is supposed to do most of the job, including `water["InitialValues"] = sample["SW"];`.

BrianCollinss:
The root type in this method is "Models.Soils.Soil" so there are no Soil children to be found.

PR #8588, opened 2024-01-29
Fix solute depth

### LINKED ISSUE #8563: If solues are missing, GUI adds solues with wrong depths
### What happened?

If Solutes are not present in the soil, when it is copied under the Paddock node, all three solutes are automatically added, but the Depth column is not the same as the physical Depth column.

image

At least in SoilDownloadPresenter, these depths are meant to be the same as physical depths:

image

Windows

BrianCollinss:
Resolves #8563 and #8564

### DIFF
--- a/Models/Core/Apsim710File/Importer.cs
+++ b/Models/Core/Apsim710File/Importer.cs
@@ -445,7 +445,7 @@ namespace Models.Core.Apsim710File
             this.AddCompNode(destParent, "Nutrient", "Nutrient");
             XmlNode newNO3Node = this.AddCompNode(destParent, "Solute", "NO3");
             XmlNode newNH4Node = this.AddCompNode(destParent, "Solute", "NH4");
-            XmlNode newUREANode = this.AddCompNode(destParent, "Solute", "UREA");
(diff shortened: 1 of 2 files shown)


BrianCollinss:
Hi @hol353 . Could you please advise on this issue? The test that is failing is this (please note the summary):

```
        /// <summary>
        /// Arguably this doesn't even belong in the converter.
        /// Nonetheless, it's not working properly at the moment so
        /// I'm just going to fix the problem and add a test. See here:
        /// https://github.com/APSIMInitiative/ApsimX/issues/6270.
        /// </summary>
        [Test]
        public void TestAddingInitialWater()
        {
            string json = ReflectionUtilities.GetResourceAsString("UnitTests.Core.ApsimFile.Soil.json");
            ConverterReturnType result = Converter.DoConvert(json, Converter.LatestVersion);
            IEnumerable<JObject> initialWaters = JsonUtilities.Children(result.Root).Where(c => string.Equals(c["Name"].ToString(), "Initial water", StringComparison.InvariantCultureIgnoreCase));
            Assert.AreEqual(1, initialWaters.Count());
            Assert.AreEqual("Models.Soils.Sample, Models", initialWaters.First()["$type"].ToString());
        }
```

PR #8589, opened 2024-01-29
Emerging

### LINKED ISSUE #8496: Error in the way phenology sets Emerged
### What happened?

When phenology increments phase, the first thing it does should be increment the phase index.   Currently it determines if the crop is emerged based on current phase (which is actually the old phase because we have not incremented the currentPhaseIndex yet.  
The result is the Emerged property is one behind where it should be.  We have worked round this error by setting IsEmerged to true on the Emerging phase for all our models which is not correct.  We need to move the setting of currentPhaseIndex to the top of the while loop as shown below and then go through all our models and change the IsEmerged on the Emerging phase of all our models from true to false.
image

Windows

HamishBrownPFR:
Resolves #8496


HamishBrownPFR:
Retest this please jenkins

HamishBrownPFR:
@hol353 I am not sure why this pull request is breaking a unit test.  Can you help me work this out

hol353:
Ahh so it looks like my last 2 pull requests actually conflicted with each other but the conflict wasn't detected because each PR modified different files. Anyway, can you change line 17 of TreeProxyTests.cs to:

```c#
public class TreeProxyTests
```

HamishBrownPFR:
@hol353 @hut104 I am going to request this pull request be merged after a careful review from you two.

Rational of Change:  Currently we have to have the IsEmerged property on the stage prior to emergence set to true to have emergence events trigger properly.  This was done to work round some forgotten bug at some stage and its quite confusing when trying to set up new models.   The code change required to fix it is minor and it in advertantly fixes a couple of bugs we had noticed in development.  An example is shown below for wheat where Stage sits unchanged on the day of development and also for a number of days in the heading phase following flag leaf appearance with the current release but these problems are gone with this PR.  
image
Nice to have those bugs fixed.  The problem is we have fitted around them in the past and there are a few diffs in phenology for Wheat, Oat, Barley and GrapeVine.

Synopsis of PO stats.
- Lots of small diffs
- Major change in yield stats for microclimate test but that is because n = 2 so it is highly sensitive to any small change
- Some diffs in phenology variables for Wheat which will require refitting of individual cultivars to fix.

Do you think we can tollerate the diffs for the sake of cleaning up the code and not having to fit round errors in the future.

hol353:
I assume this will shorten time to maturity by a couple of days for all wheat cultivars?

HamishBrownPFR:
That is correct and evident in the heading date graphs for the wheat test

hol353:
This will require some thought from @hut104.

HamishBrownPFR:
@hut104 any thoughts on this issue.

HamishBrownPFR:
@hut104 we would value your input to this

HamishBrownPFR:
@hol353 @hut104 I have done a bit more work on this pull request and it is suitable for merge now.  
Perviously phenology.cs set its Emerged properties on phase change and set events which was quite convoluted.  I have reworked it now so the Emerged property just gets the IsEmerged property from the current phase.  I had to change a few crop.json files as the property was not used properly previously and some were incorrect.
The upshot IsEmerged can be set to sensible values on each phase (the original intent of this pull request), we have tidier code to deal with the IsEmerged property and not the other "Features" in the cereal models have been reinstated.  Im not sure how they were fixed with prior commits or how the current committ unfixes them but all the diffs now a trivial so should be good to merge.

HamishBrownPFR:
@hol353 please review diffs (all Tidlers are far as I can see) and merge is happy

hol353:
There are lots of GUI diffs here that probably shouldn't be. Is this branch up to date?

HamishBrownPFR:
A number of PR's have been merged since I requested review so it has got behind.  Update now

PR #8607, opened 2024-02-05
Converter bug fix for replacing SoilNitrogen with Nutrient

### LINKED ISSUE #8558: Conversion of soil with old SoilNitrogen to one with new Nutrient
### What happened?

When opening old simulations in the current version of ApsimX, if the soil node contained SoilNitrogen, and error is generted. Apsim deletes SoilNitrogen but Nutrient is not added. Is this a 'feature' or an issue? Also, the error message created when trying to run the converted simulation is not very useful to most people for identifying the problem. 
Furthermore, if one tries to simply add Nutrient (via right click -> add model), another error appears as it seems not all of its components (pools) comes with it... One has to copy Nutrient from another simulation...

Windows

ric394:
resolves #8558 
# Problem
When converting an older apsim file (pre 166), if a Soil node contained a SoilNitrogen node it would get replaced with a Nutrient Node.
If the Soil node did not already contain a Nutrient node it would remove the SoilNitrogen node but not replace it with a Nutrient node.

# Fix
Fixed with a simple count check of soil node children of type nutrient. If zero, add a nutrient node.


ric394:
retest this please jenkins

Issue #5275, opened 2020-06-09
Don't allow duplication of variable names with JSON objects


HamishBrownPFR:
Currently it is possible to make an object in JSON that has the same name as a property of the parent object.
For example Phenology has a property called AccumulatedTT.  Currently I can also put a function on the Phenology class in the json file named AccumulatedTT.  The phenology class will not call this when running, it will use its own property.  However, if you report Phenology.AccumulatedTT Apsim will report the the Value of the Function named AccumulatedTT rather than the property that belongs to the phenology class.
This is a pretty silly thing to do but it happens (I just did it) and causes confusion.  Could we include a test that throws an error if function names duplicate the names of properties in the class they sit upon?

hol430:
Yeah, that probably shouldn't be allowed. I've added this check and I'm getting a few errors which will be quite widespread:

- GenericPhase has a child, a field, and a property all called Target
- Soil has a property and a child both called SoilWater
- Plant has a property and a child both called AboveGround

There may be others too - the error is only thrown if the property/child is accessed, and my simulations are bombing before the simulation even starts, when report does its initialisation.

hol430:
This will be a problem for any C# code which uses `Apsim.Get(...)` (this seems to happen a bit in manager scripts). I have written a test to detect this by basically just adding a few lines into the locator code which, when we find a child model with matching name, checks if a public property also exists with the same name. (Note that if the property/model is never accessed, the exception is never triggered.)

Getting my test committed and merged may be difficult.

hut104:
Can we start by helping to avoid this getting worse?

e.g. can we extend the test that naming of child nodes in the UI doesn't duplicate function names to also look for duplication of names of properties.

hol430:
We could, but that might prevent people from setting up a generic phase from scratch. E.g. it expects a child called Target (it has a child link by name), but it also has a property called Target.

hut104:
This is sounding like that earlier soil water property error, of that null reference bug, which required us to have a branch checked out for months while we slowly worked our way through every instance of the problem.
I'm assuming we finished those other two bugs????

hol430:
Yep, they should be fixed now. It's also similar in scope to the duplicate observed data issue which I notice no one has been brave enough to follow up on.

I guess the biggest problem here will be if people reference any of these models or properties in manager scripts. If we don't have to write a converter for manager scripts the problem shouldn't be too bad.

hut104:
[...]

PR #8610, opened 2024-02-06
5275 Children Matching Property Names

par456:
Working #5275 

Previous if a child had the same name as a property of it's parent, the child would be referenced instead of the property in reports, but the opposite would be true in manager scripts.

This now makes the locator check property and methods first, before then looking at children. It may break things, so this is a run to see how jenkins responds.

Along with changing the Locator code, I also broke the very large GetInternal method into several smaller parts so it can better be tested. Will start on units tests as jenkins runs.

### DIFF
--- a/ApsimNG/Presenters/ExplorerPresenter.cs
+++ b/ApsimNG/Presenters/ExplorerPresenter.cs
@@ -759,7 +759,7 @@ namespace UserInterface.Presenters
         {
             if (this.view.Tree.SelectedNode != string.Empty)
             {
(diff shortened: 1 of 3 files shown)


par456:
retest this please jenkins

par456:
That should finish up this PR if there are no problems.

We definitely need to go back and refactor that locater code to be more maintainable and easier to test. I've done a bit of work in this PR to help that, but it still needs a lot more, along with a consistent set of rules for it to follow.

PR #8613, opened 2024-02-06
Water Presenter 'Fill from top?' checkbox bug fix

### LINKED ISSUE #8543: Water Presenter 'Filled from top?' checkbox not working correctly
### What happened?

as above.

Selecting the checkbox has no affect on graph or figures.

Windows

### LINKED ISSUE #8619: Changing Water InitialValue cells to fill beyond 100% causes crashes
### What happened?

# Problem

1. Percent full cannot currently be a number above 100. When the cell value(s) change causing this number to go above 100, an exception occurs.

# Questions

1. Should percent full be allowed to go above 100?
2. If it should be allowed to go over 100, how should the distribution of water be done when either filled from top is true or false?

# Temporary solution
I'll add a check that prevents any initialValue cell from exceeding DUL for it's index.

Windows

### LINKED ISSUE #8636: Can't delete rows from Physical
### What happened?

When I try to delete a layer from physical, all the input lines are cleared, but the the calcualted SW line and PAWC line (in the cropsoil) are set to 999999 and the empty cells are left as NaNs and not deleted from the structure. Undoing the delete is also no possible as an exception is thrown.

This causes cascading errors through the soil as the layers still exist, but with NaN or 999999 values which screw everything up.

My workaround for this at the moment is to delete what I can, then to edit the file directly and remove the deleted lines from physical in the json.

Windows

ric394:
resolves #8543 
resolves #8636
resolves #8619


par456:
Can you add some units tests in for this to check that when these parameters are updated in water, it actually updates the related numbers correctly.

ric394:
retest this please jenkins

hol353:
I've pulled this branch and it works well. Two things I noticed. 

1. Percent full should be allowed to go above 100%.
2. Initial values in the grid should be able to go to SAT. Currently they are bound to DUL.
3. If you try to enter a percent full of 101% then the error below is given. It should say 'Invalid value for percent full: 101. Must be between [0, 100]'

image

APSIM.Shared/Utilities/DataTableUtilities.cs
```diff
@@ -284,16 +284,35 @@ static public int[] GetColumnAsIntegers(DataView table, string columnName)
...
+            // Checks for headers
+            //string tablesFirstValue = table.Rows[0][columnName] as string;
+            //double firstValue;
+            //if (!Double.TryParse(tablesFirstValue, out firstValue) || tablesFirstValue == null)
+            //{
+            //    hasHeaders = true;
+            //    //headerOffset = -1;
+            //    values = new double[(numValues - startRow) /*+ headerOffset*/];
+            //}
+            //else values = new double[numValues - startRow];
```
hol353, line 300: Remove commented code.

APSIM.Shared/Utilities/DataTableUtilities.cs
```diff
@@ -284,16 +284,35 @@ static public int[] GetColumnAsIntegers(DataView table, string columnName)
...
+            values = new double[numValues - startRow];
+
             int index = 0;
-            for (int Row = startRow; Row != table.Rows.Count && index != numValues; Row++)
+            for (int Row = startRow; Row != table.Rows.Count && index != numValues /*+ headerOffset*/; Row++)
             {
```
hol353, line 305: Likewise, more commented code. Also loop variables (e.g. Row) should start with a lower case.

APSIM.Shared/Utilities/DataTableUtilities.cs
```diff
@@ -284,16 +284,35 @@ static public int[] GetColumnAsIntegers(DataView table, string columnName)
...
                 if (table.Rows[Row][columnName].ToString() == string.Empty)
                     values[index] = double.NaN;
                 else
                 {
                     try
                     {
+                        //if (hasHeaders)
+                        //    values[index] = Convert.ToDouble(table.Rows[Row + 1][columnName], culture);
+                        //else
                         values[index] = Convert.ToDouble(table.Rows[Row][columnName], culture);
```
hol353, line 316: Commented code.

APSIM.Shared/Utilities/DataTableUtilities.cs
```diff
@@ -1393,5 +1412,18 @@ select new
...
+        /// </summary>
+        /// <param name="table"></param>
+        /// <returns>a <see cref="bool">bool</see></returns>
+        public static bool IsTablesFirstValueDouble(DataTable table)
+        {
+            double firstValue;
+            if (Double.TryParse(table.Rows[0][0] as string, out firstValue))
+                return true;
+            else return false;
+        }
```
hol353, line 1427: Couldn't this just be:

```c#
return Double.TryParse(table.Rows[0][0] as string, out double firstValue)
```

hol353 (changes requested):

par456:
Ah yeah, I had some questions about that as well.

How does the graph need to fill when going above 100% if filled from top or not?

Obviously when filled from top is off, it can just continue to evenly fill from LL, that's fine. But when filled from top is on, does it fill all the way to SAT before filling the next layer, or does it fill up to DUL until all layers are filled to DUL, then start filling to SAT for each layer as it goes higher?

Currently as well, the percentage filled is based on DUL, so at 100% DUL is saturated, should 200% be SAT saturated, or should it just be whatever percentage it takes to fill to SAT.

Also, are the initial values limited to SAT? So if someone puts in 1000% we can cap off the initial values to SAT.

hol353:
Up to 100% full it should use DUL. Once above 100%, filled from top should fill to SAT.

> Currently as well, the percentage filled is based on DUL, so at 100% DUL is saturated, should 200% be SAT saturated, or should it just be whatever percentage it takes to fill to SAT.

It should be whatever percentage it takes to fill to SAT. Will be different percent for different soils.

> Also, are the initial values limited to SAT? So if someone puts in 1000% we can cap off the initial values to SAT.

Yes, always cap it to SAT.

par456:
Looks it's finally run green on Jenkins. Should be good for a review @hol353 and hopefully merged.

hol353 (approved):

hol353:
Good job @ric394! I've pulled the branch and had a play. Can't find any further problems. Merging...

PR #8639, opened 2024-02-15
Fix biomass removal bugs.

HamishBrownPFR:
working on #8175
Add base winter cover for evergreen tree crops
Allow setting harvest and prune dates from STRUM instance or manager


HamishBrownPFR:
@hol353 @par456 @ric394 could someone review and merge please

PR #8643, opened 2024-02-16
8642 import issues

### LINKED ISSUE #8642: Issues with importing an APSIM Classic file
### What happened?

I've run into a few issues while importing APSIM classic files. 

1. When importing operations with a USA cultureinfo, some dates will convert incorrectly. In classic the Operations module expected dates to typically be provided in a day/month/year format (it could support other formats too, such as day-of-year year, but I am not sure how widely they were used). NextGen seems to assume the system culture info is the format Operations dates are in, causing entries with day < 13 to get their day and month flipped.
2. KS does not get copied over.
3. When importing a SWIM simulation a SwimWaterTable object will attempt to be created, but this object was removed some time back. Importing fails when this happens.

Windows

ilhuber:
Resolves #8642 

There were other date formats accepted by classic's operations module (doy year), but I am not sure how widely used they were. I split date conversion into its own method so it will be easy to add them if desired. 

Also, not sure which file format version SwimWaterTable was removed in, so I stuck its deletion after the converter handles something with SwimSolutes.

### DIFF
--- a/Models/Core/Apsim710File/Importer.cs
+++ b/Models/Core/Apsim710File/Importer.cs
@@ -1147,7 +1147,7 @@ namespace Models.Core.Apsim710File
 
                 string childText = string.Empty;
                 childNode = XmlUtilities.Find(oper, "date");
-                dateNode.InnerText = OperationDateToNGDate(childNode?.InnerText);
+                dateNode.InnerText = DateUtilities.ValidateDateString(childNode?.InnerText ?? string.Empty);
 
                 XmlNode actionNode = operationNode.AppendChild(destParent.OwnerDocument.CreateElement("Action"));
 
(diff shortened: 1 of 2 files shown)


par456:
We wrote a utility class to handle dates a little while back to help solve this exact problem, looks like you found an old bit of code that wasn't using it.

The date parsing problem in importer should be solved with just:
```c#
string childText = string.Empty;
childNode = XmlUtilities.Find(oper, "date");
dateNode.InnerText = DateUtilities.GetDate(childNode.InnerText).ToString("yyyy-MM-dd");
```

ilhuber:
Thanks for the heads up! I wasn't aware of the DateUtilites changes. 

Both Classic and NextGen are able to handle annually repeating entries in the Operations module, though, and I think DateUtilities.GetDate may not be able to maintain that by itself.

I think I may update OperationDateToNGDate to something like
```csharp
private string OperationDateToNGDate(string date)
{
    // This case occurs for an empty entry in classic's operations schedule.
    // It is common for there to be a few, as they can't be deleted through classic's UI.
    if (string.IsNullOrEmpty(date))
        return "";

    // Both classic and NextGen support annually repeating entries in the Operations module.
    string[] dateFormatsAnnual = {"d-MMM", "d_MMM"};
    if (DateTime.TryParseExact(date, dateFormatsAnnual, CultureInfo.InvariantCulture, DateTimeStyles.None, out _))
        return date;

    return DateUtilities.GetDate(date).ToString("yyyy-MM-dd");
}
```
instead.

par456:
What format are the repeating dates coming in as from classic? If it has some that we can't parse we need to add them to the utilities class so that it can.

I'm pretty sure that operations in NG require a year in the date and doesn't allow you to just have a repeating day or year or d-MMM. If I input a date without a year into the Operations GUI for example it always appends a year to the value.

hol353:
I would do the simple thing and just use DateUtilities.ParseDateString. Yes it accepts more formats than classic but the chances are pretty remote that a simulation would suddenly start working on import.

ilhuber:
If you're both fine with it then this should be ready to be merged, unless you would like some tests added to make sure operations dates get imported correctly regardless of culture info.

PR #8654, opened 2024-02-19
compiled script kept explicit

### LINKED ISSUE #8652: Can't add memo/child models to manager
### What happened?

As the manager component keeps its script component as an invisible child, any attempt to add a child component - eg a memo - is silently dropped (though still visible in the simulation tree). 

Linux

peter-devoil:
Resolves #8652

Keeps the compiled script component explicity, doesnt wipe Children[].

Needs an extra look please @par456

### DIFF
--- a/ApsimNG/Presenters/ManagerPresenter.cs
+++ b/ApsimNG/Presenters/ManagerPresenter.cs
@@ -71,7 +71,7 @@ namespace UserInterface.Presenters
                 }
             }
 
-            scriptModel = manager.compiledModel;
+            scriptModel = manager.ScriptModel;
 
             // See if manager script has a description attribute on it's class.
             if (scriptModel != null)
--- a/ApsimNG/Presenters/ManagerPresenter.cs
+++ b/ApsimNG/Presenters/ManagerPresenter.cs
@@ -71,7 +71,7 @@ namespace UserInterface.Presenters
                 }
             }
 
-            scriptModel = manager.Children.FirstOrDefault();
+            scriptModel = manager.compiledModel;
 
             // See if manager script has a description attribute on it's class.
             if (scriptModel != null)
--- a/Models/Core/ApsimFile/FileFormat.cs
+++ b/Models/Core/ApsimFile/FileFormat.cs
@@ -244,7 +244,10 @@ namespace Models.Core.ApsimFile
                 private IEnumerable<IModel> ChildrenToSerialize(IModel model)
                 {
                     if (model is Manager)
-                        return new Model[0];
+                    {
+                        // Zap any script components
+                        model.Children.RemoveAll(o => o == (model as Manager).compiledModel);
+                    }
 
                     // Serialise all child if ResourceName is empty.
                     if (string.IsNullOrEmpty(model.ResourceName))
--- a/Models/Core/ApsimFile/FileFormat.cs
+++ b/Models/Core/ApsimFile/FileFormat.cs
@@ -245,8 +245,12 @@ namespace Models.Core.ApsimFile
                 {
                     if (model is Manager)
                     {
-                        // Zap any script components
-                        model.Children.RemoveAll(o => o == (model as Manager).ScriptModel);
+                        List<Model> children = new List<Model>();
+                        foreach (Model child in model.Children) {
+                            if (child != (model as Manager).ScriptModel)
+                                children.Add(child);
+                        }
+                        return children.ToArray();
                     }
 
                     // Serialise all child if ResourceName is empty.
--- a/Models/Core/ApsimFile/FileFormat.cs
+++ b/Models/Core/ApsimFile/FileFormat.cs
@@ -246,7 +246,7 @@ namespace Models.Core.ApsimFile
                     if (model is Manager)
                     {
                         // Zap any script components
-                        model.Children.RemoveAll(o => o == (model as Manager).compiledModel);
+                        model.Children.RemoveAll(o => o == (model as Manager).ScriptModel);
                     }
 
                     // Serialise all child if ResourceName is empty.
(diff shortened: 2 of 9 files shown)


Models/Management/Manager.cs
```diff
@@ -75,6 +75,9 @@ private void SetCompiler(ScriptCompiler compiler)
             scriptCompiler = compiler;
         }
 
+        /// <summary>Which child is the compiled script model.</summary>
+        public IModel compiledModel { get; private set; } = null;
```
par456, line 79: I think we can make this a private property of Manager, as the only place it's used externally of this class is ManagerPresenter, which can find it in another way.

ApsimNG/Presenters/ManagerPresenter.cs
```diff
@@ -71,7 +71,7 @@ public void Attach(object model, object view, ExplorerPresenter presenter)
                 }
             }
 
-            scriptModel = manager.Children.FirstOrDefault();
+            scriptModel = manager.compiledModel;
```
par456, line 74: If we keep compiledModel as private, then this line can be:
scriptModel = manager.FindChild("Script");

Models/Core/ApsimFile/FileFormat.cs
```diff
@@ -244,7 +244,10 @@ public void SetValue(object target, object value)
                 private IEnumerable<IModel> ChildrenToSerialize(IModel model)
                 {
                     if (model is Manager)
-                        return new Model[0];
+                    {
+                        // Zap any script components
+                        model.Children.RemoveAll(o => o == (model as Manager).compiledModel);
```
par456, line 249: And similar here:
model.Children.Remove((model as Manager).FindChild("Script"));

par456:
This looks like it should solve the problem, though I don't think compiledModel needs to be a public variable on Manager, as it's mostly used internally. I've put some suggestions on how to do the changes to make it private in the comments above.

ApsimNG/Presenters/ManagerPresenter.cs
peter-devoil: I battled this for a while when I realised that the compiled code is not necessarily called "script"; it could be called anything - eg the EventPublishSubscribe tutorial calls them "Script1" and "Script2".

par456: Hmmm, that seems like a larger bug as there should only ever be one script node, but I could see multiple being created and that happening, and yeah, that would muck up this technique.

If we do keep it as a public property then, could you change the name to be "ScriptModel" so it's a bit more descriptive of what it's referring to. It does hold a model of type "Script", it's just that Script doesn't exist until runtime so we can't use that type in the source.

It'll also need the [JsonIgnore] flag above where it's declared so that it's not saved to file.

peter-devoil: No, there is only ever one compiled script for each model; but it's up to the script writer as to what it is called - and it doesnt appear until runtime. So figuring out which child is a "UI configuration", and which is our compiled model isn't something we can easily determine from attributes (perhaps?).

peter-devoil:
@par456 this is failing the unit test related to ["check all functions are tested"](https://github.com/APSIMInitiative/ApsimX/blob/6468632ec126797dca769a1b679821216afcdcee/Tests/UnitTests/Manager/ManagerTests.cs#L592). There's a mangled name appearing that hasn't got a test - any clues? 
image

par456:
Manager got an overhaul in its unit tests a few weeks ago so that it has a test that checks that it's properties and methods are being tested. Looks like you've found a bug in that system, and that doesn't look like an actual method name that it's found.

I've added a filter to the Reflection code to exclude methods like that and also fixed up the other unit test that was failing after the changes (was returning null instead of throwing an exception like it used to).

I've pushed that up to your branch and assuming it runs green, this should be ready to merge.

peter-devoil:
Hmm. There's a problem here that I see when opening Examples/Stock.apsimx; and opening the 1st manager component throws an exception [here](https://github.com/APSIMInitiative/ApsimX/blob/6468632ec126797dca769a1b679821216afcdcee/ApsimNG/Views/ManagerView.cs#L72) with cursor = null.

As well, the properties tabs aren't populated with properties for a considerable time after opening.

par456:
The slowdown on the manager page was caused when the cursor position saving was added, as it has to wait for the script to load before it reads now. I'll put that into a separate issue because it's not caused by these changes.

Fixed the cursor bug you found and then also worked out why that PredictedObserved validation was failing. The FileFormat changes were deleting the Script from the structure, rather than just from the list it had to pass back. We'll see how these changes go.

peter-devoil:
Oops, I dont think this was ready for a merge; I'd like to suggest a refactor of the unit test so that there aren't [4 ways of constructing](https://github.com/APSIMInitiative/ApsimX/blob/a4eefe3056bc3856be81cd4b2073254bb8f1cb4e/Tests/UnitTests/Manager/ManagerTests.cs#L31) the component - there should be 2 - one for running simulations, and one for GUI interaction; at the moment both are jumbled up..

I'll get to this later today..

par456:
That creation function is written so that I could create manager components in different levels of setup. You'll have to be careful when changing that, because a lot of the new user tests are specifically written to check how they respond to each setup.

Such as if it has a ScriptCompiler from Simulations, or if it might have code but has been compiled yet. That way I could make sure that if say, it tried to compile without a ScriptCompiler, or was missing code or didn't have a parent to find the compiler, that it would either throw or handle the problem correctly for each function call at each level of setup.

This is a newer style of unit test I've been looking to add where we test that individual methods in a class can handle a wide variety of situations without crashing, rather than the older unit tests which mainly just checked edge cases and basic uses. 

In terms of the bug that was linked, I've put a patch in to fix that. That was caused by one of my late changes where it looks like I removed a line I didn't intend to (to remove the old Script node when compiling a new one). New PR ( #8674 ) is up being tested and should be fixed very soon.

PR #8661, opened 2024-02-21
8533 Fixed "Not An Error" Exception

### LINKED ISSUE #8533: "Not an Error" exception thrown
### What happened?

I've seen this on both my simulations and @hut104 before, so I captured some info about it today when it happened. Seems to be related to simulations that have a predicted/observed in them. Crash happens when you run the simulation.

image

The data is still generated and working, just this weird error is given. When I run it again, it often returns another error such as "Unknown Error" or a longer error about a "transaction within a transaction". Will need to debug further.

I've attached the simulation I was using to test when it gave the error.

[brokenTest.zip](https://github.com/APSIMInitiative/ApsimX/files/13869816/brokenTest.zip)

Windows

### LINKED ISSUE #8625: PredictedObserved Table not generating when it starts with disabled experiment
### What happened?

Been noticing this with a file I'm working with, that if I have experiments listed in my observed file that are not currently in the simulations, that it's a lottery if the predictedobserved will actually work or not. It may also be related to #8533 since I see that error sometimes when this is happening. 

My only solution so far has been to delete the datastore and remove the unused rows from my observed file for any simulations I am not intended to run.

Windows

par456:
Resolves #8533
Resolves #8625

It was possible for one thread to try and clean an sqlite database while another was still trying to read or write causing two transactions to try and run at the same time. SQLite would throw an exception with a variety of names when this happened.

Solution was to add a locking mechanism to the Begin and End calls to only allow one thread at a time to do a transaction, and then encapsulate the uses of those functions with try...finally blocks so that the locks would always unlock if and respond with the error message if an exception was thrown.

### DIFF
--- a/APSIM.Shared/Utilities/DBMerger.cs
+++ b/APSIM.Shared/Utilities/DBMerger.cs
@@ -75,38 +75,41 @@
         public static void Merge(IDatabaseConnection source, SQLite destination)
         {
             destination.BeginTransaction();
-
-            if (source.GetTableNames().Contains("_Simulations"))
-            {
-                var sourceData = source.ExecuteQuery("SELECT * FROM _Simulations");
-                var destinationData = destination.ExecuteQuery("SELECT * FROM _Simulations");
-                var simulationNameIDMapping = destinationData
-                                                  .AsEnumerable()
-                                                  .ToDictionary(row => row.Field<string>(1),
-                                                                row => row.Field<int>(0));
-
-                var oldIDNewIDMapping = new Dictionary<int, int>();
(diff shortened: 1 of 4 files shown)


par456:
Jenkins ran four minutes slower on this build, might need to discuss this before merging.

par456:
retest this please jenkins

par456:
retest this please jenkins

par456:
@hol353 That slow down looks like it was just a jenkins hiccup, the 2nd and 3rd runs ran at normal speeds.

This should be right to merge after review.

PR #8675, opened 2024-02-22
8069 Mislabelled Columns in Barley Observed

### LINKED ISSUE #8069: Export database to excel or text throws date error
### What happened?

Try exporting the database from tests/validataion/Barley/Barley.apsimx and you will see what I mean

Windows

par456:
Resolves #8069

The Barley Observed.xlsx and the Obs_APSIMX_NPIField2019_Barley.xlsx and Obs_APSIMX_NPIField2020_Barley.xlsx, were using the columns TOS and Sowing differently, causing merge issues when apsim was combining the data.

The NPIField experiments had a factor based on the value of these columns, while as far as I could tell, the experiments in the Observed.xlsx with data in these columns did not. My solution was to switch the Sowing and TOS column headers in the Observed.xlsx and see how that goes. It doesn't crash on exporting and validation does run, but need to check first if this has effected any PO stats if something was relying on that data.


par456:
Seems to have worked without any issues.

PR #8689, opened 2024-02-25
Scrum rotation

HamishBrownPFR:
working on #7869


HamishBrownPFR:
@hol353 please review and merge if happy

ApsimNG/Presenters/GraphPresenter.cs
```diff
@@ -186,6 +186,13 @@ public void DrawGraph(IEnumerable<SeriesDefinition> definitions)
                         double yDouble = 0;
                         if (y is DateTime)
                             yDouble = ((DateTime)y).ToOADate();
+                        else if (y is String)
+                        {
+                            if ((String)y == "True")
+                                yDouble = 1.0;
+                            if ((String)y == "False")
+                                yDouble = 0.0;
```
hol353, line 194: Why is this needed? Why not use a boolean?

HamishBrownPFR: The graphs seem to work alright without this fix (ie this fix does not affect graphs appearance) but there is a non-fatal bug that keeps popping up in the debugger which is a bit annoying.  This change stops that bug

hol353: I'd like to see that bug reproduced. Can you show me tomorrow via a Teams call? I don't really want this code in the graph presenter.

HamishBrownPFR: I will pull the code change out of this pull request to keep this issue separate.  Will give you a call to show you the bug separately

PR #8693, opened 2024-02-26
Error in maxt < mint in lincoln.met

sno036:
Fix a (probably long-standing!) error in the example lincoln met file


hol353:
retest this please jenkins

PR #8698, opened 2024-02-27
8493 Changes to Lincoln9192 Leaf.Dead .N .NConc and .Wt

par456:
Working on #8493

This PR has changes to the following Wheat Observed Data:

- Wheat.Leaf.Dead.N: Values from Lincoln9192 increased by a factor of 1 (x10)
- Wheat.Leaf.Dead.NConc: Values from Lincoln9192 were stored as percentage instead of (g/g) (divided values by 100 to convert back)
- Wheat.Leaf.Dead.Wt: Values from Lincoln9192 increased by a factor of 1 (x10)

These are suspected incorrect units that were entered in for this experiment. This change brings the values into alignment with ranges recorded from other experiments in the Wheat validation.


par456:
Closing to redo with updated observed.xlsx file

Issue #8713, opened 2024-02-28
Not a legal OleAut date


par456:
### What happened?

A new bug in the graphing that I ran into: 

A date value of -693593 was tried to converted to a date in the LineSeriesWithTracker class and crashed. Click on the left-most point of the phenology graph in the Mouse experiment in Wheat validation -> South Australia.

![image](https://github.com/APSIMInitiative/ApsimX/assets/130029141/4b60c0d9-f85a-4161-9409-38c6a0e2d4b7)

### What Operating System are you on?

Windows

PR #8722, opened 2024-02-29
8717 Zadok Changes

### LINKED ISSUE #8717: Zadok Stage on day of sowing is 4.9
### What happened?

In the GerminationPhase, FractionComplete always returns 0.999, which leads to the strange situation where on the day of sowing, the plant is already at Zadok Stage 4.995

This then causes inconsistencies with observed data where people list a zadok stage of 0 the day it's sown, and apsim says it's almost completely germinated.

APSIM should probably return a zadok stage of 0 on the day of sowing, and then go through germination on the next day so that it can match the existing phenology still.

Windows

par456:
Resolves #8717 
Working on #8493

Some strange points were found in the Zadok.Stage of the wheat validation. Notably, in the MouseRemoval experiments, it had a Zadok Stage of 0 entered 35 day prior to the crop being sown, and a zadok stage of 90 after the crop had been harvested with no clock.today value which broke the graphs (see issue #8713 which will need to reproduce this now). These have been removed as errors.

The MouseRemoval experiment also had a Zadok.Stage observation recorded the same day as the crop was harvested, which meant apsim was reporting back a stage of 0 when the observation was 86. To fix this the apsim harvest was delayed to the day after.

In the Lonzee experiments, a Zadok Stage of 0 was observed on the day of the crop being sown, but apsim was reporting a score of 4.995 that day. To resolve this, the germination phase will now report back a stage of 0, as on the 2nd day of the crop it will move into the start of the emergence phase like before. This just allows for more sensible reporting on the first day. Will have to see how this impacts our stats, as it may effect other crops.


par456:
This did change the barley stats, but only because barley has 4 points where the crop has been harvested in Apsim, but the observations are on or after that date.

Basically, the comparisons of 92 observed to 4.995 predicted, are now 92 observed to 0 predicted. So they did get worse, but only for measurements that were already a problem.

PR #8723, opened 2024-02-29
:bug: Set edit-use-config and run-use-config as deprecated

### LINKED ISSUE #8704: :bug: edit-use-config and run-use-config options not following config file
## Description

I am using ApsimX by running Models in command line. The options ***--edit-use-config*** and ***--run-use-config*** are listed when I run ***Models --help*** but they are not working as expected. When I use these options, the models runs normally but don't change the parameters following the config file. I checked the source code and there is no implementation for the parameters below:

https://github.com/APSIMInitiative/ApsimX/blob/66653254abe96c3bfd9913ad74281168893208a4/Models/Options.cs#L68-L69
https://github.com/APSIMInitiative/ApsimX/blob/66653254abe96c3bfd9913ad74281168893208a4/Models/Options.cs#L77-L78

I think these parameters were replaced by **--apply** feature, but they were not deprecated like the parameter **--edit** was.

Linux

luanabeckerdaluz:
This resolves #8704 issue.

### DIFF
--- a/Models/Options.cs
+++ b/Models/Options.cs
@@ -65,7 +65,7 @@ namespace Models
         /// This property holds the path to the config file.
         /// This is identical to --edit switch. 
         /// </remarks>
-        [Option("run-use-config", HelpText = "Edit the .apsimx file(s) before running them. Path to a config file must be specified which contains lines of parameters to change, in the form 'path = value'.")]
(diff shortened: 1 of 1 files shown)


hol353:
add to whitelist jenkins

par456:
retest this please jenkins

that random documentation error has popped up again I think

ric394 (approved):

PR #8737, opened 2024-03-07
Script Classes get unique names

### LINKED ISSUE #8624: Manager doesn't work in multi-zone simulation if it has not public properties.
### What happened?

In the attached file I have set up a multizone simulation to represent an area of crop under plastic cover and an adjacent fallow area that water is shed to from the covered area.  To do this I have put a weather model in each zone and a manager that modifies Rainfall and wind to zero and increased maxT and VP in the CoveredRow zone and doubled the rainfall in the InterRow zone.

The intended modifications have worked in the example I have uploaded with this issue.  However, I have had to include a superfluous property (highlighted in yellow) for the intended set up to work.  
If you comment that property out the model does not recognize this manager (won't stop on a break point there if you include one) and it uses the weather data from the zone above (i.e. not only does it fail to apply the modification intended for this zone, it brings in the modifications applied to the other zone).

image

With superfluous property
image

With superfluous property commended out
image

While this is a simple work around, it should be fixed to avoid future confusion.

[Uploading ManagerBug.zip…]()

Windows

par456:
Resolves #8624

This turned out to be a really complicated problem, as it wasn't just a scoping issue like I first thought and had nothing to do with the zones. I spent most of yesterday debugging to work out that the problem was actually .net reflection getting the wrong Script class when two scripts had identical names and properties. It only seemed to care about properties, as adding a function or class variable still caused the problem to happen. 

I was also able to reproduce the behaviour with an operation call, where it called the wrong code even when the managers had different names.

My solution for this has been to use the fullpath of the manager to give each script a unique class name before it's compiled, that way this name conflict can never happen. While I was there, I also added an IScript interface to the script code so that we have a type we can check against to see if a child is a script or not, instead of the "has script in name" method that was used before.


peter-devoil:
I'm glad you had the patience to find this - I gave up and worked around it.

par456:
Yeah, that's what I started encountering as I dug into this problem, and I think we need a more defined way for communication between manager scripts. Something that uses the locator code that the rest of APSIM uses, instead of trying to do it by reflection in the loaded assemblies.

HamishBrownPFR:
Thanks for taking the time to resolve this.

par456:
@hol353 

I assume it's the new converter that's seriously slowing down this branch on jenkins? Having to check all the scripts for specific errors seems to be taking a long time when it has to 'update' so many files.

par456:
Retest this please jenkins

Ran a little slow still, seeing if its a one-off

par456:
@hol353 

Should be right for a review and merge now.

Issue #8760, opened 2024-03-17
The way grids work with models needs a refactor


hol353:
### Where is the code located?

Many models display data in grids and allow that data to be edited.

### What needs to be changed?

 These models implement ```interface IGridModel``` and have the property ```public List<GridTable> Tables```. Some classes also implement the method ```ConvertModelToDisplay(...)```. Classes Physical, Forages, TreeProxy are examples of this. The interface mentioned isn't really an interface because it has implementation in it. Because of that, there a multiple ways of returning data. We need a single way of doing things. Also, the code that implements this is user interface related and not model related i.e. it isn't executed when a simulation runs. I think the models assembly should only be simulation related, and not have user interface code.

Possible implementation:
The code should be moved from models into the ApsimNG assembly as a separate class for each model. These classes would be responsible for returning an ISheetDataProvider for a model:
```c#
using System.Linq;
using Models.Core;
using Models.Soils;
using UserInterface.Views;

namespace ApsimNG.SheetDataProviders
{
    public class PhysicalSheetDataProvider
    {
        /// <summary>Tabular data. Called by GUI.</summary>
        public static ISheetDataProvider GetGridTable(IModel model)
        {
            Physical physical = model as Physical;
            bool waterNodePresent = physical.FindInScope<Water>() != null;

            GridTable t = new(physical.Name);
            t.AddColumn(physical, "Depth");
            t.AddColumn(physical, "ParticleSizeSand", columnName: "Sand");
            t.AddColumn(physical, "ParticleSizeSilt", columnName: "Silt");
            t.AddColumn(physical, "ParticleSizeClay", columnName: "Clay");
            t.AddColumn(physical, "Rocks");
            t.AddColumn(physical, "BD");
            t.AddColumn(physical, "AirDry");
            t.AddColumn(physical, "LL15");
            t.AddColumn(physical, "DUL");
            t.AddColumn(physical, "SAT");

            if (waterNodePresent)
                t.AddColumn(physical, "SW", isReadOnly: !physical.IsSWSameLayerStructure);

            t.AddColumn(physical, "KS");
            t.AddColumn(physical, "BD", columnName:"Test", isCalculated: Enumerable.Repeat(true, physical.Thickness.Length));

            foreach (var soilCrop in physical.FindAllChildren<SoilCrop>())
            {
                var cropName = soilCrop.Name.Replace("Soil", "");
                t.AddColumn(soilCrop, "LL", columnName: $"{cropName} LL");
[...]

PR #8763, opened 2024-03-19
Soil defaults

### LINKED ISSUE #8757: Add in default/assumed values to the soil node
### Describe the new feature

Needed for SoilTemperature and generically useful as more models are added. 

For Clay
* If no Clay% data is supplied then assume the Clay% is 30% at all depths
* Insert this into the GUI, highlighting that it is an APSIM assumed value, encourage users to input more accurate data but allow the simulation to proceed as is
* User can then type over the supplied values in the GUI if they have better information – and in this case the highlighting would go away

For Rocks
* If Rocks% is non-blank (0 is a valid entry) then assume those values
* If Rocks% is blank then estimate the value by layer:
   * Assume rp (particle density) is 2.65 Mg /m3 
   * Estimate effective total porosity (tp) as (1-bd/rp)*0.93
   * If (1-Sat/tp) > 0.1 then take RocksFraction = (1-Sat/tp), else 0.0

Will supply similar methods for CEC later on.

hol353:
Resolves #8757

### DIFF
--- a/APSIM.Interop/Drawing/IDrawContext.cs
+++ b/APSIM.Interop/Drawing/IDrawContext.cs
@@ -6,7 +6,8 @@ namespace APSIM.Interop.Drawing
     {
         Normal,
         Selected,
-        Insensitive
+        Insensitive,
+        Calculated
     }
 
     public interface IDrawContext
--- a/APSIM.Shared/Utilities/MathUtilities.cs
+++ b/APSIM.Shared/Utilities/MathUtilities.cs
@@ -694,7 +694,7 @@ namespace APSIM.Shared.Utilities
             {
                 foreach (string Value in Values)
                 {
-                    if (Value != "")
+                    if (!string.IsNullOrEmpty(Value))
                         return true;
                 }
             }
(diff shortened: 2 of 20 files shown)


par456 (commented):
- If I change a calculated value in Physical so that it becomes black, then use undo, it doesn't go back to being red.

I also noticed when clicking around AgPasture that editing a number in forages adds an additional line to the grid. That bug is also on the master branch though, so wasn't caused by this PR.

hol353:
> If I change a calculated value in Physical so that it becomes black, then use undo, it doesn't go back to being red.

I can't find a way to fix this. It is a design problem with the grid presenter. The presenter saves (and undos) a DataTable. A DataTable only contains numbers though and no information about the colour of cells or anything else, hence the numbers are saved but the colours of cells is not. This needs to be addressed in the sheet/grid refactor (#8760) 

> I also noticed when clicking around AgPasture that editing a number in forages adds an additional line to the grid. That bug is also on the master branch though, so wasn't caused by this PR.

Yes, I've noticed that as well.

PR #8781, opened 2024-03-22
Updated soil temperature

hol353:
Working on #4080


hol353:
The changed stats are expected as the soil temperature model has changed quite a lot.

ric394 (approved):

PR #8791, opened 2024-03-28
8784 Check if string column should be date when reading from excel

### LINKED ISSUE #8784: Only set the graph axis name when it's manually set
### Describe the new feature

When a graph is created, it often sets the axis name into the axis that is then stored, even if the axis was not manually named. This then causes the name to be copied when the graph is copied and doesn't update when the X or Y axis value is changed. Leading to graphs with incorrectly named axis when you are making a lot at once and working from copies.

Should fix this so that the axis name is only saved if it was actually manually renamed, otherwise it should continue to change based on what the variable it has been set to.

Might need to also clear old graphs that have an axis with the generated name stored so that older files share the behaviour.

### LINKED ISSUE #8785: APSIM not detecting values as dates if they are listed as 'general' in excel
### What happened?

When importing an excel file (ExcelImport), if the column with dates has not been formatted to a date in excel, APSIM is treating it as a string instead.

It should have some logic to look at the value it's read in and if all of them could be parsed into a date, then it saves the column as a date instead of string type. If we don't want it to be that strict, at least have it so that Clock.Today is always a date, because it's the most common and should always be a date.

Been running into this when importing data from classic where I am using a python script to merge the outputs into an excel file.

Windows

par456:
Resolves #8785

Made it check every column with a string data type, checking until it finds an entry that cannot be parsed to a date. If all the cells can be read as dates, the column data type is then changed. Uses the DateUtilities system to determine if something is a date.


ric394 (changes requested):
These new methods will need unit tests.

ric394 (approved):

PR #8809, opened 2024-04-04
8721 Added OrderBy to DataStore presenter

### LINKED ISSUE #8519: Row order in DataStore

### LINKED ISSUE #8721: Report filters result in some years being repeated and some years being removed
### What happened?

[APSIM (Custom Build) 2024-02-29 14-06-50.zip](https://github.com/APSIMInitiative/ApsimX/files/14442682/APSIM.Custom.Build.2024-02-29.14-06-50.zip)

I ran simulations for 2010-2023. When I filter the daily report, the 2010-2011 data are repeated, and the rest of the data are not shown. Or, something is wrong with Clock-Today when it is filtered! These are the filters I tried:

`"SowDate" = "01-Jul" and "Cultivar" = "Test1" and "Population" = 50 and "AutoIrrigate" = "false"
`

`[SowDate] = "01-Jul" and [Cultivar] = "Test1" and [Population] = 50 and [AutoIrrigate] = "false"`

Windows

par456:
Resolves #8721
Resolves #8519

The rowid's were stored in a temporary table that was then used to get out a small subset of the entire data for the presenter. However, when they were stored (which happened every time the datastore was opened in the gui), they were not sorted and just broke the sorting that was supposed to happen.

I added an OrderBy to that sql code and simplified the part where it pulls out 50 rows and now it gets the correct 50 rows out for that offset into the data.

While I was there, I then also added a dropdown to the interface to allow sorting by any column. By default it will do a simulationID and Clock.Today, but this just gives the user the ability to sort on whatever column they want specifically.

These changes actually improved the performance, as now it only does the sort once at the start, instead of every time you move the scroll bar.

### DIFF
--- a/ApsimNG/Presenters/DataStorePresenter.cs
+++ b/ApsimNG/Presenters/DataStorePresenter.cs
@@ -42,6 +42,9 @@ namespace UserInterface.Presenters
         /// <summary>table name drop down.</summary>
         public DropDownView tableDropDown { get; private set; }
 
+        /// <summary>table name drop down.</summary>
+        public DropDownView orderByDropDown { get; private set; }
+
         /// <summary>Column filter edit box.</summary>
         private EditView columnFilterEditBox;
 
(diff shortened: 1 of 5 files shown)


ric394:
Order by doesn't seem to work as intended except for when it has an empty field or Clock.Today.

par456:
@hol353 

I held this back in case it caused issues while we were in NZ, but if it looks good to you, it should be good to merge now.

Issue #8822, opened 2024-04-10
Wheat Protein


hut104:
### Describe the new feature

The protein predictions from the wheat model need improvement.  Protein levels are too often close to the maximum concentration.

PR #8823, opened 2024-04-10
Wheat protein

hut104:
working on #8822


hol353:
retest this please jenkins

hut104:
@hol353 The WheatProtein pull request improves some things and makes some things not as good.

It does have extra sensibility tests to look at the timecourse of grain size and protein during grain filling.  These responses to water, n, potential grain size, etc seem to show that the trends in the model make sense now, even if we need to do some more work on the actual parameter fitting.

The form of the model is a bit more sensible, with changes to potential grain size not changing the N filling rates.  The model is more like the original model in 7.10.

PR #8831, opened 2024-04-10
8819 Root Growth Changes

### LINKED ISSUE #8819: Root Growth Using wrong layer
### What happened?

When root growth is checking the layer for water, it seems to be looking at the layer above where it currently is. Aka, if there is no water below 600mm, it's still growing to the bottom of the next layer before it stops.

Investigating further.

Windows

par456:
Resolves #8819

### DIFF
--- a/Models/Functions/SoilWaterScale.cs
+++ b/Models/Functions/SoilWaterScale.cs
@@ -37,7 +37,6 @@ namespace Models.Functions
         [Link]
         private Physical physical = null;
 
-        [Link]
         private Root root = null;
 
         private double[] sats = null;
@@ -45,13 +44,21 @@ namespace Models.Functions
         private double[] lls = null;
 
         [EventSubscribe("StartOfSimulation")]
-        private void OnStartOfSimulation(int arrayIndex = -1) {
+        private void OnStartOfSimulation(object sender, EventArgs e) {
             sats = physical.SAT;
             duls = physical.DUL;
+            root = FindAncestor<Root>();
             if (LLModel == LowerLimit.LL)
-                lls = root.SoilCrop.LL;
+            {
+                if (root == null)
+                    throw new Exception($"SoilWaterScale ({Name}) cannot use LL for lower limit as it does not have a root as an ancestor.");
+                else
+                    lls = root.SoilCrop.LL;
+            }
             else
+            {
                 lls = physical.LL15;
+            }
         }
 
         /// <summary>Gets the value of the function.</summary>
--- a/Models/Functions/SoilWaterScale.cs
+++ b/Models/Functions/SoilWaterScale.cs
@@ -42,8 +42,11 @@ namespace Models.Functions
         /// <summary>Gets the value of the function.</summary>
         public double Value(int arrayIndex = -1)
         {
+            //if (arrayIndex == -1) //Red and White Clovers both don't pass a number to this.
+            //    throw new Exception($"Soil Water Scale ({Parent.Name}.{Name}) requires an array index for soil layer");
+
             if (arrayIndex == -1)
-                throw new Exception($"Soil Water Scale ({Parent.Name}.{Name}) requires an array index for soil layer");
+                return 1;
 
             if (arrayIndex >= soilwater.SW.Length)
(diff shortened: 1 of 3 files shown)


Models/Functions/SoilWaterScale.cs
```diff
@@ -10,33 +11,77 @@ namespace Models.Functions
...
+    [Description("A simple scale to convert soil water content into a value between 0 and 2 where 0 = LL/LL15, 1 = DUL and 2 = SAT")]
     public class SoilWaterScale : Model, IFunction
     {
-        [Link]
-        IPhysical physical = null;
+
+        /// <summary> </summary>
+        [Description("SAT array internal variable: ")]
+        [Display(Type = DisplayType.None)]
+        public string SATModel { get; set; } = "[Physical].SAT";
```
hol353, line 23: Why do we need to let the user specify a SATModel? I can't think of a use case.

Models/Functions/SoilWaterScale.cs
```diff
@@ -10,33 +11,77 @@ namespace Models.Functions
...
+
+        /// <summary> </summary>
+        [Description("DUL array internal variable: ")]
+        [Display(Type = DisplayType.None)]
+        public string DULModel { get; set; } = "[Physical].DUL";
```
hol353, line 28: As above, why do we need this?

Models/Functions/SoilWaterScale.cs
```diff
@@ -10,33 +11,77 @@ namespace Models.Functions
...
 
-            double sws;
+            if (arrayIndex >= soilwater.SW.Length)
+                throw new Exception($"Soil Water Scale ({Parent.Name}.{Name}): ArrayIndex {arrayIndex} is more than SoilWater SW length {soilwater.SW.Length}.");
+
+            if (sats == null && duls == null && lls == null) 
+            {
+                sats = Locator.Get(SATModel) as double[];
+                duls = Locator.Get(DULModel) as double[];
+                lls = Locator.Get(LLModel) as double[];
```
hol353, line 58: Locator.Get is slow at runtime. Move the call to StartOfSimulation. No need to do it everyday of the sim.

Models/Functions/SoilWaterScale.cs
```diff
@@ -10,33 +11,77 @@ namespace Models.Functions
...
+
+        /// <summary> </summary>
+        [Description("DUL array internal variable: ")]
+        [Display(Type = DisplayType.None)]
+        public string DULModel { get; set; } = "[Physical].DUL";
+
+        /// <summary> </summary>
+        [Description("Lower Limit array internal variable: ")]
+        [Display(Type = DisplayType.None)]
+        public string LLModel { get; set; } = "[Physical].LL15";
```
hol353, line 33: This one makes sense. The use case is the user wants to specify one of airdry, ll15, crop ll. Rather than a string I wonder about using an enum instead. That way the user gets a drop down rather than free form text.

Models/PMF/Organs/Root.cs
```diff
@@ -180,8 +183,6 @@ public class Root : Model, IWaterNitrogenUptake, IArbitration, IOrgan, IOrganDam
         /// <summary>The N supply for reallocation</summary>
         private double nReallocationSupply = 0.0;
 
-        private SoilCrop soilCrop;
```
hol353, line 183: Why has soilCrop been made a public property of Root?

hol353 (changes requested):

Models/PMF/Organs/Root.cs
par456: That was to avoid copying the name matching code over to SoilWaterScale, as Root already has a reference to the SoilCrop.

If there is another place a reference to SoilCrop is available, I'm happy to use that, I just don't want to do the same string name matching again in SoilWaterScale.

Models/Functions/SoilWaterScale.cs
par456: I like that idea better, it'll make it easier to use, will change to enum.

Models/PMF/Organs/Root.cs
hol353: Ok that makes sense. Leave it as it is.

par456:
Should be right to merge now.

hol353 (approved):

PR #8841, opened 2024-04-11
8840 Soil Import error fix

### LINKED ISSUE #8840: Soil Import Error
### What happened?

Related to #8773

Previous fix only handled if the array was smaller, not if it was larger. Need to change the sign on the check to do both.

Windows

par456:
Resolves #8840


par456:
retest this please jenkins

PR #8844, opened 2024-04-12
Simple water uptake KL damage function

### LINKED ISSUE #8843: We need a pest/disease plant water update damage function
### Describe the new feature

This could be done simply by modifying KL.

hol353:
Resolves #8843


ric394:
Are you able to write tests for this one @hol353?

hol353:
Yes, good question. I was thinking about unit tests as well. The only bit of new code is:

```c#
        [EventSubscribe("StartOfSimulation")]
        private void OnStartOfSimulation(object sender, EventArgs e)
        {
            if (SoilCrop == null)
                throw new Exception($"No soil/crop specified in {Name}");
            SoilCrop = Soil.FindDescendant<SoilCrop>(SoilCrop.Name);
            initialKL = SoilCrop.KL.Clone() as double[];
        }

        [EventSubscribe("DoPestDiseaseDamage")]
        private void DoPestDiseaseDamage(object sender, EventArgs e)
        {
            SoilCrop.KL = MathUtilities.Multiply_Value(initialKL, reduction.Value());
        }
```

All I'm doing is setting KL by using MathUtilities.Multiply_Value method, which is already tested. Not much else to test.

KeLiu7:
@hol353 @ric394 Does this also apply to soil acid, alkaline and saline limitations on KL and XF?

hol353:
Nope. This is purely to allow a pest or disease model to modify KL in whatever way they want.

PR #8855, opened 2024-04-18
Grain nitrogen concentrogen

byzheng:
working on #3605 

Test method to calculate minimum nitrogen concentration of grain in #3606. Some improvement is observed for Australian wheat trials. Test for other crops.

### DIFF
--- a/Models/PMF/Organs/ReproductiveOrgan.cs
+++ b/Models/PMF/Organs/ReproductiveOrgan.cs
@@ -369,7 +369,25 @@ namespace Models.PMF.Organs
         {
             get
             {
-                return MinimumNConc.Value();
+                // Biomass allocation today
+                double dmAllocation = potentialDMAllocation.Storage +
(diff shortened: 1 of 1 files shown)


BrianCollinss:
This might be irrelevant to this issue. Currently accepted stats are indicators of model performance using a combination of parameters (some being genotypic, e.g., phenology) that were chosen and/or calibrated based on some observations. When there is a major change (like this) in the way the model handles biomass and N flow and deposition, don't we need to recalibrate the other parts of the models before we decide that any improvement observed is not just model behaving randomly? In other words, if for example biomass was simulated reasonably with the model before, and now we are making this change, should we also make some changes in other parameters that govern biomass components?

HamishBrownPFR:
@byzheng can this PR be closed as it has been sitting without resolution for some time and seems to be a look-see test.  We need to reduce the number of pull requests that sit open in the system

byzheng:
yes please close it.

PR #8880, opened 2024-04-29
Convert NUptake to positive values.

### LINKED ISSUE #8879: Root NUptake needs converting to positive
### What happened?

Following what is done for NUptakeLayered.

https://github.com/APSIMInitiative/ApsimX/blob/a8785f001daa3700b6078a7d71dbfb4c976f617d/Models/PMF/Organs/Root.cs#L365-L391

Windows

BrianCollinss:
Resolves #8879

Following what is done for NUptakeLayered.

https://github.com/APSIMInitiative/ApsimX/blob/a8785f001daa3700b6078a7d71dbfb4c976f617d/Models/PMF/Organs/Root.cs#L365-L391

### DIFF
--- a/Models/PMF/BiomassArbitration/RootNetwork.cs
+++ b/Models/PMF/BiomassArbitration/RootNetwork.cs
@@ -233,7 +233,7 @@ namespace Models.PMF
                 double uptake = 0;
                 foreach (NetworkZoneState zone in Zones)
                     uptake += MathUtilities.Sum(zone.NitUptake);
-                return uptake;
+                return (-uptake); // convert to positive values.
             }
         }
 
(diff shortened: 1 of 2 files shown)


hol353:
This PR makes me nervous. I would love to be able to merge this to make NUpdate consistent with NUptakeLayered but NUptake is used in some manager scripts (e.g. Validation\Chicory) and is reported a lot. Changing the sign of the values may well break the manager scripts. 

I would also like to go one step further and combine NUptake and NUptakeLayered into a single property called NUptake but this will require a converter.

BrianCollinss:
Thanks @hol353. Do you recommend I close the PR?

HamishBrownPFR:
This one makes me nervous too as I have python scripts that graph these outputs.  Would be good to tidy this up in full as you suggest @hol353

PR #8898, opened 2024-05-02
Row filters now get saved appropriately

### LINKED ISSUE #8897: Row filter keeps coming back after being deleted and file saved
### What happened?
The only way to remove it is to close and reopen the file.

https://github.com/APSIMInitiative/ApsimX/assets/39573622/4fc05029-1f0e-4dea-9913-ab92416123e0

Windows

ric394:
resolves #8897 

# Problem
- If a user clears the report filter field without pressing enter or clicking off the field, the temporarily saved report filters are not overwritten with "". This makes the deleted text reappear in the field.

# Fix
- Link the saving of report Filters to the reportFiltersFields changed event rather than leave event.

### DIFF
--- a/ApsimNG/Presenters/DataStorePresenter.cs
+++ b/ApsimNG/Presenters/DataStorePresenter.cs
@@ -1,13 +1,13 @@
-﻿using UserInterface.EventArguments;
-using Models.Core;
-using Models.Factorial;
-using Models.Storage;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


BrianCollinss:
Hi @ric394. Just wanted to let you know that I had similar problem even when I pressed Enter or clicked on another node after clearing the field.

ric394:
> Hi @ric394. Just wanted to let you know that I had similar problem even when I pressed Enter or clicked on another node after clearing the field.

@par456 had a similar problem while trying to replicate the issue. This fix should resolve that issue.

PR #8900, opened 2024-05-03
Fix invisible manager script

BigElRicho:
working on #8476 

- when copy/pasting a manager node with erroneous script it's OnCreated() method would attempt to compile the script, interrupting the adding process. This would mean the explorerView's tree wouldn't be updated but the manager node would still be added regardless.


BigElRicho:
Do not merge, causes issue with DataStore.

hol353:
add to whilelist jenkins

PR #8925, opened 2024-05-14
Fix manager overrides bug + add test.

### LINKED ISSUE #8911: --apply does not change manager parameters
### What happened?

I am unsure whether this is a bug.

In the following config commands, the first line works fine and replaces the met file name. But the second line does not change the manager script parameter, with or without quotation marks.

```
[Weather].FileName = Met/AU-Emerald.met
[Sowing].Script.SowDate = 15-Jan   ---> OR ("15-Jan")
```

Windows

ric394:
resolves #8911

## Notes
- The manager script was not updating its parameters after they were modified.
- This meant that when the file was being saved, the changes were not saved.
- A conditional statement was added to update the parameters of any manager scripts being overridden.
- A test was added to ensure this always occurs.


BrianCollinss:
I just ran the example I had shared with you and the SowDate is still 15-Aug instead of 15-Jan or 15-Sep.

image

BrianCollinss:
It's working on my example sims. Ta

hol353:
This branch is now conflicted.

BrianCollinss:
On HPC, sims failed as APSIM --apply adds "" around the SowDate:

image

PR #8934, opened 2024-05-16
8759 in memory db for cli

### LINKED ISSUE #8759: Use In Memory DB to reduce I/O on hard drive
### Describe the new feature

To run APSIMNG simulations in HPC with large number of CPUs, frequently writing report into db file is a bottleneck.

For example, I run simulations in HPC (petrichor) with 1800 cores. All data are red and wrote from high speed hard drive (i.e. scratch3). An apsimx file contains 500 single year simulations with reporting at harvest. 

A single apsimx only takes 60 s with less than 500 cores, but gradually increases to 250 s with 1800 cores. See figure below for runtimes of 12K apsimx files. 

image

This is caused by frequently I/O (i.e. write report to db file) from multiple cores. I/O will be minimised if we could use in memory DB and export report into `csv` file at the end of simulations. 

I noticed `UseInMemoryDB` is defined in the source codes and used in the unittest.

https://github.com/APSIMInitiative/ApsimX/blob/f3c769263ca33910ebb0114ba83e762fcc1b1cef/Models/Storage/DataStore.cs#L40-L53

Do we already have the ability to use in memory DB in command line (i.e. `Models`)? If not, could we implement it?

ric394:
resolves #8759 

# Notes
- The `--in-memory-db` switch is now available.
- Works with standard Model calls and when using `--apply`.
- Still generates a .db file however, it is not filled with any data.
- Documentation updated to show usage at: https://apsimnextgeneration.netlify.app/usage/commandlineusage/

# Call examples
- `Models.exe example.apsim --in-memory-db`
- `Models.exe --apply commands.txt --in-memory-db`


Models/Main.cs
par456, line 1: Remove the ChangeSimToUseInMemoryDB and CreateRunnerForInMemoryDBOption functions from here and just put the code in where they are used. Doesn't make sense to make new function that are only used in one place.

Put some comments lines into CreateRunnerForInMemoryDBOption if you want to make it clearer what each loop does once that other code is put back.

PR #8944, opened 2024-05-19
8856 Tree Proxy Revert and Fix

### LINKED ISSUE #8856: TreeeProxy temporal data entry not possible
### What happened?

The TreeProxy of the Agroforestry TreeBeltExample is not allowing data overwriting for Temporal Data by manual typing or copy-paste from elsewhere. I'd apprecaite this fixed ASAP, as I need this working to help colleagues in Vietnam with their simulations. Copying and pasting a TreeProxy node from a previous simulation that worked several months ago, also does not work. 
Here is the example freshly downloaded frrom the GUI with the latest release (7452):
[TreeBeltExample5.zip](https://github.com/APSIMInitiative/ApsimX/files/15023851/TreeBeltExample5.zip)

Windows

par456:
Resolves #8856

I accidently flipped the temporal and spatial displays in the last fix, so I've reverted that here. I also checked that the provided simulation was able to edited and ran and it seems to be correct now.

Hoping this is the final fix this needs. We might need to look at a new structure for the TreeProxy data down the line as it doesn't fit into our GridTable system very well and needs a lot of converting work at the moment to make it work. That's where a lot of these bugs are coming from.

### DIFF
--- a/ApsimNG/Presenters/TreeProxyPresenter.cs
+++ b/ApsimNG/Presenters/TreeProxyPresenter.cs
@@ -1,13 +1,7 @@
 ﻿using System.Collections.Generic;
-using System.Data;
-using System.Linq;
 using Models.Agroforestry;
-using Models.Core;
 using Models.Soils;
 using UserInterface.Views;
-using UserInterface.Commands;
-using UserInterface.EventArguments;
-using APSIM.Shared.Utilities;
 using Models.Utilities;
 
 namespace UserInterface.Presenters
--- a/ApsimNG/Presenters/TreeProxyPresenter.cs
+++ b/ApsimNG/Presenters/TreeProxyPresenter.cs
@@ -67,12 +67,12 @@ namespace UserInterface.Presenters
             List<GridTable> tables = forestryModel.Tables;
 
             spatialGridPresenter = new GridPresenter();
-            spatialGridPresenter.Attach(tables[0], forestryViewer.SpatialDataGrid, explorerPresenter);
+            spatialGridPresenter.Attach(tables[1], forestryViewer.SpatialDataGrid, explorerPresenter);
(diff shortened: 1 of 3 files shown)


ric394 (approved):

sme016:
Sorry, but I still have problems editing the data inputs to the TreeProxy. Instead I have resorted to using NotePad++ to make the edits for spatial data, which I don't really like doing due to the risk of inadvertant mistakes.
 
Temporal data: Funny things happen while I add new dates (extra lines for a while), but on saving it seems ok.

Spatial data: edits to shading or root length density numbers are impossible to complete, as too many strange things happen, e.g. values not retained after typing in and saving.

I note what you say about the GridTable system. So that overhaul is badly needed, I suspect. We have 3 PhD students waiting to use this (two in Europe working on African projects, and one in Vietnam). So prioritising it would be appreciated. 

Thanks for your efforts on this so far.

par456:
I've not been able to replicate your bugs with the file I have here. Can you give me your apsim file again, and maybe the input you are trying to put into each table?

I was able to edit existing dates and add new dates to the temporal table without extra blank lines appearing and could added shade and root depth to the spatial.

sme016:
Initially too for me this morning it seemed to work, as I edited one cell at a time, but then I tried altering depths and copying blocks of data, which landed me in more trouble. 
Attached is that simulation and a pdf explanation of what I did.
 
[TreeBeltExampleEditingIssue.zip](https://github.com/APSIMInitiative/ApsimX/files/15449395/TreeBeltExampleEditingIssue.zip)

par456:
Thanks, I didn't think to try copy-paste in there and yep, I can see all sorts of weirdness now. Will have a look and see what's going wrong.

PR #8943, opened 2024-05-17
8938 paste valid parent bug

ric394:
resolves #8938

PR #8946, opened 2024-05-20
Context menu allows pasting any model bug fix

ric394:
resolves #8943

# Notes
- Extra checks have been added to prevent this bug from occurring.
- `Models.cs` was updated to allow models with no `[ValidParent()]` attributes to be handled properly.
- Tests updated to work with new functionality.
- `Nutrient`'s  had NutrientPatchManager added as valid parent.

### DIFF
--- a/Models/Core/ApsimFile/Structure.cs
+++ b/Models/Core/ApsimFile/Structure.cs
@@ -30,12 +30,13 @@ namespace Models.Core.ApsimFile
             // Ensure the model name is valid.
             EnsureNameIsUnique(modelToAdd);
 
-            parent.Children.Add(modelToAdd);
-
-            // Call OnCreated
+            if(parent.IsChildAllowable(modelToAdd.GetType()))
+            {
+                parent.Children.Add(modelToAdd);
+            }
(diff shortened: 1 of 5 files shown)


Models/Core/ApsimFile/Structure.cs
```diff
@@ -30,12 +30,13 @@ public static IModel Add(IModel modelToAdd, IModel parent)
...
-
-            // Call OnCreated
+            if(parent.IsChildAllowable(modelToAdd.GetType()))
+            {
+                parent.Children.Add(modelToAdd);
+            }
+            else throw new ArgumentException($"A {modelToAdd.GetType().Name} cannot be added to a {parent.GetType().Name}.");
+                       
             modelToAdd.OnCreated();
-            foreach (IModel model in modelToAdd.FindAllDescendants().ToList())
```
par456, line 37: Why did this get deleted? This is making that if a model creates children, that the children have their OnCreated called?

ric394: removed by mistaken. These lines have now been readded.

ric394:
@hol353 this looks like POStats has run out of space. Are you able to assist?

par456:
retest this please jenkins

PR #8953, opened 2024-05-21
Nutrient patches: Fix solute initialisation. Fix residue decomposition.

hol353:
Working on #5327


Models/Soils/Solute.cs
par456, line 1: Just checking that there is a reason you have to convert to ppm to do the MapConcentration?

hol353: MapConcentration only works on solute concentrations which is ppm - hence the conversion. I then convert it back to kg/ha if that is how the user specified it.

PR #8955, opened 2024-05-21
Fixed clay silt sand defaults

### LINKED ISSUE #8757: Add in default/assumed values to the soil node

### LINKED ISSUE #8920: SoilTemperature is throwing an error
### What happened?

[LauristonErrorST.zip](https://github.com/APSIMInitiative/ApsimX/files/15289265/LauristonErrorST.zip)

Windows

hol353:
Resolves #8757
Resolves #8920

### DIFF
--- a/Models/Soils/Physical.cs
+++ b/Models/Soils/Physical.cs
@@ -272,19 +272,51 @@ namespace Models.Soils
             if (KS != null && KS.Length > 0)
                 KS = MathUtilities.FillMissingValues(KS, Thickness.Length, 0.0);
 
+            ParticleSizeClay ??= Enumerable.Repeat(double.NaN, Thickness.Length).ToArray();
+            ParticleSizeSilt ??= Enumerable.Repeat(double.NaN, Thickness.Length).ToArray();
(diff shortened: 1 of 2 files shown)


hol353:
Stats are all better in SoilTemperature validation because of the clay/silt typo.

PR #8964, opened 2024-05-23
8961 soil download weather bug

### LINKED ISSUE #8961: When there is no Weather object, "Download Soil" does not work
### What happened?

In case it is not intended.

image

Windows

ric394:
resolves #8961

# Notes
- No longer shows `object reference is not an instance of object` error.
- Shows suggestion to add weather node when one is missing when attempting to download a soil.
- added settings to vscode `launch.json` and `tasks.json` files.
    - This removes the loading modules outputs and makes output a lot cleaner in clean, build and run command runs.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,14 +13,6 @@
             "cwd": "${workspaceFolder}",
             "console": "internalConsole",
             "stopAtEntry": false,
-            "args":[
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
par456, line 1: Leave the changes to launch.json out of this one.

PR #8967, opened 2024-05-23
Add PAWC to Physical grid

### LINKED ISSUE #8962: LL15-based AWC
### What happened?

Can we have TAWC (total available water content) in Physical as `DUL - LL15`? Would be helpful to quickly assess soil texture and water holding capacity. 

image

If we don't have this already somewhere in the code, what does PAWC refer to in the "Download Soil" window?

image

Windows

ric394:
resolves #8962 

@hol353 Is this acceptable?

# Notes
- Allowed PAWC value to be displayed in Physical grid.

### DIFF
--- a/Models/Soils/Physical.cs
+++ b/Models/Soils/Physical.cs
@@ -159,7 +159,9 @@ namespace Models.Soils
         public double[] KS { get; set; }
 
         /// <summary>Plant available water CAPACITY (DUL-LL15).</summary>
+        [Summary]
         [Units("mm/mm")]
+        [Display(Format = "N2")]
         public double[] PAWC { get { return APSoilUtilities.CalcPAWC(Thickness, LL15, DUL, null); } }
 
         /// <summary>Plant available water CAPACITY (DUL-LL15).</summary>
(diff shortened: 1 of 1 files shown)


BrianCollinss:
I wonder if it should be called PAWC or TAWC (Total Available Water Content) or TWHC (Total Water Holding Capacity). Irrigation engineers use different terms. If PAWC is "Plant Available ...", it is incorrect. If it is "Potential Available ..." it is kind of correct.

hol353:
No, we don't want a PAWC column in the physical model. We've spent > 20 years telling users to think of plant available water as DUL - CLL. 

I also don't like the term Total Available Water Content (TAWC). Available for what? The plant models in APSIM define available water as DUL - CLL.

BrianCollinss:
I think I explained that the best term is Total Watering Holding Capacity, which is what irrigation/drainage engineers call it.

zur003:
@BrianCollinss - I must disagree. The term "Total Water Holding Capacity" to me implies all water held in the soil, whereas the critical value for APSIM purposes is the amount of water which can be extracted by plants. Admittedly this value will depend on the kind of plant, but LL15 servers as a reasonably good "generic" value for what, from a plant's perspective, is a completely "dry" soil.

BrianCollinss:
From an irrigation/drainage engineer, this is what I learned in school:

TWHC is defined as FC - PWP (permanent wilting point, usually measured at -15 bar). 

There is also (Total) Readily Available Water (RAW or TRAW) which is somewhat similar to PAWC and is defined as FC - WP (wilting point, which is the level of moisture at which the plant starts feeling stress). 

AirDry is something that is mostly used in soil evaporation and temperature equations. So, I don't remember it used in the context of crop water requirement.

hol353:
We aren't irrigation / drainage engineers. We are modellers that have a long history (> 30 years) of using the term PAWC. We aren't about to change that now.

BrianCollinss:
No one asked you to change that. The whole request was based on the fact that when user downloads the soil from APSIM GUI, you show a PAWC which is not clear whose plant it belongs to, therefore I suggested another column be added to soil physical that can be used regardless of the crop. 

The only people who talk like this are those who love to stick to their religious beliefs no matter what. Has AI turned into a cult or something?!

PR #8978, opened 2024-05-25
Fix "Fertilise when variable reaches threshold" Manager

### LINKED ISSUE #8977: "Fertilise when variable reaches threshold" of "Management Toolbox" malfunctions
### What happened?

If the target variable has a small magnitude, this manager might be triggered multiple times.

image

image

Windows

BrianCollinss:
Resolves #8977


ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
```diff
@@ -1397,30 +1397,43 @@
...
+            "\t            if (obj is double)",
+            "\t                value = (double) obj;",
+            "\t            else if (obj is IFunction f)",
+            "\t                value = f.Value();",
+            "\t            else",
+            "\t                throw new Exception($\"Unknown type of variable: {VariableName}\");\t                \t           ",
+            "\t                 ",
+            "\t            if (!wasTriggered && MathUtilities.IsGreaterThanOrEqual(value, TriggerValue)) {",
+            "\t                FertiliseDate = clock.Today.AddDays(NumDaysToWait);",
+            "\t                wasTriggered = true;                ",
```
hol353, line 1430: The logic of this looks wrong e.g. this if statement will only ever be true once per sowing.

ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
```diff
@@ -1397,30 +1397,43 @@
...
+            "\t            }",
+            "",
+            "\t            if (!wasApplied && clock.Today == FertiliseDate) {",
+            "\t            \tfertiliser.Apply(Amount: Amount, Type: FertiliserType);\t",
+            "\t            \twasApplied = true;",
+            "\t            }\t                          ",
```
hol353, line 1436: Should this if statement use wasTriggered. The previous if statement is no longer connected to this if statement.

ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
```diff
@@ -1513,6 +1526,10 @@
             "}"
           ],
           "Parameters": [
+            {
+              "Key": "EventName",
+              "Value": "[Wheat].Phenology.PlantEmerged"
+            },
```
hol353, line 1532: Where did this come from? This isn't in the tip revision on origin, nor is it used by the script.

hol353 (changes requested):

hol353:
I think a simpler approach to solve the problem is to put an else statement on the first if statement that has something like:

```
FertiliseDate = DateTime.MinValue;
```

ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
BrianCollinss: Isn't it the intention? The original IF statement would check if the value is within 0.1 of the target assuming that would happen once.

ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
BrianCollinss: No idea.

BrianCollinss:
Happy to leave this to someone more knowledgeable.

Issue #8917, opened 2024-05-12
Sorghum model senescence defect


hol353:
### What happened?

The attached simulation throws the exception:

```
Paddock.Sorghum.Leaf:                             Emergence
                                                  LAI = 0.00 (m^2/m^2)
Paddock.Sorghum.AboveGround:                      AboveGround = 1.80 (g/m^2)
.Sorghum_DEWS-30778_Northern_Soil_72_Pop9_Buster: ERROR in file: C:\Work\Temp\Heidi\Sorghum_Pop9_30778.apsimx
Simulation name: Sorghum_DEWS-30778_Northern_Soil_72_Pop9_Buster
Models.Core.SimulationException
 ---> System.Exception: Attempted to senesce more N than exists on leaf 'Leaf'
   at Models.PMF.Organs.SorghumLeaf.ApplySenescence() in C:\Work\Repos\APSIMInitiative\ApsimX\Models\PMF\Organs\SorghumLeaf.cs:line 897
   at Models.PMF.Organs.SorghumLeaf.OnDoActualPlantGrowth(Object sender, EventArgs e) in C:\Work\Repos\APSIMInitiative\ApsimX\Models\PMF\Organs\SorghumLeaf.cs:line 1102
   at Models.Clock.OnDoCommence(Object _, CommenceArgs e) in C:\Work\Repos\APSIMInitiative\ApsimX\Models\Clock.cs:line 417
   at Models.Core.Simulation.Run(CancellationTokenSource cancelToken) in C:\Work\Repos\APSIMInitiative\ApsimX\Models\Core\Simulation.cs:line 292
   --- End of inner exception stack trace ---
```

135.8 kg/ha of NO3N was added at sowing. The sorghum model is trying to senesce all leaves at emergence so my guess it is severley water stresed.

[Sorghum_Pop9_30778.zip](https://github.com/APSIMInitiative/ApsimX/files/15287949/Sorghum_Pop9_30778.zip)

### What Operating System are you on?

Windows

hol353:
@peter-devoil - can you look at this?

PR #8985, opened 2024-05-27
Avoid double accounting of N in senescence

peter-devoil:
working on #8917


Models/PMF/Organs/SorghumLeaf.cs
```diff
@@ -891,13 +891,13 @@ private void ApplySenescence()
             if (!MathUtilities.IsPositive(dltSenescedBiomass)) return;
 
             double slnToday = MathUtilities.Divide(Live.N, laiToday, 0.0);
-            DltSenescedN += DltSenescedLai * Math.Max(slnToday, 0.0);
+            double dltSenescedN = DltSenescedLai * Math.Max(slnToday, 0.0);
```
jbrider, line 894: From memory, DltSenescedN is used as a reporting/tracking variable. It can also be senesced when providing N to another organ and is therefore cumulative.  I should be confident that we weren't double counting it as it mirrored the outputs from Classic that were much easier to follow - but it's been a while since I was in that code. 
The outcome here shouldn't be an exception though - it should set the senescence to the live amount and then kill the plant - normally the reduction in LAI would catch this, but I'm guessing that the amounts that we're talking about are very small. There is an issue with how the LAI checks work due to frost removing LAI but not killing the plant completely. There's probably also a case to be made about going below the structural minimum %?

par456:
retest this please jenkins

hol353:
@jbrider @peter-devoil The stats have changed for this one - mostly improved. Are you happy for this to be merged?

peter-devoil:
I'll work it through with jason. From what I can tell, the offender is being called twice/day when it should be once.

peter-devoil:
@jbrider, I've had another look at this and suspect it is double accounting. The "double calling" isn't happening.

It is set first in [ProvideNThroughSenescence()](https://github.com/APSIMInitiative/ApsimX/blob/9751d8ad1fd60ab844d30153622b902c1e8ddc19/Models/PMF/Organs/SorghumLeaf.cs#L841) which calculates the N lost in senesced material in differing ways pre/post flowering

It's again incremented in [ApplySenescence()](https://github.com/APSIMInitiative/ApsimX/blob/9751d8ad1fd60ab844d30153622b902c1e8ddc19/Models/PMF/Organs/SorghumLeaf.cs#L894), which is using the same DltSenescedLai calculated in [ProvideNThroughSenescence() earlier](https://github.com/APSIMInitiative/ApsimX/blob/9751d8ad1fd60ab844d30153622b902c1e8ddc19/Models/PMF/Organs/SorghumLeaf.cs#L839)

I think the test, and second increment is redundant.

peter-devoil:
undo

HamishBrownPFR:
Hi @peter-devoil.  Where are you at with this PR?  Looking to reduce the number of unresolved PR that sit in the system as it has become very cluttered.  Can this one be closed and reopened when you have time to get back to it

peter-devoil:
Abandoned

PR #8987, opened 2024-05-28
Oat model validation

ekhaembah:
Working on #8503 
Re-submitting a pull request after adding new data and cultivars

PR #9003, opened 2024-06-04
Oat Model validation

ekhaembah:
Working on #8987
Re-submitting after changes


ekhaembah:
@hol353 - I've addressed the issue of zero values assigned to non-measurements. Grain number plateaued after StartGrainFill. In discussion with @HamishBrownPFR, the decision was that the final grain number is more important than the change during oat development.

PR #9007, opened 2024-06-05
PseudoPatches progress

sno036:
working on #5327 

merge this please


hol353:
@par456 I can't really merge this as the .cs files are mine. The difference in stats are all improvements.

Models/AgPasture/SimpleGrazing.cs
```diff
@@ -798,7 +774,9 @@ public void RemoveDMFromPlants(double residual, double[] speciesCutProportions)
             /// <summary>Return a value from an array that can have either 1 yearly value or 12 monthly values.</summary>
             private static double GetValueFromMonthArray(double[] arr, int month)
             {
-                if (arr.Length == 1)
+                if (arr == null)
+                    return double.NaN;
```
par456, line 778: Should this throw an exception rather than introduce NaNs into the system? I note that the methods using what this returns just assumes it's getting valid numbers back.

Models/AgPasture/UrineDungPatches.cs
```diff
@@ -0,0 +1,298 @@
...
+            AmountUrineNReturned = HarvestedN * 0.50;  // 
+            AmountDungNReturned = HarvestedN * 0.35;  // 
+            AmountDungCReturned = AmountDungNReturned * 20;
+        }
+        else if (UrineReturnType == UrineReturnTypes.SetMonthly)
+        {
+            AmountUrineNReturned = MonthlyUrineNAmt[clock.Today.Month - 1];   //  hardcoded as an input
+            AmountDungNReturned = AmountUrineNReturned / 0.50 * 0.35;  // 
+            AmountDungCReturned = AmountDungNReturned * 20;
+        } */
```
par456, line 89: Just checking you wanted to leave this commented code in.

Models/AgPasture/SimpleGrazing.cs
hol353: No we don't want an exception here because code later will look for it and do a default calculation.

Models/AgPasture/UrineDungPatches.cs
hol353: Yep I need that commented for now so that in October (our next sprint) I'll remember where I got to. It will ultimately be deleted.

PR #9008, opened 2024-06-05
Integrate latest DCaPST and Dynamic Tillering changes.

### LINKED ISSUE #7705: Update DCaPST Next Gen from Classic.

joesaddigh:
This PR addresses some fixes that were made to the Dynamic Tillering routines. It also integrates DCaPST with some tweaks to parameters, as per Alex Wu's instructions.

Working on #7705


hol353:
I notice in the stats that the sorghum model performance has changed. Does this mean that DCaPST is turned on by default for sorghum simulations? Does this slow down the runtime speed of sorghum?

The stats also show that the number of observations for sorghum has changed, in some cases 'n' has reduced.

There are also a bunch of files in this PR that probably aren't needed. Under DCaPST there is a Sorghum and Wheat directory. Are these needed?

There is also an 'old' directory under Tests/Sorghum.

joesaddigh:
This PR has been superseded

PR #9010, opened 2024-06-06
9009 Rename Checks

### LINKED ISSUE #9009: Name clash between model name and properties
### What happened?

I have found that you cannot name a model in the tree with any of the names of properties in Model or the parent Model type.
For example, I had a model in the tree structure called "Children" which leads to an odd error for the user and causes strange denial of adding further child items. This could be an issue for common names like "soil" etc.
This is because Locator.GetInternal() checks for Methods and Properties before checking if a Model. The order may need to be reconsidered or expect Method and Properties as the last part of the object path array. "Children" was not the last item in the path.
To trap this for future users APSIM either needs smarter checking to determine object or deny renaming to Parent or Model properties which not as desirable.
In debugging this issues I was getting other errors when trying to rename tree view items (os does not support serialisation) that I have seen lately but before I was using "children". The error is thrown and the rename does not take place. The error went away with a restart of APSIM.
This could be related to the finding that often the first time you click on a treeview item to rename it, make changes and hit return, it reverts to the previous name, and you need to try again. Moving the cursor around a bit seems to help.

Windows

par456:
Resolves #9009 

One bug I found when checking out this issue, there may be more.

### DIFF
--- a/ApsimNG/Views/TreeView.cs
+++ b/ApsimNG/Views/TreeView.cs
@@ -252,6 +252,28 @@ namespace UserInterface.Views
             }
         }
 
+        /// <summary>Edit the text of the current node</summary>
+        public void EndRenamingCurrentNode(string newText)
+        {
+            if (isEdittingNodeLabel == true)
+            {
+                textRender.Editable = false;
+                // TreeView.ContextMenuStrip = this.PopupMenu;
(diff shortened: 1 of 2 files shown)


par456:
retest this please jenkins

par456:
retest this please jenkins

ric394 (approved):

PR #9016, opened 2024-06-07
8407 batch cli switch

### LINKED ISSUE #8407: Improved batch running
### Describe the new feature

@hol353 set me up with the current system to run apsimx files in batches. Currently it is more designed for where there is a single experiment in the apsimx file - that that works well where there is a complete/balanced design in the experiment. When there are some soils with some locations and note others (for example) and a series of simulations within each valid soil-weather combination then the current design doesn't work well. I got around this with a script/instruction file as below with Dean's assistance:

```
load BaseCl.apsimx
[Soil]=SoilLibrary.apsimx;[Ahiaruhe_1a1]
[Weather].FileName=16864.met
[SimulationExp].Name=Sim0001
run

load BaseCl.apsimx
[Soil]=SoilLibrary.apsimx;[Ahuriri_7a1]
[Weather].FileName=19479.met
[SimulationExp].Name=Sim0002
run

load BaseCl.apsimx
[Soil]=SoilLibrary.apsimx;[Ailsa_5a1]
[Weather].FileName=19479.met
[SimulationExp].Name=Sim0003
run
```

It works ..., but is uninformative about progress (I can see the DB file size increasing is all) and is (probably) very inefficient. 

What would be great would be the ability to:
- describe the simulation variants in some form of template
- have that template insert combinations from a text file (e.g., taking the above there would be three lines specifying soils and weather files)
- have the Experiment node automatically (or use the template to assign names) rename itself
- and have something to show progress through what could be tens-hundreds of thousands of simulations

I hope this makes sense!

ric394:
resolves #8407

# Notes

As described in #8407, the batch running feature has been implemented.
There is one small change to how this is run.

Examples command are: 
`Models.exe --apply configFile.txt --batch batchFile.csv`
`Models.exe Wheat.apsimx --apply configFile.txt --batch batchFile.csv`

This feature will improve large scale runs where a variation on a base simulation are required. 
An example would be where you're changing the weather and soil models of a simulation across Australia.


ric394:
retest this please jenkins

Issue #7623, opened 2022-11-14
Migrate wheat model to simple leaf


HamishBrownPFR:
The existing leaf model is to complex leading to structural and parameter uncertainties and some unstable and unexpected behavior.

PR #9024, opened 2024-06-14
Wheat simple arbitrator

HamishBrownPFR:
working on #7623

Working on updated wheat model with simple leaf and simpler arbitrator models.  
Located in Prototypes\wheatSimpleLeaf.  
There are two .apsim files.  The one called WheatFewer.apsimx has the most advanced set up but excludes simulations that don't have biomass data.


HamishBrownPFR:
Too much stuff here.  I will break it up into smaller pull requests

PR #9026, opened 2024-06-16
Replace BinaryFormatter by using DeepCloner for cloning and JSON for …

### LINKED ISSUE #8680: Replacing BinaryFormatter
As highlighted in #8446 BinaryFormatter that we use for serialization, [is going to be deprecated this year](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md)  along with .net 6.0, so when we move .net frameworks next, this key part of APSIM will no longer work.

From what I've see looking at the uses of serialization, we use it for server socket communication and for deep cloning APSIM objects. For the server socket communication, we could probably get away with using JSON serialization as I don't think it needs to be hyper performant (correct me if I'm wrong).

For deep cloning APSIM objects however, JSON serialization will be way too slow, since it's even slower than BinaryFormatter and we clone the apsim objects a fair bit.

So I'm opening up this issue to try and gather some notes and opinions on how we tackle this. I've looked at some serializers to see if we could do a like-for-like swap, but most will require at minimum, retagging our classes throughout the project and often they can't handle circular references like we frequently have.

Serializers I've looked at:
[Protobuf-net](https://github.com/protobuf-net/protobuf-net) - Fast serialization, requires full retagging
[Memorypack](https://github.com/Cysharp/MemoryPack) - Fast serialization, struggles to handle models, poor feedback
[MessagePack-CSharp](https://github.com/MessagePack-CSharp/MessagePack-CSharp) - Fast serialization, requires full retagging, has typeless mode but can't handle models like that

Retagging the project would be a crazy amount of work however, and most likely cause a lot of conflicts since it'll touch everything in models and will take some time to do. I would really prefer to avoid that if possible.

Having a specific deep copy function might be a better solution for that, however the same problem exists, the ones I've tried either are too old of date to be any use, or simply can't handle our model structure for a file.

DeepCopy Libraries I've looked at:
[DeepCloner](https://github.com/force-net/DeepCloner) - Doesn't work on .net 5 or later
[FastDeepCloner](https://github.com/AlenToma/FastDeepCloner) - Can't handle models, poor feedback

The last option we have for deep clone would be to try and create our own so we can actually deep copy models in a sensible way, but again, would be a pretty big project to make that happen.

zur003:
…server serialisation and deserialisation.

Resolves #8680

### DIFF
--- a/APSIM.Server/V1/Commands/ReadCommand.cs
+++ b/APSIM.Server/V1/Commands/ReadCommand.cs
@@ -1,4 +1,6 @@
 using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
 using System.Collections.Generic;
 using System.Data;
 using System.Diagnostics;
@@ -18,18 +20,24 @@ namespace APSIM.Server.Commands
         /// <summary>
         /// Name of the table from which parameters will be read.
         /// </summary>
-        public string TableName { get; private set; }
+        public string TableName { get; set; }
 
         /// <summary>
         /// Parameter names to be read.
         /// </summary>
-        public IEnumerable<string> Parameters { get; private set; }
+        public List<string> Parameters { get; set; }
 
         /// <summary>
         /// The result of the ReadCommand.
         /// Contains the data 
         /// </summary>
+        [JsonConverter(typeof(DataTableConverter))]
         public DataTable Result { get; set; }
+
+        /// <summary>
+        /// Parameterless constructor for serialization
+        /// </summary>
+        public ReadCommand() { }
 
         /// <summary>
         /// Creates a <see cref="RunCommand" /> instance with sensible defaults.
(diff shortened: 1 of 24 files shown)


hol353:
Thanks @zur003 for the PR. Is this ready to go? I'm amazed it ran green on the second commit.

zur003:
Well, I think it's ready to go, but it might be good if someone could do some testing of it with APSIM,.Server. I'm reasonably happy with the cloning side of things, but less so with the JSON serialisation/deserialisation it uses with the Server commands. I've never actually used the Server myself, so wasn't sure how to go about giving it thorough testing.

zur003:
Oh, and check run times. This approach to cloning should make things a little bit faster than what we've been doing, but it would be good to do a bit more benchmarking to be sure.

hol353:
@peter-devoil or @jbrider  Can you pull this branch and test it with APSIM.Server?

peter-devoil:
No problems.

It runs about 1% faster, with slightly less process memory usage on both V1server and zmqserver
image

PR #9027, opened 2024-06-17
physical models under an experiment no longer reference base sims' water

### LINKED ISSUE #9020: Unintended interaction between factor levels when using Soil.Physical in a factor
### What happened?

It is possible (and sometimes reasonable) to put a Soil.Physical model under Experimental Factors to have an experiment with soils of different water holding capacities.  
However, the GUI enforces a constant set of values for SW across all Soil.Physical models in a simulation.
In the attached file I have set up a simple example.  The screen shot shows SW values of 0.4 all the way down for the Physical model under the Light Factor which is reflecting the values set for SW in the Heavy factor.  If I change them to 0.2 all the way down here that will change them to 0.2 in the Physical model under the heavy factor.
image

[SoilEg.zip](https://github.com/user-attachments/files/15779298/SoilEg.zip)

Windows

ric394:
resolves #9020

FIndInScope strikes again!

This was just finding the first water in scope, this is not ideal for experiment configurations.

### DIFF
--- a/Models/Soils/Physical.cs
+++ b/Models/Soils/Physical.cs
@@ -206,7 +206,7 @@ namespace Models.Soils
         {
             get
             {
-                bool waterNodePresent = FindInScope<Water>() != null;
+                bool waterNodePresent = FindSibling<Water>() != null;
                 var columns = new List<GridTableColumn>();
 
                 columns.Add(new GridTableColumn("Depth", new VariableProperty(this, GetType().GetProperty("Depth"))));
--- a/Models/Soils/Physical.cs
+++ b/Models/Soils/Physical.cs
@@ -5,6 +5,7 @@ using APSIM.Shared.APSoil;
 using APSIM.Shared.Utilities;
 using Models.Core;
 using Models.Core.ApsimFile;
+using Models.Factorial;
 using Models.Interfaces;
 using Models.Utilities;
 using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


par456:
Could we make this a little bit smarter? 

It is possible physical is sitting by itself in a factor, so if it hasn't got a water sibling, maybe then check if it's in an experiment, and if so, find the simulation and get the water from that (finddescendant on the simulation node).

ric394:
> Could we make this a little bit smarter?
> 
> It is possible physical is sitting by itself in a factor, so if it hasn't got a water sibling, maybe then check if it's in an experiment, and if so, find the simulation and get the water from that (finddescendant on the simulation node).

If you go and find the simulation's water, when you change the SW in the Factor's Physical node, it will also change it any other Physical nodes. Which is the problem @HamishBrownPFR was having.

I think because Physical is dependent on Water, you have to include both when doing experiments, especially when you're making changes to soil water.

par456:
Yeah that's fine. In Hamish's case the find sibling check will work and it'll link to the water under the factor. But in the case that I just have a physical node as a factor, then it should be able to find the water node from the simulation to link with. Otherwise I'm just going to get another "object is null" error currently.

Aka, if the simulation looks like this:
image

ric394:
At the moment, when you click on either physical node under a factor, no object is null error occurs, but SW column is not shown. 

I can make it show the simulation SW column or leave it, but I think it's probably best to leave it untouched, otherwise it gives user's the wrong impression that making changes to SW cells in the factors' physical node actual 'replace' the SW figures.

par456:
That makes sense, we don't want the SW to be edited here if it would change a water somewhere else. I didn't think of that. 

However, we also lose our water graph on the right, which should remain visible.

Is there a way to keep the SW column hidden if water isn't under the factor, but still link it to the simulation water so we can see the graph?

hol353:
It used to be that if a user edited SW in physical then it would store the values in the water node so that they never got out of sync. We don't want to remove the SW out of physical because users want to be able to see it along side the other bucket parameters (LL15, DUL etc).

hol353:
I've just reread the comments. If Physical is part of a factor, without a water node then get the values from the first water node in the simulation (there can be many) and make it read-only.

PR #9030, opened 2024-06-17
Graph/Series debugging messages will not show unless the setting is enabled in the main presenter

### LINKED ISSUE #8599: Cannot ignore the missing values in the excel file on graph
### What happened?

In the master branch of APSIMX. Generate a graph with error

```
LAI (Observed): NaN Values found in points. These may be empty cells in the datastore.
LAI (Observed): 1 points where Y is NaN, but X is valid.
```

The example node is `.Simulations.Validation.Growth Experiments.Wagga.Wagga2008.Graphs.Garnet.LAI` on Canola.apsimx (validation test).

For example, the excel file is located `Tests\Validation\Canola\Wagga2008.xlsx`.

The cell `N57` is an empty without space, tab, or other special character. But source codes is red as NaN value. 

image

 

Windows

ric394:
resolves #8599

# Notes
- Users can now enable/disable graph and series debugging messages.
- The messages are great for model builders but not so great for other users. So having the choice to enable and disable should help.

### DIFF
--- a/ApsimNG/Presenters/GraphPresenter.cs
+++ b/ApsimNG/Presenters/GraphPresenter.cs
@@ -1,20 +1,23 @@
-﻿namespace UserInterface.Presenters
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Data;
+using System.IO;
+using System.Linq;
+using APSIM.Shared.Documentation.Extensions;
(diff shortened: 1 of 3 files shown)


par456:
That is a good solution to this problem, being able to turn them on when the graphs are broken is helpful.

Issue #8997, opened 2024-05-31
Updating ISummary to DataStore when exception occurs


lie112:
### What happened?

It is difficult to determine the timing of datastore actions. 
I have some parameter and setup validation (using IValidatableObject methods and Validation Attributes) that runs in OnCommencing and StartOfSimulation and tracks all errors and warnings through the ISummary to the datastore table _Messages. Then the controlling model checks this store prior to running the simulation and throws a fatal exception if any critical issues are raised. The problem is that the database is often not updated by the time the database is checked. So, a simulation can appear to run without issue while errors were written to the database after the check. Likewise, errors can be present upon opening a simulation that were never reported to the user. Also, errors can have been addressed, but still reported. This is confusing and makes it really difficult to address suggested issues and track why the model stopped, or not stopping when it should.
I have looked deep into DataStore and feel it needs a public method that allows us to force the outstanding writes to happen so I can call this before I check for errors. I vaguely recall such an approach exists somewhere, but I think it's private. Otherwise, I have to wait for some time during the simulation or post-simulation which is too late to trap validation errors.
Also, should the _Messages table be cleared at the start of a simulation as I don't believe it is, or the same issue exists and the database has not been updated when I check?
The problem is made worse when working with Experiment-Factors across threads as no error checking happens for some runs where even in a very quick model, the code completes before the database is updated.
I don't want to throw an exception on any validation issue, as I want the full list of problems to be presented to the user before causing the exception.
 

### What Operating System are you on?

Windows

par456:
In the datastore model, there is a property called Writer, which holds the current datastore writing object. With that you can call WaitForIdle() and it will return only once writing to the datastore has been completed.

```
datastore.Writer.WaitForIdle()
```

If you put that into your code where you are waiting for the summary messages to finish writing, then I think it should do what you are looking for.

lie112:
Thanks. I have implemented this.
Does the _Messages table get cleared when a simulation is run? At lease for the current simulation name. And does this also apply to all the simulations in an experiment?
I have just done an experiment run and it threw an error which was trapped and displayed. Upon fixing the error I ran again and it threw the error from data in the datastore without throwing in the validation code. I ran the experiment again and the simulation runs so not getting my error, but still displays all the old messages in the status area. 
Emptying the datastore results in the simulation running as expected with no errors displayed.

par456:
[...]

PR #9046, opened 2024-06-21
Female.NumberOfBirthsThisTimeStep fixes

### LINKED ISSUE #9045: CLEM summary reporting of births
### What happened?

The number of births in a time-step for reporting is not working as it doesn't get reset after births.

Windows

lie112:
Resolves #9045 
Also addresses aspects of #8997

### DIFF
--- a/Models/CLEM/Activities/RuminantActivityBreed.cs
+++ b/Models/CLEM/Activities/RuminantActivityBreed.cs
@@ -312,10 +312,6 @@ namespace Models.CLEM.Activities
             // get list of all pregnant females
             List<RuminantFemale> pregnantherd = CurrentHerd(true).OfType<RuminantFemale>().Where(a => a.IsPregnant).ToList();
 
-            // Need to reset births this time step for all non pregnant.
-            // Find a place where individuals are identified on stime step after births to reset this value.
-            _ = CurrentHerd(true).OfType<RuminantFemale>().Where(a => !a.IsPregnant).Select(a => a.NumberOfBirthsThisTimestep == 0);
-
             // determine all fetus and newborn mortality of all pregnant females.
             bool preglost = false;
             bool birthoccurred = false;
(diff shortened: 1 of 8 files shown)


par456:
Is this ready to be merged in?

lie112:
Yes. When I use resolves it is right to go.
I just included the other issue id for future tracking.

Models/CLEM/Activities/RuminantActivityBreed.cs
```diff
@@ -312,6 +312,10 @@ private void OnCLEMAnimalBreeding(object sender, EventArgs e)
             // get list of all pregnant females
             List<RuminantFemale> pregnantherd = CurrentHerd(true).OfType<RuminantFemale>().Where(a => a.IsPregnant).ToList();
 
+            // Need to reset births this time step for all non pregnant.
+            // Find a place where individuals are identified on stime step after births to reset this value.
+            _ = CurrentHerd(true).OfType<RuminantFemale>().Where(a => !a.IsPregnant).Select(a => a.NumberOfBirthsThisTimestep == 0);
```
hol353, line 317: Does this actually do anything? It stores the result of the .Select into a unnamed variable (i.e. _)

lie112: Yes, it resets the NumberOfBirthsThisTimestep for each non-pregnant female, but doesn't have a return value we need to keep. It's just done in a LINQ statement rather than use a loop over the filtered group of individuals.

hol353: Hmm. You're saying ```a.NumberOfBirthsThisTimestep == 0```, a conditional statement, resets the variable ```NumberOfBirthsThisTimestep```. The reader wouldn't expect this behaviour. It is not at all obvious. I suggest a method called something like ResetNumberOfBirthsThisTimestep() that makes it obvious.

lie112: Your right! This is wrong (==) and has actually been depreciated with a correct calculation of BirthsThisTimeStep by counting the number of 0 day old sucklings. I'll commit the changes.

Issue #9053, opened 2024-06-25
Wheat Zadok Code Class would benefit from changes


hut104:
### Describe the new feature

Wheat Zadok Code Class would benefit from changes that made it easier to compare Zadok codes against model output.  Basically, to allign codes to the real meaning of current model phenology stages etc.

hut104:
Issue is now resolved. Those a related issue does point to the need to clean up the use of links within the code.

PR #9054, opened 2024-06-25
PMF Zadok improvements

hut104:
working on #9053


hut104:
@hol353 The latest change does influence the stats adversely, however, the equation is more accurate.  Published approach for wheat development should be based on Huan Stage, not leaf tips (Celestina et al 2023)

Issue #9056, opened 2024-06-25
CAMP has coded links to other models


hut104:
### Where is the code located?

Camp.cs and CalcCAMPVrnRates.cs

### What needs to be changed?

CAMP is written with fixed links to children of other models.  This could be changed to allow these to be specified by the user or specified within an interface.

### Why does the code need changing?

hard coded links in the code can be broken

PR #9057, opened 2024-06-25
CAMP model links

hut104:
working on #9056

### DIFF
--- a/Models/PMF/Phenology/CAMP/CAMP.cs
+++ b/Models/PMF/Phenology/CAMP/CAMP.cs
@@ -46,9 +46,9 @@ namespace Models.PMF.Phen
         [Link(Type = LinkType.Ancestor, ByName = true)]
         Phenology phenology = null;
 
-        /// <summary>The ancestor CAMP model and some relations</summary>
-        [Link(Type = LinkType.Path, Path = "[Phenology].HaunStage")]
-        IFunction haunStage = null;
+        /// <summary>Crop Haun Stage</summary>
+        [Link(ByName = true)]
+        public IFunction HaunStage = null;
 
         /// <summary>
         /// Calculate delta of upregulation for photo period (Pp) sensitive genes
@@ -168,8 +168,8 @@ namespace Models.PMF.Phen
         [JsonIgnore] public CultivarRateParams Params { get; set; }
 
         /// <summary>The ancestor CAMP model and some relations</summary>
-        [Link(Type = LinkType.Path, Path = "[Phenology].Phyllochron.BasePhyllochron")]
-        IFunction basePhyllochron = null;
+        [Link(ByName = true)]
+        public IFunction BasePhyllochron = null;
 
         [Link(Type = LinkType.Path, Path = "[Phenology].HaunStage.Delta")]
         IFunction DHS = null;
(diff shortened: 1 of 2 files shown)


Models/PMF/Phenology/CAMP/CAMP.cs
```diff
@@ -46,9 +46,9 @@ public class CAMP : Model, IVrnExpression
         [Link(Type = LinkType.Ancestor, ByName = true)]
         Phenology phenology = null;
 
-        /// <summary>The ancestor CAMP model and some relations</summary>
-        [Link(Type = LinkType.Path, Path = "[Phenology].HaunStage")]
-        IFunction haunStage = null;
+        /// <summary>Crop Haun Stage</summary>
+        [Link(ByName = true)]
+        public IFunction HaunStage = null;
```
hol353, line 51: Should this link be limited to just Child functions? Also, it should be private and start with a lower case letter.

Models/PMF/Phenology/CAMP/CAMP.cs
```diff
@@ -168,8 +168,8 @@ public double calcTSHS(double FLN)
         [JsonIgnore] public CultivarRateParams Params { get; set; }
 
         /// <summary>The ancestor CAMP model and some relations</summary>
-        [Link(Type = LinkType.Path, Path = "[Phenology].Phyllochron.BasePhyllochron")]
-        IFunction basePhyllochron = null;
+        [Link(ByName = true)]
+        public IFunction BasePhyllochron = null;
```
hol353, line 172: Private and start with lower case letter.

Models/PMF/Phenology/CAMP/CalcCAMPVrnRates.cs
```diff
@@ -21,9 +21,10 @@ public class CalcCAMPVrnRates : Model
         /// <summary>The parent CAMP model</summary>
         [Link(Type = LinkType.Ancestor, ByName = true)]
         private CAMP camp = null;
-        /// <summary>The ancestor CAMP model and some relations</summary>
-        [Link(Type = LinkType.Path, Path = "[Phenology].Phyllochron.BasePhyllochron")]
-        IFunction basePhyllochron = null;
+
+        /// <summary>The  phyllochron upon which phyllochron changes through the season are based</summary>
+        [Link(ByName = true)]
+        public IFunction BasePhyllochron = null;
```
hol353, line 27: private and start with lower-case

hut104: Happy to do that if link by name is case insensitive.  I had assumed that it was case sensitive

Models/PMF/Phenology/CAMP/CAMP.cs
hut104: Link is currently to another model.  This step will allow me to continue to merge while I move things around and ulitimately explore the use of a leaf interface.  Will also avoid breaking user simulations if they are using replacements on prototypes for something.

Note that the PR has "working on" rather than "resolves".  I'd imagine that the resolution will look a bit different to this.

hut104:
Let's step back from these changes for now

PR #9064, opened 2024-06-26
Stop leaf reporting a positive value if plant is not alive

### LINKED ISSUE #9063: Leaf model should return zero for height when plant is not alive
### What happened?

Currently leaf model will return a non-zero height for plants that are not alive

Windows

HamishBrownPFR:
resolves #9063


HamishBrownPFR:
@hol353 @hut104 I think this is OK to merge.  Diffs are very small in a simulation that simulated two crops in rotation.  The diffs will be because prior to this fix the models were giving positive heights in the fallow period which would have caused microclimate to make canopies for them which had a small influence on soil water and nitrogen subsequently.

Issue #9068, opened 2024-06-27
Some outputs from Wheat Model should be changed in preparation for changes to Simple Leaf


hut104:
### Where is the code located?

Wheat.JSON

### What needs to be changed?

Change output names in Wheat.Leaf to match new naming conventions used in SimpleLeaf.  This will make the transition easier, but also standardise these for users in advance.  Less change for them in the long run.

### Why does the code need changing?

Changing Wheat.Leaf from Leaf.cs to SimpleLeaf.cs has many advantages.  However, one disadvantage is that changing report variables will impact users.  By standardising these now we can avoid future changes for users, but also avoid conflicts for observed files, where report variable names are embedded.

PR #9070, opened 2024-06-27
Wheat report changes

hut104:
working on #9068


hut104:
@hol353 Please merge this in.
Variables have been renamed only.  You will see that the stats for the renamed variables match (ie the actually model behaviour is not different).  There is one output (Leaf.Tips) that now covers two different outputs (that had the same value!) and so the stats are different, but N now equals the sum of the two previous N values.

Before merging, please check the new method in the JSON converter for a graph series converter.

I haven't put resolved in as yet.  It is possible that I may find another output as I move forward.  If you want, I can close this and raise another issue if I find more output to fix.

PR #9071, opened 2024-06-27
9069 Updated DirectedGraph to new BubbleChart code

### LINKED ISSUE #9069: Nutrient page is blank
### What happened?

Nothing is showing when the Nutrient component is selected.
Reproducible using Visual Studio in debug mode with most recent code.
Reproduced on 2 different machines with at least 3 different files.

Windows

par456:
Resolves #9069

I apparently broke the DirectedGraphPresenter that Nutrient uses when updating the BubbleChartPresenter #8350 a few months ago. Went through and updated this presenter so it works on the new code.


hut104:
@par456 I think you have my reporting issue PR number in your "resolves" statement.

PR #9078, opened 2024-06-30
Created Gtk.sheet assembly and hidden sheet complexity

### LINKED ISSUE #9079: GTK.Sheet bugs
### What happened?

Issue list for Dean: (to be edited)
- I cannot add a row to tables (the empty row at the bottom is missing now)
- If I delete rows from physical (which is expected to cause errors since it needs to match water), but not all cells selected are deleted and the now empty rows stick around
- Forages has "the method or operation is not implemented" error when opened (SimpleGrazing example)
- Morris and CroptimizR Examples has the same error as Forages (looks like quite a few weird models with tables have this error)
- Hitting delete on a datastore cell causes an exception
- Can right click-copy and it works, but the shortcut does not (paste works)

hol353:
Working on #8760
resolves #9079


ApsimNG/Presenters/ProfilePresenter.cs
```diff
@@ -133,10 +134,13 @@ public void Refresh()
                 {
                     if (water != null && (model is Physical || model is Water || model is SoilCrop))
                     {
+                        if (water.Thickness.Length != physical.Thickness.Length)
+                            throw new Exception("There is a mismatch between the number of soil layers on the physical node and water nodes. Cannot create greaph");
```
ric394, line 138: graph at the end is misspelled here.

ric394 (approved):

PR #9093, opened 2024-07-02
Add output for Leaf.SpecificAreaCanopy

hut104:
working on #9068


hut104:
@hol353 Can we please merge this?
At this stage, I have provided tests for both report variable names.  In future, we can possibly remove one of these.  The two outputs provide SLA in different units, but you will see that the new output provides the same statistical fits.
There are other issues to be resolved in these files that can be undertaken after this is merged.

hol353:
retest this please jenkins

Issue #8551, opened 2024-01-17
Simple Crop Model with tabular parameterization


HamishBrownPFR:
### Describe the new feature

We need a simple crop model that can be parameterized for crops that we have limited data for.  It will need to respond to temperature, water, nitrogen and humidity and be able to represent a wide range of crops.  
the design concept is to set up a basic structure in PMF with all of the parameters specified in an external .csv file that a component reads at run time and overwrites them into the model.  The idea being it is easier to extrapolate and justify parameter estimates for data porse crops when the parameters are laid out in a table and it is easy to make comparisons to other similar crops.  It also means the parameterization becomes explicitly the responsibility of the user making it possible to quickly set up rough crop models that do not have to go through release.

PR #9100, opened 2024-07-02
Adding classes for parallel phenology

HamishBrownPFR:
working on #8551

I have needed to do parts of the phenology in parllell for DEROPAPY.  The problem is that we use phenology to mark key events for the canopy and for biomass partitioning.  However, DEROPAPY represent a broad range of crops and key events (e.g start reproductive growth vs full canopy) do not alway happen in the same sequence.

The files in this PR represent an intum solution where I use the main phenology model to represent events in the canopy (start growth, full canopy, start senescence, canopy death) and I have added a concept of parallel phases that can start at a certian point and run through a seperate sequence to represent partitioning events (vegetative, flowering, reproductive growth).

This has got things working well enough to progress with the DEROPAPY prototype and I need to get this locked in to complete the first phase of development.

A better long-term solution would be to change Phenology a crop can have two (or more) phenology models driving different aspects of phenology.  I did try to do this but it got complicated quickly.  Now I have something that does what I want functionally, I will come back to it and refactor into something more tidy.


HamishBrownPFR:
@hol353 @par456 @ric394 can someone please review and merge when happy.  I am not sure why the stats have improved for the clover tests but the improvements are very small.

Models/PMF/Phenology/Phases/GotoPhase.cs
```diff
@@ -49,23 +50,37 @@ public string End
         [Display(Type = DisplayType.CropPhaseName)]
         public string PhaseNameToGoto { get; set; }
 
+        /// <summary>
+        /// The type of biomass removal event
+        /// </summary>
+        [Description("Type of biomass removal.  This triggers events OnCutting, OnGrazing etc")]
+        public BiomassRemovalType RemovalType
```
hol353, line 57: Why is biomass removal stuff being put into a GotoPhase function?

Models/PMF/Phenology/Phases/GrazeAndRewindPhase.cs
```diff
@@ -35,7 +35,7 @@ public string End
         {
             get
             {
-                return phenology?.FindChild<IPhase>(PhaseNameToGoto)?.Start;
+                return "Rewind phase has no end";//phenology?.FindChild<IPhase>(PhaseNameToGoto)?.Start;
```
hol353, line 38: Remove commented code.

Models/PMF/Phenology/Phases/GotoPhase.cs
HamishBrownPFR: The go to phase broadcasts a PhenologyDefoliate event that has an arguement of RemovalType.
The BiomassRemovalEvent subscribes to this event and makes the appropriate biomass removal happen when ever the goToPhase is reached.  This mimics what the GrazeAndRewind phase does and means we can deprocate that phase as it is now redundant.  I still need to replace the GrazeAndRewind phase in SPRUM with the more generic GoToPhase before I can deprocate the class though.

Models/PMF/Phenology/Phases/GrazeAndRewindPhase.cs
HamishBrownPFR: That class is marked for deprecation.  I will remove the commented code in my next pull request to save Jenkens the job of running the tests again

HamishBrownPFR:
@hol353 are these Reponses OK.

PR #9101, opened 2024-07-03
XYPairView refactor

### LINKED ISSUE #9097: XY Pairs View needs refactor
## Problems

- [x] Changing x value causes graph to refresh slowly (~ 2sec delay).
- [x] Description is wrong. Spelling and grammar need work. Make it pull from the summary, if not already.
- [x] The view should be changed to one where the screen is split vertically. XY grid on left, graph on right. This is to avoid ugliness when screen size changes.
- [x] Make vertical split 33% for xy grid, 66% for graph.

ric394:
resolves #9097

- XY Pairs view now aligns the grid table and graph vertically.
- Description text grammar fixed
- Grid table object null errors fixed by correcting some incorrect casting statements.


zur003:
@ric394 - The use of HPaned and VPaned has been deprecated since Gtk 3.16. You can declare the variable as `Paned`, and construct using `hpaned = new Paned(Orientation.Horizonal);` to generate equivalent behaviour.

BrianCollinss:
I wonder of this PR fixes that wide gap between the table and the graph:

image

ric394:
@BrianCollinss This PR won't fix that particular issue but removing that gap would be nice. I'll create an issue for that shortly.

Issue #9102, opened 2024-07-03
Incorrect Data in Wheat Validation Set


hut104:
### What happened?

Specific Leaf Area data within the wheat validation set do not represent the two values from which they can be derived.

### What Operating System are you on?

Windows

hut104:
Furthermore, the Wheat validation set still has data with Stem N% as 0 (observed!)

bj-far:
And... the wheat model has picked up some new cultivars (added to the "New Zealand" folder) which probably should _only_ be in the validation set. 
The current release APSIM has Voltron, Sorrial, Cesario, Stockade, CRW247, Zyatt, Graham and Kerrin which won't behave the way users expect them to (outside the Lincoln 2021 and 2023 simulations).

PR #9120, opened 2024-07-04
Wheat validation data

hut104:
working on #9102


hut104:
@hol353 Please merge this for now.  Will come back later if/when we find more issues.
The decrease in N comes from the removal of observed zeros.
In improvement in haunstage comes from removing predicted zeros.

Issue #9131, opened 2024-07-07
Wheat model sensibility test for N dynamics


hut104:
### Describe the new feature

The wheat model has no sensibility test for internal N dynamics (accumulation, retranslocation, etc).

Such a test has proven essential in other crop models to ensure that the timing and amounts of N changes matches expectations.

hut104:
Need to add graphs for general arbitration results and time trends

PR #9132, opened 2024-07-08
Wheat n dynamics test

hut104:
working on #9131


hut104:
@hol353 Merge this please.  I may make further changes if we find them needed for the simple leaf work.

PR #9135, opened 2024-07-08
Removing wheat outputs from validation set

hut104:
working on #9068


hut104:
@hol353 Please merge.  There is one output that has been removed because very few data sets ever have this measurement, and it is not available in newer leaf models.

PR #9136, opened 2024-07-08
Adding NextState to RotationManager

### LINKED ISSUE #9111: RotationManager does not report the last state
### What happened?

In the example Rotation.apsimx, it seems the last state is not reported. I tried to replace `[RotationManager].Transition` with `[RotationManager].TransitionToX` or `[RotationManager].TransitionToY` (according to [this](https://apsimnextgeneration.netlify.app/usage/rotationmanager/)) but that event has not been defined.

**Report (last state is Summer Fallow):**
image

**TransitionReport (last state is Wheat):**
image

Windows

BrianCollinss:
Resolves #9111 

- Ready to merge, if approved.
- First, tested a more aggressive approach, which might not be ideal, so reverted the changes.
- This is a simple workaround for the issue (adding NextState to RotationManager). That is, both current and next state can be reported, helping the user better understand transitions.


BrianCollinss:
@peter-devoil Do you see any problem with this change?

peter-devoil:
I do: Suppose we are reporting (state, yield) on transition, and decide to harvest a crop of wheat. We'd expect the report output line to look like "Wheat, 2042.0" but this change would mean we get a line like "fallow, 2042" - ie yield & state disconnected.

The current behaviour (using the rotation sample) is
image

What you're seeing in the daily output file is the state at the end of the day, after the operation has taken place. 

As with the plant phenology models, it's important not to confuse phases (a period of time) with events - points in time. It's particularly nasty that we have some processes happening in the middle of the day (eg DoManagement sowing a crop), and others that arent valid until the end of the day (eg SW content), and putting all of these into a single output is hard: the logic in this [PhaseAccumulator](https://github.com/APSIMInitiative/ApsimX/blob/702e8b1fa5d50ae96a80413f3b72333e04b33a0c/Tests/Simulation/MultiZoneManagement/MultiPaddock.apsimx) script mixes some climate/ soil & crop variables into a single table. It's not pretty.

BrianCollinss:
Thanks Peter.

image

This is what I get now. You are right, yield value now sits next to fallow. However, is it not the correct state? That is, we have wheat yield when it is harvested and current state is fallow. 

What do you think of having three event handlers instead of one, so user can choose precisely what to report?

BrianCollinss:
I think this approach is safer.

image

par456:
This looks like a suitable solution to me (nextState is updated before the transition event fires, then currentState is updated to the same place after).

Does that work for your uses @peter-devoil ?

peter-devoil:
That's fine; I think there was a "transition_from_A" and "transition_to_B" event in there as well..

PR #9139, opened 2024-07-09
changed error messages

### LINKED ISSUE #9138: Error message points to wrong name
### What happened?

Error messages in RetranslocateNonStructural.cs talk of an error computed for "Name".

However, name is the name of the function.  The name of the plant organ would be more helpful.

Perhaps change the error messages to point to the organ name AND the function name.

Windows

hut104:
resolves #9138


hut104:
@hol353 I think this can be merged in now and closed

PR #9140, opened 2024-07-09
Add extra graphs for arbitration

hut104:
working on #9131


hut104:
@hol353 These can be merged for now.

Issue #9142, opened 2024-07-10
Sensescence can make pools go negative in wheat


hut104:
### What happened?

It is possible for the wheat model to senesce more material in a generic organ pool than is still there if there are competing flows of DM.

### What Operating System are you on?

Windows

LenLon:
This also happened in my maize simulations, though it refers to the Sorghum model in the error message for some reason:

```
Simulating...
----------------------------------------------
ERROR in file: /home/jansen/input-data/WestAfrica/sim/WestAfrica_C4Maize_MIROC6ssp126.apsimx
Simulation name: Simulation
Models.Core.SimulationException
 ---> System.Exception: Attempted to senesce more biomass than exists on leaf 'Leaf'
   at Models.PMF.Organs.SorghumLeaf.ApplySenescence() in /p/projects/agrica/APSIM/APSIMX/ApsimX_077556/Models/PMF/Organs/SorghumLeaf.cs:line 889
```

PR #9144, opened 2024-07-10
Error in Wheat arbitration

hut104:
working on #9142


hut104:
@hol353 I think this is a minimal change.  There is a real issue somewhere else, however this will limit the impact of whatever that other error is.  In theory the error should only ever be small.  This change doesn't break mass balance because it is applied to the delta before being applied.  It is just designed to avoid -ve values that may arise from an error elsewhere.

PR #9145, opened 2024-07-10
Update all glade files to 3.24, and replace all deprecated Gtk features.

### LINKED ISSUE #9125: Remove deprecated Gtk features

User interface code in ApsimNG

### What needs to be changed?

Gtk features which have been deprecated should be replaced with more modern equivalents.

This does not strictly need to be done at this stage, but should make future updating a bit less complicated. We are currently using GtkSharp 3.24.24.38, but updating to 3.24.24.95 treats use of some of these deprecated feaures (e.g., HBox constructor) as errors. 

This should be regarded as a "sub-issue" of #9088 (question: does github provide any formal mechanism for treating an issue as a series of smaller sub-issues?)

zur003:
Resolves #9125 

This should be OK, although changes to the UI do not readily lend themselves to unit testing. This updates all the Gtk stuff to version 3.24. There are quite a few changes, and I've tried to test everything (and fixed a couple of minor bugs along the way), but I can't guarantee that there are no errors.

I might note that we still can't easily update the GtkSharp nuget packages from our currently used 3.24.24.38 to the latest stable version 3.24.24.95, since doing so doesn't play well with the nuget version of Oxyplot.GtkSharp3.


hol353:
Thanks @zur003 for the PR. I've pulled your branch and found a problem. To reproduce:

Open Tests\Validation\Wheat\Wheat.apsimx. Click on the Tests model under DataStore - throws Exception. Something to do with Gtk.Label.

hol353:
The same thing happens if you open the AgPasture example and click on Forages under AgPastureExample simulation.

zur003:
@hol353 - Thanks for picking that up. It should now be working as it had previously (including some oddities).

zur003:
I think I will resubmit this PR after "flattening" the commits.

PR #9148, opened 2024-07-11
Remove More deprecated reports

hut104:
working on #9068


hut104:
@hol353 I think this can be merged.  I'm not sure if there are more outputs to be removed, but we will see over the coming week or so.

PR #9160, opened 2024-07-16
9155 Sugarcane Editting Script and Presenter Change

### LINKED ISSUE #9155: No genotype parameters for Sugarcane
### What happened?

Hi,
Following examples of how to access genotype parameters for wheat <https://www.youtube.com/watch?v=N5HGqRO5Krc> by right clicking the Plant Model and asking for showing Model Structure. When you right click Sugarcane plant module, this option is already selected and there is nothing there. 

Kind regards,
Isis

Windows

par456:
Resolves #9155

Script from previous issue was no longer working, so I rewrote it to include all the cultivar parameters for sugarcane and added the script to the management toolbox so it's not lost in an example from an issue.

I also changed the Sugarcane model to use the Generic Presenter and wrote in the summary that the script could be found in the management toolbox. This should hopefully help other people trying to do something with sugarcane find these resources.

### DIFF
--- a/ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
+++ b/ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
@@ -1,13 +1,12 @@
 {
   "$type": "Models.Core.Simulations, Models",
-  "Version": 174,
+  "Version": 176,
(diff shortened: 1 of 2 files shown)


par456:
Changing to better solution #9167

PR #9161, opened 2024-07-17
9137 Soil Download Issue

### LINKED ISSUE #9137: Exception thrown after soil is downloaded from APSoil or ISRIC
### What happened?

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=ApsimNG
  StackTrace:
   at UserInterface.Presenters.ProfilePresenter.PopulateChemicalGraph(GraphView graph, Double[] thickness, Double[] pH, PHUnitsEnum phUnits, IEnumerable`1 solutes) in ApsimNG\Presenters\ProfilePresenter.cs:line 282
```

image

Windows

par456:
Resolves #9137 

Changed to https for apsoil requests, also did some checking for other bugs encountered when testing.

### DIFF
--- a/ApsimNG/Presenters/SoilDownloadPresenter.cs
+++ b/ApsimNG/Presenters/SoilDownloadPresenter.cs
@@ -523,7 +523,9 @@ namespace UserInterface.Presenters
         /// <param name="soil"></param>
         private static void InitialiseSoil(Soil soil)
         {
-            soil.Children.Add(new CERESSoilTemperature());
+            CERESSoilTemperature temperature = new CERESSoilTemperature();
+            temperature.Name = "Temperature";
+            soil.Children.Add(temperature);
             var physical = soil.FindChild<Physical>();
             if (physical != null)
             {
@@ -822,6 +824,9 @@ namespace UserInterface.Presenters
                     wheat.Name = "WheatSoil";
                     waterNode.ParentAllDescendants();
 
+                    CERESSoilTemperature temperature = new CERESSoilTemperature();
+                    temperature.Name = "Temperature";
+
                     newSoil.Children.Add(waterNode);
                     newSoil.Children.Add(soilWater);
                     newSoil.Children.Add(nutrient);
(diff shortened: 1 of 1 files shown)


ApsimNG/Presenters/SoilDownloadPresenter.cs
```diff
@@ -348,10 +348,33 @@ private async Task<IEnumerable<SoilFromDataSource>> GetApsoilSoilsAsync(IProgres
...
-                double longitude = Convert.ToDouble(longitudeEditBox.Text, System.Globalization.CultureInfo.InvariantCulture);
-                double radius = Convert.ToDouble(radiusEditBox.Text, System.Globalization.CultureInfo.InvariantCulture);
+                double latitude = 0;
+                double longitude = 0;
+                double radius = 0;
+                try 
+                {
+                    latitude = Convert.ToDouble(latitudeEditBox.Text, System.Globalization.CultureInfo.InvariantCulture);
+                }
+                catch (Exception e)
```
hol353, line 358: Rather then using try / catch to determine if a string can be converted to double, use ```double.TryParse``` method

hol353 (changes requested):

ApsimNG/Presenters/SoilDownloadPresenter.cs
par456: Yep, that's a better idea, updated.

ApsimNG/Presenters/SoilDownloadPresenter.cs
```diff
@@ -348,10 +348,26 @@ private async Task<IEnumerable<SoilFromDataSource>> GetApsoilSoilsAsync(IProgres
...
             {
-                // fixme: Shouldn't this be using the current culture?
-                double latitude = Convert.ToDouble(latitudeEditBox.Text, System.Globalization.CultureInfo.InvariantCulture);
-                double longitude = Convert.ToDouble(longitudeEditBox.Text, System.Globalization.CultureInfo.InvariantCulture);
-                double radius = Convert.ToDouble(radiusEditBox.Text, System.Globalization.CultureInfo.InvariantCulture);
+                bool canParse;
+                
+                double latitude;
+                canParse = double.TryParse(latitudeEditBox.Text, out latitude);
+                if(!canParse) {
```
hol353, line 355: This can be combined onto one line like this:
```c#
if (!double.TryParse(latitudeEditBox.Text, out double latitude)
```

Simpler?

par456: yes, yes it can.

hol353 (approved):

PR #9164, opened 2024-07-17
9052 ExcelInput Multiline

### LINKED ISSUE #9052: Change ExcelInput to multiline
### What happened?

The properties on the ExcelInput should be multiline entries instead of single line. Make sure that the formatting for that makes sense as well and that the buttons still work.

Windows

par456:
Resolves #9052

Makes the input fields on the ExcelInput model into multi-line text boxes. Much easier to read now than one big long list in something like the Wheat validation.

I tried to avoid changing the model as much as possible here and do it all through GUI changes, as the underlying structure was already suitable for this, it just needed to be displayed better.

Might be worth pulling and testing as a GUI change though before it's merged.

### DIFF
--- a/ApsimNG/Views/PropertyView.cs
+++ b/ApsimNG/Views/PropertyView.cs
@@ -340,7 +340,6 @@ namespace UserInterface.Views
                     component = dropDown.MainWidget;
                     break;
                 case PropertyType.File:
-                case PropertyType.Files:
                 case PropertyType.Directory:
                     //case PropertyType.Directories:
                     // Add an Entry and a Button inside a VBox.
@@ -353,8 +352,6 @@ namespace UserInterface.Views
                     fileChooserButton.Name = property.ID.ToString();
                     if (property.DisplayMethod == PropertyType.File)
                         fileChooserButton.Clicked += (o, _) => ChooseFile(o as Widget, false, false);
-                    else if (property.DisplayMethod == PropertyType.Files)
-                        fileChooserButton.Clicked += (o, _) => ChooseFile(o as Widget, true, false);
                     else if (property.DisplayMethod == PropertyType.Directory)
                         fileChooserButton.Clicked += (o, _) => ChooseFile(o as Widget, false, true);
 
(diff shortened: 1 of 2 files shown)


ApsimNG/Views/PropertyView.cs
```diff
@@ -353,15 +352,41 @@ private Widget GenerateInputWidget(Property property)
...
                     Box container = new HBox();
                     container.PackStart(fileNameInput, true, true, 0);
                     container.PackStart(fileChooserButton, false, false, 0);
                     component = container;
+                    break;
+                case PropertyType.Files:
+                    string filenamesText = property.Value?.ToString();
+                    filenamesText = filenamesText.Replace(",", "\n");
+                    filenamesText = filenamesText.Replace("\n ", "\n");
+                    if (!filenamesText.EndsWith('\n'))
```
hol353, line 367: Rather than all this handling of \n characters, why not just cast ```property.Value``` to an array of strings? I assume it is because users may have commas given that is how it used to be. Why not write a converter to strip the commas and do it properly? This will also simplify UpdateFilenamesText below.

ApsimNG/Views/PropertyView.cs
```diff
@@ -532,6 +557,41 @@ private void UpdateText(object sender, EventArgs e)
             }
         }
 
+        /// <summary>
+        /// Called when a multiline filenames has been modified.
+        /// </summary>
+        /// <param name="sender">The entry which has been modified.</param>
+        /// <param name="e">Event data.</param>
+        [GLib.ConnectBefore]
+        private void UpdateFilenamesText(object sender, EventArgs e)
```
hol353, line 566: If you write a converter, can this disappear?

hol353 (changes requested):

hol353:
I also notice the tool tip needs updating:
image

ApsimNG/Views/PropertyView.cs
par456: The current code has a conversion somewhere between where this returns and the model in the property gets updated to convert the string this returns into an array using commas as separators. I'll have a look and see just how tangled that converter is with current code, I had trouble finding it before so I just stuck to rearranging the string.

par456: We don't need a converter here as the filenames are already stored as a string array in excel input, the problem is that by the time property.Value gets to this part of propertyview, it has already been converted from a string array to a string seperated by commas.

However I can use one of our utility functions to convert it back to an array again as a one liner here, which makes this much easier to read.

ApsimNG/Views/PropertyView.cs
par456: This can disappear if I'm a bit smarter with how I use the existing event. Will change.

PR #9167, opened 2024-07-18
9155 Converting Sugarcane to use Cultivar system

### LINKED ISSUE #9155: No genotype parameters for Sugarcane

par456:
Resolves #9155

The aim of this PR is to have Sugarcane use the apsim Cultivar system for adjusting sugarcane varieties instead of its own custom built system.

It would be good if we had a "default value" for each cultivar parameter, that way a user wouldn't have to define all the cultivar parameters for every cultivar they write, which currently leads to a lot of repetition. Other crops have default values, so the cultivar then only has to change the stuff that is different. **I just need someone with a bit of sugarcane knowledge that knows a good base cultivar to use for the defaults.**

While we can't get rid of CultivarConstants and CropConstants, I'm going to do a rename on them to SugarcaneCultivar and SugarcaneCrop so that it's really obvious what they actually are. It also has its own MathUtil class that should be replaced by our existing math utils so that it produces the same results.


BrianCollinss:
If Sugar moule has been ported from Classic without any significant change, I would say most of the 'default' parameters are calibrated for Q117.

hol353:
Yep, the sugar example in 7.10 uses Q117. Let's go with that.

BrianCollinss:
I just put this here until someone gets to look into it:

image

Models/Resources/Sugarcane.json
```diff
@@ -46,7 +46,7 @@
       "NH4_min_ub": 10000.0,
       "NH4_min_lb": 0.0,
       "plant": {
-        "$type": "CropConstants, Models",
+        "$type": "SugarcaneCrop, Models",
```
BrianCollinss, line 49: This might be unimportant. Some frameworks/models (e.g., DSSAT) distinguish these two groups of parameters by naming them 'species' and 'cultivar' parameters (DSSAT also has 'ecotype'!). I thought perhaps this can be changed to 'SugarcaneSpecies' to avoid the word 'Crop'.

hol353: Good idea. I don't like the word Crop either. How about:

SugarcaneSpeciesConstants 

Another thought that just occurred. @par456 Do we need a converter so that existing user simulation files still work? I suspect because we're changing $type and the cultivar mechanism in the json file they won't work anymore.

hol353 (changes requested):

Models/Resources/Sugarcane.json
par456: I might need to do some more checking, but I don't think we'll need a converter?

The two classes we renamed aren't models, and they are basically used as structs to hold properties for the .cult and .plant properties of sugarcane. The only references I've found to them were in the sugracane class and json file, otherwise users just reference it as [sugarcane].cult.leaf_size (for example).

You also can't do a replacements on sugercane in apsim, so those classes should never appear in a saved structure either.

hol353 (approved):

Issue #9174, opened 2024-07-18
DataStoreWriter can be very slow and consume a lot of RAM in certain factorial simulations.


ilhuber:
### What happened?

When running a factorial (~550 simulations), the data store writer seems to get clogged, taking a considerable amount of time to finish writing after all simulations have completed. Additionally, it can consume an obscene amount of RAM:
![thumbnail_Outlook-4w44tz0l](https://github.com/user-attachments/assets/130f1684-83e4-46d2-8e56-068061c21c76)
The culprit looks like it is the use of a list to track datastore jobs within DataStoreWriter. Removing a job from the front of the list causes a large copy. This can be pretty effortlessly swapped to a queue, eliminating the issue.

Oddly, this doesn't seem to be problematic for other large simulations. For instance, the Wheat validation simulations' writer is able to keep pace. It also isn't problematic if I either set Summary's verbosity to 0 or make year a factorial, rather than running a 34 year simulation with reset. This seems to imply there might be issues with Summary in long term simulations in addition to the dequeueing inefficiency mentioned above.

### What Operating System are you on?

Windows

par456:
I wonder if this is also related to #8948

hol353:
Is this now resolved @ilhuber? If so, can you close it?

ilhuber:
The weird issues with the Summary component still exist even after the change, though the amount of memory consumed is an order of magnitude smaller. The summary issues are probably better brought up in a separate issue. I'll see about preparing a small example simulation that demonstrates the issue.

PR #9175, opened 2024-07-18
Change from a list to a queue in DataStoreWriter.

ilhuber:
Working on #9174 

Swaps the list for a queue, which helps with both memory and running time. I wonder if it might also improve the #8759 situation. Doesn't do anything to address Summary slowing long term simulations down.


par456:
This seems to have sped up jenkins runs by several minutes, really good find.

PR #9189, opened 2024-07-25
9187 Unsquash Rotation Manager Presenter

### LINKED ISSUE #9187: RotationManager's Bubble Chart is squashed
### What happened?

image

Windows

par456:
Resolves #9187

Not sure why GTK has suddenly stopped respecting the size of the drawing area in the paned window (maybe has to do with #9149 changing to the newer Paned object instead of HPaned and VPaned).

This change just gives the drawing area a default size so that it doesn't get squished.

### DIFF
--- a/ApsimNG/Views/BubbleChartView.cs
+++ b/ApsimNG/Views/BubbleChartView.cs
@@ -96,7 +96,7 @@ namespace UserInterface.Views
             objectBox = (Gtk.Box)builder.GetObject("object_box");
             Box instructionsBox = (Gtk.Box)builder.GetObject("instructions_box");
 
-            graphView = new DirectedGraphView(this);
+            graphView = new DirectedGraphView(this, horizontalSplitterPosition);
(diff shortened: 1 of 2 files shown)


zur003:
@par456 - I think what was needed here was just the addition of `drawable.Expand = true` in the constructor of DirectedGraphView. That will allow it to expand to fill the available space, and avoids the problems associated with specifying an arbitrary width. I ran into this "no longer expanding by default" problem somewhere else when working on #9149, but didn't test the bubble charts.

BrianCollinss:
Is this the intended behaviour?

https://github.com/user-attachments/assets/9a3a48a1-213c-4100-8033-72954ace23e3

par456:
Yeah, that was an existing bug in rotation, though it used to have a bit more space before it happened. This was more a patch job in a busy week so it would at least be useable again.

I'll give Eric's solution a try on Monday and see if that produces better results.

PR #9210, opened 2024-07-30
9207 Allows adding an expression to a reporting event

### LINKED ISSUE #9207: Harvesting Event doesn't work with date filter
### What happened?

We should be able to write:
```
[Wheat].Harvesting && [Clock].Today >= new DateTime(2000, 1, 1)
```

To only report harvesting events after the 1/1/2000, however at the moment that does not compile, I get an exception " Invalid report frequency found"

Windows

par456:
Resolves #9207

So apparently this functionality didn't exist, but wasn't too hard to patch in. The only problem I ran into was that I had to assume the first part of the line was always the event, as I couldn't see a good way to have it figure out if a given string was a valid event or function expression.

So this works:
```
[Wheat].Harvesting && [Clock].Today.Year >= 2000
```

~~But this doesnt:~~
And so does this now:
```
[Clock].Today.Year >= 2000 && [Wheat].Harvesting
```

~~Should I try and expand this to allow the event to be placed anywhere in the line? or do I just put some more errors to detect when that's not the case and tell the user that.~~


ric394:
Either solution seems ok, but I'd prefer if you could place it anywhere, so you don't have to remember formatting (or be reminded with an error message when its formatted incorrectly). 

Is it possible/practical to use reflection to see if each non '&&' string sub section is an event or not?

hol353:
The order on the line shouldn't matter. Expand the parsing to work with any ordering of expression / event.

par456:
This can be reviewed now. It will now take the event and conditions in any order.

PR #9217, opened 2024-07-31
Refactor grid

hol353:
Working on #8760


hol353:
@peter-devoil - I'm changing the way models work with the grid control. I've modified FarmMachinery and FarmPaddock in this PR. Can you pull the branch and check I haven't broken anything. There was some code in FarmMachinery.cs that I didn't know what is was doing so I deleted it, because that's what programmer do :).
```c#
        /// <summary>
        /// Ensure any new child components added by user are present in the arrays
        /// </summary>
        public DataTable ConvertModelToDisplay(DataTable dt)
        { 
            var newCombos = new List<string>();
            foreach (var t in getTractorNames()) 
               foreach(var i in getImplementNames()) 
                  newCombos.Add(t + "." + i);

            var dtCombos = new List<string>();
            for (var row = 0; row < dt.Rows.Count; row++) 
               dtCombos.Add(dt.Rows[row]["Tractor"].ToString() + "." + dt.Rows[row]["Implement"].ToString());

            foreach (var combo in newCombos)
               if ( ! dtCombos.Contains(combo)) {
                  DataRow newRow = dt.NewRow();
                  newRow["Tractor"] = combo.Split(".")[0];
                  newRow["Implement"] = combo.Split(".")[1];
                  dt.Rows.Add(newRow);
               }
            return(dt);
        }
```

peter-devoil:
I'll have a look. That code is there so that when new child components are added below the grid, they appear in the grid. It may be superfluous now

peter-devoil:
Hmm. This isnt working. After a new child component is added, where can we hook into this [Display] thing to update the 4 properties that appear [here](https://github.com/APSIMInitiative/ApsimX/blob/6a876ec1bb8184c41ca6dc31c1f7a0c213603e95/Models/Management/FarmMachinery.cs#L157)? 

There's a sample in Tests/Simulation/MultiZoneManagement/MultipaddockEcon.apsimx:
image

hol353:
So, are you saying that when you add a new machine as a child then the FarmMachinery grid updates itself. Is that grid editable by the user?

peter-devoil:
Yes?

hol353:
@peter-devoil: OK hopefully fixed now - can you check it again?

peter-devoil:
Looks to be OK. Did you notice that the manager script  is readonly?

par456:
The only thing I was able to break was the select all function on the grid. It selects the header rows (which are normally not selectable) and misses the last 1 or 2 rows. Then if you try to copy or paste with the headers selected, it throws an error.

Otherwise it looks good.

par456:
How did this manage to modify stats?

hol353:
If you look at the stats they haven't really changed. Obviously there are small diffs in lower decimal places. Some changes have been made for Forages and SimpleGrazing so it is possible that there are slight calculation diffs.

This is ready to be re-reviewed and merged. Note, TreeProxy has changed significanly since the last peer-review. This will need testing by pulling the branch and playing with TreeProxy.

Issue #8683, opened 2024-02-22
Values of Tav and Amp calculated by APSIM are not reliable


sno036:
### What happened?

The values of Tav and Amp that APSIM would calculate (i.e. when the rows are removed from the file) for AmesSoilTempAllYears.met are as below and are *rather* wrong. They should be Tav 9.4 and Amp 14.7. 

@sarchontoulis - note that the Amp of 29.64 in the file is wrong also.

![image](https://github.com/APSIMInitiative/ApsimX/assets/12176269/62803809-5f72-41ee-939f-bce64b1d3117)

### What Operating System are you on?

Windows

ilhuber:
I think there may still be some problems with the current approach. Years with no data are still being included in the number of years, lowering the Tav and Amp averages in sparse met files. It may also be a good idea to only include complete years when automatically computing Tav and Amp, since a year with data for only one month contributes a value of 0 for Amp average and probably throws Tav off too.

ric394:
@sno036 may I have a copy of `AmesSoilTempAllYears - NoTavAmp.met` for testing please?

hol353:
Is there still a problem here?

PR #9221, opened 2024-08-02
Make calculating of tav and amp smarter.

ric394:
working on #8683

# Notes

- now ignores months and years where maxt and mint values are zero in mean calaculations.
- multi-year tav and amp calculations will now ignore years where a specific year has a tav and amp of zero.

# Where to go from here
- need to handle years and months where data is sparse. This requires sensible boundary values for years and months


rcichota:
The calculation of TAV+AMP probably should throw an error if the amount of data supplied is too small, i.e., less than a few years (certainly if less than a year)

PR #9226, opened 2024-08-05
922 cli add command parent with child node bug fix

### LINKED ISSUE #9225: models.exe add command parent node with child bug
### What happened?

When using the add command, if a user was to use the command:

`add [Simulation].Zone Report`

or similar, a model will not be added.

Windows

ric394:
resolves #9225

# Notes
Add command now works correctly when a parent node has a child e.g. `[Simulations].Field`


ric394:
I'll submit another PR to fix this one.

PR #9238, opened 2024-08-13
9228 Batching Summary messages to improve write speeds

### LINKED ISSUE #9228: Why Is APSIM Consuming Excessive Resources on Google Cloud's Linux Environment?
### What happened?

Hi,

I am running APSIM version 7.2.8.1 on a Linux system in Google Cloud. My setup includes optimized cloud computing with 16 cores and 64GB of memory. Despite this, running APSIM with the attached file (and in many other scenarios) takes over 15 minutes.

During this time, I observed that APSIM uses all 16 cores at 100% utilization. This seems highly unreasonable. Could there be an issue with the APSIM files, the weather files, or is there another way to optimize its operation?

[Soy_Minnesota.zip](https://github.com/user-attachments/files/16508535/Soy_Minnesota.zip)

Windows

par456:
Resolves #9228

 It seems the Linux slowdown was being caused by the summary messages being written one at a time leading to them getting queued up and taking too long. I've set messages to batch into groups of 50, so we'll see what the performance impact is.

There still is a difference between windows and linux, but this at least gets them a bit closer (50s vs 75s on the simulation I was testing with).

### DIFF
--- a/Models/Clock.cs
+++ b/Models/Clock.cs
@@ -513,6 +513,7 @@ namespace Models
                 EndOfSimulation.Invoke(this, args);
 
             Summary?.WriteMessage(this, "Simulation terminated normally", MessageType.Information);
+            Summary?.WriteMessagesToDataStore();
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


par456:
Looks like this change has knocked 5 minutes off the jenkins tests run, so that seems like a good result.

PR #9254, opened 2024-08-22
S_RUM upgrade - enabling EndCrop for ScrumInstance

rcichota:
Working on #7869 
Set public methods to enable calling Harvest and EndCrop for a ScrumInstance from manager scripts.
Currently the two actions were triggered internally, OnDoManagement, based on criteria given at initialisation. This change enables calling them from managers, giving more flexibility to the model. 

This PR has also some commits for updates to the dataset used in SCRUM Validation... These will still be update further later on.


rcichota:
@hol353 
As this P.R. went green, could you merge it, please. This update will help us setting crop rotations for a couple of projects her. Hamish, Edith, and co. still working to get SCRUM past the point to be fully on release, but we want to use (and  test) it already.

PR #9259, opened 2024-08-23
9258 Commands system Filepath for value

### LINKED ISSUE #9258: Commands system can take in filepath for value
### Describe the new feature

Need a way to feed in the contents of a file to a property value so that newline characters can be used with commands.

par456:
Resolves #9258

To provide a filepath instead of a value in the commands txt, you can set the value like so:

```
[Simulation].Name=mytextfile.txt;
```

It will look for a semi-colon at the end (like when you do a replace action with a node from another file) and if a file with that name exists. If both are true, instead of using the text in that line, the text from the file is read in instead and entered as the value.

This lets you provide multi-line string inputs and larger values from a separate file instead of having to get it all into the commands.txt

A unit test was also created for this behaviour.

### DIFF
--- a/Models/Core/ConfigFile/ConfigFile.cs
+++ b/Models/Core/ConfigFile/ConfigFile.cs
@@ -76,9 +76,16 @@ namespace Models.Core.ConfigFile
                 Match firstSplitResult = rxOverrideTargetNode.Match(commandSplits[0]);
                 if (firstSplitResult.Success)
                 {
+                    string property = commandSplits[0];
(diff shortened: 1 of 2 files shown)


Models/Core/ConfigFile/ConfigFile.cs
ric394, line 1: I think the instruction print statements should be rearranged to show the original command. This would make it easier to track down the offending line in the config file.

ric394 (approved):

PR #9269, opened 2024-08-28
Put protection in for when there is only one row in the DataStore when setting column widths

### LINKED ISSUE #9267: DB presenter hiding data
### What happened?

Perhaps we could adjust the column width to the data and the header, or make the columns adjustable?

image
vs
image

Windows

ric394:
resolves #9267

# Cause of bug
 
When setting the column widths in Sheet, if there was a single data row, the data rows columns were not included in the process of determining column width.

This meant that if you had a fairly short column header like 'Es' data would be hidden, if the data was longer than the header.

# Fix

I made sure that when there is a single data row, that it always gets included for column width setting.

### DIFF
--- a/Gtk.Sheet/Sheet.cs
+++ b/Gtk.Sheet/Sheet.cs
@@ -564,20 +564,11 @@ namespace Gtk.Sheet
             if (autoCalculateColumnWidths && DataProvider != null)
             {
                 int visibleRows = FullyVisibleRowIndexes.Count() + NumberHiddenRows;
-                if (visibleRows >= DataProvider.RowCount)
-                {
-                    // Prevents visibleRows from going to zero when there is only one data row.
-                    if (DataProvider.RowCount > 1)
-                    {
-                        visibleRows = DataProvider.RowCount - 1;
-                    }
-                }
-
                 ColumnWidths = new int[DataProvider.ColumnCount];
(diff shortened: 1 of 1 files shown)


par456:
Looking at the code, couldn't you achieve the same effect by having it like this:

```
/// <summary>Calculte the column widths in pixels.</summary>
        /// <param name="cr">The current draing context.</param>
        private void CalculateColumnWidths(IDrawContext cr)
        {
            if (autoCalculateColumnWidths && DataProvider != null)
            {
                int visibleRows = FullyVisibleRowIndexes.Count() + NumberHiddenRows;
                ColumnWidths = new int[DataProvider.ColumnCount];
                for (int columnIndex = 0; columnIndex < DataProvider.ColumnCount; columnIndex++)
                {
                    int columnWidth = GetWidthOfCell(cr, columnIndex, 0);
                    for (int rowIndex = NumberHiddenRows; rowIndex < visibleRows; rowIndex++)
                        columnWidth = Math.Max(columnWidth, GetWidthOfCell(cr, columnIndex, rowIndex));

                    ColumnWidths[columnIndex] = columnWidth + ColumnPadding * 2;
                }
                RedrawNeeded?.Invoke(this, new EventArgs());
            }
        }
```

PR #9283, opened 2024-09-05
New Soil temperature tidy up

rcichota:
working on #4080 
Several changes to tidy up the code (in general and to make integration with Crop2ML easier)
 - adding summary tags
 - deleting redundant checks
 - remove dictionary (soil properties)
 - ensure model uses value from other models directly (e.g. weather), not copies


rcichota:
The stats are slightly different for the NZ simulation. This is because the weather data has values for wind speed, while the others don't (thus default is always used). As the diffs are small and reading the wind speed from weather data is the right thing to do, these changes can be accepted, and this Pull Request can be merged
NOTE: the sensitivity of soil temperature to wind speed should be better checked in the near future (will Talk with Val), the simulation 'InitialisingSoilCarbonPools' in the Examples folder crashes at times due to soil temperature being estimated outside bounds. That simulation uses ```WeatherSampler``` and the crash happens depending on the weather 'selection' (changing the seed). So, this might be caused by some specific value or combination of values (windy vs no wind in consecutive days). Perhaps ```SoilTemperature``` needs to apply some bounds to wind to avoid this... I will add this to the list of tasks to finish off the changes to this model for release

Models/Soils/SoilTemp/CERESSoilTemperature.cs
```diff
@@ -131,9 +131,27 @@ private void Clear()
...
+        /// <summary>Average soil temperature for each layer (oC)</summary>
+        public double[] MinimumSoilTemperature { get { return st; } }
+
+        /// <summary>Average soil temperature for soil surface (oC)</summary>
+        public double MinimumSoilSurfaceTemperature { get { return double.NaN; } }
+
+        /// <summary>Average soil temperature for each layer (oC)</summary>
+        public double[] MaximumSoilTemperature { get { return st; } }
+
+        /// <summary>Average soil temperature for soil surface (oC)</summary>
```
BrianCollinss, line 152: Daily maximum soil temperature

Models/Soils/SoilTemp/CERESSoilTemperature.cs
```diff
@@ -131,9 +131,27 @@ private void Clear()
...
+        /// <summary>Soil temperature for each layer (oC)</summary>
         public double[] Value { get { return st; } }
 
+        /// <summary>Average soil temperature for each layer (oC)</summary>
+        public double[] AverageSoilTemperature { get { return st; } }
+
+        /// <summary>Average soil temperature for soil surface (oC)</summary>
+        public double AverageSoilSurfaceTemperature { get { return double.NaN; } }
+
+        /// <summary>Average soil temperature for each layer (oC)</summary>
```
BrianCollinss, line 143: Daily minimum soil temperature

rcichota: We thought about using the term "Daily" in the name, but it would make it long, and Dean remarks that all variables in APSIM are on a daily basis... I will correct the info in the summary tag in the next commit...

PR #9302, opened 2024-09-09
Manager read only bug fix

### LINKED ISSUE #9278: Management script not edittable until saved
### What happened?

I ran into this issue modifying the management toolbox scripts.  This is also an issue if add a management script using the add model functionality as well.
I was unable to change anything in the script editor until it was compiled/saved.

Windows

ric394:
resolves #9278

This was a read-only issue.

This checks for a vertical scrollbar in the managers editor (meaning the script is quite long) and makes it do buffer checks before making text writable.
If however a manager's editor does not have a vertical scrollbar it will skip buffer checks and make the text writable.

### DIFF
--- a/ApsimNG/Views/ManagerView.cs
+++ b/ApsimNG/Views/ManagerView.cs
@@ -79,7 +79,7 @@ namespace UserInterface.Views
             }
 
             //default value for no cursor
-            if (cursor.ScrollH.Valid == false)
+            if (cursor.ScrollV.Valid == false)
             {
                 cursor = scriptEditor.Location;
                 notebook.Drawn -= OnDrawn;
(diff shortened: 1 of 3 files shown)


ApsimNG/Views/ManagerView.cs
```diff
@@ -66,29 +62,42 @@ public void OnPageChanged(object sender, EventArgs e)
...
-                scriptEditor.ReadOnly = false;
+                cursor = scriptEditor.Location;
                 notebook.Drawn -= OnDrawn;
-                if (cursor != null && this.TabIndex == TAB_SCRIPT)
+                return;
+            }
+
+            if (this.TabIndex == TAB_SCRIPT)
+            {
+                if (cursor.ScrollH.Upper == scriptEditor.Location.ScrollH.Upper)
```
ric394, line 91: This seems like the wrong orientation to me. Should it be ScrollV?

par456: right yeah...it should be

ric394: That works really well!

PR #9312, opened 2024-09-12
Integrate latest DCaPST and Dynamic Tillering changes.

joesaddigh:
This PR addresses some fixes that were made to the Dynamic Tillering routines. It also integrates DCaPST with some tweaks to parameters, as per Alex Wu's instructions.

Working on https://github.com/APSIMInitiative/ApsimX/issues/7705

This PR also supersedes: https://github.com/APSIMInitiative/ApsimX/pull/9008


joesaddigh:
Just replying to Dean Holzworth's original comment as this PR is a new version of the originally reviewed PR #9008. 

2024-09-11 13_26_59-Integrate latest DCaPST and Dynamic Tillering changes  by joesaddigh · Pull Requ

- I notice in the stats that the sorghum model performance has changed. Does this mean that DCaPST is turned on by default for sorghum simulations? Does this slow down the runtime speed of sorghum?

DCaPST is not turned on by default. You can only enable DCaPST by adding it to the model under the Paddock.  
The stats might've changed due to some changes to tillering and the canopy for Sorghum. However, we also removed some experiments that we deemed as unfit for the validation. We also increased the N by allowing certain experiments through the validation that didn't quite make it to Harvest. There were a number that got very very close but didn't reach the harvest stage due to a freeze event.. This has been reviewed by Graeme Hammer.
We are going to profile DCaPST in a separate PR in the near future to try and speed it up and address any bottlenecks. 

- The stats also show that the number of observations for sorghum has changed, in some cases 'n' has reduced.

As explained above :-)

- There are also a bunch of files in this PR that probably aren't needed. Under DCaPST there is a Sorghum and Wheat directory. Are these needed?
- There is also an 'old' directory under Tests/Sorghum.

These files have now been removed.

Please let me know if you have any other feedback and we will be happy to address.

Thanks,
Joe

PR #9315, opened 2024-09-12
Move LeafCohortParameters class from Leaf.cs into its own file

### LINKED ISSUE #9314: LeafCohortParameters Class is in same file as Leaf Class

Leaf.cs

### What needs to be changed?

LeafCohortParameters needs to be moved to it's one file. 

Two classes should not occupy the one class.

ric394:
resolves #9314


ric394:
retest this please jenkins

Issue #9325, opened 2024-09-17
Add and make few variables to be settable


rcichota:
### Describe the new feature

Weather reads in a series of values from the .met file, most of these can be given as daily values or as single constants. Some of these can be set also (from a script, for example), but others are only gettable. This issue proposes a few improvements to make this model even more useful, particularly when testing specific models (basically it would be great to be able to set environmental factors with values given alongside the weather data or set via scripts, instead of allowing then to be solely computed by APSIM):
 - allow more/all variables to be settable (e.g. latitude, and including MeanT)
 - allow AirPressure to be set as well as to be calculated internally (currently it is only a constant)
 -  add to the list of variables to include potential evapotranspiration, plus potential and actual evaporation
These changes were needed to test SoilTemperature models under fixed/simple conditions, but should be useful elsewhere. Enabling variables to be settable should not change the model behaviour as the 'default' values should be returned unless the user actively set their values (reading in MeanT might have some risks as it is possible met files may have such values, but this is rare and most times the values are the same as computed in APSIM (average between MinT and MaxT). Making this settable allows to simulate variations of this paradigm if needed.
One can note a few typos plus need to standardise and document the code in few places. These can be tackled here too.

PR #9326, opened 2024-09-17
Add/set weather variables

rcichota:
Working on #9325 
updating the code to:
 - fix a few typos and minor organisation of the code
 - allow a few weather variable to be settable
 - add a few extra variables

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ Prototypes/SCRUM/.ipynb_checkpoints/ParameterListBuilder-checkpoint.ipynb
 Prototypes/SCRUM/Crop.xml
 Prototypes/SCRUM/SCRUM.apsimx.orig
 Prototypes/Grapevine/Observations/PhenoDataObservation - Copy.xlsx
+Prototypes/**/Outputs*/
 lib/
 .vs/
 .Rproj.user
--- a/Models/Climate/Weather.cs
+++ b/Models/Climate/Weather.cs
@@ -125,7 +125,7 @@ namespace Models.Climate
         /// <summary>
         /// The index of the diffuse radiation fraction column in the weather file
         /// </summary>
-        private int DiffuseFractionIndex;
+        private int diffuseFractionIndex;
 
         /// <summary>
         /// The index of the day length column in the weather file
@@ -300,9 +300,6 @@ namespace Models.Climate
                 }
             }
         }
-
-        /// <summary>Flag whether we're one the first day of simulation</summary>
-        private bool First = true;
 
         /// <summary>Gets the number of days since the winter solstice</summary>
         [Units("d")]
@@ -555,7 +552,7 @@ namespace Models.Climate
             this.vapourPressureIndex = 0;
             this.windIndex = 0;
             this.co2Index = -1;
-            this.DiffuseFractionIndex = 0;
+            this.diffuseFractionIndex = 0;
             this.dayLengthIndex = 0;
             if (AirPressure == 0)
                 this.AirPressure = 1010;
@@ -582,15 +579,6 @@ namespace Models.Climate
 
             foreach (var message in Validate())
                 summary.WriteMessage(this, message, MessageType.Warning);
-        }
-
-        /// <summary>
-        /// Performs the necessary initialisation at the start of simulation
-        /// </summary>
-        [EventSubscribe("StartOfSimulation")]
-        private void OnStartOfSimulation(object sender, EventArgs e)
-        {
-            First = true;
         }
 
         /// <summary>Overrides the base class method to allow for clean up task</summary>
@@ -649,7 +637,7 @@ namespace Models.Climate
             if (this.PreparingNewWeatherData != null)
                 this.PreparingNewWeatherData.Invoke(this, new EventArgs());
 
-            if (First)
+            if (clock.Today.Date == clock.StartDate.Date)
             {
                 //StartDAWS = met.DaysSinceWinterSolstice;
                 if (clock.Today.DayOfYear < WinterSolsticeDOY)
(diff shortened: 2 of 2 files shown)


hut104:
Our Weather models seem to be getting pretty complex (with several models and lots of public variables and/or setters).
What are some of the use cases here? e.g. why is latitude settable?
If someone changes latitude, I assume that daylength is recalculated, as are dates of seasons, etc.  There are lots of interdependencies between weather variables.  If one gets changed, do we handle the dependencies?

I also see repeated patterns of if statements all through the model suggesting a refactoring may be needed.

rcichota:
Good points @hut104 , I am aware that there are many interdependencies, and so trying to keep that in mind when doing any changes. I will set a few tests regarding the potential flow on effect of setting Latitude. I will revert if needed or advised on that.
I've got into doing this as me and @hol353 needed to set some of the weather variables (including latitude, mean T and PET) for our work with soil temperature models in Montpellier. The case to make this general is that is allows (as is needed for that work) to set controlled conditions and isolate more effectively some models for testing and sensitivity analysis. Looking at the code I noticed that the weather variables were defined/set in somewhat inconsistent manner, so I am trying to standardise things a bit (also add a few variables that may be useful, eg. PET, but as this would only be made available to models, would not change SoilWater). I have used the same structure that was in the model, I agree there are repeated patterns, it would be good if someone with more coding skills had a look at it.
Certainly will need a good revision of this before merging

rcichota:
My idea is to standardise the behaviour of setting the variables so that they can firstly be given daily values (as a column), if not found then check if there is a single value, finally providing a 'default' value (which is 'null' for require data (Rain, MaxT, etc.), a constant (for Wind or CO2) or compute a value using a function (DiffuseFraction, VP, etc).
Obviously, this does not apply to values expected to be constants (like Latitude)

hol353:
You’re on the right track. I think what you’re coming up against is the poor design by the weather components brought about by a design that has evolved over time.

What we need is a single place (model) that stores and checks weather data and makes it available to other models that need it. It would be quite a simple model (called 'weather' like it is now). We could then have another model (as a child model maybe) that reads weather data in APSIM format, perhaps called WeatherReader (and maybe another in WTH format - called WthWeatherReader) and stores the values in 'weather'. No need for ControlledEnvironement as that would be a weather component without a reader. No need for SimpleWeather either I suspect. I think Hamish wrote SimpleWeather so might need to check with him. WeatherSampler would just sample from 'Weather'. The current GUI weather graphs would also work with 'Weather'.

This might be a step too far for this PR.

rcichota:
I like the idea of refactoring and making sure that things don't get duplicated. I will be 'finishing' this soon by making sure WeatherSampler has the same variables / same behaviour. I will then open a new issue for refactoring, but will need some help from someone with more coding skills to separate the code properly...

rcichota:
I need a bit of help here,
Jenkins is saying that there's an error when running the unit test for WeatherSampler (it seems the variable 'SplitDate' has not been set, somehow - although it should have a default). I had a good look at the code and couldn't find any clue. I do not know how to run the UnitTests so cannot debug locally...

ric394:
@rcichota for vs code:

run_unit_tests

for visual studio:

run_unit_tests_vs

hol353:
@rcichota - Fixed.

rcichota:
Thanks @ric394 , I can now run the units test locally.
And thanks @hol353 for fixing the issue with the WeatherSampler UnitTest.
I have another issue, I am not able to replicate the results from validation tests (AgPasture.apsimx). There seem to be a significant issue with one of the sites in the validation (Ranui, see RngAnnualHarvestedWt), but when I run locally, the diffs I get are very small (almost negligible, e.g. my RMSE for that graph is 2689.70). Not sure how this is possible, will continue looking into...

rcichota:
retest this please Jenkins

rcichota:
@hol353 
I haven't been able to replicate the results from Jenkins, and neither could Xiumei. This is evident on the AgPasture simulation. Could be that the test system is doing something odd?
I just don't know what action I could try to solve the diffs if I cannot replicate the runs...

hol353:
@rcichota: I can reproduce the results on jenkins if I use a linux computer (which is what Jenkins is using). I've spent a couple of hours trying to work out what isn't working correctly on linux and I can't. I've tried lots of things. The problem is that this PR has a lot of changes (rearranged code and some more semi-major changes) in a lot of commits and it has never run green on Jenkins.

I can't spend more time on this. You might have to go back to the beginning and reintroduce the changes slowly and carefully.

As I've said several comments above, my interest is separating out the weather file reading code into a separate class and refactor out the helper code (e.g. calculating things like tav/amp/daylength/airpressure etc) into a separate weather utilities class. Perhaps we abandon this PR and start again.

rcichota:
Thanks @hol353 
Much appreciated the time you spend on this. Still odd that the results in Linux are so different than in Windows (I get 3.6t of biomass, while Jenkins gest more than 10t, for the Ranui site). The code has run green many times (I am sure), only when I changed how AirPressure is output, it went red (and mostly only for AgPasture, as the value now respond to Altitude and that simulation actually has values for it). Funny it is showing all commits with crosses now...
Regardless, I quite agree with you on starting anew (there were lots of changes indeed). But then it would be better to start 'properly', with the code set up as you proposed a few comments earlier. I am happy to put a bit of time and get this over the line (mostly would be checking/copying what was done here), but I need help to get started, get the basic models/files/classes in place then I can put in the actual code. Do you, or any of the guys there, could help with that?

peter-devoil:
My apologies for introducing simpleweather. The context was a ML application where the simulation is continually restarted, but not re-initialised (ie after link resolution), sometimes using a different weather file. I couldnt see any sense in a linked list that caches yesterdays data, some other application must be using it, so I left it in place. It is also a hit if we want to have one enormous weather file and randomly sample from it.

HamishBrownPFR:
@rcichota Where are you at with this development?  Looking to reduce the number of pull requests that sit open on the system.  This could be closed and reopened when it becomes a priority again?

rcichota:
I am closing this PR as this needs some re-thinking and time to find solutions for the diffs that arose from the changes made. We'll re-open when possible.

Issue #9327, opened 2024-09-17
SWIM spends a lot of time figuring out that timecode "00:00" is minute of day 0.


ilhuber:
### Where is the code located?

SWIM

### What needs to be changed?

Call TimeToMins less frequently. Having an internal representation of the time should eliminate most uses.

### Why does the code need changing?

DateTime.TryParseExact is a little costly. I think this is related to a mostly vestigial part of SWIM where it was able to support time steps other than daily. There does not appear to be any capacity to make apsim_time anything but 00:00 and apsim_timestep anything but 1440.

PR #9328, opened 2024-09-17
Switch SWIM's time internal representation.

ilhuber:
Working on #9327 

SWIM now avoids calling TimeToMins unnecessarily.

### DIFF
--- a/Models/Soils/SWIM/Swim.cs
+++ b/Models/Soils/SWIM/Swim.cs
@@ -196,7 +196,7 @@ namespace Models.Soils
         private double apsim_timestep = 1440.0;
         private int start_day;
         private int start_year;
-        private string apsim_time = "00:00";
+        private int _apsimTimeMinutes = 0;
         private bool run_has_started;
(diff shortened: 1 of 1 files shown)


Models/Soils/SWIM/Swim.cs
```diff
@@ -1803,18 +1805,18 @@ private void GetOtherVariables()
...
             double duration = 0.0;
             double intensity;
+            int timeOfDay;
             if (string.IsNullOrWhiteSpace(rain_time))
             {
-                time = default_rain_time;
+                timeOfDay = 0;
                 duration = default_rain_duration;
             }
             else
```
ilhuber, line 1817: I am not sure that this else branch is reachable. Should this method be more extensively refactored?

hol353: Probably unreachable at the moment but there is talk about allowing rain_time to be settable. Let's leave it in for the moment. In general SWIM needs a refactor but that is for another day.

PR #9333, opened 2024-09-23
make graph series error bars enable/disable with their matched series

### LINKED ISSUE #9310: Error bars still visible after series are disabled
### What happened?

In case it is not intentional. 

image

Windows

ric394:
resolves #9310

# Notes
- Error bars for a series are actually a separate series. Special handling is required to have them toggle visibility inline with their matched scatter plot.

### DIFF
--- a/APSIM.Shared/APSIM.Shared.csproj
+++ b/APSIM.Shared/APSIM.Shared.csproj
@@ -24,7 +24,6 @@
     <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="2.88.6" />
     <PackageReference Include="System.Drawing.Common" Version="6.0.0" />
     <PackageReference Include="Docker.DotNet" Version="3.125.13" />
-    <PackageReference Include="OxyPlot.Core" Version="2.1.2" />
   </ItemGroup>
   
   <ItemGroup>
--- a/APSIM.Shared/APSIM.Shared.csproj
+++ b/APSIM.Shared/APSIM.Shared.csproj
@@ -24,6 +24,7 @@
     <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="2.88.6" />
     <PackageReference Include="System.Drawing.Common" Version="6.0.0" />
     <PackageReference Include="Docker.DotNet" Version="3.125.13" />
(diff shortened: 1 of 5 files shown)


hol353:
retest this please jenkins

APSIM.Shared/APSIM.Shared.csproj
```diff
@@ -24,6 +24,7 @@
     <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="2.88.6" />
     <PackageReference Include="System.Drawing.Common" Version="6.0.0" />
     <PackageReference Include="Docker.DotNet" Version="3.125.13" />
+    <PackageReference Include="OxyPlot.Core" Version="2.1.2" />
```
hol353, line 27: Do we need to add this dependency to APSIM.Shared?

APSIM.Shared/Graphing/NamedScatterErrorSeries.cs
```diff
@@ -0,0 +1,23 @@
+using OxyPlot.Series;
+
+namespace APSIM.Shared.Graphing;
+
+/// <summary>
+/// A Nameable ScatterErrorSeries.
+/// </summary>
+public class NamedScatterErrorSeries: ScatterErrorSeries, INameableSeries
```
hol353, line 8: Hmm this doesn't seem to do much. Why is it necessary?

If it is needed then move it to where it is needed (not in APSIM.Shared). One day I'd like to remove all the graphing code out of APSIM.Shared.

APSIM.Shared/APSIM.Shared.csproj
ric394: In it's current implementation it's required but I may be able to move it to ApsimNG

APSIM.Shared/Graphing/NamedScatterErrorSeries.cs
ric394: @hol353 Alternatively, would you suggest making a separate assembly for graphing in this pull request to start the process of moving graphing related files to their own assembly?

hol353: No, don't create a new assembly. That's a bigger job for another day.

PR #9337, opened 2024-09-24
Fixed delete row on grids using ClassWithOneListProperty

### LINKED ISSUE #9293: CroptimizR parameter table cannot be edited
### What happened?

When I remove a row, it returns when I save. Also I have this in the APSIM file now:

```
        {
          "$type": "Models.Sensitivity.Parameter, Models",
          "Name": "",
          "Path": "",
          "LowerBound": "NaN",
          "UpperBound": "NaN"
        }
```

https://github.com/user-attachments/assets/5f7fd4ad-b2cf-4609-9bf6-8322039cb2a5

Windows

hol353:
Resolves #9293

### DIFF
--- a/Gtk.Sheet/Data/ClassWithOneListProperty.cs
+++ b/Gtk.Sheet/Data/ClassWithOneListProperty.cs
@@ -31,7 +31,7 @@ class ClassWithOneListProperty : IDataProvider
                 {
                         this.properties.Add(properties);
                 });
-        }        
+        }
     }
 
     /// <summary>An event invoked when a cell changes.</summary>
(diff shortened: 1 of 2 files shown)


ric394:
Let me know if this is unrelated but I get the below behavior which is similar to the original issue. 

https://github.com/user-attachments/assets/0d1d2d63-6606-49d9-ac42-1e6df2c68d01

hol353:
I can't reproduce what the screen recording is showing. I'm using the wheat example.

ric394:
> I can't reproduce what the screen recording is showing. I'm using the wheat example.

Funnily, all is well after retesting this. I'm assuming something was the issue in my local environment 🤔

PR #9356, opened 2024-09-27
New model to let users specify CO2

### LINKED ISSUE #8470: Unable to find where I can modify the CO2 conc. in model?
### What is your question?

Hi Folks, I want to change the CO2 conc. levels in model to check the impact on crop performance. Previously it can be modified in "ClimateController" but now it's not showing the CO2 conc. Could you please clarify where I can modify this in model?

hol353:
Resolves #8470


ric394 (changes requested):
The only thing I can suggest is maybe a simple test that makes sure the CO2 values change with the addition of the CO2 model

hol353:
I've added a CO2 tutorial that has a test in it that will fail if the CO2 model doesn't work.

ric394 (approved):

PR #9369, opened 2024-10-03
9367 ISO Date fix

### LINKED ISSUE #9367: Operations should accept one-digit day and month
### Describe the new feature

- 2015-10-03 and 2015-09-03 work.
- 2015-10-3 and 2015-9-03 do not work.

Many users may not know this.

par456:
Resolves #9367 

The regex used for detecting an ISO date was not flexible enough to see it when 2000-1-1 zeros were missing from the month and day. Fixed that and added unit tests for that edge case.


par456:
retest this please jenkins

ric394 (approved):

PR #9383, opened 2024-10-16
Fixed Cotton Boll data

hut104:
working on #4631


hut104:
@hol353 Merge this please.  This only affects the prototype and fixed errors in Obs data.

PR #9391, opened 2024-10-21
Fixed the issue with urine being added at the wrong depth in the soil

### LINKED ISSUE #9390: SimpleGrazing is adding urine AT the given depth rather than DOWN TO the given depth
### What happened?

Explained in the title

Windows

sno036:
resolves #9390


sno036:
Please accept these stats - they are expected after fixing an error in the urine return to the soil

PR #9403, opened 2024-10-25
Graphs can be copied to clipboard in two different sizes

### LINKED ISSUE #9352: GraphView's ExportToClipboard produces too small images
### Describe the new feature

Changing the default export size from 800x600 to 1200x800 makes it a bit better. Is there a better way to do it (e.g., using the real dimensions or aspect ratio of the graph)?

image
image

BrianCollinss:
Resolves #9352 

Do you approve my approach? It gives flexibility to choose the suitable size depending on legend presence and positioning, number of series, etc.

### DIFF
--- a/ApsimNG/Interfaces/IGraphView.cs
+++ b/ApsimNG/Interfaces/IGraphView.cs
@@ -342,7 +342,7 @@
         /// <summary>
         /// Export the graph to the clipboard
         /// </summary>
-        void ExportToClipboard();
+        void ExportToClipboard(int width=1200, int height=800);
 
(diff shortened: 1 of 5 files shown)


par456:
Yeah, this has bugged me for a while too, but I think there is a better solution than hardcoding a new size.

I've opened another PR for this here #9424 so that the size can be configured by users.

BrianCollinss:
Great @par456. Feel free to close this one if yours was successful. Ta

ric394:
This issue has been resolved with #9424. I'll close this now.

PR #9407, opened 2024-10-25
Add 'concat' function to Report

### LINKED ISSUE #9406: Concat function in Report
### Describe the new feature

- This function 'joins/combines' all values over the aggregation period into a single string.
- This is particularly useful when user does not want to have daily reports in very large simulations.
- The use case is in environmental characterisation where daily stress indexes are needed to cluster large number of environments into finite number of groups.

BrianCollinss:
Resolves #9406 

Please merge if you approve my updated approach.


par456:
Having a look at this, can you provide an example of how/why this is used. I get that you are looking to combine multiple numbers into a single string in a column, but not sure how those numbers are referenced to do that (you mention it replacing daily reports).

Maybe an apsimx file using this branch that demonstrates it?

BrianCollinss:
In Envirotyping, we usually only need a time- or TT-based index matched with a daily stress index, so we can cluster simulations. With this, user can only export, for example, zadoks stage in one string and daily stress index in another string without the need for a separate daily report.

BrianCollinss:
@par456 Andrew, can I provide more info?

par456:
The main problem I'm having when looking at your concat function is that it doesn't fully match the existing functions that it is copying from. For example, the sum command is used like that, however it can also be used to combine layer arrays together which this currently doesn't. This is also currently a very specific use case that requires code to handle it specifically from all the other aggregations since it deals with strings instead of numbers.

I feel like this would make more sense to be written in as a manager script that combines the values together into a property that the report can then print out directly as a string. We could include that in the management toolbox then, and that would also let people format the output how they required as well (commas, spaces, rounding to decimal places etc). 

I'll have a look at writing that up and see if it will fit your needs.

BrianCollinss:
@par456 Andrew, you're breaking my heart! 😉 I leave it to you to choose the best approach.

BrianCollinss:
@par456 Andrew, can we have this merged at least for now? It doesn't affect the code much and can be referred when you get time to create a manager script instead.

BrianCollinss:
Andrew @par456 I created this new function that does what I need. Do you approve this approach? It is based on AccumulationFunction, but returns a string as well. 

```
        /// <summary>Called by Plant.cs when phenology routines are complete.</summary>
        /// <param name="sender">Plant.cs</param>
        /// <param name="e">Event arguments</param>
        [EventSubscribe("PostPhenology")]
        private void PostPhenology(object sender, EventArgs e)
        {
            if (ChildFunctions == null)
                ChildFunctions = FindAllChildren<IFunction>().ToList();

            if (phenology.Between(startStageIndex, endStageIndex))
            {
                string DailyIncrement = "";
                foreach (IFunction function in ChildFunctions)
                {
                    if (DailyIncrement.Length > 0) 
                        DailyIncrement += ",";
                    DailyIncrement += Math.Round(function.Value(), RoundDigits).ToString();
                }

                if (AccumulatedValue.Length > 0)
                    AccumulatedValue += ",";
                AccumulatedValue += DailyIncrement;
            }
        }

        /// <summary>Gets the value.</summary>
        /// <value>The value.</value>
        public string ValueString()
        {
            return AccumulatedValue;
        }
```

par456:
Sorry for the delay, we've had a few busy weeks.

Can you test if this manager script does what you need:
```c#
using System;
using Models;
using Models.Functions;
using Models.Core;
using Models.Interfaces;
using Models.PMF;
using APSIM.Shared.Utilities;

namespace Models
{
    [Serializable]
    public class Script : Model
    {        
        List<string> values;
        
        [Description("Variable to Combine: ")] 
        public string VariableName { get; set;}
        
        [Description("Seperator Symbol ")] 
        public string Seperator { get; set;} = ",";
        
        public string Value { 
            get {
                string output = "";
                foreach (string val in values)
                    output += val + Seperator;
                if (output.Length > 0)
                {
                    output = output.Substring(0, output.Length-Seperator.Length);
                }
                return output;
            }
        }
        
        [EventSubscribe("StartOfSimulation")]  
        private void OnStartOfSimulation(object sender, EventArgs e)
        {
            values = new List<string>();
        }

        [EventSubscribe("PostPhenology")]
        private void OnPostPhenology(object sender, EventArgs e)
        {
            values.Add(this.Locator.Get(VariableName).ToString());
        }
    }
}
```

I have it set up like this:
image

And in report, I can report it out like so:
```
[ConcatenateVariable].Script.Value
```

BrianCollinss:
Thanks @par456 I will test this and get back to you. I wonder what you think of my updated approach? It will remove the need for multiple manager scripts.

BrianCollinss:
Hi @par456 Andrew. I'd appreciate your comment here so we can move on with this. Thanks.

ric394:
> @ric394 Julian, I wonder if you can comment on my updates as Andrew seems very busy. Ta

Hey @BrianCollinss, I believe @par456 is investigating #9456 before moving forward on this one.

BrianCollinss:
Hey guys. If this is not approved, please let me know so start working on my manager script. I appreciate your time.

par456:
I think it's worth going for the manager script approach for now, just because this is conflicting with a few issues around accumulator that I am looking into. The 2nd solution you have is better, but is replicating some bad code from accumulator that we already know is not working correctly. 

Not only is there the weirdness you found with it not adding up the same, but we also have #9434 where it is reseting prior to reporting at harvest. Considering how widespread it's use currently is, none of the changes can be done quickly because it's going to have fairly widespread effects and so we actually need to plan this before rolling out a fix.

Sorry for the delays on it, it's been a busy time of year.

PR #9412, opened 2024-10-29
9411 Added naming to Apply commands

### LINKED ISSUE #9411: Add ability to rename node when created with Add/Copy/Duplicate command functions
### Describe the new feature

Need way to rename a newly created node with the apply commands so that it can be referenced in further lines without the type conflicting with the name.

par456:
Resolves #9411

Also did a refactor of parts of this code to make it easier to expand and removed bunch of unnecessary code that we had utilities functions for.

### DIFF
--- a/APSIM.Shared/Utilities/ApsimTextFile.cs
+++ b/APSIM.Shared/Utilities/ApsimTextFile.cs
@@ -527,7 +527,8 @@ namespace APSIM.Shared.Utilities
                 }
                 else
                 {
-                    Headings = StringUtilities.SplitStringHonouringQuotes(HeadingLines[0], " \t");
+                    Headings = new StringCollection();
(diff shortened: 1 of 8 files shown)


BrianCollinss:
Would be very helpful to have an example (just here as a comment perhaps). Ta

par456:
Yep, I need to update the documentation for the commands and add some unit tests for it. In brief:

Add and Copy can now take an extra input that renames the node when it's created. So

```
add [Experiment].Factors.Site CompositeFactor SiteA
```

Will add a CompositeFactor to the Site node called SiteA. Which can then be used in further commands like so:

```
add [Experiment].Factors.Site.FARAwardTrial_1 Clock
add [Experiment].Factors.Site.FARAwardTrial_1 Weather
```

To further build the file.

ric394 (approved):

PR #9414, opened 2024-10-30
Assorted improvements for Crop2ML project

hol353:
Working on #4080

* Fixed bug in auto-doc - not finding models.exe
* Added useful DataTableUtilities.ReadDataTable method for manager scripts.
* Made small number of weather variables settable.
* Added some units to various interfaces.
* Fixed a bug in SaturatedFlow that was throwing when KS is an empty array.

### DIFF
--- a/APSIM.Documentation/Models/AutoDocumentation.cs
+++ b/APSIM.Documentation/Models/AutoDocumentation.cs
@@ -147,7 +147,7 @@ namespace APSIM.Documentation.Models
 
             DefineFunctions().TryGetValue(model.GetType(), out Type docType);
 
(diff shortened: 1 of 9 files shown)


par456:
I assume you have some sort of auto-formatting on in vsc that's removing unneeded whitespace? 

I don't mind that, except in AutoDocumentation.cs it's going to cause merging hell when we merge that into the next AutoDoc branch since it's touching so many lines in that file. Could you redo that file and so the only changes are the 3 lines you intended?

hol353:
Thanks for this @par456. I've changed AutoDocumentation.cs as suggested.

PR #9424, opened 2024-11-04
9352 Graph Size options

### LINKED ISSUE #9352: GraphView's ExportToClipboard produces too small images

par456:
Resolves #9352

Adds a graph size option into the settings menu (settings are in the top bar on the first screen in apsimng). This then defines how big the graph should be when copied out of apsim.

Considered a few options for this, but this is the easiest way to make it completely customizable for whatever size and formatting users want in their graph, and still making it consistent every time they copy a graph. It will still default to 800x600 currently, but that can now be changed.

### DIFF
--- a/ApsimNG/Utility/Configuration.cs
+++ b/ApsimNG/Utility/Configuration.cs
@@ -2,6 +2,9 @@
 using System.Collections.Generic;
 using System.Drawing;
 using System.IO;
+using System.Linq;
+using APSIM.Shared.Documentation.Extensions;
+using APSIM.Shared.Utilities;
 using Models.Core;
 
 namespace Utility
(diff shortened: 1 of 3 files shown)


par456:
retest this please jenkins

BrianCollinss:
I wonder if this approach provides enough flexibility when user has a wide range of graph configurations (number of series, legend size and positioning, etc) each needing a different size to produce a readable image.

par456:
In cases where someone needs to graph a significant amount of data, we generally advise to export the data from apsim and use another tool to do that graphing. The graphing tools in apsim are not as fully featured as other libraries in R or python that people would normally use for graphing.

We did discuss having the size of the window determine the size of the graph, however that leads to the opposite problem of then making it hard to consistently get graphs of the same size, so we didn't go that route.

Considering previously the size was hard set, I think we'll wait for the use case where someone has a lot of different sized graphs they want to output before worrying about the user experience doing that.

PR #9433, opened 2024-11-08
9397 HTML Documentation

### LINKED ISSUE #9397: Write code to render tags to HTML
Issue for changing over the documentation system to produce HTML files for the new web based documentation instead of static pdfs.

Markdig already supports doing this, so we just need to convert our doco tag system to markdown and handle inserting memos (which are already markdown) into that structure correctly.

Will make new branch off of the refactor branch for this issue, so that it can be reviewed and merged while we work on the next part.

par456:
Resolves #9397


ric394:
retest this please jenkins

PR #9444, opened 2024-11-13
Change column name to soybean.leaf.nodenumber

### LINKED ISSUE #9443: Main Stem Node Number Data in Soybean Test Set does not appear in the Statistics
### What happened?

The data in OBS data sets has a different name to that in the model output and so these data are not merged by the OBS vs Pre stats.
The data should use the standard APSIM reporting name.

Windows

hut104:
resolves #9443


hut104:
Merge please @hol353 .  No changes to the model. Just new data

PR #9461, opened 2024-11-22
Update the geocoding key

### LINKED ISSUE #9457: Geolocator api key is expired
### What happened?

When looking for a soil via place name, the user is told "Cannot find a latitude/longitude from the specified place name.", no matter what town they type in.

It appears the error from google relates to an invalid key:
image

The key is defined [here](https://github.com/APSIMInitiative/ApsimX/blob/b2546ea47df443ba7384f7b11408c78453d06b8f/ApsimNG/Presenters/SoilDownloadPresenter.cs#L34)

Linux

ric394:
resolves #9457

Updated the geocoding key


zur003:
I believe this also needs to be updated in Utility\WeatherDownloadDialog.cs

PR #9474, opened 2024-11-26
Improve WeatherSampler

### LINKED ISSUE #9472: WeatherSampler could be improved
### Describe the new feature

- add an option to give a range of specific years (as well as specifying specific years)
- have Tav and AMP calculated if not in the weather file 
- change the text-string file path to allow the user to browse to the file and make it a relative path 

Also note that the current "Specific Years" is incorrect and is taking the year before the intended year.

sno036:
- allow specification of a range of years
- fixed an issue where the years were offset by 1 under some conditions
- added an output for actual date sampled

resolves #9472

### DIFF
--- a/Models/Climate/WeatherSampler.cs
+++ b/Models/Climate/WeatherSampler.cs
@@ -23,6 +23,7 @@ namespace Models.Climate
                  "There are three sampling methods.\r\n" +
                  "1. RandomSample - whole years of weather data will be sampled randomly and independently until the duration specified in Clock has been met.\r\n" +
                  "2. SpecificYears - specific years can be specified. Weather data will be taken from these years in the order specified. Once all years have been sampled, the model will cycle back to the first year until the duration specified in Clock has been met.\r\n" +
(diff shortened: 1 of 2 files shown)


Models/Climate/WeatherSampler.cs
```diff
@@ -54,6 +55,9 @@ public enum RandomiserTypeEnum
             /// <summary>Specify years manually.</summary>
             SpecificYears,
 
+            /// <summary>Specify the first and last year manually.</summary>
+            SpecificYearRange,
+
```
hol353, line 60: Val, if you move this to after RandomSample, Jenkins might run green. The reason is that the .apsimx stores these as an integer e.g. 0, 1, 2. You have redefined what index 1 means. It used to be RandomSample. Now it means SpecificYearRange.

sno036:
[like]  Val Snow reacted to your message:

sno036:
Thanks @hol353. I think I would prefer to fix the tutorial though. The order in that list is what is presented to the user and the order (to me anyway) makes conceptual sense. There is not much usage of this component (yet) so maybe make it look nice now?

PR #9477, opened 2024-11-28
Remove Simulation as valid parent for SlopeEffectOnWeather model

### LINKED ISSUE #9449: Slope effects on weather module does not work when implemented at the simulation level above separate zones
### What happened?

If implemented at the simulation level, with multiple zones having different aspects and slopes, slope effects does not distinguish these params between the zones. If those different aspects/slopes are implemented at the zone level, within separate simulations, it works OK. The attached sim file, with two sims and zones using similar slope/aspect params, demonstrates this. The overlap in output can be seen by viewing the charts using 'vary by sim' or 'vary by zone'.
[test1.zip](https://github.com/user-attachments/files/17759507/test1.zip)

Windows

ric394:
resolves #9449

- A converter has been created to automatically make offending files conform to avoid future bugs.


BrianCollinss:
Is this intentional?

- myWindSpeed=0 by default, but it is set to be equal to weather.Wind (line 329) before being used.
- However, myVP=20 by default, but it is never adjusted using dVapPressure (line 366). That is, weather.VP is always equal to myVP.

par456:
Have you confirmed that this solves the linking problem with multiple zones in a simulation (I just don't trust the Locator)? 

Could really use a unit test here to make sure of that and future proof it.

par456:
For the converter, if it only has 1 zone and doesn't slope under it, move the slope from simulation to under the zone.

If it has multiple zones, copy the slope to each zone.

That way it should function the same.

PR #9485, opened 2024-12-01
change return value of divide function if denominator is zero

### LINKED ISSUE #9484: Divide function returning zero when denominator is zero
### What happened?

The divide function has a test for when the denominator is zero, but it is returning zero it the test is true. This is wrong, it should be expected to return an error or infinity. To avoid crashes, the function could return the maximum value for a double (```double.MaxValue```). This is a much more sensible value than returning zero...

Windows

rcichota:
Resolves #9484
Returning the maximum value for double for the cases when denominator of the divide function is zero. This should be much more sensible and don't break expectations for users (not too much at least). The current return of zero is just not right.


hut104:
A few comments.

1) What if the first value is zero?  what is zero/zero?
2) What if you have three values?  1/0/0.5  I think this will still return zero because the first dived with give double.maxvalue and the second divide will return zero again.

rcichota:
Good questions Neil
That probably should be a catch and return error if zero/zero (or maybe even any division by zero). But it seems this will break a few simulations... Not sure what to do about them.
However, if a value is to be returned, I think that returning MaxValue is less problematic (less 'wrong') than returning zero.

rcichota:
Added a check for the nominator being zero, which then simply stops the function and outputs zero (this means 0/0 will return 0.0). Also added account for the sign of the number to return in case of divided by zero (now it will be -double.maxValue of +double.maxValue).
Need to decide what to do with a sequence of values in which one is zero. Currently if the first is zero, then the results is zero, if the last is zero, the return is +- double.maxValue, but if zero appears in the middle, the results less 'predictable' (i.e. calculated, e.g. 2/0/2 will return half of double.maxValue)...
Zeroes are hard to handle. No wonder they were outlawed in the middle ages...

rcichota:
This seems to work now. Very small diff in sorghum only... it helps a lot with the Plantain model.
What you guys think?
@hol353  @par456 @hut104  @HamishBrownPFR

par456:
Retest this please jenkins

Just going to re-run, the stats change may not have been from this PR.

rcichota:
@hol353 @par456 
Any chance of merging this one?
I have a PR for PlantainForage that I want to complete but is dependent on this one...

rcichota:
@hol353 @par456
Sorry to be a bother, any chance of merging this one?

par456:
We had a chat about this on what the wider effects might be, and I think most of this will work. However, I did some reading up on Double.MaxValue and it makes more sense to instead return Double.PositiveInfinity. 

The reason being that if you do any operation on Double.MaxValue that makes it larger, you end up with Infinity anyway and once it's infinity it also protects users from doing maths on the result and getting real values when they really shouldn't.

In terms of your case where you need to use the value elsewhere, you would then have the difference you need to tell actual 0 and infinity apart, and handle the infinity in a sensible way.

rcichota:
Thanks @par456 I will change that and we'll see if things will work.
I was a bit uncomfortable with simply returning the max, as you said one can then try to add to it, which would cause a crash anyway... I am happy with that, better than retuning a zero :)

PR #9487, opened 2024-12-02
9476 Added Set Emergence script to management toolbox

### LINKED ISSUE #9476: Add Emergence Date property to Wheat Example Sowing Script
### Describe the new feature

We can set the emergence date for Wheat crops and do so in the Wheat validation file, however the script in the Wheat Example only uses the default sowing script which doesn't have that property. We should update that so people don't need to keep finding that setting in the Wheat Validation file.

par456:
Resolves #9476

Decided to add this to the toolbox instead of the wheat example since it may be usable in other crops as well.


par456:
retest this please jenkins

par456:
retest this please jenkins

ric394 (approved):

PR #9489, opened 2024-12-02
Fix message conditions when checking for csv files

### LINKED ISSUE #9488: Remove csv not succcessful message on csv switch in Models run
### What happened?

If a csv file is not created that matches the files name a message is output that says the output is unsuccessful, when it normally does create a csv file.
Instead it should check that a csv is created that includes any report names in the csv file name.

This bug needs to be removed as it's misleading.

Windows

ric394:
resolves #9488


par456:
retest this please jenkins

par456:
retest this please jenkins

PR #9496, opened 2024-12-04
9434 Accumulate Function Resetting at Harvest

### LINKED ISSUE #9434: Accumulate Function Harvesting event resets value to 0 before Reporting happens
### What happened?

In the case of Soybean, if you try and report NodeNumber at harvest, it will always report 0 because the Accumulated Value that holds NodeNumber is changed prior to the report being run.

This is obviously designed for pasture plants, so that you can harvest them without ending them and have the accumulated values reduce, but with soybean (and I assume other crops) the harvest factor is set to 100% removal, so it's cleared to 0 instead.

Windows

par456:
Resolves #9434


par456:
retest this please jenkins

par456:
So the solution I've written up here is to add a "PostHarvesting" event (using the naming convention that "PostPhenology" has) which is called after the "Harvesting" event when reporting happens. 

The PostHarvesting event can then be listened for by organs which have a biomass removal step that happens at harvest, but after the Harvesting step (previously this was done in a loop in the Plant Harvest function where each organ's harvest function was called). I've then also attached the Accumulators to PostHarvesting for their removal step, which previously would happen at the same time as reporting (which lead to inconsistent results due to event timing).

Only question I have is if we keep the event as "PostHarvesting" or if a better name would be good here. Considering all these are biomass removal functions, maybe something like "BiomassRemovalAfterHarvesting" might be a better name if we wanted something more descriptive.

BrianCollinss:
Can they simply be named PreHarvest and PostHarvest?

par456:
I wouldn't change the name of the Harvesting event as that is used by everyone in their report frequencies, so it would be a difficult change for everyone to stop using Harvesting and starting using PreHarvest, which also doesn't make as much sense in the reporting context.

par456:
retest this please jenkins

PR #9500, opened 2024-12-04
Add generation of geojson files

jmadden173:
The package `fiona` was not able to append collection of objects to the same geojson file so each sensor is separated into separate `.geojson` files. Example command generated below.

```bash
oasis metompkin MetompkinFarm_WC_ST_SC_220719-231109.csv/MetompkinFarm_WC_ST_SC_220719-231109.csv bla.geojson
```

Confirmed that it works with `qgis` timeseries functionality.


jmadden173:
Meant as a PR to another repository

Issue #9503, opened 2024-12-05
Update from .NET 6 to .NET 8


par456:
.NET 6 hit it's end of life [last month](https://devblogs.microsoft.com/dotnet/dotnet-6-end-of-support/) so it's time once again to move up our .NET version to .NET 8

This will have to be done very carefully, as we have a lot of packages and such that apsim relies on that will also need updating to the .net 8 versions. Once completed, we should be able to undo #9501 so that the warnings become errors again for the future.

Hopefully this process is a lot less painful than the 3.1 to 6.0 update from a few years ago.

In the meantime until we do this update, track any versioning problems with packages or visual studio here so that we can know to test them again once updated.

PR #9504, opened 2024-12-05
Initial conversion to Net 8.0

zur003:
Working on #9503 

This version compiles, but doesn't include package updates. Those are yet to come.

### DIFF
--- a/APSIM.Cli/APSIM.Cli.csproj
+++ b/APSIM.Cli/APSIM.Cli.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net6.0</TargetFrameworks>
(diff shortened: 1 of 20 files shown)


zur003:
It looks like the docker image that Jenkins uses to build ApsimX will need to be update to use .Net 8.0 before this will work.

par456:
Looking at the docker scripts for this, it's based on a 6.0 image, so it'll be a pain to add both 6 and 8 in at the same time, and if we change it to net 8, then all the other PRs won't run until this is merged. We could potentially install 8 as part of the build process, but that'll also blowout the build times while it's active like that.

Considering it's halfway through a friday, we might need to sit on this PR for another day where we do a planned switch across to 8 and get the conversion done in one jump and have time for all the followup testing it'll require.

Thanks for getting a start on this Eric, much appreciated.

zur003:
Resolved with #9592

Issue #9458, opened 2024-11-20
AccumulationFunction and Report's Sum function give different values


BrianCollinss:
### What happened?

[Wheat 2.zip](https://github.com/user-attachments/files/17823736/Wheat.2.zip)

I am creating a new function based on `AccumulateFunction` to address #9406 and noticed that it does the summation (in `PostPhenology`) twice at emergence. It could be the reason for this difference.

### What Operating System are you on?

Windows

BrianCollinss:
@ric394 If I am right about this issue, this might have some impact on the validation stats and many users' works.

BrianCollinss:
In my example, the difference between the two functions is 0.024, which is exactly equal to the value of LAI at Emergence. 

![image](https://github.com/user-attachments/assets/7f342d96-6e0a-4032-8c01-a6e68818f80a)

par456:
To follow up on this, I changed the end point to flowering instead, and then also added a line where I used the dates for sowing and flowering instead. And now I have 3 ways of measuring that period, and all of them are coming out different:

![image](https://github.com/user-attachments/assets/46432d1b-c960-4484-b2a4-be8d7b6802d8)

Something weird is going here with the different ways these work, they should all be reporting the same value.

BrianCollinss:
As I mentioned, it seems `PostPhenology` is triggered twice at emergence. Beside that, it can be related to StartOfDay and EndOfDay being used for each of these algorithms.

par456:
So have checked and PostPhenology does only run once at emergence, you can also replicate the same differences by comparing between Flowering and Maturity for example, which doesn't include emergence.

The difference between the events and the dates are is a timing problem. The events fire during the day, while the date measure at the end. Sum between Events is inclusive of the first day, but excludes the final day, and sum between dates is inclusive of the start and end day.

For example, in the case of this:

```
Sum of [Wheat].LAI from [Wheat].Sowing to [Wheat].Flowering as SumPhenology
Sum of [Wheat].LAI from 11-June to 26-Sep as SumBetweenDates
```

The events will measure from the 11th of June to the 25th of September as on the 26th of September, the Flowering event will happen meaning it will stop measuring when it comes to process the report at the end of the day.


But this is not the case for an accumulator. If I measure something like Weather, it behaves consistent with the sum between events (inclusive of first day, exclusive of last day), however when measuring something like LAI, it adds up a different set of values. 

[...]

PR #9514, opened 2024-12-06
9458 Change Accumulator update event

par456:
Working on #9458

At the moment the Accumulator function updates in the PostPhenology event which is part-way through the day, this leads to inconsistent results compared to "sum of" in reports as sometimes the value being accumulated is modified after that event.

This is a test of changing it to the DoUpdate event nearer to the end of the day to see what effect it has on stats.


par456:
Closing this for now, as it changes too much in the stats. But we should consider refactoring this in the future due to the original issue.

PR #9521, opened 2024-12-08
Forage plantain

### LINKED ISSUE #9520: Improvements to forage plantain
### Describe the new feature

There are a few improvements that need to be brought into the release version:
Revise the parameters for a few processes (growth limiting factors, N concentration, biomass partitioning)
Update the test simulation, adding new dataset(s).

rcichota:
Resolves #9520 
Check model performance after a series of changes to the model, and new dataset


rcichota:
After a number of changes to the model (and some cleaning up to the validation simulation), I think that this version is now ready to be merged, as the stats are all better than were before.
There will be some more work on this model, with focus on multi-species simulations, but this will be done slow over the year...

PR #9524, opened 2024-12-09
Allow DCaPST AmbientCO2 value to be read from weather component.

### LINKED ISSUE #9523: Need to update concentration of carbon dioxide (CO2) in the atmosphere from 360ppm to 420ppm for Sorghum and Wheat.
### What happened?

The APSIMNG current default value of the concentration of carbon dioxide (CO2) in the atmosphere is 360ppm.
Update the current value from 360 to 420ppm for use in DCaPST (AirCO2 in SorghumCropParameterGenerator.cs WheatCropParameterGenerator.cs) and CO2Modifier in Sorghum.json.

Windows

joesaddigh:
Resolves #9523 

DCaPST currently uses a default value for the concentration of carbon dioxide (CO2) in the atmosphere, of 363ppm. This value is out of date. Rather than simply change the harcoded value to match the current value of 420ppm, DCaPST will be udpated to use the Weather.CO2 value. This Weather.CO2 can be set via the CO2Value component. DCaPST will now error if no CO2Value component has been added to the Weather component.


BrianCollinss:
Some simulations in the DCaPST validation file are set to run for years in the 1990s and 2000s. Is this change appropriate? Shouldn't a CO₂ setter be included in each of those simulations to set CO2 at the appropriate level?

joesaddigh:
> Some simulations in the DCaPST validation file are set to run for years in the 1990s and 2000s. Is this change appropriate? Shouldn't a CO₂ setter be included in each of those simulations to set CO2 at the appropriate level?

Hi Brian. Thank you for your comment. It is a good point. We are looking at implementing a better solution that allows more flexibility than a hardcoded value. Once implemented and tested, I will provide some more information on this Bug report. Hope that's ok?

BrianCollinss:
This might be helpful: https://github.com/APSIMInitiative/ApsimX/pull/9356.

hol353:
retest this please jenkins

PR #9531, opened 2024-12-10
Pcds

### LINKED ISSUE #9530: Make wheat model phenology consistent with PCDS
### Describe the new feature

A recent paper by Celestina etal 2023 has described the Population Culm Development Scale (PCDS) which provides a protocol for recording phenology in the field that is compliable to the population based stages that models predict.  
We should ensure the stage and phase names in APSIM are consistent with the PCDS so observations recorded following this protocol are directly comparable to APSIM outputs.

HamishBrownPFR:
resolves #9530


HamishBrownPFR:
@par456 @ric394.  This PR is failing on a documentation unit test and I am not sure how to diagnose that.  Any help appreciated

par456:
In the new system we have a unit test that checks if the documentation of a model changes (sections added or removed) so that unintended changes are detected. The error should have actually mentioned that, so I'll push a PR to address the error message.

But to resolve the change, we just have to run some code to update the resource files that it uses to check against. To do this:

1. Open up the DocumentTests.cs unit test file
2. Change the REGENERATE_FILES flag to true (line 26)
3. Run the TestDocumentationStructure test (either from the tests window, or right click on [Test] and choose "Run Test").
4. Set REGENERATE_FILES back to false

This will remake the test files and you can add them to your PR so that the test will run true during validation. This is still a bit complicated, we need to work out an easier way to do this like we do for updating resource files.

HamishBrownPFR:
Thanks Andrew.  I will follow that approach this time.
Perhaps that can be included in the actions that occur when you hit the update resources button in the gui so it does that and then updates the doc

par456:
retest this please jenkins

HamishBrownPFR:
@par456 @ric394.  I have reviewed this with @hut104 and we are happy the diffs are all minor and justified.
The improvement in MaturityDAS is because we removed the maturity phase and now maturity DAS reports days to harvest ripe.
The slight change in finalbiomass and other variables is because a couple of functions in the wheat model accumulated up to the maturity phase.  This has been changed so they accumulate all the way through to the harvest ripe phase now which has increased growth a little s some situations at the end of the crop.  Resultant diffs are small and these changes are generally helpful so the diffs can be tolerated.

PR #9584, opened 2025-01-14
9545 Model Descriptions Fix

### LINKED ISSUE #9545: Models do not display their descriptions anymore
### What happened?

image

Windows

par456:
Resolves #9545

This resolves a few issues that were either caused by the changes to documentation, or were actually broken before. Generic Presenter now has it's code restored and shows the markdown from the documentation, and DocumentationPresenter now has it's summary and remarks back for both the class and the methods/properties etc.


ric394:
Is this ok to merge with the changed stats?

par456:
It shouldn't be changing stats, would like to see another PR run first and see if it's the stats being out of date or something on this branch

PR #9600, opened 2025-01-20
Changes to make controled environment interpolate its data out to hou…

### LINKED ISSUE #9599: Hourly weather interpolation methods don't work with controlled environment
### Describe the new feature

Hourly weather interpolation methods assume a specific pattern to extrapolate daily min and max temp out to hourly values.  These are not suitable to controlled environment where temperature patterns follow a square pattern.

Solution is to have Controlled environment calculate its own hourly data that is appropriate given the CI settings and have interpolation methods access this.

HamishBrownPFR:
…rly values so hourly interpolation methods can be used with controled invironment experiments

resolves #9599


HamishBrownPFR:
@hol353 @par456 @ric394 can you please review and comment or merge if happy.
Thanks

PR #9603, opened 2025-01-20
Tt remaining

### LINKED ISSUE #9602: Create object to calculate how much Tt remains to pass between two stages
### Describe the new feature

A common pattern we have developed in our plant models is to calculate how much Tt remains between stages as shown in the screen shot below.

This approach is a bit cumbersome and error prone when copying functions to duplicates. 

Would be better to have a dedicated function to calculate the Tt remaining.  Would only need to have inputs of the stage names (with drop downs so it is not possible to get them wrong) and the rest can be done in code so there is no possibility to make config errors in the json model set up.

Image

HamishBrownPFR:
resolves #9602

### DIFF
--- a/Models/PMF/Phenology/Phenology.cs
+++ b/Models/PMF/Phenology/Phenology.cs
@@ -38,7 +38,8 @@ namespace Models.PMF.Phen
         /// -------------------------------------------------------------------------------------------------
 
         /// <summary>The phases</summary>
-        private List<IPhase> phases = new List<IPhase>();
+        [JsonIgnore]
+        public List<IPhase> phases = new List<IPhase>();
 
         /// <summary>The current phase index</summary>
         private int currentPhaseIndex;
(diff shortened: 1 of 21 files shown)


HamishBrownPFR:
@hol353 @par456 @hut104 @ric394 please review and comment or merge if happy

Models/PMF/Phenology/Phenology.cs
```diff
@@ -408,6 +409,22 @@ public IPhase PhaseStartingWith(String start)
...
+        /// Helper function to check if a particular phase is present between specifice start and end stages.
+        /// </summary>
+        /// <param name="startStage"></param>
+        /// <param name="endStage"></param>
+        /// <param name="checkPhase"></param>
+        /// <returns></returns>
+        public bool PhaseBetweenStages(string startStage, string endStage, IPhase checkPhase)
+        {
+            if ((checkPhase.Index >= StartStagePhaseIndex(startStage)) && (checkPhase.Index <= EndStagePhaseIndex(endStage)))
+            {
```
hol353, line 422: Shouldn't need to add an Index property to all Phases. You should be able to rewrite the test above to iterate through all phases looking at the start and end stage to determine if the phase is between the start and end stage.

HamishBrownPFR: @hol353 good call.  I have implemented the changes you suggest which tidies it up

HamishBrownPFR: @hol353 please check and comment further or merge if happy.

Issue #9613, opened 2025-01-26
Consolidate accumulation functions


HamishBrownPFR:
### Describe the new feature

We have 6 different accumulate functions that duplicate large amounts of code.

These could be consolidated into a single function

HamishBrownPFR:
Capturing discussion from #9614 

We need to:
1. Add a user interface that helps users select which type of accumulate function they want and hides irrelevant property fields
2. refactor General accumulation function to facilitate 1
3. Write a converter to move existing accumulate functions to all use the general function
4. deprecate the multitude of existing accumulate functions.

PR #9614, opened 2025-01-27
Add general accumulate functions 

HamishBrownPFR:
that can be configured to do the job of most the other accumulate functions

working on #9613

Still need to write a converter to move the existing accumulate functions onto this generic function then deprecate the other

### DIFF
--- a/Models/Functions/AccumulateFunctionGeneral.cs
+++ b/Models/Functions/AccumulateFunctionGeneral.cs
@@ -1,0 +1,310 @@
+﻿using System;
+using System.Collections.Generic;
+using Models.Core;
+using Models.PMF.Phen;
+using System.Linq;
+using APSIM.Shared.Utilities;
+
+
+namespace Models.Functions
+{
+    /// <summary>Accumulates a child function between a start and end stage or start and end events.
+    /// </summary>
+    [Serializable]
+    [Description("Adds the value of all children functions to the previous accumulation each time the Specified Accumulate Event occurs.  " +
+        "Option for crop models is to specify start and end stages for accumulation.  " +
+        "Option for all models is to specify start and end events for accumulation.  " +
+        "Option for all models is to specify start and end dates for accumulation.  " +
+        "If none are specified will accumulate every time the accumulate event occurs.  " +
+        "Optional full or partial removal of accumulated values can occur on specified events, stages or dates")]
+    [ViewName("UserInterface.Views.PropertyView")]
+    [PresenterName("UserInterface.Presenters.PropertyPresenter")]
+    public class AccumulateFunctionGeneral : Model, IFunction
+    {
+        ///Links
+        /// -----------------------------------------------------------------------------------------------------------
+
+        /// <summary>Link to an event service.</summary>
+        [Link]
+        private IEvent events = null;
+
+        [Link]
+        private Clock clock = null;
+
+        /// Private class members
+        /// -----------------------------------------------------------------------------------------------------------
+     
+        private double AccumulatedValue = 0;
+
+        private bool AccumulateToday = false;
+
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@par456 @hol353 @ric394 can you please review and comment or merge if happy

Models/Functions/AccumulateFunctionGeneral.cs
```diff
@@ -0,0 +1,310 @@
...
+        public string StartStageName { get; set; }
+
+        /// <summary>The end stage name</summary>
+        [Description("(optional for plant models) Stage name to stop accumulation")]
+        [Display(Type = DisplayType.CropStageName)]
+        public string EndStageName { get; set; }
+
+        /// <summary>The start event</summary>
+        [Description("(optional for any model) Event name to start accumulation")]
+        public string StartEventName { get; set; }
```
hol353, line 63: Because most of these properties are optional, you end up with a lot of if statements below checking for whether the user has put a value into the various property. It also lets the user put in nonsensical values e.g. a StartStageName and a StartEventName and a StateDate. 

So I'm unsure about having a general accumulate function vs a collection of specific functions. There are pros and cons to both approaches. I'm leaning towards lots of specific functions that obvious to the user in what they do and how to parameterise them. We probably need to refactor them all to make them easier to use and rename them to make it more obvious what they do. But I could be talked into a more generic approach like Hamish has presented here.

What do others think?

HamishBrownPFR: I started off with yet another accumulate function to add as there was not one already that worked how I needed it to.  Understand that this one could be confusing to parametrize and have tried to add clear descriptions to the parameter fields.  Some checks to make sure parameterizations make sense could be useful.  Or perhaps a multi-level view where users first choose what sort of accumulate function they want and then the gui only shows the parameters relevant to that.  I would need some help to implement something like that

par456: If we want to keep the functions separate, would it make sense to instead use some static utility functions to remove the duplicated code?

Otherwise yeah, this generic accumulate function would need a much more advanced presenter/view so that it's not showing all the properties for all the use cases at all times, which would be confusing for uses.

HamishBrownPFR: I am happy to implement what ever is needed but need a decision to forage ahead.

1.  Keep existing numerous accum functions and add another to do the job I need
     pros:  Quick to do.  Maintain existing pattern
     cons:  Maintaining more code than needed.  Not apparent to the general user which function does what 
2. Write a more sophisticated ui component for the generic function and deprocate the duplicates
    pros:  Smaller code base to maintain.  UI will help guide users through what accumulate method to use and   
    how to parameterise it.  Model code implementation already written (as protype at least).
    cons:  someone will need to provide me a bit of guidance to get the UI component set up.  New working 
    pattern 
3. Keep existing accum functions but write static utility functions to reduce duplicate code.
    pros:  Keep similar working pattern.  Remove duplicate code
    cons:  background programming needs doing and I would need some pointers to get started

Which path do we want to take and if it is 2 or 3 who is willing to help me get it done?

par456: I think for the short term here, go with the first option you've listed as it keeps to the existing pattern. Then we can create an issue to log this as a problem that needs refactoring.

I think option 2 is the best solution, but it would take some time to redesign that and make a new GUI view for it and I don't see it getting to the top of the to do list for a bit. It would need a proper design session to make sure we cover the edge cases and scope correctly. And if we are going to spend the time to refactor these functions, we should do it the correct way rather than option 3 which is only a partial fix.

PR #9622, opened 2025-01-28
Remove hard coded references from classes 

HamishBrownPFR:
working on #7623

Need to remove hard coded paths from source code to facilitate model restructuring


par456:
The changes to the linking seem fine, but I do wonder if wheat.json and croptimizer .apsimx files are supposed to be in this PR?

HamishBrownPFR:
@par456 yes, they need to be there.  Removing the hard code paths from the links mean we then need a reference function in the .json so the class can still find the necessary references.  For croptomyzer, it has a wheat model in replacements with a zadok class in it so it also needs a reference function included so it can still find the necessary reference

PR #9635, opened 2025-01-31
9593 Documentation View Fixes

### LINKED ISSUE #9593: DocumentationView Public Methods table no longer displays.
### What happened?

It looks like #9545 may have exposed a lurking bug when drawing a markdown table with entries that contain newlines. The entire table will fail to display if there is an entry containing a newline as seen below (the description of the methods FindByPath and FindAllByPath from Model both contain multiple newlines): 

Image

Windows

### LINKED ISSUE #9594: DocumentationView is slow to display

APSIM.Shared/Documentation/CodeDocumentation.cs

Specifically GetDocumentationElement.

### What needs to be changed?

XmlDocument.SelectSingleNode misses are quite expensive when working with large files. Now that potentially the entire inheritance tree is walked looking for remarks and summaries it can take quite a while to create the documentation table. A way of caching the relevant content of the XmlDocument on load for much quicker lookups could solve the problem.

Takes approximately a second to display.

par456:
Resolves #9593
Resolves #9594

The fixes to get the Documentation View working again made it a little slow, so this improves the experience a bit. A cache is now used for method/variable lookup so it only needs to go loading the xml once, and the page load is split into 4 steps so it load gradually instead of waiting for everything and loading all at once.

Also fixed a bug in the datatable to markdown code that was causing newlines to break tables in the view.

### DIFF
--- a/APSIM.Shared/Documentation/CodeDocumentation.cs
+++ b/APSIM.Shared/Documentation/CodeDocumentation.cs
@@ -132,14 +132,15 @@ namespace APSIM.Shared.Documentation
 
             if (typeLetter == 'P' || typeLetter == 'F' || typeLetter == 'E')
             {
+                string cacheName = fullName+tagName;
                 string description = "";
-                if (modelDescriptionCache.TryGetValue(fullName, out description))
+                if (modelDescriptionCache.TryGetValue(cacheName, out description))
                     return description;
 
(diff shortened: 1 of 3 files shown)


ric394:
I noticed that a remarks column has the same text as the description column in some models even though the model doesn't have remarks 🤔 
image

par456:
Good catch, fixed that bug.

PR #9636, opened 2025-01-31
Quick fix to prevent Manager completion crash.

### LINKED ISSUE #9631: Completions cause crashes while editing manager scripts.
### What happened?

Following the upgrade to dotnet 8, manager scripts completions can now cause APSIM to crash unexpectedly. The error message comes from GtkSharp, and is hidden when launched via the executable, but when launched via calling dotnet on the .dll you get the following error message:
```
> dotnet .\bin\Release\net8.0\ApsimNG.dll
Exception in Gtk# callback delegate
  Note: Applications can use GLib.ExceptionManager.UnhandledException to handle the exception.
System.ArgumentException: 'ApsimNG.Resources.Function.png' is not a valid resource name of assembly 'GtkSourceSharp, Ver   at Gdk.PixbufLoader.InitFromAssemblyResource(Assembly assembly, String resource)
   at Gdk.Pixbuf..ctor(Assembly assembly, String resource, Int32 width, Int32 height)
   at GtkSource.CompletionProposalAdapter.GetIcon_cb(IntPtr inst)
   at GLib.ExceptionManager.RaiseUnhandledException(Exception e, Boolean is_terminal)
   at GtkSource.CompletionProposalAdapter.GetIcon_cb(IntPtr inst)
   at Gtk.Application.Run()
   at UserInterface.UserInterface.Main(String[] args) in C:\Users\ilhuber\NextGen\ApsimX\ApsimNG\UserInterface.cs:line 38
```
Oddly, the code works properly initially and the crash is only caused after some time has passed. Further, the crash doesn't happen at all when built for debug. It's a release-specific issue. The code for the constructor that throws the exception can be found [here](https://github.com/GtkSharp/GtkSharp/blob/b7303616129ab5a0ca64def45649ab522d83fa4a/Source/Libs/GdkSharp/Pixbuf.cs#L110). It appears to my uneducated eye that the jit is causing the behavior of Assembly.GetCallingAssembly() to change via inlining.

The problematic APSIM code is found [here](https://github.com/APSIMInitiative/ApsimX/blob/d184d553b395b69ec06d7ef1337f13c28ba5942e/ApsimNG/Classes/CustomScriptCompletionProposal.cs#L50). Adding the NoInlining decorator to this getter is enough to prevent the error from occurring (supporting the jit inlining is causing problems theory), but that is a solution that I am not a fan of. It is probably better to make the assembly reference explicit (also probably wasteful to be loading these images every time).

Another area this constructor is used is [here](https://github.com/APSIMInitiative/ApsimX/blob/d184d553b395b69ec06d7ef1337f13c28ba5942e/ApsimNG/Views/IntellisenseView.cs#L353). This usage seems more immune to getting inlined into another assembly, but it may be wise to similarly make the assembly reference explicit.

Windows

ilhuber:
Resolves #9631 

This prevents manager completions from crashing the UI. The completion images still fail to properly display (as they did before) but I don't think that is an APSIM issue.

I favor making the assembly reference explicit here rather than relying on the behavior of null as first argument even though this wouldn't get inlined.

### DIFF
--- a/ApsimNG/Classes/CustomScriptCompletionProposal.cs
+++ b/ApsimNG/Classes/CustomScriptCompletionProposal.cs
@@ -42,18 +42,15 @@ namespace UserInterface.Intellisense
             }
         }
 
+        private static readonly Pixbuf functionPixbuf = new(typeof(CustomScriptCompletionProposal).Assembly, "ApsimNG.Resources.Function.png", 16, 16);
+        private static readonly Pixbuf propertyPixbuf = new(typeof(CustomScriptCompletionProposal).Assembly, "ApsimNG.Resources.Property.png", 16, 16);
+
         /// <summary>
         /// Gets the GdkPixbuf for the icon of proposal.
         /// </summary>
         public Pixbuf Icon
         {
-            get
-            {
-                // tbi
-                Pixbuf functionPixbuf = new Pixbuf(null, "ApsimNG.Resources.Function.png", 16, 16);
-                Pixbuf propertyPixbuf = new Pixbuf(null, "ApsimNG.Resources.Property.png", 16, 16);
-                return isProperty ? propertyPixbuf : functionPixbuf;
-            }
+            get => isProperty ? propertyPixbuf : functionPixbuf;
         }
 
         /// <summary>


par456:
> This prevents manager completions from crashing the UI. The completion images still fail to properly display (as they did before) but I don't think that is an APSIM issue.

Yep, the gtk library we use for the completions doesn't give us a lot of control over the menu element which is annoying.

Issue #9641, opened 2025-02-04
Add ability to model enhanced-efficiency fertilisers


hol353:
### Describe the new feature

# Background

Enhanced-efficiency nitrogen (N) fertilizers are of interest to agricultural industries because of their potential to reduce N loss to the environment and to provide agronomic benefits through increased N use efficiency, increased yield, or both (Trenkel 2010). The most commonly used EEFs that target nitrate leaching losses or nitrous oxide emissions either keep the N in forms less susceptible to loss (e.g., ammonium N using nitrification inhibitors (NI)) or slowly release the N to the soil (e.g., polymer-coated, controlled-release fertilisers (CRF)). Urease inhibitors (UI) delay or slow down urea hydrolysis, thereby potentially reducing ammonia N volatilisation losses due to keeping the ammonium N concentration lower and limiting the pH increase accompanying hydrolysis. Depending on timing of N cycling they could also impact on subsequent losses of nitrate N. 

# Requirements

Release of existing forms of N, P (or K in future): 

* Allow different mathematical forms to describe a range of release patterns (linear constant, 3-stage, sigmoidal release). 
    - These should be able to be described in such a way that they result in a fraction of the total remaining amount being released.  
    - At least initially we would want the user to be able to code the release description, or do we want to prepare a range of different formulations?
* Allow this fraction to be responsive to factors such as soil temperature, soil water (or matric potential) and allowing these to be specified in different ways (‘standard’ APSIM way relative to opt and to LL, DUL, SAT, or using Q10 relative to default temperature and e.g. water-filled pore space like the modifications used for denitrification) 
* Allow a lag phase that does not release and the duration of which may be temperature or water dependent. 
* Include a point at which the remainder is released completely (e.g. when 0.995 released) 
* Allow multiple slow-release pools to be tracked – e.g. if the previous CRF has not fully released yet, or if using both rock phosphate and a CRF N product 
* Allow specification of the released product 
    - In case of mixtures, e.g. N, P, K, these should probably be treated as separate products as they are likely to have different release rates (hence another reason to allow multiple pools to be tracked)

hol353:
This has been done - #9642

PR #9642, opened 2025-02-04
Added ability to add slow release fertilisers

hol353:
Working on #9641


par456:
So in this new process for fertiliser, calling 'Apply' in an operation or manager just puts the fertiliser into a pool that is then released during the DoFertiliserApplications event? Rather than immediately putting it into the soil like it previous did.

How might this effect simulations (like grapevine) that have both a fertilise top-up and a fertilise on fixed date, which are both running in DoManagement, but don't take effect until DoFertiliserApplications, so the check in fertilise top-up is always a day behind.

I don't think this is a problem based on the results of the validation and POStats, but thought it was worth mentioning.

hol353:
The DoFertiliserApplications event is immediately after DoManagement (which is what the manager and operations use). So it won't be offset by a day. In fact there should be very little difference.

Issue #9647, opened 2025-02-04
Indexed Expression Function


ilhuber:
### Describe the new feature

As part of a soybean phenology project, we've needed to implement a variant of the Wang-Engel temperature function that has parameters that vary with plant stage. This function also does not have the same structure as the Wang-Engel already present in APSIM (used by grapevine): there is no cardinal temperature and an extra 0-1 power smoothing factor. 

Rather than creating a second Wang-Engel function or further complicating the one that exists, I think it may be a good approach to create a version of the expression function that can be used along with subdaily interpolation. The user could specify what variable in the expression is to be treated as index, and the remaining variables can come from the model (in our case several PhaseLookupFunctions). This could offer a lot of flexibility in defining new subdaily responses in general, not just for our specific case.

HamishBrownPFR:
@ilhuber I find the expression function becomes difficult to work with when expressions are long.  would it be possible to modify the view so expressions can be entered over multiple lines

PR #9648, opened 2025-02-04
Add new IndexedExpressionFunction.

ilhuber:
Working on #9647 

Not ready for merge yet, will need to do some more testing.

Is there a way to EventSubscribe to every daily clock tick? I want to avoid repeatedly Convert.ToDoubling or IFunction.Valuing stuff whose results are going to be the same (when I didn't avoid this it was VERY slow), but also don't want to miss anything.

There is a lot of similar code to ExpressionEvaluator, it could make sense to split this off into some static helper class (may be useful to the report as well, which uses the ExpressionEvaluator).


ilhuber:
The function appears to work as expected in most cases. This is ready for review (if it is something that is wanted at all).

PR #9649, opened 2025-02-05
Wheat tidyup

HamishBrownPFR:
working on #7623

Removed some more index columns from observed files and renamed a number of column headers that should match an APSIM variable but had an incorrect name


HamishBrownPFR:
@hol353 @par456 I have removed some columns that had index data which is why they are showing up as missing in the stats.  I have also renamed some columns that had the wrong header name so the stats are picking them up now.
Recommend accepting stats and merging as the stats are picking up more data now.  Please review and merge if happy.

BrianCollinss:
Juts noting that MaturityDAS now has an RMSE of 21.97 days.

HamishBrownPFR:
Thanks @BrianCollinss.  There were a few columns in obs sheets named Wheat.Phenology.ReadyForHarvestDAS which is not an APSIM wheat output.  I have renamed these columns to Wheat.Phenology.MaturityDAS so the stats now pick them up which is what caused the change.  There is one simulation that is giving a very bad fit causing the large RMSE but my preference is to fix that as part of a seperate pull request and keep this one to just renaming of columns

PR #9652, opened 2025-02-05
Fix soil water references in wheat validation

### LINKED ISSUE #9651: Variable [Wheat].SoilWater.Es needs to be changed to [ISoilWater].Es
### What happened?

The Maricopa tests in the wheat simulation use SWIM because the protocol for configuring the experiments required thin layers.  We output ET for a few simulations as ([Wheat].Leaf.Transpiration + [Wheat].SoilWater.Es + [MicroClimate].PrecipitationInterception) but this does not work for the Maricopa test that do not have SoilWater.  Need to change to ISoilWater so it works across all soil water models.

I am not sure why the build does not throw and error for the Maricopa experiments at the moment.  I came across this because APSIM started throwing an error for these experiments locally.

Windows

HamishBrownPFR:
resolves #9651


HamishBrownPFR:
@hol353 @par456 please review and merge if happy.

PR #9655, opened 2025-02-06
Fix wrong variable name

HamishBrownPFR:
working on #7623 

Fixed a wrong variable name in two obs files


HamishBrownPFR:
@hol353 @par456 @ric394 please review and merge if happy.  I have added some outputs to the harvest report so all the obs data in the test set is now compared to a model output

PR #9661, opened 2025-02-09
Add APSIM.Workflow project

ric394:
resolves #9661

Introduce the APSIM.Workflow project for use with the WorkFlo API for submitting jobs to Azure


ric394:
retest this please jenkins

rcichota:
Hi @ric394 
I don't seem to be able to build APSIM at the moment, VisualStudio complains about a reference to this new feature:
```Unable to find project information for 'C:\ApsimX\APSIM.Workflow\APSIM.Workflow.csproj'. If you are using Visual Studio, this may be because the project is unloaded or not part of the current solution so run a restore from the command-line. Otherwise, the project file may be invalid or missing targets required for restore.```
It seems that workflow is referred to in UnitTests but not in ApsimX.sln, so the IDE does not recognise it. Am I missing something? Do I need to add/load this proj into the solution or should it be coming from APSIM dev?

par456:
Looks like something is wrong with the configuration of that new project file, we'll get that fixed right away since that's obviously a pretty big issue.

rcichota:
Thanks @par456 
I have added to the solution here (ApsimX.sln) and it compiles... so should be an easy fix

PR #9662, opened 2025-02-10
remove unused daily outputs

HamishBrownPFR:
working on #7623


HamishBrownPFR:
@hol353 @par456 Please review and merge if happy.  
I have made a few more changes to tidy up the test set here:
1: Removed alot of variables from DailyReport that have not corresponding observations.  This reduces the size of this report substantially.  There will be some observation free graphs in the gui that no longer have the variable they are trying to plot.  However, it is not good practice to put variables that don't have matching obs into this table.  There are few sets of detailed graphs and outputs that can be copied round the set if we need to look at other things that are not in the DailyReport table.  
2: I have added a few variables that had observations but no corresponding prediction being reported
4:  I have removed some of the leaf variables from the Harvest report as they did not have much data and leaf data is not informative at the time of harvest anyway.
5:  I have added one variable to the harvest report that has additional data.

The stats some of the variables I have added are not that great and I think there could be some issues with the obs data there.  But I think it is better to merge them at this point and come back to tidy up the data later now we can see clearly where issues are.

PR #9681, opened 2025-02-14
Fix rounding errors in SLA data and add NDVI report

HamishBrownPFR:
working on #7623


HamishBrownPFR:
@hol353 @par456 @ric394 please review and merge if happy.  Stats have changed because variables had been rounded to much.  Retuning necessary precision to values causes them to deviate from the accepted values.

PR #9687, opened 2025-02-18
fix: add the APSIM.Workflow project to the ApsimX sln file

### LINKED ISSUE #9686: Unable to find project information for APSIM.Workflow
### What happened?

> I don't seem to be able to build APSIM at the moment, VisualStudio complains about a reference to this new feature:
> Unable to find project information for 'C:\ApsimX\APSIM.Workflow\APSIM.Workflow.csproj'. If you are using Visual Studio, this may be because the project is unloaded or not part of the current solution so run a restore from the command-line. Otherwise, the project file may be invalid or missing targets required for restore.
> It seems that workflow is referred to in UnitTests but not in ApsimX.sln, so the IDE does not recognise it. Am I missing something? Do I need to add/load this proj into the solution or should it be coming from APSIM dev?

Issue from merging PR #9661

Windows

ric394:
resolves #9686


ric394:
retest this please jenkins

PR #9692, opened 2025-02-18
fix: make sheets allow full table view via horizontal scrollbar

### LINKED ISSUE #9509: Report table horizontal scrollbar is not long enough
### What happened?

Is the length of the Report table's horizontal scrollbar intentionally limited? If there is a long string stored in a column, most of the columns will not be visible unless arrow keys are used.

image

Windows

ric394:
resolves #9509

# Notes

- this fixes an issue where the report table was not able to be fully seen because the horizontal scrollbar would not allow navigation to the end of the table.
- Note that if your column is too large for the scale of the program it will still be truncated as only whole columns can be viewed with GTK

### DIFF
--- a/Gtk.Sheet/Sheet.cs
+++ b/Gtk.Sheet/Sheet.cs
@@ -163,6 +163,11 @@ namespace Gtk.Sheet
         /// <summary>The number of rows to paint in the grid. If zero, then the data provider will determine the number of rows in the grid.</summary>
         public int RowCount { get; private set; } = 0;
 
+        public int ColumnCount 
+        {
+            get => ColumnWidths.Length;
+        }
+
         /// <summary>A collection of column indexes that are currently visible or partially visible.</summary>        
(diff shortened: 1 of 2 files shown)


Gtk.Sheet/Sheet.cs
```diff
@@ -163,6 +163,11 @@ public int NumberHiddenRows
         /// <summary>The number of rows to paint in the grid. If zero, then the data provider will determine the number of rows in the grid.</summary>
         public int RowCount { get; private set; } = 0;
 
+        public int ColumnCount 
+        {
+            get => ColumnWidths.Length;
```
par456, line 168: Can I get you to set up the ColumnCount to be similar to RowCount, where it's set in SetDataProvider() and Refresh() based on the DataProvider value.

Don't want to have the ColumnCount be reliant on the column width calculation having been run.

ric394: This has been corrected

PR #9695, opened 2025-02-19
9694 Renamed all Nconc to NConc

### LINKED ISSUE #9694: Inconsistent naming of NConc Variables
### What needs to be changed?

Throughout apsim we have a mix of NConc and Nconc variables and it's causing linking issues when people use the wrong on in observed files. Going to try and standardize them to NConc, and use a converter to fix reports and scripts.

par456:
Resolves #9694

Need to see if this effects stats or if it break validation.


par456:
Stats are the same, just renamed to the new correct names. Barley, Mungbean and Soybean also have additional stats that are now linking thanks to the names being correct.

par456:
Thanks Brian, yeah not going to merge this in until I can get these new POStats cleaned up somewhat. I assume when they were added to observed files they weren't cleaned since they weren't appearing, but want to double check that first.

par456:
I've gone and had a look at the data for those problem points, so I fixed, others I left for when the model is worked on next. Description below for each.

Mungbean
- Removed 0 from Pod.NConc
- The other strange point does calculate correctly from N and biomass, so there may be an entry error here. Not fixing in this PR, but should be addressed with a cleanup later.

Oats
- Leaf.Dead.NConc ranges from 0 to 1.7, which suggests it is in the wrong units.  Not fixing here however as its an existing PO stats.

Soybean.Pod.NConc
- Points at 0 on the Predicted are due to APSIM having not developed pods by the matching date (observation on the 10th, apsim pods appear on the 11th). These points are correct, and the model is not matching observed.
- Ames points seem to have the wrong units, but that's a particular dataset that only the modeller should touch.

Wheat.AboveGround.NConc
- Fixed unit error in FAR observed data

PR #9706, opened 2025-02-25
Add ability to specify subdaily temperature values.

### LINKED ISSUE #9705: Add possibility to specify subdaily temperatures for controlled environment.
### Describe the new feature

The current implementation of controlled environment assumes MaxT for daylight hours and MinT otherwise, it would be nice to be able to specify our own values if these are known. The only function in APSIM that uses these values that I am aware of is Models.Functions.HourlySinPpAdjusted.

ilhuber:
Resolves #9705

Keeps the same behavior as before if the SubDailyTemperature field is unset.

### DIFF
--- a/Models/Climate/ControlledEnvironment.cs
+++ b/Models/Climate/ControlledEnvironment.cs
@@ -21,6 +21,10 @@ namespace Models.Climate
         /// </summary>
         [Link]
         private IClock clock = null;
+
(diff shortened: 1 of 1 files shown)


ric394 (commented):
This one looks good code-wise. Do you have a moment to review @hut104 or @hol353?

PR #9712, opened 2025-02-26
Sub-daily irrigation time information

### LINKED ISSUE #9711: Add irrigation time to Apply method
### Describe the new feature

The Apply allows for the application of irrigation provided an amount of water, but also can take an irrigation duration to allow sub-daily irrigation events.  A time should be added to allow irrigation to be applied at a certain time of the day.
This will only make sense for SWIM and so an exception should be thrown if the apply method is used with waterbalance with sub-daily time info.

hut104:
Resolves #9711


BrianCollinss:
This can be used to improve runoff simulation when SoilWat is used.

par456:
The unit test that failed is just notifying that your changes have changed the documentation structure. You can use the "Upgrade Resources" button in apsim to update the unit test and push up the change to have it run green.

hut104:
> This can be used to improve runoff simulation when SoilWat is used. A similar approach can be implemented for rainfall.

The current implementation will throw an exception if you try to use sub-daily irrigation with WaterBalance (because it is a daily model).

PR #9714, opened 2025-02-26
Fixed soils with rocks in some simulations

### LINKED ISSUE #9713: Some soil physical properties invalid in test suite due to rocks
### What happened?

The percent sand, silt, clay, rocks adds up to > 100% in some cases where rocks% is specified. They need to be adjusted.

Windows

hol353:
Resolves #9713


hol353:
Change in stats is expected because the soils have been altered to be compatible with rocks.

PR #9725, opened 2025-02-28
Fix CN reduction for tillage

### LINKED ISSUE #9724: CN reduction for tillage is applied twice
### What happened?

1. On line 95 using `CNReductionForTillage`. However, `TillageCompleted` is not implemented so I am not sure if those variables are ever set. Also in `CNReductionForTillage`, CN reduction is calculated based on `weather.Rain`, not `soil.PotentialRunoff`. 

2. On lines 103-105 (using soil.PotentialRunoff, which accounts for canopy interception). 

Image

Windows

BrianCollinss:
Resolves #9724

Ready to merge. I removed the code from `RunoffModel` and tried to complete `CNReductionForTillage` in a way that does not need an extra event (`TillageCompleted`) to be published. If the event is implemented, these parameters can be removed from `RunoffModel` and moved to `CNReductionForTillage`. 

```
        /// <summary>Cumulative rainfall below which tillage reduces CN (mm).</summary>
        public double TillageCnCumWater { get; set; }
        /// <summary>Reduction in CN due to tillage()</summary>
        public double TillageCnRed { get; set; }
        /// <summary>Running total of cumulative rainfall since last tillage event. Used for tillage CN reduction (mm).</summary>
        public double CumWaterSinceTillage { get; set; }
```

In one example I used, this did not change results. The reason could be that `CNReductionForTillage` was never doing anything and all the reduction was done inside the `RunoffModel`. Now, at least the order and logic of things are correct.

### DIFF
--- a/Models/Soils/WaterModel/CNReductionForTillage.cs
+++ b/Models/Soils/WaterModel/CNReductionForTillage.cs
@@ -63,8 +63,14 @@ namespace Models.WaterModel
             {
                 // Tillage Reduction is biggest (tillageCnRed value) straight after Tillage 
                 // and gets smaller and becomes 0 when reaches tillageCnCumWater.
-                double tillage_fract = Math.Min(1, MathUtilities.Divide(cumWaterSinceTillage, tillageCnCumWater, 0.0));
-                double tillage_reduction = tillageCnRed * tillage_fract;
+
+                // We want the opposite fraction (hence, 1 - x). 
+                // If cumWaterSinceTillage = 0, tillage_reduction = tillageCnRed.
+                // If cumWaterSinceTillage = 0.3 x tillageCnCumWater, tillage_reduction = 0.7 x tillageCnRed.
+                // If cumWaterSinceTillage >= tillageCnCumWater, tillage_reduction = 0 (there won't be a continued reduction).
+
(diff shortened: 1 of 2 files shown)


BrianCollinss:
This is ready to merge if approved @par456

Models/Soils/WaterModel/CNReductionForTillage.cs
```diff
@@ -37,38 +37,46 @@ public class CNReductionForTillage : Model, IFunction
...
+            {
+                runoffModel = parent as RunoffModel;
+
+                cumWaterSinceTillage = runoffModel.CumWaterSinceTillage;
+                tillageCnCumWater = runoffModel.TillageCnCumWater;
+                tillageCnRed = runoffModel.TillageCnRed;
+            }
+            else
+            {
+                throw new Exception("Parent model of CNReductionForTillage must be RunoffModel.");
```
ric394, line 59: Instead of using this if statement, change line 18 from `WaterBalance` to `RunoffModel`.

ric394 (commented):
I'm happy with this code-wise. Is someone available to review from a scientific standpoint?

BrianCollinss:
@ric394 Thanks for the review. All I did was to clean up the runoff class so it does not do the same calculation twice. The only reason it did not use to run it twice (mistakenly) was that `TillageCompleted` event is not implemented.

hol353:
This looks fine to me. Go ahead and merge @ric394

PR #9729, opened 2025-03-04
Add in subdaily DeWit function.

### LINKED ISSUE #9728: Add in De Wit 1978 subdaily temperature interpolation method
### Describe the new feature

In our ongoing work on soybean phenology, this model has emerged as a favorite option for subdaily interpolation. You can find the equation described at https://doi.org/10.1016/0168-1923(94)90098-1 in section 2.2 (I can't find a more direct citation online). The existing temperature interpolation methods (ceres 3-hourly and Gourdriaan 1-hourly) produced some dubious results that this model was more likely to avoid.

We can allow the number of subdaily timesteps to be configured by the user.

ilhuber:
Resolves #9728 

This uses the max and min temperatures of the adjacent days, similar to the Gourdriaan method.

### DIFF
--- a/Models/Functions/SubdailyDeWit.cs
+++ b/Models/Functions/SubdailyDeWit.cs
@@ -1,0 +1,74 @@
+using System;
+using System.Collections.Generic;
+using Models.Climate;
+using Models.Core;
+using Models.Interfaces;
+
+namespace Models.Functions;
+/// <summary>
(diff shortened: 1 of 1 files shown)


hut104:
How is this different to the sub-daily temperature function added by Hamish brown? It seems to do the similar things. I wonder if we should evaluate if this new approach could replace earlier work or vice versa

PR #9734, opened 2025-03-07
Changed WaterBalance to move all solutes

### LINKED ISSUE #9733: Change WaterBalance to work with all solutes in simulation
### Describe the new feature

Currently NO3, NH4, Urea and Cl are hard coded. Any other solutes added to the simulation are ignored.

hol353:
Resolves #9733


Models/Soils/Nutrients/NFlow.cs
```diff
@@ -77,7 +81,7 @@ public void DoFlow()
             {
                 double nitrogenFlow = 0;
                 if (source[i] > 0)
-                    nitrogenFlow = rate.Value(i) * source[i];
+                    nitrogenFlow = rate.Value(i) * reduction.Value(i) * source[i];
```
par456, line 84: Might need to just check that reduction isn't null here since it's an optional function.

hol353:
@par456 This is now ready for final review and merging if ok.

PR #9738, opened 2025-03-09
Add rainfall start time and duration to the UI for SWIM

### LINKED ISSUE #9737: Make rain duration in SWIM settable and on the UI
### Describe the new feature

Resorting to original version (as @hut104 originally intended!)

sno036:
resolves #9737

### DIFF
--- a/Models/Soils/SWIM/Swim.cs
+++ b/Models/Soils/SWIM/Swim.cs
@@ -117,16 +117,6 @@ namespace Models.Soils
         /// </summary>
         private const double slcerr = 0.000001;
 
-        /// <summary>
-        /// default time of rainfall (hh:mm)
-        /// </summary>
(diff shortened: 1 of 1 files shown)


sno036:
Changed stats (most improved) with the rainfall duration reverting to 12 hours a day as a default - this was the default in Classic. Please accept these.

Tao1991:
Why does the SWIM3 soil water simulation not respond to the start time of rainfall?

I set up two rainfall scenarios, one with rainfall occurring during the night and the other during the day. I expected that rainfall timing would affect evaporation and therefore soil water dynamics, but the simulation showed no response to the start time of rainfall. Why is SWIM3 insensitive to rainfall timing in this case?
[test.zip](https://github.com/user-attachments/files/26623324/test.zip)

PR #9739, opened 2025-03-10
fix: make converter initialise values and thicknesses for solutes when converting

### LINKED ISSUE #9727: System.NullReferenceException when converting Classic soils
### What happened?

[Test.zip](https://github.com/user-attachments/files/19062108/Test.zip)

The `Chemical` model expects some Double values, which may not be present. Should the convertor add depths to Solutes, add zeros for N values, or the error be more informative? 

```
   at UserInterface.Presenters.ProfilePresenter.PopulateChemicalGraph(GraphView graph, Double[] thickness, Double[] pH, PHUnitsEnum phUnits, IEnumerable`1 solutes) in ApsimNG\Presenters\ProfilePresenter.cs:line 277
   at UserInterface.Presenters.ProfilePresenter.Refresh() in ApsimNG\Presenters\ProfilePresenter.cs:line 166
```

Image

Windows

ric394:
resolves #9727

Solutes should no longer have null thickness and initial value values. 
These are now given the defaults:

* Thickness is taken from the Soil's Water thickness value
* InitialValues is now given a double array the same length as the Soil's Water thickness value which is initialised with zeroes.

### DIFF
--- a/Models/Core/ApsimFile/Converter.cs
+++ b/Models/Core/ApsimFile/Converter.cs
@@ -7,9 +7,7 @@ using Models.PMF;
 using Models.Soils;
 using Newtonsoft.Json.Linq;
 using System;
-using System.Collections;
 using System.Collections.Generic;
-using System.Drawing.Text;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


ric394:
retest this please jenkins

par456:
Closing due to PR #9748

PR #9740, opened 2025-03-10
Display model type in DocumentationPresenter and GenericPresenter views

### LINKED ISSUE #9697: GUI Documentation/Generic Model view has wrong name
### What happened?

When loading up a model that just has the documentation view on the right, it uses the name of the node instead of the name of the class in the documentation. This can make it hard to look up the class for the model, since we can change the name of the model in apsim.

Perhaps both should be shown, name of the model and then the class, or the other way around, need to have a think. But the class name needs to be listed somewhere there.

Windows

ric394:
resolves #9697

Add a line at the top of the view to show the simple model type, enhancing clarity for users.

@par456 what do you think about this?

### DIFF
--- a/ApsimNG/Presenters/DocumentationPresenter.cs
+++ b/ApsimNG/Presenters/DocumentationPresenter.cs
@@ -44,25 +44,27 @@ namespace UserInterface.Presenters
             try
             {
                 //Default text while loading
+                string classInfo = "Model type: " + model.GetType().Name + $"{Environment.NewLine}";
+
                 markdownSummaryAndRemarks = DocumentSummaryAndRemarks(model);
-                view.Text = markdownSummaryAndRemarks.ToString().Replace("<", @"\<");
+                view.Text = classInfo + markdownSummaryAndRemarks.ToString().Replace("<", @"\<");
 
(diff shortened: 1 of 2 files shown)


lie112:
This functionality was available through the CLEM presenter that has a heap of additional functionality ready to be provided to other APSIM models (with some generic updates to the codebase) if needed. I might need to remove this aspect from CLEM or turn off this new feature when nested within the CLEM presenter to avoid duplication of the model name.
The CLEMPresenter provides tabs for the Properties, a pretty descriptive summary as MarkDown of the settings, internal model versioning that I think I'll remove, and some additional options for some models such as the formatted list of errors and warnings from the CLEM Zone component. This is handled through the CLEM presenter which identifies a CLEM component and wraps the desired presenter with the CLEM presenter. 
This also includes the path to the model and the description provided in the summary tag of the class. 
CLEM also provides some other presenters such as a repeated grouping presenter that will display the properties all repeated children if a model, and a categorical presenter that provides a grouping tree to the left that groups large lists of properties using an attribute provided with the property. 
I was looking to add a simple search bar that would filter the properties presented by a case insensitive contains search as the ruminant model has a vast number of properties that need to be filtered. 

image

PR #9747, opened 2025-03-11
S rum review prep

rcichota:
Working on #7869
Exposing (turning public) some variables to give more flexibility on set up crops
Tidying up the code a bit (make some variable names more explicit and some documentation)
Fixed a few small issues around calculation of N demand and handling dates in BiomassRemovalEvents
This also incorporate a change to the effect of N supply on canopy by Hamish


rcichota:
@hol353 @ric394 
This can be merged for now, so we can use this in some projects. There are a few more checks/update coming up in next weeks, but they can be in separate Pull Requests

rcichota:
Hi @hol353
I resolved the conflict, brought in the changes Andrew made last week. I think it would be good to merge this before we get other conflicts (also I want to use these changes in a project)

DeepCloner.Core/DeepCloner.Core.csproj
```diff
@@ -41,7 +41,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="JetBrains.Annotations" Version="2023.3.0" />
+    <PackageReference Include="JetBrains.Annotations" Version="2024.3.0" />
```
ric394, line 44: If this update was intentional can you separate this change into another pull request?

Models/Management/BiomassRemovalEvents.cs
```diff
@@ -26,119 +24,88 @@ namespace Models.Management
...
         public string StageToSet { get; set; }
 
-        /// <summary>
-        /// Dates to trigger biomass removal events
-        /// </summary>
-        [Description("Removal Event Dates (comma seperated dd/mm/yyyy")]
-        public string RemovalDatesInput { get; set; }
+        /// <summary>List of dates to trigger biomass removal events.</summary>
+        [Description("List of dates for removal events (comma separated, dd/mm/yyyy or dd-mmm):")]
+        public string[] RemovalDates { get; set; }
```
ric394, line 54: A converter will need to be created for these changes and affected .apsimx files removed from this pull request.

DeepCloner.Core/DeepCloner.Core.csproj
rcichota: gosh that one commit from Edith keep on giving... It's fixed as it as a mistake originally

Models/Management/BiomassRemovalEvents.cs
ric394: Converter looks correct to me. Once the apsimx file changes are reverted, your pull request should be good to merge.

rcichota: Thanks @ric394 
Just wondering why I need to revert those files, if all I did was what the converter will do anyways?
Will be working on that anyway

par456: Most of the time it won't cause a problem, but if someone else happens to be working on these apsimx files on a branch at the moment, their version may not be updated when they go to merge in the changes you've added. 

Git is pretty good at merging files together, but they may end up with a file that is only 'partially' converted depending how good a job git does at it.

It's mainly just a problem with the apsimx files within our repo, as git doesn't understand the formatting and only reads line-by-line. So in a pull request like this, we generally just want to reduce the potential conflicts that could occur with other people's work, especially with files that don't need to be in the PR that were just used for testing.

rcichota: Thanks, all clear

rcichota: All sorted out, I hope.

ric394 (approved):

PR #9748, opened 2025-03-11
9727 Apsim Importer can handling missing soultes/initialwater

### LINKED ISSUE #9727: System.NullReferenceException when converting Classic soils

par456:
Resolves #9727 

Refactored the importer solutes code to handle if solutes is missing and give the new solutes the default values and layer structure from Water.

Also added default value of 100% full for initialwater if that node is missing from the apsim file. All defaults are based on what Apsim 7.10 uses as defaults for components when added new.

### DIFF
--- a/Tests/UnitTests/ImporterTests.cs
+++ b/Tests/UnitTests/ImporterTests.cs
@@ -423,22 +423,22 @@ namespace UnitTests
         }
 
         [Test]
-        public void TestInitialEmptySolutesInitialisesSolutes()
+        public void TestImporterCreatesSolutesAndWaterBalancesIfMissing()
         {
-          string xml = ReflectionUtilities.GetResourceAsString("UnitTests.Core.ApsimFile.SoilMissingSolutesAndInitialWater.apsim");
-          var importer = new Importer();
-          Simulations simulations = importer.CreateSimulationsFromXml(xml, e => Assert.Fail());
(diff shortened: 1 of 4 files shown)


Tests/UnitTests/ImporterTests.cs
```diff
@@ -421,5 +421,24 @@ public void EnsureMemoImports()
             Assert.That(memo, Is.Not.Null);
             Assert.That(memo.Text, Is.EqualTo("hello there"), "Failed to import memo message from .apsim file");
         }
+
+        [Test]
+        public void TestInitialEmptySolutesInitialisesSolutes()
```
ric394, line 426: The only thing I would suggest is to change the name of this to what it actually tests. Something like 'TestImporterInitialisesSolutesAndWaterBalances'

ric394 (approved):

PR #9755, opened 2025-03-13
Throw exception if LayerStructure is deeper than Physical.Thickness.

### LINKED ISSUE #9709: If LayerStructure goes deeper than the soil physical properties an error should be thrown.
### What happened?

At present it seems that the extra depth is assigned the same properties as the deepest described layer. The initial water content seems to be set to LL15. Could address the water content but might be better to throw an error for the user to address the issue.

Windows

hol353:
Resolves #9709


par456:
retest this please jenkins

par456:
retest this please jenkins

par456:
Is there a reason the STRUM prototype has lost 100K lines?

hol353:
A converter must remove something out of the STRUM .apsimx file. I opened it in the tip revision and the lines also disappeared some this isn't related to this PR. The number of simulations doesn't change (32) though so must be something else.

retest this please jenkins

par456:
retest this please jenkins

Issue #9763, opened 2025-03-14
Add simulation of NH3 volatilisation into Nutrient


sno036:
### Describe the new feature

This has been a feature missing from the SoilN/SoilNitrogen/Nutrient since approximately forever. This work will convert an old APSIM Classic manager script that estimated volatilisation into a proper model under Nutrient.

PR #9764, opened 2025-03-14
Volatilisation model for Nutrient

sno036:
Working on #9763 

Please merge this so that the wider team can assist with testing and parameterisation.


hol353:
retest this please jenkins

par456:
Only comment is that you have a section with public outputs, but all the output parameters have both a getter and setter on them meaning someone could change the output value not realising that it won't do anything to your model (at least I'd hope it doesn't).

Considering this is still under development, happy to merge still and get that fixed later, but might want to think of restricting the outputs to avoid users getting them confused with the user accessible parameters.

sno036:
Thanks Andrew - I will do that when the next set of improvements is ready to go.

Mauri ora, Val

PR #9778, opened 2025-03-18
Set parameters per simulation

### LINKED ISSUE #9777: Set model parameters in specific simulation from values listed in seperate file
### Describe the new feature

Sometimes when testing models or for specific applications it is necessary to overwrite parameters in the released model.  This can be done by putting models in replacements but that overwrites the parameter value with the same new value for all simulations in the .apsimx file

Propose adding a component that can set specific parameters to specific values for individual simulations.  Would provide parameter overwrite values in a csv file with a flat table structure with SimulationName as a row index and Parameter Names as columns index.  Component would read file into code and for each parameter set the corresponding value for the current simulation.

A similar outcome can be achieved in manager scripts but the proposed approach has the following advantages.
- Easier to lay out parameter set values in a table format so quicker to set up simulation specific overwrite values.
- Clear that the model is using non custom values as a specific component is included in simulations to make this happen.

HamishBrownPFR:
Adds code component and test for setting model parameter values in individual simulations from values specified in a .csv file

resolves #9777


HamishBrownPFR:
@hol353 @par456 @ric394 please review and comment or merge if happy.

HamishBrownPFR:
@hol353  @par456  @ric394.  Has anyone had a chance to look at this one?

HamishBrownPFR:
@hol353 @par456 @ric394. Am I missing something?  Is there any reason why this PR has not been reviewed

Models/Utilities/SetModelParamsBySimulation.cs
```diff
@@ -0,0 +1,211 @@
...
+            return dt2;
+        }
+
+        /// <summary>
+        /// Writes out changes from the grid to the csv file
+        /// </summary>
+        public DataTable ConvertDisplayToModel(DataTable dt)
+        {
+            return new DataTable();
+        }
```
hol353, line 148: I don't think 'Tables', 'ConvertModelToDisplay' and 'ConvertDisplayToModel' are used. Can be deleted.

Models/Utilities/SetModelParamsBySimulation.cs
```diff
@@ -0,0 +1,211 @@
...
+                    CurrentSimParams = getCurrentParams(setVars, CurrentSimulationName, varName);
+                    object Pval = 0;
+                    try
+                    {
+                        Pval = CurrentSimParams[CurrentSimulationName];
+                    }
+                    catch
+                    {
+                        throw new WarningException(Name + " was not able to set any model parameters because simulatoin name " +
+                                                 CurrentSimulationName + " is not present in the SimulationName column in " + FullFileName);
```
hol353, line 203: Incorrent spelling 'simulatoin'

hol353 (commented):
One downside to using a .csv file to set parameters is that converters now need to check them whenever a variable is renamed.

Also need to add an icon for the new model.

Models/Utilities/SetModelParamsBySimulation.cs
HamishBrownPFR: Correct.  Thanks, they were working but broke.  The table view was not essential so left it for now.

Models/Utilities/SetModelParamsBySimulation.cs
HamishBrownPFR: fixed

HamishBrownPFR:
@hol353 I am not sure what to do about the converter issue.  It is the same issue we have with variable names as column headers in observed excel files.  When we change a variable name do we have a system that trolls all our observed files for the old name and replace it with the new name?  (we should if we don't).  Could this be expanded to include .csv files too.  I could change the utility so it reads an excel file instead of a csv but this creates some annoying behaviors in the work flow.  Specifically, if you have the excel file open to edit then run the simulation it throws an error because the excel file is open and can't be read.  This problem does not occur with text files. Using Rons Data Edit allows you to have a grid editor view of a csv so it feels like excel without the annoying control excreted on the file. 

Suggestions appreciated

Models/Utilities/SetModelParamsBySimulation.cs
```diff
@@ -0,0 +1,211 @@
...
+    /// Each parameter column must have a header name that matches a full parameter address (e.g [Wheat].Phenology.Phyllochron.BasePhyllochron.FixedValue)
+    /// Set desired parameter values down the column to match the SimulationName specified in each row.
+    /// </summary>
+    [Core.Description("Reads a .csv file containing parameter values and sets those into the specified model for the specified simulatoin only.  \n" +
+                      "Must have a column with a header called \"SimulatoinName\" containing simulation names that match the simulation you are wanting to overwrite parameters in.  \n" +
+                      "Contains additional columns for each model parameter that is to be overwritten.  \n" +
+                      "Each parameter column must have a header name that matches a full parameter address (e.g [Wheat].Phenology.Phyllochron.BasePhyllochron.FixedValue)  \n" +
+                      "Set desired parameter values down the column to match the SimulationName specified in each row.")]
+    
+    [ValidParent(ParentType = typeof(Zone))]
```
ric394, line 27: If this is per Simulation, can I suggest changing the parent to Simulation?

ric394:
> When we change a variable name do we have a system that trolls all our observed files for the old name and replace it with the new name? (we should if we don't). 

I'm not aware of one at the moment, but there may be a converter function that does this. Would be nice to have!

On a different note, what is the use case for this functionality? Can replacements be used instead? My thinking is, we currently have two ways of changing parameters at the moment, models.exe options and via replacements in the GUI. 
We might want to be careful adding and supporting multiple ways of doing this, especially if an existing part exists that can do the same thing or can be modified to do so.

Models/Utilities/SetModelParamsBySimulation.cs
HamishBrownPFR: Thanks @ric394.  

The class needs to link to the appropriate zone to set its changes into.  It is safest to have the valid parent as zone so the link to zone can be by Ancestor and then it always finds the right zone.  
Perhaps it could be named better.  The idea is you can arrange model parameter overwrites in a tabular format for specific simulations.  They could be individual simulations or factored simulations in an experiment.  
ParameterOverwritesFromTable might describe what it is better.  Other ideas welcomed

HamishBrownPFR:
@ric394 The use case is pretty narrow and most likely limited to model development and testing where we want to remove the effects of errors in one part of the model to see how accurate the model is down stream of those errors.  The specific use case which I developed this for is testing the wheat models predictions of biomass production for different cultivars.  We want to remove the confounding coming from errors in prediction of development between cultivars.  So I set the phenology parameters specific to each simulation (different cultivars at different sowing dates and different locations) so we get the timing of flag leaf and flowering correct to the day and can then look at how the model is doing describing growth differences.  
We can't use replacements because that only allows a single overwrite value for all simulations in the same .apsimx file.  The models.exe functionality is not so well suited to this use case as the work flow is to look at results in the GUI and tweek the values for parameter overwrites in the table until the desired outcome is achieved.  
This same functionality can be achieved from a manager script and there are some tests in the build where model parameters are set in that way.  It is quite a bit of mucking around getting it set up that way so I thought I would write this class to make quicker and so other developers could use this.
I don't mind if we don't think this is a good idea to go into release.  Let me know either way so I can plan how to progress things.

hol353:
Sorry @HamishBrownPFR for the delay. Your PR has sparked a discussion about how best to implement your use case.

I can think of lots of ways to change a parameter in a simulation:

* Cultivar
* Experiment factor
* Replacements
* --apply on command line
* manager scripts
* the model in this PR.

All of these are for different use cases so I get the need to different mechanisms.

I wonder whether we can roll them up into a single, unified approch, something that is flexible. I wonder if a new 'replacements' model could be designed that is dropped at the top level of the simulation. It's right hand side view could be flexible enough to allow the user to configure parameter/model replacements (for all or a subset of sims), factorials of parameter replacements (for all or a subset of sims), loading parameter values from a .csv, applying new cultivars etc. Perhaps it could use a natural language like how --apply does. 

If we go down this path, we would have to start by identifying all the use cases. Then do some design on what the user sees. I would wrap all the code into a self contained assembly, with a limited set of public classes/methods and unit test it thoroughly. 

This would be quite a bit of work. For now, do we merge this PR and create a new GitHub issue with a rough plan, that links to this PR so we don't forget where this new model sits?

HamishBrownPFR:
@hol353 thanks for your comments.  It would be good to unify how things are set in the model.  It can get quite confusing working out what values the model is actually using in some situations and it would be nice if it were simpler.

I am happy with the suggestion of merging this pull request as a place holder for this use case and linking it to a new issue to try to tidy this space up.

HamishBrownPFR:
@hol353 @par456 @ric394 I have added an issue (#9818) pointing to this PR to capture the discussion regarding rationalizing the approaches we have for setting parameters in simulations.

Should be good to merge this PR now

PR #9781, opened 2025-03-19
Add NotYetSown phase to beginning of phenology

### LINKED ISSUE #9775: CurrentPhaseName is 'Germinating' when no crop is on the ground
### What happened?

Is this intentional in wheat?

Windows

HamishBrownPFR:
resolves #9775

@BrianCollinss I have only done a rough test on this.  Please pull this branch and check that this resolves the issue you have raised here or suggest further refinements.


BrianCollinss:
@HamishBrownPFR Thank you. I now have `StoredSeed` before planting and after harvest.

HamishBrownPFR:
@hol353 I am getting errors on this pull request relating to rock content in the peanut tests.  I have made a minor change to phenology which should only influence the name of current stage output when the crop is not in the ground.  The error being generated here On Commence which would be before any of the changes I made do anything.  I have had the same rock error popping up elsewhere recently.  Any suggestions on how best to fix this issue.  

This PR is also failing on the swim tests which I am having trouble working out.  But in that case the simulation has the maize model in it which has been touched by this PR so I will keep investigating that one.

HamishBrownPFR:
retest this please jenkins

hol353:
Hmm the test in the soil temperature model that is tiggering your error in peanut uses soil water content so I guess your change in phenology is somehow changing the simulation somehow. The test in soil temperature may be too sensitive and might need a rethink. Perhaps @rcichota and @sno036 can have a look at it again. On my computer (simulation: Kingaroy1819), the test is failing on 1/3/2019 (2 months after the clock start date) during the DoSoilTemperature call.

The problem this PR is trying to fix is to stop the model from having a phenology phase of 'germinating' when the plant hasn't been sown. Isn't it simplier to just have an if statement to return null for CurrentPhaseName when plant.IsAlive = false? Also do the same for other phenology outputs. I guess this might be a lot of if statements everywhere. Looking at the code in Phenology, I think I can see why you didn't go down this path. Lots of code and outputs for something that really should be simple. Needs a refactor (like a lot of things in APSIM).

I'm not against adding an extra phase, but I wonder whether we add it explicitly to the plant .json files rather than in code so that users can see it when they look at the model structure in the GUI. The downside to this is a converter would need to be written. 

I also don't like calling the new phase ('StoredSeed'). How about 'NotSown' or 'NotAlive'.

Also, you have .json and .apsimx files in this PR that don't need to be there.

HamishBrownPFR:
Closing this PR because it is superseded by 9775 (Thanks @par456)
Of no relevance now but I didn't use 'NotSown' because it makes sense prior to sowing but not after harvest.  Avoided 'NotAlive' because seed is technically alive, just dormant.  Andrews solution is better to return nothing then we can avoid the technical arguments about what word to use.

PR #9807, opened 2025-03-26
9775 Phenology Phase returns null when not alive

### LINKED ISSUE #9775: CurrentPhaseName is 'Germinating' when no crop is on the ground

par456:
Resolves #9775

Rather than the approach Hamish used in #9781 to add a new phenology phase, ~~I've simply adjusted Phenology to return 'null' when the linked plant to it is not alive. It already returned null when phases weren't set or when the phase index was too big, so this shouldn't have too many knock on effects.~~

Then when other functions like getting the name, or emergence are called, they return sensible results if the phase is null.

Will test in validation and see what happens.

### DIFF
--- a/Models/PMF/Phenology/Phases/BlankPhase.cs
+++ b/Models/PMF/Phenology/Phases/BlankPhase.cs
@@ -1,0 +1,31 @@
+﻿using Models.Core;
+
+namespace Models.PMF.Phen
+{
+    /// <summary>
+    /// A phase that represents no phase. Used when a plant has not been sown yet.
+    /// </summary>
+    public class BlankPhase : Model, IPhase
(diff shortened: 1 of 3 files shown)


par456:
Since so many manager scripts use GetCurrentPhase to then try and do Comparisions and calculations, I ended up creating a BlankPhase that has no name and returns 0, false or blank string for each property. This seems to satisfy the validation files without causing issues, and now means that you don't get a Phenology phase name when the plant is not alive.

I also had to edit ZadokPMF as with BlankPhase it was falling through to an IF statement that it should not have been. Did a general tidyup of the class while I was there.

PR #9811, opened 2025-03-26
Ensure detached and removed biomass are zeroed

### LINKED ISSUE #9809: Values for removed and detached biomass should be zeroed daily
### What happened?

PMF organs have outputs for the biomass that is detached (goes to surfaceOM / FOM) and removed (generally harvested, goes off field). These helps following the fluxes of biomass in each organ. The issue is that they are currently not zeroed properly, the values are daily only, so should be cleaned up at the start of each day (and at end of crop...
Also noticed that these values (Called LitterDeposition and RootDetached) are not filled in for AgPasture at the end of the crop (they are ok on daily values)...

Windows

rcichota:
Resolves #9809 
Organs have two Biomass component (Detached and Removed) to record daily values of biomass that are detached (going back to residue or FOM) or removed (harvested, taken off field). These were not being zeroed at the end of crop and thus making difficult to record these variables to track biomass fluxes.
I added a variable (hasJustBeenEnded) that is set to true on CropEnding and then used to zero the two Biomass components (OnDoDailyInitialisation).

### DIFF
--- a/Models/AgPasture/PastureSpecies.cs
+++ b/Models/AgPasture/PastureSpecies.cs
@@ -287,9 +287,6 @@ namespace Models.AgPasture
                 root.Dead.DetachBiomass(RootWt, RootN);
             }
 
-            // set flag to zero all transfer variables 'tomorrow'
-            hasJustEnded = true;
-
             // reset state variables
             Leaf.SetBiomassState(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0);
             Stem.SetBiomassState(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0);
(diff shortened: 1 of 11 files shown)


rcichota:
Thanks @hol353 
I cannot disagree that using this flag is not ideal. The whole thing would work with simply having the variables being set to zero every day, so no need to test OnDailyInitialisation (including IsAlive). This would mean zeroing the variables everyday, even if the plant is not active. Not sure if this would compromise performance??
And sorry about not adding the 'private' declaration, one does dumb things when trying to finish a commit before dinner...

rcichota:
I was having a read in StackExchange and it seems I am falling victim of over micro-optimisation... It seems setting up a value (zeroing in our case here) is generally not expensive, having the code checking whether it should or not change may be more costly. So, as we want/need those variables to be zeroed every day, then we can simply have that be done so, even at times when the plant is not active... This would mean getting rid of the check in on DailyInitialisation and simply having the instruction to clean up the variables... 
What do you think? I can do that here (in which case, we are admitting that the test for IsAlive is the problem, and removing that solve our issue).

hol353:
Good call @rcichota. I much prefer this approach i.e. zero every day. We shouldn't let percieved performance issues (which may not eventuate) affect our design decisions. I'm happy for this to be merged.

PR #9817, opened 2025-03-27
Wheat tidyup

HamishBrownPFR:
Working on #7623

- Removed - Leaf.SpecificArea because it duplicates Leaf.SpecificAreaCanopy but with different units
- Added - Leaf.TotalSpecificArea to obs files which is SLA calculated with Leaf.Wt as opposed to Leaf.SpecificAreaCanopy which uses Leaf.Live.Wt.  Allows us to utilize a bit more data.  Current model does not output this but simple leaf model can
- Renamed any Leaf.Tips to Phenology.HaunStage in .apsimx and moved Leaf tips data into haunstage column in obs files as the two are the same thing (approximately) and analysis is more powerfull if we combine them rather than leave them seperate.
- Calculated Leaf.Wt where Leaf.Live.Wt and Leaf.Dead.Wt were present

also resolves one issue raised in ##9682 by @par456 "BCVT_Mt._Madden_1994-06-13Cultivarjanz has a . in the simulation name that is causing it not to match the sim name in apsim."


HamishBrownPFR:
@hol353 @par456 @ric394 please review and merge.
Three variables are missing.  This is intended.  
- Leaf.Ligules will not be calculated by the simple leaf model and it tells us the same thing as Phenology.HaunStage so removing to reduce duplicaiton and confusion
- Leaf.tips is the same thing as Phenology.HaunStage.  I have removed Leaf.Tips as an output and merged observed data into Phenology.HaunStage to reduce duplication and confision
- n ha increased for Phenology.HaunStage because of this
- Leaf.SpecificArea is the same as Leaf.SpecificAreaCanopy (x1000000).  To reduce confusion I have removed the former.
- n has increased for Leaf.SpecificAreaCanopy as I have merged the obs from Leaf.SpecificArea in with it
- n has increased for Leaf.Wt because I have filled some gaps where Leaf.Live.Wt and Leaf.Dead.Wt were present but Leaf.Wt was not

HamishBrownPFR:
dont merge yet.  I have mucked something up on an additional commit

HamishBrownPFR:
In additions to above comments on stats:
- 	Wheat.Structure.TotalStemPopn is missing because I have removed it.  It was duplicating    	Wheat.Leaf.StemNumberPerPlant	
- n has gone up for  Wheat.Leaf.StemNumberPerPlant because obs data from Wheat.Structure.TotalStemPopn has been moved into this variable

HamishBrownPFR:
@hol353 @ric394 @par456 
Good to go for review and merge again

ric394 (approved):

PR #9855, opened 2025-04-08
apsim software and file versions and quirks documentation

ric394:
resolves #9823

### DIFF
--- a/Docs/content/TechnicalInformation/Versioning.md
+++ b/Docs/content/TechnicalInformation/Versioning.md
@@ -1,0 +1,22 @@
+---
+title: "APSIM file versions & APSIM Software versions"
+draft: false
+---
+
+There are two different versions used in APSIM Next Gen with different meanings and uses. They are:
+
+1. The build number or the APSIM program version (as seen in upgrade). This tracks every time the APSIM Next Generation codebase changes and is packaged for all users (GitHub pull requests with hash resolved) and is used to identify when any fix was made or a new feature added (Git commit history). You report this version in publications for repeatability and transparency as you can ensure the simulation will always give the same results when run with this same version and your simulation file provided with supplementary data, and
+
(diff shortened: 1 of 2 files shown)


BrianCollinss:
One issue I face quite often is that APSIM that's in a docker or singularity on HPC fails to run simulations that I created on my laptop. Those HPC APSIM installations are usually updated much less frequently than the one on my laptop. Moreover, most changes in the file version are of no effect on most simulations (e.g., they are related to AgPasture while I run Wheat). 

So, I wonder if it's better to let APSIM try to run, and print a warning (in GUI and summary) to remind user of the old version being used?

ric394:
@BrianCollinss these are fair points.

Relaxing the processes around converters and therefore the apsim file version would create the need for more numerous and specific error checking and unnecessarily complicates the code base.
As outlined in the documentation in this PR there are ways around the issues you're facing such as reducing the apsim file version number to one compatible with your apsim HPC containers. You could also do the same when running your files locally.

par456:
> One issue I face quite often is that APSIM that's in a docker or singularity on HPC fails to run simulations that I created on my laptop. Those HPC APSIM installations are usually updated much less frequently than the one on my laptop. Moreover, most changes in the file version are of no effect on most simulations (e.g., they are related to AgPasture while I run Wheat).
> 
> So, I wonder if it's better to let APSIM try to run, and print a warning (in GUI and summary) to remind user of the old version being used?

The core apsim model doesn't really differ between when it's run in a GUI and when it's run directly from the console (for example), so it's hard to have it halt when running in the console and the versions don't match, but let it keep going in the GUI on the same error. It's also bad practice to let it run differently in the GUI and console, as the GUI shouldn't effect the model or output, it's supposed to just be a display.

And the main reason we do halt on a version mismatch is that a newer file may still run without errors in an older version, but we cannot know for sure if what it produces is actually correct anymore. And it's very dangerous for users if they aren't getting errors and think the output is correct when it's actually not. As you note most of the time it won't effect a simulation, but occasionally there are some big converters (like the temperature one recently) that effects everybody.

Generally in your case where you're running on a HPC that only updates it's image occasionally, I'd be using a set Release version to keep the simulation to a specific version and only moving up versions when a new feature is required. Or if you are needing to use the tip revision, doing the manual version number change back after reviewing what the inbetween converters were doing (to make sure they don't effect your simulation).

PR #9861, opened 2025-04-08
9860 Sim names on graphs works with permutations

### LINKED ISSUE #9860: Simulation names on graphs not working with permutations
### What happened?

Related to #9837

The solution is not working with permutations and labelling points incorrectly.

Windows

par456:
Resolves #9860

The graphing code produces Nan values in the x/y columns sometimes, and specifically with permutation experiments being graphed. When this happens, the Nan values get filtered out, and therefore the captions need to as well, otherwise the indexs don't line up.

This PR should fix it.


par456:
Retest this please jenkins

ric394 (approved):

Issue #8413, opened 2023-10-30
Graphing and reporting work flows could be improved


HamishBrownPFR:
### Describe the new feature

I have been doing some development on the wheat model and wheat.apsimx is now a mess of reports spitting out mostly redundant variables and graphs that were time consuming to set up and are often broken.  Some ideas on how we could improve things.

There are two types of graph that are most useful and we may be justified on creating some reporting and graphing objects that serve these two types well but don't provide the full level of flexibility that is present in the current graphing and reporting objects.

Observed vs Predicted scatter plots for quantifying overall model performance.  These are needed for the model performance testing system and are very instructive of the overall accuracy of a model.  Currently we have to make sure we are reporting all the variables we want to compare to observations and then make each graph.  We tend to report much more that what is being compared to observations.  I suggest we make a Performance Stats report object.  This would have a reporting functionality built into the background that pools the observed report files and creates a simulation report that has all the variables that have a matching observation but only for simulations and dates that have observed values.  This would reduce the amount of redundant reporting on the build.  In the .apsimx file this object could have a view object where it creates the typical obs vs pred graph for all the simulations in scope with a drop down to choose which variable to display.  In the build we would render a graph for each variable as we currenlty do. 

The other graph type that I use all the time for testing and debugging is a scatter plot with clock.today as the x axis and one or more other variables as the y axis.  These graphs may or may not have observations plotted as well.  I think we could make a time coarse object that served this type of graph more efficiently by addressing the functinality below. 

1.  Adding Observations onto graphs is cumbersome but is normally wanted if the observations exist.  Can we just build the graphing of observations into the background of the proposed time coarse graph so it finds and displays observations if they are present.  

2.  The graphing approach gives lots of flexibility which is good but in most cases it is not necessary and creates the need for lots of clicking to get a graph working.    For the proposed time coarse graph, we introduce some defaults to speed things up.  In most cases if we draw a scatter plot we would want lines for simulated values and symbols for observations.  This could be coded as a default for this graph type.  Most the time we don't care which factor is used for demarking color and which is used for demarking line or symbol type.  Could we have it so by default the top level factor is used to demark color, and the second factor level is used to demark marker or line type. 

3. The filter expression is powerful but often trips you up when you copy a graph from one experiment to another.  Could we make it so we have the filter expression on a seperate object and the time coarse graph applies a filter experssion if it finds one in scope.  That way we could copy a folder of graphs and just change the expression filter once to get all the graphs working in the new location.

[...]

PR #9867, opened 2025-04-09
8413 ObservedInput

par456:
Working on #8413 

This is the first part of the ObservedInput and ObservedReport system that we've been working on for a while. The aim of this is simplify generating PO stats and give more feedback about errors within the observed data.

I've pulled the ObservedInput work out of the original branch as we are looking to use its functionality first to help clean up observed data in the Wheat validation, along with generating values that can be derived from the existing observed data (eg, NConc from N and Wt, Totals from Live and Dead etc).

I've stripped out everything that ObservedInput doesn't need from the original branch to hopefully simplify a review of this. This is still a work in progress however, so it's not as tidy as it could be and further work is needed as I add back and complete features.


par456:
Github acting funny, killing PR and trying again.

PR #9869, opened 2025-04-09
8413 Observed Input

par456:
Working on #8413

This is the first part of the ObservedInput and ObservedReport system that we've been working on for a while. The aim of this is simplify generating PO stats and give more feedback about errors within the observed data.

I've pulled the ObservedInput work out of the original branch as we are looking to use its functionality first to help clean up observed data in the Wheat validation, along with generating values that can be derived from the existing observed data (eg, NConc from N and Wt, Totals from Live and Dead etc).

I've stripped out everything that ObservedInput doesn't need from the original branch to hopefully simplify a review of this. This is still a work in progress however, so it's not as tidy as it could be and further work is needed as I add back and complete features.


par456:
This is ready for a review/merge so @hut104 can start to user test the functionality.

ApsimNG/Presenters/ObservedInputPresenter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// Detach the model from the view.
+        /// </summary>
+        public void Detach()
+        {
+            propertyPresenter.Detach();
+            columnsGridPresenter.Detach();
+            derivedGridPresenter.Detach();
+        }
```
ric394, line 49: doc string is incorrect

ApsimNG/Presenters/ObservedInputPresenter.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// Detach the model from the view.
+        /// </summary>
+        public void CreateGridTab(string property, IModel model, GridPresenter gridPresenter, ContainerView viewContainer)
+        {
+            gridPresenter.Attach(new DataTableProvider(new DataTable()), viewContainer, explorerPresenter);
+            gridPresenter.AddContextMenuOptions(new string[] { "Cut", "Copy", "Paste", "Delete", "Select All" });
+
+            IDataProvider provider = DataProviderFactory.CreateUsingDataTableName(model, property, (properties) => {});
+            gridPresenter.PopulateWithDataProvider(provider);
+        }
```
ric394, line 61: doc string is incorrect

ApsimNG/Presenters/ObservedInputPresenter.cs
par456: This one is correct, the model is just wrapped up within those subpresenters rather than explicitly being detached. The other summary does need correcting.

ric394: Ignore this comment. I've selected the wrong block I believe

ric394 (approved):

hol353:
This needs a refactor. There are a couple methods that are ~ 100 lines long.

PR #9884, opened 2025-04-14
Wheat tidyup

HamishBrownPFR:
working on #7623


HamishBrownPFR:
@hol353 @par456 @ric394 @hut104.  Please review and merge
Diffs are expected as have set phenology for the Maricopa experiments so we can focus on the biomass production.  Most diffs are ticks.  The crosses are all small.

PR #9897, opened 2025-04-21
Wheat tidyup

### LINKED ISSUE #9682: Tidy up NDVI and Leaf.Wt and SLA data in wheat test
### Describe the new feature

Some of the validation sets from NZ have CoverGreen data derived from NDVI.  Now we are modelling NDVI dirrectly this data should be replaced with raw NDVI data for doing model stats.
Need to review the use of Leaf.Wt vs Leaf.Live.Wt as these two seem to be used interchangeably in the observed set.  Need to make sure they are what they say they are.
Check if there are any cases where we have leaf wt LAI data and can calculate SLA.

HamishBrownPFR:
Resolves #9682


BrianCollinss:
Are these new Tav correct? They seem quite low.

HamishBrownPFR:
@BrianCollinss, no they are not correct.  They keep getting changed to these errant values in the background.

HamishBrownPFR:
@ric394 @par456 @hol353 Please review and merge.  
I have added some data, moved some data that was under the wrong variable name (from CoverGreen to NDVI and some from leaf.live.wt to leaf.wt) and deleted some duplicate data (Spike wt)

PR #9899, opened 2025-04-22
9849 Tav Amp Weather fix

### LINKED ISSUE #9849: Tav and Amp are saved to apsimx file
### What happened?

The TAV and AMP values for weather are saved into the apsimx file as they are a property on weather, these should just be read from file or calculated at runtime, not saved into the file.

There also seems to be an error in the calculation, with Maricopa met file in the Wheat Validation coming in much lower than it should be.

Was found in this PR #9826

Windows

par456:
Resolves #9849 

Implements the changes discussed in the issue. Tav and Amp calculations will now throw if there is less than 2 years of data, and are no longer stored in the apsimx file.

Maricopa met files have had Tamp renamed to Amp so that it is read by the weather model.


par456:
Invalid Weather Files:
Peanut\Gainsville.met

Wheat\met\WaggaWagga2019NPIfield.met
Wheat\met\Callington2019NPIfield.met
Wheat\met\dale2019NPIfield.met
Wheat\met\yanyean2019NPIfield.met
Wheat\met\2020NPI_WaggaWagga.met
Wheat\met\2020NPI_Urrbrae.met
Wheat\met\2020NPI_Dale.met
Wheat\met\2020NPI_YanYean.met
Wheat\met\miles96.met
Wheat\met\emer96.met

Barley\WaggaWagga2019NPIfield.met
Barley\Callington2019NPIfield.met
Barley\dale2019NPIfield.met
Barley\yanyean2019NPIfield.met
Barley\2020NPI_WaggaWagga.met
Barley\2020NPI_Urrbrae.met
Barley\2020NPI_Dale.met
Barley\2020NPI_YanYean.met

par456:
After discussing with @hut104, I've gone through and added Tav and Amp to the weather files in our validation set that didn't have Tav and Amp and it couldn't be calculated because the time was under two years. This was done by using the nearest LongPaddock point to each of those sites, with the choice documented in the met file.

HamishBrownPFR (approved):
Not sure if my credentials allow me to do reviews.  
These changes look good to me

ilhuber:
If you [JsonIgnore] could be added to Latitude as well that would help speed up saving simulations that read met data from excel workbooks. Longitude is get only for now, but maybe to guard against similar problems in the future it should have one added too.

par456:
That's a good point, it would have been accidently included into the file when Longitude was given a setter function.

I've added jsonIgnore to both Longitude and Latitude so that if Latitude is also given a setter in the future, it won't then accidently start appearring in the file.

par456:
retest this please jenkins

Models/Climate/Weather.cs
```diff
@@ -1054,6 +1058,9 @@ private void ProcessMonthlyTAVAMP(out double tav, out double amp)
             DateTime last = this.reader.LastDate;
             int nyears = last.Year - start.Year + 1;
 
+            if ((last - start).TotalDays < 730)
+                throw new Exception("Tav and Amp cannot be calculated from less than than two years of met data.");
```
ric394, line 1062: remove extra than

ric394 (approved):

PR #9903, opened 2025-04-23
9866 Use correct layers, fix message

### LINKED ISSUE #9866: More helpful information from CheckSoil would be useful
### What happened?

CheckSoil would be more helpful if it returned a little more information about where the error was found. The soil attached is giving an index out of bounds but no information about where (and I can't find it! scanning the UIs) it is failing the check.

[Example.zip](https://github.com/user-attachments/files/19659898/Example.zip)

Image

Windows

par456:
Resolves #9866

The check soil code was assuming that organic and water had the same soil layer structure as physical, when they can be different, so that problem was fixed.

It also had a summary message that should have been written like the other messages in this code, so fixed that as well.

### DIFF
--- a/Models/Soils/Soil.cs
+++ b/Models/Soils/Soil.cs
@@ -234,31 +234,33 @@ namespace Models.Soils
                 if (organic.Carbon.Length == 0)
                     message.AppendLine("Cannot find OC values in soil");
                 else
-                    for (int layer = 0; layer != physical.Thickness.Length; layer++)
+                    for (int layer = 0; layer != organic.Thickness.Length; layer++)
                     {
                         int layerNumber = layer + 1;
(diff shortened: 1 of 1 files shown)


par456:
After changing the Organic carbon checks to use the structure from Organic instead of Physical, it appears to be finding a few failed soils in Wheat and Swim where LayerStructure is being used alongside Swim.

The errors are indeed correct, the Organic.Carbon for some layers of the soil are set to 0 instead of the minimum 0.01 that it checks for.

I suspect because it was previously using the physical indexs, it wasn't actually going deep enough into the Organic layers to find those layers with 0 carbon. Might need to investigate this problem further next week. @hol353

par456:
retest this please jenkins

PR #9904, opened 2025-04-23
9864 Simulation names don't conflict under replacements

### LINKED ISSUE #9864: Exception thrown when Experiment's base scenario is to be replaced
### What happened?

I expected the following configuration to work:

image

but I got this error:

`Duplicate simulation names found: Simulation`.

Also, when I have a Simulation outside the Experiment and the Experiment is disabled, I still get the error. 

image

In a kind-of related observation, when I disable the Simulation under Replacements and run the model while selecting Simulations node, GUI counts 17 simulations to run, while I have only 16 factorial sims. 

image
image

Windows

par456:
Resolves #9864 

Rather than just solving this problem by putting yet another replacements check in where it was required, I did a small refactor instead.

I've added three functions under Folder that do the checking if a model is a folder that can be a Replacements, so that all the error checking and constraints we have on replacements is done in one place. 

New functions are:
- IsModelReplacementsFolder
- FindReplacementsFolder
- IsUnderReplacementsFolder

These are the 3 use cases in apsim for either checking if a folder is replacements, or to find the replacements folder, so now all references to replacements have to go through these functions. Not sure if this is the best place for it, I could see this in Structure instead, but that class is already quite large, and Replacements are just special folders, so we can discuss that.

At least with this, all the requirements for how we determine if a folder is Replacements is now in one class, instead of being spread across a bunch of files. In the future when we refactor this, it should be much easier now.

### DIFF
--- a/APSIM.Documentation/Models/Types/DocSimulations.cs
+++ b/APSIM.Documentation/Models/Types/DocSimulations.cs
@@ -159,7 +159,7 @@ namespace APSIM.Documentation.Models.Types
                 { 
                     tags.Add(new Section(child.Name, DocumentTutorial(child as Simulation)));
                 } 
-                else if(child is Memo || child is Graph || !Folder.IsModelReplacementsFolder(child))
+                else if(child is Memo || child is Graph || (child is Folder && !Folder.IsModelReplacementsFolder(child)))
                 {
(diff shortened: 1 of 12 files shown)


hol353:
Nice! We need to do more refactoring when we see something messy.

PR #9906, opened 2025-04-23
Fix cultivar used in Lincoln1992

HamishBrownPFR:
working on #7623 
Reviewing some graphs in Lincoln1992 I noticed the cultivar sown was "BattenWinter" which is the other member of an isogenic pair with "BattenSpring".  "Batten" was released in 1990 and later noticed that the population showed 1:3 segregation for non-flowering when sown late in warm environments.  The plants that did not flower were selected out and named "BattenWinter" which was used in small scale phenology trials but never released.  The isolation of the two flowering types happened later in the 1990s so this trial would have been sown with the released material of mixed flowering types which would be 2/3 spring type.  Further, the winter type behaves like the spring type when sown in Lincoln (Brooking and Jamieson 2002).  So the BattenSpring cultivar is more appropriate to use for simulating this experiment


HamishBrownPFR:
Further review of the Lincoln1992 set up.  Batten cultivars had grain number overwrites which were causing under predictions so have removed them so the wheat default value is used.
Observed LAI was measured with an LAI2000 which measures total area not green area (which is what LAI is in the model).   I have adjusted observed LAI based on the proportion of live leaf (so LAI's decrease where there was dead leaf in the canopy.  Also recalculated SLA with new LAI data.
Removed TotalSpecificArea because it is a useless variable

par456:
retest this please jenkins

PR #9907, opened 2025-04-23
9843 PredictedObserved Ignores case when matching columns

### LINKED ISSUE #9843: Incorrect case in Report variable can stop predObs from matching
### What happened?

If you have a report with Soil.Water.mm written in, it will cause PredictedObserved from correctly matching the result against Soil.Water.MM in an observed file, as it will label the column in the datastore incorrectly.

**But much worse**
 
If you have multiple reports with the same name, but some use .mm and some use .MM, you get a race condition where the name of the column is based on whatever report runs first. So you end up from run to run, either matching the data correctly with PredictedObserved, or not because of the difference in case.

When a report variable is not aliased, it should really be basing the name of the variable on the name of the property, rather than on how it was written in the report line. 

Windows

par456:
Resolves #9843


par456:
@hut104 @hol353 

Lots of new validation points have been picked up since doing this, will need a review.

HamishBrownPFR (approved):
Looks good to me.  Unlocks some more data from our observations that the tests had not been considering.  Changes to stats are all due to addition of new tests resulting from removal of case depencency

par456:
retest this please jenkins

par456:
retest this please jenkins

par456:
retest this please jenkins

par456:
retest this please jenkins

ric394 (approved):

PR #9909, opened 2025-04-23
Marianas expt

HamishBrownPFR:
working on #7623 

Removed Leaf.TotalSpecificArea from observed files because it is not use and confuses things


HamishBrownPFR:
@par456 @ric394 @hol353 please review and merge.

Just deleted a redundant columns from observed files so should be easy

PR #9910, opened 2025-04-24
Set Enabled property of MenuDescriptionArgs to "true" by default.

### LINKED ISSUE #9908: Can no longer change units from ppm to kg/ha to set initial N concentrations
### What happened?

I have upgraded NextGen (version 2025.4.7727.0). When setting initial NO3 and NH4 in a soil, the option to change the unit from ppm to kg/ha no longer exists when you click on the header. Now the box is just blank and cannot be selected. 
Image

Windows

zur003:
Resolves #9908.
Also fixes popup menus generally.

### DIFF
--- a/ApsimNG/Interfaces/IMenuView.cs
+++ b/ApsimNG/Interfaces/IMenuView.cs
@@ -38,7 +38,7 @@
         public string ShortcutKey;
 
         /// <summary>Is the item enabled</summary>
-        public bool Enabled;
+        public bool Enabled = true;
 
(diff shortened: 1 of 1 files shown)


par456:
From having a look at IMenuView, that whole interface and class needs a bit of work. IMenuView is an interface that is only used by a single class, and MenuDescriptionArgs should be defined in its own file.

Not only that, but this problem has come about because MenuDescriptionArgs doesn't have a constructor and when it's used, only some variables are set. Rather than just solving this bug, I might just do a quick reactor here and clean it up fully.

par456:
Solved in #9914

PR #9916, opened 2025-04-28
Fixing sowing populations in southern australian wheat tests

HamishBrownPFR:
Some experiments in the Southern Australian tests have populations that are quite different to the observed populations.  These have been corrected in this PR.  
A few other tidy ups as well

working on #7623


HamishBrownPFR:
@par456 @ric394 @hol353 @hut104 please review and merge.  
Minor changes to stats expected as have fixed sowing population in a lot of simulations.
n has decreased for StemNumberPerPlant because I have removed the marked plant data for this variable for Lincoln2024 and replaced it with the random stem data which does not have a systematic observer effect

PR #9927, opened 2025-05-01
Add ability to close drains.

### LINKED ISSUE #9926: Add ability to close subsurface drainage in SWIM.
### Describe the new feature

Restore this popular feature from classic.

ilhuber:
Resolves #9926


par456:
I just want to check on this change, as I'm not overly familiar with the SWIM code. Is this the only location in the class where having the drain open or closed will have an impact?

For example, you still have variables like SubsurfaceDrainNO3 that are being calculated each day, should those be influenced by closing the drain? 

Also at about line 916, SWIM has code for doing some reporting of the drain status, the open or closed state should also be reported here so it's state is known.

par456:
retest this please jenkins

PR #9941, opened 2025-05-06
Cotton model

TiemenRhebergen:
working on #4631


ric394:
retest this please jenkins

hut104:
@ric394 or @TiemenRhebergen Can someone check why N>400 for some phenology observations.  We don't have that many simulations.  It may be matching every day in the daily output stats.  We usually only do this in the harvest stats.

hol353:
Closing this - old. Create a new PR when ready.

PR #9943, opened 2025-05-07
Stop AgPasture roots from reaching layers with XF>0 that are below la…

### LINKED ISSUE #9942: AgPasture should pay more attention to XF values in setting the root depth
### What happened?

When there is a XF>0 below layers with XF=0, the roots should not be able to reach those deeper layers

Windows

sno036:
…yers with XF=0

resolves #9942


ric394:
retest this please jenkins

sno036:
It is looking like some of the AgPasture simulations in the test suite have bad XF values (at least that is the most likely explanation as this change should only have prevented roots growing through XF=0 layers into deeper layers where XF is more hospitable).

sno036:
Accept these new statistics please and merge.

PR #9966, opened 2025-05-16
Renamed solute degradation to decomposition. Changed to 1-degradation.

### LINKED ISSUE #9934: Specification of degradation of solutes
### Describe the new feature

Currently degradation is modelled as the fraction remaining rather than fraction degrading/decomposing, as is used for the decomposition of organic pools. Suggest it would be more intuitive and similar to approaches for organic pools to define degradation as the fraction remaining. "Behind the scenes" this may involve multiplying the existing pool by (1- degradation rate). Also suggest to stick with one term - decomposition.

hol353:
Resolves #9934

### DIFF
--- a/Models/Core/ApsimFile/Converter.cs
+++ b/Models/Core/ApsimFile/Converter.cs
@@ -26,7 +26,7 @@ namespace Models.Core.ApsimFile
     public class Converter
     {
         /// <summary>Gets the latest .apsimx file format version.</summary>
-        public static int LatestVersion { get { return 193; } }
+        public static int LatestVersion { get { return 194; } }
 
         /// <summary>Converts a .apsimx string to the latest version.</summary>
         /// <param name="st">XML or JSON string to convert.</param>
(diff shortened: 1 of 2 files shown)


Models/Core/ApsimFile/Converter.cs
```diff
@@ -6397,7 +6397,7 @@ private static void UpgradeToVersion191(JObject root, string _)
                                                                  .Where(m => JsonUtilities.Name(m) == "NitrificationInhibition"))
             {
                 var oldParent = JsonUtilities.Parent(nitrificationInhibition) as JObject;
-                var newParent = JsonUtilities.Parent(oldParent) as JObject;;
+                var newParent = JsonUtilities.Parent(oldParent) as JObject; ;
```
ric394, line 6400: double semi-colon can be removed

hol353: Fixed.

PR #9970, opened 2025-05-16
Rename TTremaining and add functionality

### LINKED ISSUE #9969: ThermalTimeRemaining report can be expanded to provide more useful functionality
### Describe the new feature

https://github.com/APSIMInitiative/ApsimX/blob/master/Models/Functions/ThermalTimeRemaining.cs

This class was designed to calculate how much thermal time is left to run over multiple phases.  However, it could be usefully expanded to also return how much tt has already accumulated, the target for the phases specified and the ammount accumulated and remaining in relative values.  We have alot of function nests in crop models that do these 5 jobs and this could tidy that up.

This follows on from #9602 where the class being altered here was initially created

HamishBrownPFR:
resolves #9969


hol353:
Can you change the spelling to 'aggregate'?

HamishBrownPFR:
Fixed, thankyou @hol353

PR #9973, opened 2025-05-16
Add functionality so biomass removals affect LAI and cover

### LINKED ISSUE #9972: Simple leaf does not loose LAI or cover with biomass removals
### Describe the new feature

It should

HamishBrownPFR:
resolves #9972

### DIFF
--- a/Models/PMF/Organs/SimpleLeaf.cs
+++ b/Models/PMF/Organs/SimpleLeaf.cs
@@ -4,6 +4,7 @@ using APSIM.Numerics;
 using APSIM.Shared.Utilities;
 using Models.Core;
 using Models.Functions;
+using Models.Functions.DemandFunctions;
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@hut104 can you review this please.  Working on getting SCRUM into release.  Found biomass removals do not affect LAI in simple leaf so added some functionality.  Have not added code to affect cover directly as cover is being calculation in functions in SCRUM and I pick up the LAIremoved and use it there.  Do you think I should add code to reduce cover for cases in other crop models where cover is calculated in this class.

hol353:
This doesn't change LAI though. It just creates a couple of outputs from SimpleLeaf. Am I missing something?

HamishBrownPFR:
The LAI model is out in the .json file in an accumulate function.  It picks up the removal value and deals with it

Issue #9975, opened 2025-05-16
A few potential bugs in RootNetwork


BrianCollinss:
### What happened?

I was not sure if these are just intentional or bugs, so I chose to post it as an issue instead of a discussion.

In `RootNetwork.cs`, lines 445-465, 

- Where is the impact of root distribution across layers? In the simplest form, should we not have `KL` somewhere as a proxy for root distribution, or another water stress function be used instead of `SoilWaterScale` to account for the extractable soil moisture instead of total soil moisture in each layer?

- Does `maxNUptake` show the maximum N uptake, or maximum uptake of NO3 and NH4, separately? If the former is the case, should we not change line 458 and line 462 to something like this (assuming NO3 is more readily available for uptake)?
```
NO3Supply[layer] = Math.Min(zone.NO3N[layer] * kno3 * NO3ppm * SWAF * factorRootDepth, (maxNUptake - NO3Uptake - NH4Uptake));

NH4Supply[layer] = Math.Min(zone.NH4N[layer] * knh4 * NH4ppm * SWAF * factorRootDepth, (maxNUptake - NH4Uptake - NO3Uptake));
```

- `RWC` is calculated but never used. 

![Image](https://github.com/user-attachments/assets/ad57e8e4-3412-4b99-a7bf-e8cc49d84c52)

### What Operating System are you on?

Windows

BrianCollinss:
Thanks heaps @HamishBrownPFR. It was very insightful. The Zone- and Roo-related classes are very complex and the documentation is not helpful whatsoever. See https://github.com/APSIMInitiative/ApsimX/discussions/9974 for two examples. I hope someone can spare some time to improve these docs. It would be a great contribution to the user community.

PR #9976, opened 2025-05-18
Test layered KNO3 and KNH4.

BrianCollinss:
Working on #9975

### DIFF
--- a/Models/PMF/Organs/Root.cs
+++ b/Models/PMF/Organs/Root.cs
@@ -761,7 +761,6 @@ namespace Models.PMF.Organs
                         NH4Uptake += NH4Supply[layer];
                     }
                 }
-
             }
         }
 


BrianCollinss:
@HamishBrownPFR I just wanted to see the effect. I thought you might be interested:

image

HamishBrownPFR:
Hi @BrianCollinss.  Couple of things here.  the constant values of KNO3 and KNH4 are generally smaller than kl in the top soil layer (where most of the mineral N normally is) so just substituting Kl for these two seems to be making N uptake to much.  Try putting a multiplier so you can reduce the magnitude but still get the same shape as kl.

There are some key N response experiments that we would need to have a close look at the effects of this change.  APS6 and Lincoln2015 have time coarse data of N uptake with different N treatments.  If you check point the release results and then make the change, we can get a good idea of the impact it has on N uptake predictions.

BrianCollinss:
Thanks @HamishBrownPFR I'll keep playing with this. Re the KN values, I assumed the KL values weighted with layer thickness would be roughly equal to 0.02. I may try to calibrate these two if there's good data in those two exp.

BrianCollinss:
Scaled by 0.8:

image

BrianCollinss:
Scaled by 0.6:

image

BrianCollinss:
Scaled by 0.5:

image

BrianCollinss:
@HamishBrownPFR what do you think of these comparisons? It seems scaling by 0.4-0.6 generally improves the results (especially in `DailyObsPred`). More specifically, the effect on `StemPopulation` at harvest is interesting. Perhaps the current effect of N stress on `BranchingRate` is too harsh:

image

BrianCollinss:
Scaled by 0.4:

hol353:
Is this PR needed? Please close if not.

PR #9981, opened 2025-05-19
Test some changes in SaturatedFlow

BrianCollinss:
Working on #9979

PR #9984, opened 2025-05-19
Fix SaturatedFlow for KS=0

### LINKED ISSUE #9983: If KS=0, SaturatedFlow doesn't work properly
### What happened?

In `SaturatedFlow.cs', line 82, we have:

```
if (soilPhysical.KS == null || soilPhysical.KS.Length == 0)
```

This means when KS is not `null` but is all zero, the model uses KS to calculate saturated flow. This is the case in some of the simulations in Rice prototype and Mungbean validation. I am not sure how these models are working, but they shouldn't!

Windows

BrianCollinss:
Resolves #9983

This might not be the best way to handle such errors. We might need a way to warn users of having KS=0 instead or throw if more than one layer have KS=0 (which would lead to many sims failing).

The best solution (as proposed in https://github.com/APSIMInitiative/ApsimX/pull/9981) is to change this line:
`double excess_down = Math.Min(soilPhysical.KS[i] - w_drain, w_excess);`


BrianCollinss:
Retest this please jenkins.

ric394:
@BrianCollinss you should be able to change the labels in far right of the page. See image.

image

ric394:
@BrianCollinss I'll see if I can give you the required permissions to do this.

hol353:
Rather than leaving the zero KS values in the sims, can you make it throw instead? Then you'll have to remove the zeros in the simulations in our test suite that have them so that Jenkins goes green.

BrianCollinss:
@hol353 KS=0 is not inherently incorrect. Some users might want to test the effect of an impermeable layer (which will require the other PR merged later). IMO, it must throw if more than one layer has KS=0 which means they might be making a mistake (e.g. Using KS=0 for all layers). 

Also, GUI doesn't allow editing these cells (see #10030). It needs to be done with a text editor. An issue to look into by someone more familiar with that part of the code. 

Removing the zeroes needs to be done by the teams responsible for developing the models. Removing zeroes changes the stats in some cases significantly which means models need to be re-evaluated and perhaps recalibrated (in case of Mungbean IMO).

BrianCollinss:
BTW @hol353. Jenkins will never go green even if we remove zeros. The models have been tested with these zeroes in place, which have led to SaturatedFlow doing some weird stuff (https://github.com/APSIMInitiative/ApsimX/pull/9981)

hut104:
Hi all,
Ks of zero makes little sense for soils, and will cause errors for SWIM.  It might be best to deal with the errors in the build system (ie files that have been committed with Ks = 0) and accept the improved results.

BrianCollinss:
@hol353 @hut104 Thanks Neil. For majority of the models affected, stat changes are minor, except for Mungbean. It would be safest to move it back to prototype and ask someone to have a second look.

hol353:
We've had a chat about this. We think the solution is to 
* put a check in Soil.Check (soil.cs) to throw when KS < 0 in any layer
* remove your change to SaturatedFlow.cs in this PR - not needed.
* remove the zero KS values from all simulations in our test suite EXCEPT for Mungbean.
* set the KS values for Mungbean to very low values e.g. 0.01. This should very similar stats for Mungbean.
* raise a GitHub Issue stating that Mungbean has invalid KS values and needs to be refitted. Hopefully the mungbean model owner will find time to do this.

BrianCollinss:
Thanks @hol353. I think if #10030 is fixed, it'd be much easier to remove zeroes from gui.

BrianCollinss:
I close this to follow Dean's proposed approach.

PR #9990, opened 2025-05-20
Try to avoid recursion when setting values of Objects from strings.

### LINKED ISSUE #9977: Untrapped crash from Experiment
### What happened?

I am using the latest APSIM master code in development mode and something in the Experiment component is causing APSIM to crash and close without any errors reported. 

The last output from debug in Visual Studio is 
`The program '[40356] ApsimNG.exe' has exited with code 3221225477 (0xc0000005) 'Access violation'.`

If I shift the simulation out of Experiment into the Simulations children it works so I do not believe this is an issue with the actual CLEM simulation where errors are all trapped and reported in the UI. I was able to get this to work this morning, but the problem has returned and I cannot determine the cause. Initially I thought it may have been an incorrect name in the factor compoenent, but this seems ok. My computer has just updated to Windows 11 and I have done a clean all/rebuild all, as well as restarting Visual Studio to try the usual tricks for such issues. 

This apsimx simulation file will reproduce the issue when the experiment is run and will work ok when the StandAlone simulation is run.

[CLEM_CropEnabledCattle.zip](https://github.com/user-attachments/files/20276564/CLEM_CropEnabledCattle.zip)

Windows

zur003:
Resolves #9977

### DIFF
--- a/Models/Core/VariableProperty.cs
+++ b/Models/Core/VariableProperty.cs
@@ -726,9 +726,13 @@ namespace Models.Core
                 {
                     this.Value = Enum.Parse(this.DataType, value);
                 }
+                else if (this.DataType == typeof(Object))
+                {
+                    this.property.SetValue(this.Object, value, null);
+                }
                 else
                 {
-                    this.Value = value;
+                    throw new ApsimXException(null, "Invalid property type: " + this.DataType.ToString());
                 }
             }
         }


par456:
Good catch on this one, that was a tricky problem. I had assumed when it said that no exception was being thrown that it was going to be another empty try-catch statement somewhere hiding the error, but a stack overflow wouldn't be picked up by exception code.

PR #9994, opened 2025-05-21
Some more tidy ups to AccumulateFunctionGeneral

### LINKED ISSUE #9993: AccumulateFunctionGeneral is easy to confuse order of accumulate and reduce events
### Describe the new feature

Need to force all reduction events to happen on endOfDay so accumulate events happen first.  
Throw if accumulate event is set to endOfDay
Make reduceDate handle comma seperated values so multiple dates can be added.

HamishBrownPFR:
Add some more flexability to reductions and push all reductions to endOfDay so they happen after accumulation.  
Add exception if endOfDay chosen to accumulate on.  
Add default for accumulate event
Make ReductionDate able to accept multiple dates, comma seperated

resolves #9993

### DIFF
--- a/Models/Functions/AccumulateFunctionGeneral.cs
+++ b/Models/Functions/AccumulateFunctionGeneral.cs
@@ -48,7 +48,7 @@ namespace Models.Functions
         ///Public Properties
         /// -----------------------------------------------------------------------------------------------------------
         /// <summary>The event that accumulation happens on</summary>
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
@hol353 @par456 @ric394 please review.  I have added an example file to both test all the functionality of the accumulate function is working and provide people with some ideas of how to use it

par456:
That example file was good, might need to wrap some unit tests up with that at a later date, I'll log an issue for that.

Tried to break the inputs and nothing crashed or acted particularly weird, so happy to merge.

PR #10007, opened 2025-05-26
Append "GoTo_" to EndStage name for GotoPhase

### LINKED ISSUE #10006: GoToPhase.EndStage name is not unique
### What happened?

GoToPhase.EndStage is set as the start stage of the phase it is going to.  This causes problems with the dictionary of stage names (Phenology.stageDict) as it has duplicate stage names. 
Need to append something to the EndStage name so it remains unique.

Windows

HamishBrownPFR:
resolves #10006

### DIFF
--- a/Models/PMF/Phenology/Phases/GotoPhase.cs
+++ b/Models/PMF/Phenology/Phases/GotoPhase.cs
@@ -35,7 +35,7 @@ namespace Models.PMF.Phen
             {
                 if (phenology == null)
                     phenology = FindInScope<Phenology>();
-                return "GoTo_"+phenology.FindChild<IPhase>(PhaseNameToGoto)?.Start;
+                return phenology.FindChild<IPhase>(PhaseNameToGoto)?.Start;
             }
         }
         /// <summary>Is the phase emerged from the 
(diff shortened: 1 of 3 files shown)


par456:
I get why you're adding the prefix to stop the name clash in the dictionary Phenology builds, but this seems like a clunky solution to what is actually a problem in how that dictionary is built.

Might just have a think about if we can solve the root problem instead of just appending a name change.

HamishBrownPFR:
@par456 I think it is better to have the Goto_ appended to the stage name.  As well as fixing the issue I needed to fix here, it also means we don't have duplicate stage names in the drop down selector for crop stage name.  That got confusing at times.

HamishBrownPFR:
The simulations in the SPRUM test trigger this problem.  
Branch: https://github.com/HamishBrownPFR/ApsimX/tree/SPRUM_release 
ApsimX\Prototypes\SPRUM
Simulation: UserGrazeTest
I have merged this branch into that and run it locally.  There was one minor problem which I fixed otherwise this is working OK so I am happy for this to be merged once you guys are

par456:
retest this please jenkins

not sure why that failed, got a github timeout error from postats, just going to try again

PR #10012, opened 2025-05-29
Add cultivar support to AgPasture

### LINKED ISSUE #9400: Allow a cultivar model to be dropped on an AgPasture model
### Describe the new feature

To more easily vary species parameters (and they actually do change!)

Dean to set up
Val to add to the example simulation

hol353:
Resolves #9400

### DIFF
--- a/Examples/AgPasture.apsimx
+++ b/Examples/AgPasture.apsimx
@@ -1,6 +1,6 @@
 {
   "$type": "Models.Core.Simulations, Models",
-  "Version": 173,
+  "Version": 195,
   "Name": "Simulations",
   "ResourceName": null,
   "Children": [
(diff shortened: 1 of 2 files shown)


par456:
retest this please jenkins

par456:
I notice in the RenewPasture example that the RenewPasture script is disabled, is that supposed to be like that?

ric394 (approved):

ric394:
> I notice in the RenewPasture example that the RenewPasture script is disabled, is that supposed to be like that?

It's an included alternative that's disabled by default

hol353:
Yeah. It's mentioned in the memo in that example sim. It gives the user another way to time events in the simulation i.e. operations vs manager script.

PR #10019, opened 2025-05-30
Fixed soil check code.

### LINKED ISSUE #9866: More helpful information from CheckSoil would be useful

hol353:
Resolves #9866

### DIFF
--- a/Models/Soils/Soil.cs
+++ b/Models/Soils/Soil.cs
@@ -134,7 +134,7 @@ namespace Models.Soils
             var soil = Apsim.Clone(this) as Soil;
             soil.Sanitise();
 
-            Check(summary);
+            soil.Check(summary);
         }
 
         /// <summary>


par456:
When I looked into that issue I found it not reading the correct layers (and therefore not properly checking some layer structures) in #9903

You might need to address that problem before closing that issue.

hol353:
This PR is the correct way to do it. It standardises the layer stucture across all the soil nodes. You don't need #9866 so don't merge it. The bug was that it was standardising the layer structure in a cloned soil but then checking the non-cloned soil. This PR fixes that.

PR #10020, opened 2025-05-30
Added extra soil checks

### LINKED ISSUE #9865: Some additional tests for bad soil inputs would be helpful
### Describe the new feature

Suggesting:

- KL between 0 and 1
- XF between 0 and 1
- sum(XF) > 0
- SWCON between 0 and 1 
- Organic C >0

Probably others

hol353:
Resolves #9865

### DIFF
--- a/Models/Soils/Soil.cs
+++ b/Models/Soils/Soil.cs
@@ -1,8 +1,10 @@
 ﻿using System;
+using System.Linq;
 using System.Text;
 using APSIM.Numerics;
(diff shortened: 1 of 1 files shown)


BrianCollinss:
@hol353 I wonder if there should be some checks for KS=0 as well (https://github.com/APSIMInitiative/ApsimX/pull/10020). BTW, currently the grid view doesn't allow empty cells for KS, therefore some users may be tempted to type in zero instead.

PR #10025, opened 2025-05-30
SPRUM release

### LINKED ISSUE #7075: Simple pasture model

HamishBrownPFR:
Resolves #7075

This pull request adds a SPRUM.json file to resources, a SprumCropInstance.cs to source code which make the new simple pasture model operational.  
It also includes a validation of the model against a lucerne vs Paster experiment, some sensibility test to ensure it is responding to parameters as intended and a set of examples to help users get started.


Models/PMF/SimplePlantModels/SprumPastureInstance.cs
```diff
@@ -14,123 +16,277 @@ namespace Models.PMF.SimplePlantModels
...
         /// <summary>Years from planting to reach Maximum dimension (years)</summary>
-        [Description("Years from planting to reach Maximum root depth (years)")]
-        public int YearsToMaxDimension { get; set; }
+        [Description("Age At Start of Simulation (0-100 years)")]
+        public double AgeAtSimulationStart 
+        {
+            get { return _AgeAtSimulationStart; }
+            set { _AgeAtSimulationStart = constrain(value, 0, 100); }
+        }
+        private double _AgeAtSimulationStart { get; set; }
```
hol353: These look ugly. We have a [Bounds] attribute that you can add to the properties. Can you then add a GitHub issue that asks to change PropertyPresenter to use these [Bounds] attributes. The Issue can then reference this file as an example.

Also, can you put all the privates at the top of the class. The privates should look like:
```c#
private double _ageAtSimulationStart;
```
i.e. not a property and starts with a lower case.

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
```diff
@@ -14,123 +16,277 @@ namespace Models.PMF.SimplePlantModels
...
 
         /// <summary>Maximum growth rate of pasture (g/MJ)</summary>
         [Separator("Pasture growth")]
-        [Description("Radiation use efficiency (g/MJ)")]
+        [Description("Radiation use efficiency (0.1 - 3.0 g/MJ)")]
         public double RUE { get; set; }
+        private double _RUE
+        {
+            get { return _RUE; }
+            set { _RUE = constrain(value, 0.1, 3.0); }
```
hol353: This one looks wrong. The private _RUE references itself.

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
```diff
@@ -14,123 +16,277 @@ namespace Models.PMF.SimplePlantModels
...
         /// <summary>"Does the crop respond to water stress?"</summary>
         [Description("Does the crop respond to water stress?")]
         public bool WaterStress { get; set; }
 
+        /// <summary>The clock</summary>
+        [Link(Type = LinkType.Scoped, ByName = true)]
+        private Clock clock = null;
 
         /// <summary>The plant</summary>
         [Link(Type = LinkType.Scoped, ByName = true)]
```
hol353: Put [Link]s at top of class

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
```diff
@@ -157,6 +313,7 @@ public class SprumPastureInstance: Model
         /// <summary>The cultivar object representing the current instance of the SPRUM pasture/// </summary>
         private Cultivar pasture = null;
 
+        private DateTime establishDate;
```
hol353: All privates to top of class.

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
```diff
@@ -316,13 +483,49 @@ public Cultivar CoeffCalc()
...
+
+        /// <summary>
+        /// Provides an error message to display if something is wrong.
+        /// Used by the UserInterface to give a warning of what is wrong
+        /// 
+        /// When the user selects a file using the browse button in the UserInterface 
+        /// and the file can not be displayed for some reason in the UserInterface.
+        /// </summary>
+        [JsonIgnore]
+        public string ErrorMessage = string.Empty;
```
hol353: I'm not sure how this error message works. Property presenter wouldn't see it or do anything with it. Why not just have your constrain class throw an Exception. Hopefully the property presenter would show this in the window at the bottom.

At the very least the public ErrorMessage field should be a property, not a field.

hol353 (changes requested):

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
HamishBrownPFR: I have done the same thing in SCRUM.  will add an issue and cite the two as examples.  The idea with the constrain function was so I could have it throw a UI error to alert users that the value they entered has been constrained at set up (rather than at run time).  However, I have not been able to get the error working yet.

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
HamishBrownPFR: It was wrong.  Thanks

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
HamishBrownPFR: done

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
HamishBrownPFR: done

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
HamishBrownPFR: It doesn't work and I couldn't find an example to follow.  How do you get the GUI to write the little red error down the bottom as you are working there.

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
HamishBrownPFR: Have added Bounds attributes to all properties that are doubles.  Also added an issue to build this constraint checking into PropertyPresenter in the future #10035

Models/PMF/SimplePlantModels/SprumPastureInstance.cs
HamishBrownPFR: I have changed the constrain method to throw when out of bounds values entered.  Didn't realize it was so easy to get the UI errors to show up.

HamishBrownPFR:
@hol353, I am happy with how this is all working now.  Can you let me know if it needs any further changes from your perspective.

PR #10038, opened 2025-06-04
Add check for KS<=0 to Soil.Check

### LINKED ISSUE #9983: If KS=0, SaturatedFlow doesn't work properly

BrianCollinss:
Resolves #9983

**Ready to be merged.**


BrianCollinss:
@par456 some help here please?

BrianCollinss:
retest this please jenkins

BrianCollinss:
@hol353 Hi Dean, I’ve updated the converter to replace KS with null when all values are either zero or NaN. I also revised all files where these problematic KS arrays were present. Details of the steps I followed based on your comment on the old PR are documented here: https://github.com/APSIMInitiative/ApsimX/issues/9983.

hol353:
Good work. Now you need to manually open Mungbean.apsimx and change the KS values to small numbers e.g. 0.01 and then commit that file to this PR. That way the stats hopefully won't change much for mungbean.

Also, there are a bunch of .apsimx files in this PR under prototypes that don't need to be there (I suspect). We try and limit the number of changed files in a PR to lessen the chance of a user getting conflicts because they are also modifying the .apsimx.

BrianCollinss:
Thanks @hol353 . The reason I had to update prototypes is that the new Soil Check would complain and APSIM would throw during the build process.

I'll update the Mungbean now.

BrianCollinss:
@hol353 Just remembered: even with KS=0.01, the stats won’t change. APSIM used to use a zero KS value to calculate downward flux—which is both physically and mathematically incorrect (hence the need for https://github.com/APSIMInitiative/ApsimX/pull/9981). Setting KS=0.01 still forces the model through the same flawed process, just with an extremely small value of KS instead of zero, so the resulting stats remain largely unchanged.

BrianCollinss:
@hol353 I’ve just updated the Mungbean soils by setting KS = 0.01 across the board. That said, I still believe the best course of action is to move Mungbean to prototype until someone can properly review and potentially recalibrate it. The current model was tested using a soil module that simulated saturated moisture flux quite inaccurately. 

Moreover, please note that in SaturatedFlow.cs, we have:
`double excess_down = Math.Min(soilPhysical.KS[i] - w_drain, w_excess);`
Which produces negative values when KS is small, hence the need for https://github.com/APSIMInitiative/ApsimX/pull/9981.

hol353:
> Which produces negative values when KS is small, hence the need for https://github.com/APSIMInitiative/ApsimX/pull/9981.

So we should wait until #9981 is merged before merging this one? Yes?

BrianCollinss:
@hol353 this one can be merged now.

hol353:
I've had another look that this. Something funny going on. I diffed your mungbean.apsimx with the released version. The screenshot below is looking at the soil in TOSYear3 (which is very different in the stats). On the left is the released version. On the right is the version from this PR. The released version has KS of null which is valid. Why are these changed?

image

I also notice in the same soil that the SWCON values are missing. Might need to relook at what is happening in this PR.

image

BrianCollinss:
@hol353 was your screenshot taken before or after APSIM was run (via GUI or not)?

hol353:
I didn't run APSIM (GUI or command line). I just downloded the .apsimx file from this PR and the one from the tip revision on GitHub. I used a diff tool to get the screenshot.

BrianCollinss:
OK. This is from `Experiment_1_Berken`. NaN are converted to zeros. It's the case for all `Thomas Dissertation` experiments.

image

image

I might have mistakenly replaced KS=0.01 in other places. I will fix those instances.

hol353:
We've rethought the need for the 0.01 values for Mungbean as well. Can you not put them back in. Instead just go with null KS values.

BrianCollinss:
@hol353 Sure :)

hol353:
> OK. This is from Experiment_1_Berken. NaN are converted to zeros. It's the case for all Thomas Dissertation experiments.

So, is this PR converting NaN to zeros or is it a previous converter? Either way we need to stop that happening.

BrianCollinss:
The one that was already in place. I did not touch that part of the coverntor. I opened a separated issue for that Issue a few days ago.

hol353:
Some of the .apsimx files don't need to be in this PR e.g. Mungbean, rice, pinus. They either have KS of NaN or null which your converter should fix on the fly. You only need those files that you manually need to edit e.g. KS = 0 or KS < minimum. Can you check all of them and remove the ones that aren't needed.

Also, your branch is now conflicted so you'll need to pull latest branch.

hol353:
I also notice mungbean.apsimx has NaN for particle size sand,silt and clay. Can you also do us a favour and add an extra soil check for these 3 variables and modify your converter to convert an array of NaN to null? If you're busy, that's ok, we'll find time to do it.

BrianCollinss:
@hol353 Not a problem Dean.

BrianCollinss:
@hol353 I reverted the SCRUM files to the most recent master and just upgraded the resource files again (to avoid Jenkins yelling!). I am unsure why there are still two conflicts. 

BTW, sorry that I wasn’t aware that the preferred approach was to let the converter handle the changes, rather than updating the files manually. I’ve now removed all unnecessary files from this PR.

hol353:
Some notes on where I've got to in determining why mungbean is so different when the fixes in this PR are applied (I may not get back to this today).

In the screenshot below (SaturatedFlow.cs), the released version goes through the green else statement because KS = NaN in the .apsimx file which gets converted to KS = 0 somewhere in APSIM. It then calculates an ```excess_down``` value that often goes negative (will be fixed in #9981 when merged).

This PR rightly changes the NaN values to null which makes SaturatedFlow go through the yellow if statement (which is what we want). The problem is that mungbean, eucalyptus and pinus stats are quite different because of this fix. The models must have been calibrated around this problem. The models need to be refitted with this change.

image

BrianCollinss:
@hol353 @HamishBrownPFR I am afraid I can't fix SCRUM and SPRUM conflicts/issues. Could you please help me here?

BrianCollinss:
@hol353 could this error be related to one of the recent merges? I did not get such an error a couple of hours ago. It complains about a `null` parent in `Structure.cs`.

BrianCollinss:
@hol353 I am afraid I can't get rid of this conflict in STRUM!

BrianCollinss:
@HamishBrownPFR could you please help me resolve this issue with STRUM/SPRUM?

```
System.Exception
  HResult=0x80131500
  Message=0.99 is higher than maximum allowed.  Enter a value less than 0.98
  Source=Models
  StackTrace:
   at Models.PMF.SimplePlantModels.SprumPastureInstance.constrain(Double value, Double min, Double max) in APSIMNextGen\Models\PMF\SimplePlantModels\SprumPastureInstance.cs:line 559
   at Models.PMF.SimplePlantModels.SprumPastureInstance.set_MaxCover(Double value) in APSIMNextGen\Models\PMF\SimplePlantModels\SprumPastureInstance.cs:line 264
   at Newtonsoft.Json.Serialization.ExpressionValueProvider.SetValue(Object target, Object value)

```

BrianCollinss:
@hol353 It's impossible to resolve these issues unless I get some help from STRUM/SPRUM and Lucerne teams.

Prototypes/STRUM/STRUM.apsimx
```diff
@@ -31327,11 +34756,10 @@
                   "MaxRD": 700.0,
                   "MaxHeight": 300.0,
                   "MaxPrunedHeight": 100.0,
-                  "MaxCover": 0.97,
+                  "MaxCover": 0.99,
```
hol353: Could this be why you're getting an error on Jenkins?

BrianCollinss:
@hol353 It is. But I am not sure if I need to change the threshold in the model or change the parameter values. And I wonder how the model works now as I am not changing anything except for upgrading the resource file.

BrianCollinss:
I am lost! For some reason, SPRUM was added to STRUM prototype.

BrianCollinss:
@hol353 @par456 this is ready now.

BrianCollinss:
Can we please have this reviewed and merged? Other pulls have been submitted which will have conflict with this one.

hol353:
Yes, we want this merged as well. The problem is that 2 models (pinus and mungbean) become unvalidated if this gets merged. We can't unrelease them so we're looking for a way forward. We're also busy on a whole bunch of other stuff atm. We'll try to get to this soon (probably not this week though).

BrianCollinss:
Thanks @hol353. FYI, I emailed Mungbean's developers many weeks ago, but did not hear back.

PR #10041, opened 2025-06-04
Add check for BD<=0.1 to Soil.Check

### LINKED ISSUE #10040: BD needs a lower bound
### Describe the new feature

Currently user can input even negative values for soi BD. Lowest values of BD is observed in pit soil and can be as low as 0.1 gr/m3.

BrianCollinss:
Resolves #10040


hol353:
Can you remove the .apsimx file please?

BrianCollinss:
It uses the wrong BD values (someone copied AirDry onto BD by mistake) so the new Check catches it.

Issue #9939, opened 2025-05-06
Remove the need for models in inherit from IModel/Model


hol353:
### Where is the code located?

Primarily in Models.Core

### What needs to be changed?

Eventually IModel.cs and Model.cs need to be deleted. The parent / child relationship (the primary reason for having IModel) should be maintained elsewhere i.e. perhaps in NodeTree.cs and Node.cs classes.

### Why does the code need changing?

We now have APSIM.Soils and APSIM.Graphs assemblies that contain POCO i.e. simple classes with no inheritance from from Model. Both assemblies have classes (e.g. Soil, Physical, Graph, Series) that need to be represented in a simulation and the GUI.

PR #10044, opened 2025-06-05
Moved scope to APSIM.Core and made internal.

hol353:
Working on #9939


APSIM.Core/Nodes/Node.cs
```diff
@@ -82,6 +82,29 @@ public IEnumerable<Node> Walk()
                 yield return childNode;
     }
 
+    /// <summary>Walk nodes in scope (depth first), returing each node</summary>
+    public IEnumerable<Node> WalkScoped()
+    {
+        if (scope != null)  // can be null if called while Nodes are being created.
```
par456, line 88: Does this return null if scope is null?

APSIM.Core/Nodes/Node.cs
```diff
@@ -82,6 +82,29 @@ public IEnumerable<Node> Walk()
...
+        if (scope == null)  // can be null if called while Nodes are being created.
+            return null;
+        return scope.FindScopedParentModel(this);
+    }
+
+    /// <summary>Is this node in scope of another node.</summary>
+    /// <param name="node">The other node</param>
+    public bool InScopeOf(Node node)
+    {
+        return scope.InScopeOf(this, node);
```
par456, line 105: Does this also need to check if scope is null?

ApsimNG/Presenters/SummaryPresenter.cs
```diff
@@ -97,8 +98,7 @@ private void OnSimulationNameChanged(object sender, EventArgs e)
         private void SetSimulationNamesInView()
         {
             // populate the simulation names in the view.
-            ScopingRules scope = new();
-            IModel scopedParent = scope.FindScopedParentModel(summaryModel);
+            IModel scopedParent = summaryModel.Node.ParentScoped().Model as IModel;
```
par456: ParentScoped doesn't read very well for what it is doing. .Parent get the model's direct parent, and .ParentScoped reads a similar when it's getting potentially something much further up the hierarchy. Not sure of a better name though. Almost like a ParentOfScope or TopOfScope or something.

Models/Core/Events.cs
```diff
@@ -33,7 +32,7 @@ public void ConnectEvents()
             modelsToInspectForSubscribers.AddRange(relativeTo.FindAllDescendants());
 
             // Get a list of models in scope that publish events.
-            var modelsToInspectForPublishers = scope.FindAll(relativeTo).ToList();
+            var modelsToInspectForPublishers = relativeTo.Node.WalkScoped().Select(n => n.Model as IModel).ToList();
```
par456: That is a complex line to replace what was a simple function call. Is there a way we could wrap that in a function again just so we aren't writing lambas whenever we want a list of everything in scope.

APSIM.Core/Nodes/Node.cs
hol353: It will never return null. Because 'yield return' is used it essentially returns an empty collection.

APSIM.Core/Nodes/Node.cs
hol353: Likewise, Scope.InScopeOf will never return null.

Models/Core/Events.cs
hol353: I've create a NodeExtensions class to wrap up some of the lambda expression (i.e. the Select).

ApsimNG/Presenters/SummaryPresenter.cs
hol353: I've changed it to ScopedParent which is more accurate. It looks for a parent that implements IScopedParent.

PR #10057, opened 2025-06-09
DataStore array type column display fix

### LINKED ISSUE #9991: Summing over partial array from report puts wrong variable name in datastore
### What happened?

The attached example is just the wheat example and I have added an additional output
sum([Soil].Water.MM[1:4])
Simulation runs fine and report gets the calculation right for this variable.  However, it puts it into the datastore under the header sum([Soil].Water.MM(1)) when it should be sum([Soil].Water.MM(1:4))

I am not sure the square brackets change to round brackets in the report variable but I can live with that.  But we loose information in the variable name by not having the range of layers included in the variable name so that part needs fixing.

This can be worked around with an alias but they obscure what the variable actually is and I am trying to avoid them.

Image

Image

[test.zip](https://github.com/user-attachments/files/20354765/test.zip)

Windows

ric394:
resolves #9991

This ones simply removes code that changed array report variables from using the first index number in parentheses rather than the full variable name as the datastore column name.
Tests were also updated to accomodate this change.


ric394:
retest this please jenkins

ric394:
@hol353 is this still needed? I recall discussions that you were working in the same area several weeks ago.

hol353:
Yeah I'm working in this space atm. I'll deal with this Issue. You can close this PR.

PR #10064, opened 2025-06-10
10056 Water Presenter Refactor

### LINKED ISSUE #10056: SW column in Soil Physical not updating according to Water node (copied soil)
### What happened?

The SW column in physical does not update with the initial SW in Water node, although the figure in Physical correctly displays the initial SW.
The summary file Field.Soil.Physical reflects the wrong SW, whereas Field.Soil.Water is correct.
Simulation output reflects that simulation uses the correct initial SW from Field.Soil.Water
Opening and closing sim does not fix it.

Removing the other (not enabled) soil variant from which this soil was copied and rerunning after emptying DB no change.
However, subsequently closing this file and reopening fixed it.
So somehow it kept a link to the Water node of the soil it was copied from, which was only removed once the file was closed after that node was removed. Zip file has word doc with details and apsim file

[

[ExampleSWproblemJune25.zip](https://github.com/user-attachments/files/20647870/ExampleSWproblemJune25.zip)

]

Have had problems with SW before  when I was adapting the soil profile and APSIM would get in a tangle (freeze) due to a mismatch of layers that it couldn't resolve. 
Do we need the SW column in the Physical table?

Windows

par456:
Resolves #10056

This PR removes the specialised water presenter and view, and connects it up with the more generic profile presenter that solutes, organic, chemical and physical all use. The error handling from the presenter for the input values has been added to the water model instead (where it should be) and some adjustments were made to the layout for ProfilePresenter so that it looks better across all the models using it now.

I also had to added some more flexibility into the Sheet library so that the model can throw an error when changing a property that is viewed in a grid. It now has error handling that feeds back to the GUI (instead of crashing gtk) and restores the previous value if an exception was thrown (so the GUI and model don't get out of sync).

I also added a SetCellContent function to the IDataProvider that basically just acts as a 1 cell wrapper for SetCellContents which wants arrays. I noticed that we nearly always want to change only one cell when editing, so instead of wrapping the single values in arrays each time, I made a dedicated function that does that instead.

Since this has touched a few views, **I would suggest when reviewing to pull this branch and explore around the GUI** and try and break the ProfilePresenter views. Water, Physical, Solute, Organic and Chemical.

We should also look to simplify the ProfilePresenter at some point too, it's got a bit of edge case code that might be able to be tidied up, and a bunch of graphing building code that really should be stored somewhere else (if it's not already in the new graphing library).


par456:
Need to fix LL15 on SoilCrop view

par456:
Need to put adjustments to starting position of sliders on SoilWater since it doesn't have a graph

ric394:
Some odd things:
- [x] Grid cells are able to be set to an invalid type such as a string. The grid should warn you and reset the cell to a previous valid value.
- [x] pressing the combination of ctrl + z twice while on the physical node brings up the next siblings model's grid. If another sibling model of physical is viewed before selecting the physical node, the model's grid will be shown when that combination is pressed.

PR #10067, opened 2025-06-10
Multipy speedup

### LINKED ISSUE #10066: Speed up multiply function when children return zero

Multiply.cs in the Functions namesapece

### What needs to be changed?

have the function return zero as soon as any child has returned a zero to save iterating through other children functions unnecessarily.  

Doesn't need changing.  But its an easy change to make and could give some small speed gains

HamishBrownPFR:
resolves #10066


HamishBrownPFR:
@par456 @ric394 @hol353 not sure if there is any merit in this change.  Up to you guys.

Issue #10070, opened 2025-06-10
Alternate approach for Excel weather data.


ilhuber:
### Where is the code located?

ExcelUtilities and ApsimTextFile

### What needs to be changed?

Try a different approach for reading OpenXML excel files.

### Why does the code need changing?

A recent style of project organization @sarchontoulis has favored is to store all met data for a set of experiments in a single Excel file, with each sheet containing weather data for a location. When reading these files, ExcelDataReader spends a significant amount of time creating the table via its internal APIs (especially if the excel file is hosted on some sort of cloud storage, like Box). These experiences were what prompted #9911. 

I was curious if using DocumentFormat.OpenXML to read the content of the excel sheet then just use the same code as if we had read a text based .met file to handle the weather data would be a beneficial approach. This change was able to cut the running time of our 1100 simulation dataset in half (4 minutes -> 2 minutes). The code in the pull request I am about to make probably isn't ready to be merged yet, but I would be interested in feedback (as well as seeing if I've added any errors).

lie112:
Relates to #6629 The read excel class should be generic to allow a specified sheet in an excel file to be opened into a data table. The met version of this should be a special case and not the other way around. I have had to implement some gymnastics code to use the apsim xls reader and in some CLEM code I have my own open xls section as there's a lot of met file specific code. Overall the class naming is confusing when met functionality is the default.

ric394:
resolved by #10071

PR #10071, opened 2025-06-10
Add new approach for OpenXML excel weather.

ilhuber:
An alternate reader for excel weather data.

Working on #10070 

Ready for review.


ilhuber:
Looks like the header can lack an equals sign for constants in previously allowed excel input. Also, units before value. Will have to think about how to handle that.

ric394:
@ilhuber Do you have an example excel met file and perhaps some unit tests to show it working? I'd like to have a look at this one.

ilhuber:
@ric394 There are a few scenarios in the validation that use excel weather, and one unit test too. Pretty useful for demonstrating my initial idea of reading the excel sheet as text and treating it as we would a met file wasn't going to be viable. 

This does not currently handle excel date formats (which are unfortunately a numeric style rather than distinct datatype), but should work for the format described in the Next Gen documentation if you want to take a look. This PR is mostly a speedup (2x on my 1100 sim dataset, 4 minutes -> 2 minutes) compared to how these were handled previously. 

I am leery about extending this beyond reading weather data, though. There might be user scripts that depend on previous behavior of ReadExcelFileData that I don't want to break.

ric394:
This seems to work well however I noticed a small issue. I have a file that spans two years, but the start and end dates don't match the files.

image

**Excel met file**

image


**Example archive**
[wheat_example.zip](https://github.com/user-attachments/files/20856855/wheat_example.zip)

ilhuber:
@ric394 I gave this file a try and it seems like it has issues on the main branch as well. 

https://github.com/APSIMInitiative/ApsimX/blob/85dce43bc296cb9193584c40aadb2722993275a9/APSIM.Shared/Utilities/ApsimTextFile.cs#L744-L760

It looks like line 753 is effectively a no-op that static analysis doesn't pick up, so we get through this loop with year set but no month or day. There's also an issue on US culture info systems where the DateTime.ToString call will return a string that APSIM doesn't recognize.

ilhuber:
I've added a way of handling OA dates if they appear in a date column for excel weather, would it be a good idea to create a simple test for this as well?

PR #10074, opened 2025-06-11
Fix bug in wheat phenology

### LINKED ISSUE #10068: Decreasing growth stage in wheat model
### What happened?

@HamishBrownPFR I noticed a very strange behaviour for CAMP Wheat model for a long term simulation in Goulburn, NSW with sowing dates on 20 June for cultivar Strzelecki from 1941 to 2024.

The final leaf number (CAMP.FLN) is keeping increasing since 1971.

Image.

This is caused by negative growth stage in 1971. 

Image

If I set Clock.Start as 1970-01-01, this error is still existed 

Image

However I set Clock.Start as 1971-01-01, this error is disappearing

Image

The decreasing or negative growth stage is happening during `SpikeletsDifferentiating` Phase.

Not sure how to figure out this problem. Happy to further debug if you point me to the right directions. 

Attached the example apsimx file. 
[Goulburn.zip](https://github.com/user-attachments/files/20665294/Goulburn.zip)

Windows

HamishBrownPFR:
resolves #10068


HamishBrownPFR:
@hol353 @par456 @hut104 @ric394 please review and merge.

This fixes a bug that @byzheng discovered with continuous wheat simulations.  Note that fixing this error will cause predication of slower development for some cultivars in some years when they are sown in a long-term simulation.  This is becasue the MaxVrn value was not being zeroed on sowing so the constrain it applied to development (for some cultivars) was only working for the first sowing.

PR #10091, opened 2025-06-15
Remove all Cloud/Azure support.

### LINKED ISSUE #6371: 'Run on cloud' feature needs work
The run on azure feature could be improved:

- More options for VM type
- Run in docker image
- Better error reporting when azure quotas are reached
- Tidy up configuration file
- Use PropertyView in UI
- Use Linux VMs to save costs

zur003:
Resolves #6371 - in a rather brutal way.

Note that this PR removes the entire DeploymentSupport\global folder, on the assumption that these files were used only to support Azure. Can anyone verify whether that assumption is correct?


zur003:
retest this please jenkins

hol430:
That DeploymentSupport/global file tree was definitely used by the Azure runner, and I'm pretty sure it wasn't used by anything else. If it was, that was surely a mistake and could/should be fixed anyway.

ric394 (approved):

PR #10104, opened 2025-06-19
Use correct namespace for JsonIgnore

### LINKED ISSUE #10087: Bubble chart extra/hidden nodes
### What happened?

Are these extra/hidden nodes there for a reason?

https://github.com/user-attachments/assets/88e82a24-dbfe-4535-b95c-675f13017c68

Windows

zur003:
Resolves #10087

Some units were using System.Text.Json.Serialization rather than Newtonsoft.Json as the namespace for JsonIgnore.

### DIFF
--- a/APSIM.Shared/Graphing/Arc.cs
+++ b/APSIM.Shared/Graphing/Arc.cs
@@ -1,7 +1,7 @@
 ﻿using System.Collections.Generic;
 using System;
 using System.Drawing;
-using System.Text.Json.Serialization;
+using Newtonsoft.Json;
 
 namespace APSIM.Shared.Graphing
 {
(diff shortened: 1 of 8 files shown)


lie112:
How are we deciding which Json library to use in APSIMx?
[Reddit Discussion on pros and cons of various JSON libraries](https://www.reddit.com/r/csharp/comments/tnmgs4/newtonsoftjson_vs_systemtextjson_vs_jsonnet/)
I see I accidentally applied the System.Text.Json to a recently built section of CLEM

Are there reasons we are sticking with Newtonsoft.Json with an additional Nuget package to maintain rather than shifting to System.Text.Json which is maintained and shipped with .Net? Are we using the unsupported Polymorphic (De)Serialization?

zur003:
@lie112 - If I recall correctly, polymorphic deserialization was the reason, but I'm not 100% certain.

PR #10126, opened 2025-06-30
Reduce redundant code calling in SoilArbitrator.

### LINKED ISSUE #10125: Reduce redundant code calls in SoilArbitrator
### Describe the new feature

Soil arbitrator makes a list of all plants in the simulation a startOfSimulation and then each day makes 4 estimates of uptakes with different levels of soil water and nutrient supply to arbitrate uptakes in the case of competition.  However, it there is only one live plant in the simulation it only needs to do the first uptake estimate (with the current soil status).

I suspect we can reduce the amount of redundant code calling by updating the list of plants to arbitrate each day, limiting it to those that are alive, and only doing the initial uptake estimate if there is only one live plant in the simulation.  This will be the majority of cases

HamishBrownPFR:
resolves #10125

### DIFF
--- a/Models/Agroforestry/TreeProxy.cs
+++ b/Models/Agroforestry/TreeProxy.cs
@@ -11,6 +11,7 @@ using Models.Soils.Arbitrator;
 using APSIM.Shared.Utilities;
 using System.Globalization;
 using APSIM.Numerics;
+using Models.PMF;
 
 namespace Models.Agroforestry
 {
@@ -43,6 +44,10 @@ namespace Models.Agroforestry
         IWeather weather = null;
         [Link]
         IClock clock = null;
+
+        /// <summary>The Live status of the crop</summary>
+        [JsonIgnore]
+        public bool IsAlive { get { return true; } }
 
         /// <summary>
         /// Tree proxy spatial data.
--- a/Models/Interfaces/IUptake.cs
+++ b/Models/Interfaces/IUptake.cs
@@ -10,6 +10,9 @@ namespace Models.Interfaces
     /// </summary>
     public interface IUptake
     {
+        /// <summary> Flag if the plant is alive</summary>
+        bool IsAlive { get; }
+        
         /// <summary>
         /// Calculate the potential sw uptake for today. Should return null if crop is not in the ground.
         /// </summary>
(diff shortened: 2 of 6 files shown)


HamishBrownPFR:
@hol353 @par456 @ric394 @hut104 @sno036 
Interested in your comments on this PR, if it is worth progressing?  I have had to disable one simulation in the wheat set that has SWIM because is throwing a fatal and the test scripts in PatchyMcPatchHead as they are throwing too.  Excluding them, the changes in stats are all tidlers.  I suspect we could get performance gains on a lot of simulations where there is only one plant actively growing so we don't have to do a full rangi cutta.  If you think it is worth progressing I will need some help to fix the errors in the SWIM and Patch tests.

BrianCollinss:
@HamishBrownPFR I'm curious why the stats change at all. Without competition, I would expect both approaches to yield the same outcome.

HamishBrownPFR:
Could just be floating point errors accumulating when passing through a longer calculation scheme?  Runge-Kutta is an approximation so there could be some small errors coming through it too?  Could be I have not implemented the shortcut correctly.

HamishBrownPFR:
@hol353 @hut @par456 @sno036.  Anyone able to comment on if this is worth pursuing?

sno036:
I'll try to look later this week.

HamishBrownPFR:
@sno036 I think the reason the patch tests are failing is due small differences in results that are triggering an exception in the test script.  We are seeing small variations in the other testing stats too.

sno036:
Thanks @HamishBrownPFR  - I get where you are coming from with this and agree that the small changes in those patch characteristics are not really an issue (the new stats could be accepted). What does worry me is that the intention was to include soil water evaporative demand in this arbitration so that most of the time there would be the soil and a single plant. So merging this as is would create a barrier to doing something better with soil water. 

How about instead checking for the number of demands? That way, when soil water evaporation comes along, the solution will hold.

Given the centrality of SoilArbitrator, a through code review would be good as well.

HamishBrownPFR:
Closing this PR.  The fix is not urgent and it would help to reduce the amount of stuff building up in the PR system

PR #10128, opened 2025-07-01
Nutrient constrained growth

byzheng:
Working on #3605

The nutrient constrain growth check the new minimum nitrogen concentration in each organ to limit daily actual growth. 

The minimum nitrogen concentration is a measured trait for whole organ, instead of new organ. 

The proposed changes apply nitrogen constrain to whole organ. Check the performance for all validation test.


hol353:
@hut104  and @HamishBrownPFR: Can you look at this PR? It is only a small change but might have a big impact.

byzheng:
We had a discussion with **@hut104**, **@HamishBrownPFR**, and **@EnliWang** about this change. Here is my current understanding (which might not be entirely correct):

- The **original approach** is based on a *physiological perspective*:  
  The parameter **minimum nitrogen concentration** represents the minimum nitrogen requirement for the **developing portion** of an organ. The developing organ may require different minimum nitrogen concentrations at different stages.  
  For example, in wheat, grains initially fill with proteins (which require higher nitrogen), and later with starch (which requires less nitrogen).

- My **modification** is based on an *observable (empirical) perspective*:  
  The parameter **minimum nitrogen concentration** represents the minimum nitrogen requirement for the **entire organ**, which can be measured directly from field experiments.

To address the issue in **#3605** for wheat and canola, an **alternative approach** proposed by **@hut104** and **@HamishBrownPFR** would be to make the *minimum nitrogen concentration* **stage-dependent** — using a curve instead of a constant, with much smaller values toward the end of grain filling.

par456:
The bot shouldn't have closed this yet, reopening.

PR #10147, opened 2025-07-09
Don't add zeroes as "default" KS values

### LINKED ISSUE #10030: GUI doesn't allow removing zeros from KS column
### What happened?

https://github.com/user-attachments/assets/65c3b573-01a2-43b5-9a85-f7dcf7fb6de3

Windows

zur003:
Resolves #10030


BrianCollinss:
Is #10038 different?

zur003:
@BrianCollinss - Yes, I think it's different. Basically this removes the line in SoilStandardise.InFill (Physical) which fills missing KS values with 0.

However, given this PR failed the tests on Jenkins. I suspect #10038 needs to be applied first to correct the KS problems in the Test simulations. This is intended to prevent 0 values from being re-introduced when viewing values in the GUI.

PR #10162, opened 2025-07-14
Convert SPRUM to new organ/arbitrator class

### LINKED ISSUE #10161: Convert SPRUM to new arbitrator/organ classes
### Describe the new feature

I have been slowly developing and testing a new class structure for doing organ/arbitration (classes in ...ApsimX\Models\PMF\BiomassArbitration) which are working well enough now to use them as the basis for released models.  
As a first test, I will convert SPRUM, which is quite a simple model and not widely used model, to use the new arbitrator classes in a released model.

HamishBrownPFR:
resolves #10161


HamishBrownPFR:
@hol353 @par456 @ric394 please review and merge if happy.  
Diffs in stats are all minor and due to subtle differences in the way the new arbitrator works relative to the existing on.  There are three stats missing and three new which are due to variables being renamed.  The stats are the same for these.

hol353:
This appears to be the first use of the new arbitrator in a released model (SPRUM). Is this correct? I can see it in a couple of prototypes. 

Can we (the software team) have a chat about the future direction of the new arbitrator and review the design? I'm uneasy about some models using a different arbitrator. We want to get the design right first otherwise we may need to write converters at a later date if we decide to change the design.

HamishBrownPFR:
@hol353 you are correct, this will be the first released model using the new arbitrator.  Linking in to the issue that prior arbitrator work has been done under #6763.  
Happy to run the software team through where this is at.  If you recall I did the initial programming of this during lockdown and we had a couple of sessions where you reviewed the design.  I have been chipping away at it subsequently where I have time, implementing prototype models using this and have done a lot of testing and improvement.
My main motivation to getting SPRUM onto this framework is to work in with STRUM which I have also implemented using this arbitrator.  And my motivation for doing that was because there were some faults in the way things scaled with zone size in the existing arbitrator but I found it extremely difficult to track them through its code base.

hol353:
Sorry @HamishBrownPFR, a bug in APSIM has caused resource files to be incorectly upgraded and this PR is affected. Can you re-upgrade all the resources and push to your branch? You'll have to revert the resource files back to the tip revision and then re-upgrade them. This branch is also conflicted which will need to be fixed.

HamishBrownPFR:
@hol353 should be good to go now

PR #10165, opened 2025-07-15
Moved locator to APSIM.Core assembly

hol353:
Working on #9939


APSIM.Core/Data/ApsimConvert.cs
```diff
@@ -0,0 +1,94 @@
...
+            else if (targetType == typeof(float)) // yuck!
+                return System.Convert.ToSingle(data, CultureInfo.InvariantCulture);
+            else if (targetType == typeof(int))
+                return System.Convert.ToInt32(data, CultureInfo.InvariantCulture);
+            else if (targetType == typeof(string))
+                return data.ToString();
+            else if (targetType == typeof(bool))
+                return System.Convert.ToBoolean(data, CultureInfo.InvariantCulture);
+            else if (targetType == typeof(DateTime))
+                return System.Convert.ToDateTime(data, CultureInfo.InvariantCulture);
```
par456: This should use DateUtilities

APSIM.Core/Data/ApsimConvert.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// <summary>
+    /// Is the specified type an Array or List?
+    /// </summary>
+    /// <param name="type">The type of the Array or List (e.g. int[] or List<int>)</param>
+    /// <returns>True if Array or List.</returns>
+    private static bool IsIList(Type type)
+    {
+        return type.IsArray || type.Name == "List`1";
+    }
+}
```
par456, line 94: This feels very much like a utilities type class, just some functions to convert objects to a type and provide some checking. I wonder if this would be better in Shared rather than Core? Do we already have something like this in shared it would fit with?

APSIM.Core/Data/DataArrayFilter.cs
```diff
@@ -0,0 +1,152 @@
...
+                    endIndex = SpecifierToInt(end) - 1;  // turn into 0-based index
+                else
+                    endIndex = int.MaxValue;
+            }
+
+            if (startIndex == -1)
+                throw new Exception($"Array indexing in APSIM (report) is one based. Cannot have an index of zero. Array specifier: {specifier}");
+        }
+    }
+
```
par456, line 117: This class could be simplified a bit, and needs more error checking because it's user input.

1. Conversion to 0-base should happen at the end, rather than in 3 places
2. We do a startIndex check here, and then in other functions do checks if it's in range. I'm not sure of the context for how this is used, but if we put all the error checking in one function and set the start/end indexes, we shouldn't need to then check them again before they are used?

APSIM.Core/Data/DataArrayFilter.cs
```diff
@@ -0,0 +1,152 @@
...
+    /*
+        if (Object is IModel model)
+        {
+            var physical = model.FindInScope<Physical>();
+            if (physical == null)
+                throw new Exception($"Cannot find a Physical node when evaluating array specifier {arraySpecifier}");
+            return SoilUtilities.LayerIndexOfDepth(physical.Thickness, depth) + 1; // Add 1 because the return index is 1 based, not 0 based.
+        }
+    */
```
par456: Old code?

APSIM.Documentation/Models/Types/DocFrostSenescenceFunction.cs
```diff
@@ -22,7 +23,7 @@ public DocFrostSenescenceFunction(IModel model): base(model) {}
         public override List<ITag> Document(int none = 0)
         {
             Section section = GetSummaryAndRemarksSection(model);
-            
+
             FrostSenescenceFunction frost = model as FrostSenescenceFunction;
             IFunction frostKill = model.FindChild<IFunction>("frostKill");
             IFunction frostKillSevere = model.FindChild<IFunction>("frostKillSevere");
```
par456, line 29: Did this file need to be touched?

ApsimNG/Presenters/XYPairsPresenter.cs
```diff
@@ -7,9 +7,9 @@
 using Series = Models.Series;
 using UserInterface.Views;
 using Models.Utilities;
-using APSIM.Documentation.Models;
 using Gtk.Sheet;
 using APSIM.Shared.Utilities;
+//using APSIM.Core;
```
par456: commented out line?

Models/Core/Overrides.cs
```diff
@@ -102,7 +104,7 @@ public static IEnumerable<Override> Apply(IModel model, string path, object valu
                     // e.g. if path = Data[3:4] then the replacementValue needs to be the full
                     // array and not just the values that are going to be used
                     // This gets around a design decision in VariableProperty.Value.set.
-                    replacementValue = ConvertValueToFullArray(model, path, replacementValue);
+                    //replacementValue = ConvertValueToFullArray(model, path, replacementValue);
```
par456: Commented out line?

Models/Core/Overrides.cs
```diff
@@ -289,34 +291,73 @@ private static string TryConvertToAbsolutePath(string valueAsString)
...
+                    string modelName = path.Substring(1, posCloseBracket - 1);
+                    path = path.Remove(0, posCloseBracket + 1).TrimStart('.');
+                    matches = model.FindAllInScope(modelName);
+                    if (!matches.Any())
+                    {
+                        // Didn't find a model with a name matching the square bracketed string so
+                        // now try and look for a model with a type matching the square bracketed string.
+                        Type[] modelTypes = ReflectionUtilities.GetTypeWithoutNameSpace(modelName, Assembly.GetExecutingAssembly());
+                        if (modelTypes.Length == 1)
+                            matches = model.FindAllInScope().Where(m => modelTypes[0].IsAssignableFrom(m.GetType()));
```
par456, line 297: I really dislike that this is duplicating some of the process of Locator by checking for a model with a name, then a model with the type. I don't think that goes in this PR, but it might be something we need to address later somehow.

Models/Resources/SCRUM.json
par456, line 1: What happened to scrum in the converter?

Prototypes/C4Maize/C4Maize.apsimx
par456, line 1: Is there a reason C4Mazie changes are in this PR?

APSIM.Documentation/Models/Types/DocFrostSenescenceFunction.cs
par456: There ended up being a lot of files like this, where it was just adding Core into the includes with no other changes, I'm guessing they were needed?

APSIM.Core/Data/ApsimConvert.cs
hol353: Yeah it is a utilities type class but it relies on DataAccessor, DataArrayFilter and IDataProvider so they would all have to be moved to APSIM.Shared. I don't see any advantage though. There is enough stuff in APSIM.Shared already.

APSIM.Documentation/Models/Types/DocFrostSenescenceFunction.cs
hol353: Yes it is needed because of the reference to IFunction, which is now in APSIM.Core. I'll check all files to ensure they are needed in the PR>

Prototypes/C4Maize/C4Maize.apsimx
hol353: This needs to be in the PR as I couldn't write a converter to convert it's manager script.

Models/Core/Overrides.cs
hol353: Yeah but locator doesn't return multiple matches. At least now this method is private in Overrides, the only place it it used. I would recommend a complete rework of the multiple overide mechanisms in APSIM.

PR #10169, opened 2025-07-15
Tree row tidy up

### LINKED ISSUE #10168: Tidy up tree row radiation interception calculations in microclimate
### What happened?

When doing a multi zone simulation with a tree canopy it is possible for the area of the canopy to be different to the area of the zone.  For instance, in an orchard the tree canopy will overhang the grass alley way.  In other situations, the tree canopy may be smaller than its current zone.  In calculating potential EP, the advective term assumes the canopy is the same size as its zone which under-estimates PET in the former situation and overestimates it in the latter.  Need to be able to adjust PETa for the size of the canopy relative to the size of the zone.  This is only relevant for tree row zone simulations.
The code for determining which radiation interception model to use is also a bit clunky and a better method of choosing the radiation interception model is needed.

Windows

HamishBrownPFR:
resolves #10168

### DIFF
--- a/ApsimNG/Classes/Property.cs
+++ b/ApsimNG/Classes/Property.cs
@@ -417,6 +417,14 @@ namespace UserInterface.Classes
                     if (water != null)
                         DropDownOptions = water.AllowedRelativeTo.ToArray();
                     break;
+                case DisplayType.StrumTreeTypes:
+                    DisplayMethod = PropertyType.DropDown;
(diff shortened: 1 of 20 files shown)


HamishBrownPFR:
@hol353 @par456 @ric394 I am not sure why the stats are better for grape vine but it is only a very small improvement.  Please review and merge if happy

PR #10180, opened 2025-07-16
10171 Barley ObservedInput

par456:
working on #10171


par456:
5 new derived variables appearing as expected, can be merged.

Issue #10170, opened 2025-07-15
New barley model based on the WinterCereal model


PengchengHU:
### What needs to be changed?

Adding a new barley model

PR #10186, opened 2025-07-16
Barley observed data clean up again

PengchengHU:
Working on #10170


par456:
Removed a duplicate from Barley.Grain.Wt, fixed height units

PR #10195, opened 2025-07-17
Optimise fertiliser to not create FertiliserPool instances

### LINKED ISSUE #10194: The fertiliser model needs optimising

Fertiliser and FertiliserPool classes

### What needs to be changed?

Fertiliser needs to be changed to only create the FertiliserPool when a slow release fertiliser is applied which isn't a common situation.

Currently it creates a new FertiliserPool instance every fertiliser application. This in turn causes the infrastructure to clear the locator cache which has a performance penalty.

hol353:
unless necessary.

Resolves #10194


hol353:
@par456 The difference in stats in this PR is unexpected. I think it came from an earlier PR today - see #10169. For some reason, the POStats changes didn't stick when I accepted the stats. I recommend accepting stats again and merging this PR.

PR #10200, opened 2025-07-18
Arbitrator data structures

HamishBrownPFR:
working on #6789

I have made a data structure (PlantWaterOrNDelta) that holds a plants N or Water demands (to pass to the soil arbitrator) or uptakes (allocated back from the soil arbitrator) to provide a consistent way for handling these deltas and aggregating them across zones.  I have added properties for these demands and allocations into IWaterNitrogenUptakes but have not connected these to any code in this PR.  
I have also added a class that aggregates other variables across zones in the same way (ZoneAgregate.cs) to provide a consistent approach for agregating multi zone data.


HamishBrownPFR:
@hol353 please review and merge if happy.  I am happy to have a discussion about where I am heading with this.

PR #10218, opened 2025-07-24
Added functionality to locate portions of arrays with depth strings.

### LINKED ISSUE #8686: Reporting of variables that are depth arrays - make it based on a specified depth/depth range rather than by index
### Describe the new feature

e.g. [Soil].sw.100mm rather than [Soil].sw.{2]
- deal with depth ranges (that don't necessarily align with simulation layers)
- deal with aggregation of concentration-type variables as well as amount-type
- still allow aggregation across timesteps (as is currently there)

### LINKED ISSUE #9991: Summing over partial array from report puts wrong variable name in datastore

hol353:
Resolves #8686
Resolves #9991


hol353:
@HamishBrownPFR  FYI

This PR fixes #9991 which means reports over partial arrays have had their column names changed. This has impacted the stats for SCRUM.

In ObservedDataSVS.xlsx I changed column headings for 4 columns that were all named the same

        (sum(Soil.OutputNutrientLayers.NH4(1))+sum(Soil.OutputNutrientLayers.NO3(1)))
        to
        (sum(Soil.OutputNutrientLayers.NH4(1:3))+sum(Soil.OutputNutrientLayers.NO3(1:3)))
        (sum(Soil.OutputNutrientLayers.NH4(1:4))+sum(Soil.OutputNutrientLayers.NO3(1:4)))
        (sum(Soil.OutputNutrientLayers.NH4(1:5))+sum(Soil.OutputNutrientLayers.NO3(1:5)))
        (sum(Soil.OutputNutrientLayers.NH4(1:6))+sum(Soil.OutputNutrientLayers.NO3(1:6)))

        sum(Soil.OutputWaterLayers.SWmm(1))
        to 
        sum([Soil].OutputWaterLayers.SWmm[1:3])

        

In MaizeObservation.xlsx I changed column heading 
        sum(Soil.Water.MM(1))
        to
        sum(Soil.Water.MM(1:8))

Issue #10221, opened 2025-07-24
Tidy up a few minor things in Microclimate


HamishBrownPFR:
### Describe the new feature

-add ET components into ICanopy interface so they can be reported
-fix scaling of albedo

HamishBrownPFR:
Closing.  Not trying to put PET components into Icanopy now.  Other part complete

PR #10107, opened 2025-06-20
Canopy temperature

BrianCollinss:
Resolves #10106

PR #10222, opened 2025-07-24
Add PET a and r into I canopy

HamishBrownPFR:
working on #10221

@hol353 @par456 @hut104 @ric394 adding these ET components to I canopy is the only way to report them given the current structure of microclimate.  Arguably they do not really need reporting but I have found it essential to review these components as part of the model sensibility checking for STRUM.

### DIFF
--- a/Models/AgPasture/PastureSpecies.cs
+++ b/Models/AgPasture/PastureSpecies.cs
@@ -170,6 +170,17 @@ namespace Models.AgPasture
             set { myWaterDemand = value; }
         }
 
+        /// <summary>The pe tr</summary>
+        [Units("mm")]
+        [JsonIgnore]
+        public double PETr { get; set; }
+
+        /// <summary>The pe ta</summary>
+        [JsonIgnore]
+        [Units("mm")]
+        public double PETa { get; set; }
+
+
         /// <summary>Light profile, energy available for each canopy layer (W/m^2).</summary>
         private CanopyEnergyBalanceInterceptionlayerType[] myLightProfile;
 
--- a/Models/G_Range/G_Range.cs
+++ b/Models/G_Range/G_Range.cs
@@ -164,6 +164,17 @@ namespace Models
         [JsonIgnore]
         public double PotentialEP { get; set; }
 
+        /// <summary>The pe tr</summary>
+        [Units("mm")]
+        [JsonIgnore]
+        public double PETr { get; set; }
+
+        /// <summary>The pe ta</summary>
+        [JsonIgnore]
+        [Units("mm")]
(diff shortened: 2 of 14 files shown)


HamishBrownPFR:
@hol353 @par456 @ric394 please review

HamishBrownPFR:
@hol353, let me know if you don't want to merge this one and I can close the PR

hol353:
@HamishBrownPFR, @BrianCollinss: This is very similar to PR #10107 in that both of you want to add something to ICanopy purely for reporting purposes. ICanopy is meant to describe how a plant and MicroClimate communicate. Adding reporting variables is something different.

The software team want to have a chat about what an alternative mechanism might be. The problem is a lack of time. I'll try to bump this up the priority list.

A possible solution maybe to just have an array of canopy structures in MicroClimate for reporting. Leave both PRs open for the moment.

HamishBrownPFR:
@hol353 no problem.  I think there is a need to give microclimate a refactor.  I have struggled alot tying to implement the tree canopy radiation interception/ET model.  I will be happy to meet with the SEG to discuss better ways of doing it and take a lead on implementing a refactor.

hol353:
@HamishBrownPFR and @BrianCollinss We've had a chat about how to do this. Let's not put these variables into ICanopy. Instead I notice that MicroClimate has this:

```c#
 private List<MicroClimateZone> microClimatesZones;
```

MicroClimateZone already has a PETA and PETR and it has a list of canopies. If the above private field became a public property then it's members could be reported. e.g.

```
[MicroClimate].Zones[0].PETA
```

We wonder whether this would be sufficient. As you point out above, we also noticed that MicroClimate needs a refactor / redesign. For canopy temps these could be added to MicroClimateZone as an array with an element for each canopy. Not elegant to report though:

```
[MicroClimate].Zones[0].MinCanopyTemperature
```

which would return a series of values, 1 for each canopy.

HamishBrownPFR:
Thanks @hol353.
The problem with your suggestion for PETa and PETr is that these properties in MicroclimateZone aggregate across all ICanopy  instances in that zone.  It is not possible to distinguish between the ICanopy components which is what I was wanting to do.

HamishBrownPFR:
@hol353 I am happy to close my PR.  Now I have finished debugging I don't really need to be reporting these.  Can leave doing a better job on this until we can organize a redesign.

BrianCollinss:
Hi @HamishBrownPFR and @hol353. Thank you for the instructions. I will have a second look to see how I can move the canopy temperature stuff into `MicroClimateZone`. Ta

HamishBrownPFR:
Closing this

PR #10223, opened 2025-07-24
TreeRow radiation and albedo scaling fixes

HamishBrownPFR:
working on #10221


HamishBrownPFR:
@hol353 @par456 @ric394 please review and merge if happy.

@junqi108 the stats are different for grapevine because this PR has fixed a bug that was affecting the tree row interception model but also the vine row interception model.  I note there is another error in the vine model causing negative albedos.  I added an error test that identified this issue but have disabled it until we can work out how to resolve it.  I think we should sit down and work out how we can harmonize the tree and vine row interception models, we should not need to have two.

PR #10237, opened 2025-07-28
Scrum kl

### LINKED ISSUE #10236: SCRUM should override CropSoil.KL
### Describe the new feature

SCRUM represents are range of crops with arbitrary root depth.  To give a sensible kl pattern it uses an exponential decline from a surface layer value to zero below max root depth.   However, for this to work properly it requires the user to set Kl to the surface layer value in all layers.  There is to much that can go wrong here.  Instead, SCRUMcrop instance should link to its soil and set kl in all layers to 1.0 and then apply its KLmodifier to give sensible patterns.  This way SurfaceKL can be made a property set in the SCRUMcrop instance which tidies things up and ensures the kl modifier is to being applied on top of user specified kl values.

HamishBrownPFR:
resolves #10236 

Added functionality so SCRUM sets KL to necessary values


HamishBrownPFR:
@hol353 @par456 @ric394 Please review and merge.  Stats have changed because two of the experiments in the test set (Hawksbay3 and 4) had KL set up wrong for the use of SCRUM.  This is the exact problem this change is intended to fix.

Issue #7381, opened 2022-07-14
Port the pasture model from AusFarm / Grazplan to APSIM


hol353:
During the 2022/23 FY we will port the pasture model from AusFarm / Grazplan (https://grazplan.csiro.au/) to APSIM. The model is currently in Pascal. This will be changed to C#, tested with a validation set and made available for review by the Reference Panel.

yashvirchauhan:
Was interested in seeing the demos from the website to learn about the models' capabilities. They seem to have been taken off?

hol353:
Which videos are you talking about? Links to all videos are here: https://www.apsim.info/support/videos/.

PR #10239, opened 2025-07-29
Updated Ausfarm pasture model

hol353:
Working on #7381


par456:
Note for Future: Currently written to match the results out of AusFarm with as few code changes as possible so that it can be compared side by side. Will need future work to tidy up and remove some of the hacks added to make results match, see the notes.md file.

PR #10250, opened 2025-08-01
Change phase

HamishBrownPFR:
working on #8175

STRUM is the only model using the DAWSPhase.  It has a parameter of the days after winter solstice to progress to the next phase.  It is fiddly having to parameterise things in DAWS so I have changed the phase to use dates instead.  It works in the same way, progresses on the date specified by the parameter but it is easier to work out which parameters to use when dates are used dirrectly.


HamishBrownPFR:
This needs other bits.  Will close this PR and put it up as part of another that has all the bits needed

Issue #10248, opened 2025-07-31
Oxygen Deficient Maize Datasets


ilhuber:
### Describe the new feature

As part of our oxygen deficient stresses project, we will be adding simulation datasets (and later updates to the models) relating to the 5 crop types that were being worked on (Maize, Soybean, Wheat, Barley, Canola). This issue is for maize. We will be adding datasets from 3 publications:

Kaur et al., 2017. https://doi.org/10.2134/agronj2016.07.0411

Kaur et al., 2020. https://doi.org/10.3390/agronomy10030378

Tian et al., 2020, https://doi.org/10.1016/j.eja.2020.126054

The initial PR will just be the maize datasets and new cultivars in the maize resource file. A separate issue will be created to bring in the model changes, which are handled entirely by PMF. A converter for cultivars will need to be written, as some values have been moved deeper into the model tree.

BrianCollinss:
@Keith-Pembleton FYI.

PR #10253, opened 2025-08-01
Maize Waterlogged Data

ilhuber:
Working on #10248 

The fits for these new sims are going to be very bad because the model changes to facilitate oxdef stresses are not a part of this PR. You can probably speculate what those changes are by looking at the commented out lines of the new cultivars. If we see differences in any other sim then something has gone wrong.


ilhuber:
@hol353 @hut104 @sno036  Would it be better to put a version of this model in prototypes or similar (with the replacements maize -- maybe in addition to this data being added to the maize validation)? This is an area of the APSIM initiative's process that I am not super familiar with.

sno036:
Great to see this coming along!

From my point of view, ideal would be that the changed code/routines/??? are applied to all maize crops in the test suite and that your new validation simulations are added to the current ones. Whether this is via replacements or something else I can't comment as I don't know what your changes are. Also don't forget that the example simulation will need some attention.

ilhuber:
@hol353 @hut104 @sno036 If updating this PR with the model changes is what is wanted I am going to hold off on doing so until #10260 has been merged.

ilhuber:
I've gone forward with adding the new stuff to this PR, but it should be easy enough to revert and reapply the diff in a different PR if needed.

ilhuber:
That unit test passes locally. I am going to try reverting the Documentation\TestFiles json and see if that goes through.

ilhuber:
For whatever reason I was under the impression that we were presenting this work to the reference panel next week. I've got no idea when this is actually going to be presented. I can revert to acf5f80 if need be, doesn't seem like the next steps are a big job for maize.

hol353:
I'm happy to merge this. The change in stats are mostly positive. Let me know what you want to do.

ilhuber:
The model hasn't been reviewed yet, so it might be best to leave it out for now. I'd be concerned about dealing with the tedium of merge conflicts with other PRs that make converter changes. It might be easiest to just revert then redo later.

PR #10256, opened 2025-08-01
Alternative approach for calculating Eo

### LINKED ISSUE #10254: Potential evaporation is based on Green Cover
### What happened?

Was this intentional or an oversight?

Context: `MicroClimateZone.Albedo` is calculated from the (green?) canopy albedo and the light extinction coefficient, accounting for the pattern of light interception across the canopy. In contrast, the original code calculates the albedo of the soil–plant system as the average of soil and canopy albedo, with canopy albedo weighted by `coverGreen`.

Image

Windows

### LINKED ISSUE #10255: Canopy cover reduces soil evaporation twice
### What happened?

Once:

Image

Then, in `CalcEoReducedDueToShading`:

Image

Windows

BrianCollinss:
Resolves #10254 
Resolves #10255 

**Note:** all major commits were reverted eventually. 

- Use Zone's albedo instead of canopy cover and soil albedo.
- ~~Remove canopy effect from calculation of Eos by `EvaporationModel`.~~

### DIFF
--- a/Tests/Simulation/SoilNitrogenPatch/PatchyMcPatchFace.apsimx
+++ b/Tests/Simulation/SoilNitrogenPatch/PatchyMcPatchFace.apsimx
@@ -9041,19 +9041,19 @@
                 },
                 {
                   "Key": "AcceptedSlope",
-                  "Value": "1"
+                  "Value": "1.02"
                 },
                 {
                   "Key": "AcceptedIntercept",
-                  "Value": "424.7"
+                  "Value": "413.77"
                 },
                 {
                   "Key": "AcceptedRMSE",
-                  "Value": "587.32"
+                  "Value": "712.05"
                 },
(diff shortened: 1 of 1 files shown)


BrianCollinss:
- I am not familiar with `RITTemplate2025` simulations. 

- FYI, when I implemented the changes in "[Remove canopy effect from calculation of Eos by EvaporationModel.](https://github.com/APSIMInitiative/ApsimX/pull/10256/commits/789b5e3d72ad4d64deab11ce7e9f276bcb2a1afd)", the stats in the Nutrient.apsimx tests improved significantly. However, yield and other plant-related outputs were negatively impacted. I’m unsure whether this is due to the way yield and other variables were calibrated while evaporation was being simulated incorrectly, or if there are other underlying factors contributing to this outcome

PR #10257, opened 2025-08-04
Strum fixes

### LINKED ISSUE #8175: Make orchard model more general

HamishBrownPFR:
Resolves #8175

@hol353 @hut104 @par456 @ric394 Merging this pull request will put STRUM into release.  PR includes

- STRUM.json added to resources including documention.
- example for STRUM
- Test set in validation with some sensibility tests and a number of evaluations against observed data
- Necessary weather and observation files

Note there are a few source code changes on this PR:
- Change to ZoneAgregate.cs is fixing a variable.  STRUM validation is currently the only thing using this so should be OK here
- Removed DAWSphase and replaced it with DatePhase.  It was to confusing trying to parameterise phase transitions in days after winter solstice so I change it to specify a date.  The reason for using DAWS was so it worked in both hemispheres.  However, I have changed STRUM to be parameterised the same as SCRUM and SPRUM (with a model that forces simulation specific parameters onto the model) so the dates of phases have to be set for each simulation and users need to ensure sensible values.  So it is cleaner to use date instead of DAWS.  STRUM is the only model using DAWSPhase.  I tried to put this change in as a seperate PR but it broke the version of STRUM in prototypes.  As the prototype is deprecated with this PR is is easier to put the two bits together.
- Change to Phenology.cs is a minor needed because of the removal of DAWSPhase


HamishBrownPFR:
@hol353 @par456 @ric394 Stats failing because of new tests added for STRUM.  Ready for review.

PR #10260, opened 2025-08-04
Check cultivar paths for validity when applying.

### LINKED ISSUE #10259: Invalid paths in cultivars are no longer reported as an error.
### What happened?

Might be related to the recent locator changes.

Windows

ilhuber:
Resolves #10259

### DIFF
--- a/Models/PMF/Cultivar/Cultivar.cs
+++ b/Models/PMF/Cultivar/Cultivar.cs
@@ -91,8 +91,7 @@ namespace Models.PMF
                 if (!Overrides.PathHasMatches(model, incoming.Path))
                     throw new Exception($"Can't find {incoming.Path} in cultivar {Name}.");
             }
-            if (Command != null)
-                undos = Overrides.Apply(model, incomingOverrides);
+            undos = Overrides.Apply(model, incomingOverrides);
         }
 
         /// <summary>Undoes cultivar changes, if any.</summary>


Models/Core/Overrides.cs
```diff
@@ -183,6 +183,12 @@ public static IEnumerable<Override> ParseStrings(string[] lines)
             }
         }
 
+        /// <summary>Evaluates whether or not a prospective override path matches up with anything.</summary>
+        /// <param name="relativeTo">The model to check relative to.</param>
+        /// <param name="path">The path to be checked against relativeTo.</param>
+        /// <returns>True if there is at least one hit.</returns>
+        public static bool PathHasMatches(IModel relativeTo, string path) => FindAllByPath(relativeTo, path).Any();
```
hol353, line 190: Why is this in overrides? Why not make it private in Cultivar where it is being called from?

ilhuber: Overrides is where FindAllByPath was moved to as a private member (it formerly had been a Model method). This method is called by the cultivar path through Overrides.Apply. An alternate approach that does not add a new public method to Overrides could be to change Cultivar Apply to use the singular Overrides.Apply for each incoming change and throw if the undos returned from that are ever empty.

However, it may be useful for other areas of APSIM in which overrides are mass applied to have similar checking. If I accidentally type .Sinulations rather than .Simulations in my config file that should probably be an error. Maybe it would be better for Overrides.Apply to throw if variables is empty when matching on NameAndType. I am not sure if there are legit reasons match nothing with NameAndType, but there are with Name (e.g. a model under replacements might not match with anything and that is fine).

PR #10263, opened 2025-08-05
Use NuGet package Microsoft.Data.Sqlite to access sqlite databases. T…

### LINKED ISSUE #5317: We should use official sqlite package

zur003:
…idy inconsistent use of "TOS" in Canola validation. 

Resolves #5317.

Note that this removes the need to explicitly add a platform-native copy of sqlite3.dll (or sqlite3.so or sqlite3.dylib) outside of the normal publishing procedure. This now gets handled automatically by the publishing mechanism, and uses the name e_sqllite3 to avoid possible conflicts.

I've made this enforce column types a bit more strictly when reading data than was the case previously. This revealed that the Canola validation tests were using "TOS" in two different ways (this is very similar to old issue #8069 for barley). TOS is a Factor in many of the simulations, but in some it was a factor with date values (e.g., 1-Apr,16-Apr,3-May), whereas in other cases it was the parent of CompositeFactors (so that "1" might point to 1-Apr, "2" to 16-Apr, etc.) This meant that in the database, the TOS column was sometimes holding a date string and other times a number. This inconsistency also carried over into imported observed data. I've made this consistent by using "TOS" when the factor is a set of dates, and "TOS_index" when the factor has CompositeFactors under it.


ric394:
Before I merge this one @zur003 I'll test to see if an apsimx image can be built without installing the apt sqlite package.

PR #10270, opened 2025-08-07
Fix intellisense bug and add unit test

### LINKED ISSUE #10268: Nodes do not properly show their children in cultivar or report.
### What happened?

Relating to comments on #10259 from @yacobbeletse, when displaying completions for a node that is in scope we only see the Node properties and not any of the model's children that had formerly been displayed. This is likely due to the recent Model <-> Node split. 

Something like 

```cs
if (o is Node)
    o = (o as Node).Model;
```
at the beginning of ExamineObjectForContextItems in NeedContextItemArgs.cs could be a quick fix (I don't imagine someone editing a cultivar or report would ever care about the node), but it may be better to address the underlying issue of finding the wrong thing in scope.

Windows

hol353:
Resolves #10268


par456:
That was a good cleanup of some nasty looking code.

Issue #10269, opened 2025-08-07
APSIM crashing and disappearing without a sign


KirstenVer:
### What happened?

2025.7.7821.0
This has hit me several times in the last few days where I have APSIM open with several simulations open (and not necessarily all saved) and then I think it is because I go to open another APSIM from Explorer (double click on APSIM file), the open APSIM is completely gone.

I also think that in the past when I opened APSIM files by double clicking in Explorer it would open a new APSIM instantiation, but now it adds it as a tab in existing open APSIM. So that has worked (for me to notice that), but sometimes it doesn't.

It is possible this goes wrong when opening an APSIM file that was from an earlier version??

I probably liked it better when they each opened into their own instantiation (if that's the right word).

### What Operating System are you on?

Windows

zur003:
Whether simulations open in a new tab or a new instance should be made a user-settable option.

I don't have any insights into why APSIM would be crashing, though. Is there a particular .apsimx file that triggers the issue?

par456:
> Whether simulations open in a new tab or a new instance should be made a user-settable option.

I was just thinking the same thing, should be settable in the settings menu.

> I don't have any insights into why APSIM would be crashing, though. Is there a particular .apsimx file that triggers the issue?

In order to open in the same window, apsim has to do some system calls and call a function on a running process, there could be an exception being thrown and not handled as part of that which is causing this crash.

Would be really useful to know if there is a particular file that causes this, or if it happens when you're doing something in particular with the running apsim window.

KirstenVer:
I will explore a bit further - I initially thought it was just me accidentally closing APSIM. Was opening and closing a lot of files to learn from / copy manager code to create new manager code. Some files I opened through Explorer and some from within APSIM (which didn't always work - related?).

KirstenVer:
A quick play with APSIM opening multiple files couldn't reproduce it at will. And this time the Explorer opened files did open in new APSIM windows. 
Maybe an unlucky glitch? (Murphy's Law - ahead of tight deadline?)

I'll keep an eye on it and see if I can reproduce.

zur003:
@KirstenVer - Getting multiple new APSIM windows from Explorer-opened files should not normally have happened with the current version. Thinking about it, though, I suppose it would if you opened an Apsim window, opened another (directly rather than by clicking on an .apsimx file), then closed the original window.

KirstenVer:
[...]

PR #10276, opened 2025-08-08
Whether double-clicking opens a new tab or new instance made user-set…

zur003:
…table.

Also explicitly release the Mutex when the hosting instance of the application is closed.

Working on #10269 

I have not been able to replicate the original Issue, although explicit release of the mutex should help prevent possible AbandonedMutex exceptions.


zur003:
retest this please, jenkins

PR #10279, opened 2025-08-08
#8784 Graph title resolved

### LINKED ISSUE #8784: Only set the graph axis name when it's manually set

JPaul-Vasa:
Resolves #8784 
Resolved graph titles issues.

1. Automatically resets the axis titles from manually set titles to the dropdown X and Y axis when selected,
2. Removed hardwired Axisposition set to position 'Bottom'


ric394 (approved):

JPaul-Vasa:
Retest this please Jenkins

PR #10283, opened 2025-08-10
When another application modifies our .apsimx file, give the user the…

### LINKED ISSUE #9558: Reload apsim files in the GUI
### Describe the new feature

If a file is changed using a text editor while it's open in the GUI, the only way to load the updated file is to close it, avoid saving, and reopen it, which is time consuming and the updates may be lost if the old version is accidentally saved.

zur003:
… option of re-loading it.

Resolves #9558

### DIFF
--- a/ApsimNG/Presenters/ExplorerPresenter.cs
+++ b/ApsimNG/Presenters/ExplorerPresenter.cs
@@ -1,15 +1,11 @@
 ﻿using APSIM.Core;
 using APSIM.Shared.Utilities;
-using DocumentFormat.OpenXml.Drawing.Diagrams;
-using Gtk;
 using Models;
 using Models.Core;
-using Models.Core.ApsimFile;
 using Models.Core.Run;
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.IO.Abstractions;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;


ApsimNG/Presenters/ExplorerPresenter.cs
```diff
@@ -1178,11 +1217,110 @@ public static (bool, string) CheckIfIconExists(string iconName)
...
+        /// <summary>
+        /// Received if and when our associated file has been changed by another application.
+        /// There is an oddity here in that some editors (such as Notepad++) save a file in
+        /// several steps, and we get notified each time, but we only want to give the user a single
+        /// dialog, not several. Hence we ignore any notifications that come within 1/10 second of
+        /// the first.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void Watcher_Changed(object sender, FileSystemEventArgs e)
```
hol353: No underscores please Eric.

ApsimNG/Presenters/ExplorerPresenter.cs
```diff
@@ -1178,11 +1217,110 @@ public static (bool, string) CheckIfIconExists(string iconName)
...
+                return;
+            _lastTimeFileWatcherEventRaised = DateTime.Now;
+            Gtk.Application.Invoke(delegate
+            {
+                int result;
+                string message = $"The file {StringUtilities.PangoString(this.ApsimXFile.FileName)} has been modifed by another application."
+                                + "\n \nClick \"Yes\" to reload the file or \"No\" to continue."
+                                + "\n\nWARNING: Any changes you have made here may be lost!";
+                using (MessageDialog md = new MessageDialog((view as ExplorerView).MainWidget.Toplevel as Window, DialogFlags.Modal, Gtk.MessageType.Question, ButtonsType.YesNo, message))
+                {
```
hol353: Rather than using MessageDialog and a Gtk.Application.Invoke can you just call MainView's method: 
```c#
QuestionResponseEnum AskQuestion(string message);
```

ApsimNG/Presenters/ExplorerPresenter.cs
zur003: Ok. I'm blame that on Visual Studio being overly helpful.

ApsimNG/Presenters/ExplorerPresenter.cs
zur003: It makes sense to use the AskQuestion method, but I think the Invoke in still needed to ensure that execution occurs on the main thread, as Gtk requires.

PR #10289, opened 2025-08-11
Interpolate U and Cona using a half-cosine function.

### LINKED ISSUE #10288: Interpolate U and Cona
### Describe the new feature

Currently, U and Cona sharply change when a transition from summer to winter or vice versa occurs. We can use a cosine function to interpolate these values instead.

Image

BrianCollinss:
Resolves #10288 
see #10290 and #10291


hol353:
@BrianCollinss I would classify this as a science change that the Reference Panel (RP) needs to discuss. I know a bunch of people would like to get rid of cona/u and replace it with something else. I think we need direction from the RP before proceeding.

BrianCollinss:
Thanks Dean @hol353. We (CSAS) and NSW DPI are working on a proposal for an upcoming grant. Our focus will be on soil moisture monitoring/modelling and management and we have been keen to find a way to contribute to APSIM as well. So far, our ideas have been around improving APSIM soil evaporation. If, as you mentioned, there is interest in replacing CONA/U (Ritchie's) with another approach, we are keen to include it in our proposal. Would you please advise how we can get some direction from the RP on this matter?

hol353:
Talk to your RP rep and get it tabled at the next meeing or if more urgent then perhaps your rep can arrange with the chair to do an out-of-session email to RP members.

BrianCollinss:
For now, I'll close this.

PR #10316, opened 2025-08-18
Simplified dung / urine return in SimpleGrazing

### LINKED ISSUE #10103: A simpler approach to dung and urine N partitioning for SimpleGrazing

Dung and urine N partitioning in SimpleGrazing. This will also affect SimpleCow.

### What needs to be changed?

- DungWt to be calculated from from intake and digestibility as currently
- Apply a C:DM ratio to get the carbon content. Default to 0.45 but allow this to be reset within the simulation
- Apply a set N concentration in the dung. Default to 2.6 g N / 100 g DM in the dung but allow this to be reset within the simulation
- Specify the fraction of intake N that exits the simulation (product, lanes etc.) as is the current case. Allow this to be reset during the simulation (and so the user can deal with that varying within the year if they wish). 
- Urine N is intake N - N exited simulation - dung N 
- Remove the "Fraction of defoliated Biomass ..." as an option
- Add an "Advanced Excreta Options" parameter that by default hires the last two "Send some fraction ..." parameters so that only specialists can see them/ These should default to 0.

The attached spreadsheet for the records/reminders. The option chosen is the third option on the third worksheet.

@rcichota 

[DungPartitioning.zip](https://github.com/user-attachments/files/20788639/DungPartitioning.zip)

To simplify the partitioning. There are currently too many options with complex logic.

hol353:
Resolves #10103


hol353:
@sno036 and I are happy with the changed stats which is due to a new algorithm in SimpleGrazing.

Issue #7759, opened 2023-02-06
Sesame model


andomariot:
UniSQ Centre for Sustainable Agricultural Systems is leading a project in Understanding Sesame with the Broadacre crop Initiative program (BACI) and partnership with QDAF.
The project is in its second year now and have collated field data  from central Queensland region ( 2017-2022) and is continuing  to collect data for the coming season 2022 to 2024 in collaboration with CQU.
Sufficient data may come then available to develop a sesame model with APSIM-X .
The approach will use the mungbean plant architecture as  model template but will learn more from oilseed plant modelling for processes, yield formation and oil content.
@APSIMInitiative/other-crops 
@APSIMInitiative/legumes

hut104:
Hi
I wonder if the soybean model may be more appropriate given its oil content.
I would also recommend using newly developed ways to link these plant templates to avoid duplication of efforts.  I'd suggest looking into this once model development commences.

tefera21:
UNOFFICIAL

Hi,

I also agree with Neil that a template of the soybean model is more appropriate for the Sesame-APSIM model development/parameterisation.

hol353:
@APSIMInitiative/other-crops: FYI

BrianCollinss:
More information will be soon provided regarding the parameterisation and validation datasets.
Contact: Brian Collins ([brian.collins@unisq.edu.au](mailto:brian.collins@unisq.edu.au), @BrianCollinss).

BrianCollinss:
@hol353 the development of the Sesame model is under way. Some field data has been used to test the model and more is being collected. The model performs reasonably well so far. May I please know if it can be added to the Prototypes?

hol353:
Absolutely. Prototypes is the place to put models that are being constructed.

PR #10351, opened 2025-08-25
Update LAIncrementWithBranching function.

BrianCollinss:
Working on #7759

### DIFF
--- a/Models/Functions/DemandFunctions/LAIncrementWithBranching.cs
+++ b/Models/Functions/DemandFunctions/LAIncrementWithBranching.cs
@@ -1,16 +1,26 @@
-﻿using System;
+﻿using Models.Core;
+using Models.PMF;
+using System;
 using System.Linq;
 using APSIM.Core;
-using Models.Core;
 
 namespace Models.Functions.DemandFunctions
 {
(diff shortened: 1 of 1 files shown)


BrianCollinss:
@par456 May I request that this be merged? The APSIM container on our UniSQ HPC is built from the master branch, and we require this merge to meet a Sesame project milestone. Thank you.

par456:
Just having a look at this. Is there a reason that this new function can't be split off into its own function instead of having the old code and new one in the same function and split by a flag?

BrianCollinss:
@par456 It's just used in Sesame so the old function is also my work for the sesame prototype. Both functions account for Branching in calculating leaf area but that flag/option adds a level of complexity, if needed.

par456:
Generally we try to avoid having flags that drastically change the functionality of a model/function, because it is very easy to accidently run it in the wrong mode. Whereas if this was implemented as two different functions it becomes much harder for that to happen. 

While at the moment you are the only one using this, it is possible in the future someone else does and it may not be clear then what the flag is for.

If you were to split this new functionality off to something like "LAIncrementWithBranchingComplex", or move the previous functionality out to its own function (if that makes more sense), then it should be suitable to merge.

BrianCollinss:
Thanks @par456. Please review.

PR #10357, opened 2025-08-25
Added `IrrigationEvents` to Irrigation class.

### LINKED ISSUE #10356: Add Output to the Irrigation model for the Number of irrigation events triggered in a season
### Describe the new feature

It could be helpful to avoid the need for more complex output variables or manager scripts.

BrianCollinss:
Resolves #10356

### DIFF
--- a/Models/Management/IIrrigation.cs
+++ b/Models/Management/IIrrigation.cs
@@ -9,6 +9,9 @@ namespace Models
         /// <summary>The amount of irrigation actually applied (mm).</summary>
         double IrrigationApplied { get; }
 
+        /// <summary>The number of irrigation events triggered.</summary>
+        double IrrigationEvents { get; }
+
         /// <summary>The depth at which irrigation is applied (mm).</summary>
         double Depth { get; }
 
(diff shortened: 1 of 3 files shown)


par456:
We don't really want to start adding in event counters like this to APSIM, when really we should be providing a way for reports to output this like APSIM Classic could. If we started adding these in, then next it would be expected that harvesting, or sowing or fertilizing should also have counter properties which starts getting messy. 

I've opened up an issue #10415 to get the functionality added to report like we'd expect it, but for access in manager scripts, it's better if the script itself tracks the events, because generally you'll want to filter that more than just the number of times it's happened in total in the simulation.

In the short term I would recommend continuing to use a manager script to output a counter like this to a report (if that's what you need), and we can add the functionality to report in the future.

BrianCollinss:
Thanks @par456. Just a note: Number of irrigations in a season, especially when AutoIrrigation exists, is usually as important as total irrigation in-crop. Hence, it is a bit different from the other events you mentioned. But, if you think this kind of codes can clutter the source code, I will close it. I agree that the code from Classic will be quite helpful.

Issue #7505, opened 2022-09-14
Soil erosion Model


hol353:
Initially this will be a port of the soil erosion model from APSIM 7.10. The first version won't have soil profile reduction code because we need to determine if this is desirable and how to implement it in Next Gen.

hol353:
@APSIMInitiative/soil: FYI

edvlmaas:
Could there please be an update on this model? I have version 2026.2.7986.0 and I'm interested in knowing if the Erosion model is ok to use and how accurate it is. It would be nice to have an entry for it in the documentation at https://docs.apsim.info/ .

PR #10368, opened 2025-08-26
Set erosion cover using surfaceOrganicMatter and microclimate covers

HamishBrownPFR:
working on #7505

erosion cover property is declared and used but never set.  Adding code to set it.


HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
@ric394 @par456 @hol353 This is failing because it cant find an observed file in one of the tutorials.  Pretty sure what I have changed shouldn't have broken that.  

What is the polite phrase we use to get the build system to re-test now Jenkins in no longer in charge

ric394:
@HamishBrownPFR your PR looks fine on POstats2. I don't see any errors. 
At the moment there is no phrase to rerun the github workflow. But you can rerun it from the actions tab above:

**Click actions**

image

**Find your PR validation workflow**

Then find your PR in the list. Making sure the it's the model validation workflow and not the build and test workflow.

image

**Rerun the workflow**
Hopefully everyone has permissions to rerun these but please let me know if you don't have this option.

image

HamishBrownPFR:
@ric394 thanks for your reply.  I see it is not show an error now.  It was this morning when I checked.  I wonder why it has changed now?

Anyway, good to know how to trigger a retest.  Thanks

HamishBrownPFR:
@hol353 @ric394 @par456 please review and merge if happy.

Models/Soils/Erosion.cs
```diff
@@ -158,6 +165,10 @@ private void OnDoSoilErosion(object sender, EventArgs e)
             soil_loss_bed = 0.0;
             soil_loss_susp = 0.0;
 
+            double g_contact_cover = surfaceOrganicMatter.Cover; //+ g_basal_cover;
+            double visible_contact_cover = g_contact_cover * (1.0 - microClimate.CanopyCover * 0.44); // A_to_evap_fact = 0.00022 / 0.0005 = 0.44� confirmed from APSIM classic developers
+            erosion_cover = visible_contact_cover + microClimate.CanopyCover * 0.44;
```
hut104, line 170: Is there a reason why we need the 0.44 term in here?  That is a conversion factor between Joe Ritchie's empirical coefficient for evaporation and the specific area of stubble (from memory).

Models/Soils/Erosion.cs
```diff
@@ -24,6 +25,12 @@ public class Erosion : Model
         [Link]
         private ISoilWater waterBalance = null;
 
+        [Link(Type=LinkType.Scoped)]
+        private ISurfaceOrganicMatter surfaceOrganicMatter = null;
+
+        [Link(Type = LinkType.Ancestor)]
```
hut104, line 31: Can these two links just be [Link]?

[Link(Type=LinkType.Scoped)] is the same as [Link]

[Link(Type = LinkType.Ancestor)] - some simulations seem to have microclimate in zones still.

par456:
Just working on our new process for this, once this is good again on your end, put the review tag back on to let us know.

Models/Soils/Erosion.cs
HamishBrownPFR: No problem

Models/Soils/Erosion.cs
SteRee: We were trying to duplicate how this was implemented in Classic.  I don't have a lot of understanding of Classic but was told the below was how it is set out in Classic with the `p%crop_cover_wtg` being equal to 0.44. Possibly I am misunderstanding something.

`g%contact_cover = g%resid_cover + g%basal_cover + g%cover_extra` 

` visible_contact_cover = g%contact_cover * (1.0 - g%crop_cover * p%crop_cover_wtg)`

` g%erosion_cover = bound(visible_contact_cover + g%crop_cover * p%crop_cover_wtg, 0.0, 1.0)`

As our main reason for needing this was for it to work with SCRUM and STRUM we also ignored the basal_cover as this is  not produced by SCRUM and STRUM. If other plant models produce this output it might need to be included in a way that doesn't error if basal cover isn't produced.

hut104: I think that code was commented out in the 7.10 version.
https://github.com/APSIMInitiative/APSIM710/blob/1a5379b97030d0c217ca3cb7cf6f65ade17ff107/Model/Erosion/erosion.for#L752C1-L759C58

I think that the 7.10 model just uses SurfaceCoverRunoff from SWIM or Soilwat.

I also think the new code's reference to A_to_evap_fac is totally irrelevant.  It is not part of the erosion algorithm but the evaporation model.

HamishBrownPFR: @SteRee I can implement changes suggested here by @hut104 so Erosion gets its cover value from the soil water models rather that calculating its own cover.  Makes sense not to duplicate the way things are done.  Will this work for what you are needing to do with the erosion model

SteRee: Yes I think that makes sense and will work for us.  Happy for you to change it to use the `cover_surface_runoff`.

hut104: Thanks for that.  The model will need to reproduce the science in the existing model that was previously tested.
We will then need to look to see what datasets we have to check how it is working as well.

HamishBrownPFR: @hut104 Who implemented Erosion in the classic?  I.e who do I talk to check how the science is implemented in the old model and find some tests?

HamishBrownPFR: @hut104 this is not straight forward.  SoilWat has a IFunction called CNReductionForCover which gets covers from all canopies and determines a total cover.  It seems SWIM does similar calculations but internally.
Would it make sense if Microclimate did the calculations to determine total cover and the Erosion, SoilWat and Swim got their values from there.
Proposed coarse of action:  
1. Implement total cover calculations in Microclimate
2. Have Erosion get those values
3. Raise a new issue to remove these duplicate calculations from SWIM and SoilWat and have them reference Microclimate
4. Do the programming needed to resolve the 3.

Is this an appropriate way forward or can you recommend something better.

Models/Soils/Erosion.cs
HamishBrownPFR: Rather than depend on the scoping rules (because I don't really understand what they do) I have changed it so surfaceOrganicMatter is always found relative to its parent zone and microClimate is always found relative to its parent simulation.

hut104:
UNOFFICIAL

The covers of each canopy is discounted for height (ie tall canopies do not influence runoff or erosion as much as short canopies).
It may be easier to just use the existing code and we can flag this for possible refactoring later.
It may be that we change the formulation to use canopy strata  from microclimate, but it may be quicker to just implement the Classic code for now.

hut104:
BTW everyone watching this PR....

I think I have sediment data that we can add to one of the long term simulation test sets already in the system.  It was a catchment study with 20-30 years of runoff/sediment data.  I think we also have the NO3 content of the runoff water and so will have nitrate export as well.

I think there are other datasets available online for the PERFECT model that could be converted to APSIM

sno036:
Perhaps of interest. @GhimireChandra and I (mostly Chandra!) are compiling some NZ datasets at the moment. I have on my list to come back to you @hut104 for that longer-term dataset. If anyone wants a conversation just let me know.

PR #10369, opened 2025-08-26
Add workflow condition to stop triggers

### LINKED ISSUE #10346: Patching Intermittent sqlite error
### What happened?

Not sure if this is related to the change over to the sqlite nuget package the other day, but have been getting a random sqlite exception in validation over the last few days.

Windows

### LINKED ISSUE #10367: Stop GitHub Workflows from triggering on forks
### What happened?

Currently the new build system's GitHub workflow is triggered on a user's fork when pull requests are created. 
Workflow runs should only be triggered on the main fork (apsiminitiative/ApsimX) to reduce noise from GitHub notifications and to avoid confusion.

The intention is to allow this feature in the future for those user's who would like to do there own runs on their own Azure account, but right now it is not a priority.

Windows

ric394:
resolves #10367
resolves #10346

### DIFF
--- a/.github/workflows/run-model-validations.yml
+++ b/.github/workflows/run-model-validations.yml
@@ -14,6 +14,7 @@ on:
 jobs:
   run-model-validations:
     name: run-model-validations
+    if: github.repository == 'APSIMInitiative/ApsimX'
     runs-on: ubuntu-latest
     env:
       ALL_APSIM_FILES: ""
--- a/.github/workflows/run-model-validations.yml
+++ b/.github/workflows/run-model-validations.yml
@@ -14,7 +14,7 @@ on:
 jobs:
   run-model-validations:
     name: run-model-validations
-    if: github.repository == 'APSIMInitiative/ApsimX'
+    if: github.repository_owner == 'APSIMInitiative'
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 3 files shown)


ric394:
retest this please jenkins

par456:
Looks like jenkins threw the sql error again

ric394:
> Looks like jenkins threw the sql error again

I'll give this one more try as I'm not changing any sql code.

ric394:
retest this please jenkins

par456:
This branch threw the sqlite error again, but on a different using. Using same pattern as before to fix that error and adding it to this branch.

ric394:
retest this please jenkins

ric394:
@par456 this one should be right to review

par456:
Good catch, I didn't even think of that, that might be the root cause of this problem.

Previously these were both wrapped in a using, but SqliteDataReader only has a DisposeAsync, no normal Dispose function, so was randomly throwing an exception on jenkins.

I split that off into a specific call to DisposeAsync, and then SqliteCommand started giving the same random error, so I did the same.

But you're right, since these are Ansync calls, they should have an await. My theory was before when we moved to the nuget package that the version might have changed and SqliteCommand lost it's normal Dispose, and the using statement was now using the wrong function call.

PR #10371, opened 2025-08-26
Ensure Gtk has adequate time to display descriptions correctly.

zur003:
Resolves #10360.

The problem was not with Dropdown widgets, but rather with excess space being used by the description above. There was already an attempt in ExplorerView.cs to deal with this, but because some layout in a Gtk TextView occurs using "idle" handlers rather than "event" handlers, that earlier fix did not always work.

PR #10379, opened 2025-08-27
Improve behaviour of DropDownView when dealing with long lists of items.

### LINKED ISSUE #10378: Initial popup display of a DropDownView is poor if there are many items
### What happened?

The DropDownView, which is essentially a wrapper for GtkComboBox, does not initially function well when there are many items in the display list. The first time the list is dropped down, it displays slowly, may not display the complete list, and may not display the scrollbar. Subsequent use of the list behaves as expected.

A simple example of the problem may be seen in the sowing rule manager for the wheat example. Selecting the wheat cultivar as a parameter often exhibits this problem, as the number of wheat cultivars is rather large.

Windows

zur003:
Resolves #10378 

The initial display is still rather slow when there are many items, but a wait cursor assures the user something is happening.
All items are listed and the scrollbar displays correctly.

### DIFF
--- a/APSIM.Shared/Utilities/SQLite.cs
+++ b/APSIM.Shared/Utilities/SQLite.cs
@@ -176,7 +176,7 @@ namespace APSIM.Shared.Utilities
                     }
                 }
                 reader.Close();
-                reader.DisposeAsync();
+                reader.Dispose();
             }
             return table;
         }
--- a/APSIM.Shared/Utilities/SQLite.cs
+++ b/APSIM.Shared/Utilities/SQLite.cs
@@ -176,7 +176,7 @@ namespace APSIM.Shared.Utilities
                     }
                 }
                 reader.Close();
-                reader.Dispose();
+                reader.DisposeAsync();
             }
             return table;
         }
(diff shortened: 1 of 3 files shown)


zur003:
retest this please, jenkins

zur003:
retest this please, jenkins

zur003:
retest this please, jenkins

ric394:
@zur003 I believe this was originally using dispose. Oddly the DisposeAsync method seems to not throw random exceptions, which is what we were seeing lately in Jenkins runs.

zur003:
@ric394 - I've run into those Exceptions myself in Jenkins runs (e.g., PR #10371), but I would wager that using DisposeAsync this way isn't going to really fix the issue, and might create some new problems. Calling DisposeAsync without awaiting is only asking for trouble.

Those errors are rather odd, though. They're occurring only when executing the unit tests, at points where the Sqlite database is being closed, and they don't seem to happen consistently. I've never seen the problem when running the unit tests locally.

ric394:
Old sqlite dll fetching method removed here as no longer needed. See #10263

Issue #10336, opened 2025-08-20
Export to Excel is not working


sno036:
### What happened?

Missing DLL?

Could not load file or assembly 'ClosedXML.Parser, Version=1.0.0.0, Culture=neutral, PublicKeyToken=1d5f7376574c51ec'. The system cannot find the file specified.

### What Operating System are you on?

Windows

cranju:
Any updates on this one as it is still not working. I am encountering the same error: "Could not load file or assembly 'ClosedXML.Parser, Version=1.0.0.0, Culture=neutral, PublicKeyToken=1d5f7376574c51ec'. The system cannot find the file specified."

zur003:
Does the file ClosedXML.Parser.dll exist in the bin folder? 

Is this on Windows? I just updated my Windows version today, and the dll installed as expected. I have no idea why it might be unavailable, unless possibly some corporate policy is blacklisting it

sno036:
Hi @zur003 - thanks for looking at this! I've looked in the four directories (Debug and Release, net8.0 and net6.0) and there is a ClosedXML.dll but no ClosedXML.Parser.dll. I am on Windows and I think @hol353 mentioned it was the same for him.

@cranju - @hol353 pointed out that the export to CSV is working and then you can get it into Excel from there as a workaround until this is fixed.

zur003:
@sno036 - I think I found the problem. APSIM.Workflow and ApsimNG were pulling in different versions of ClosedXML. the newer version puts some logic in OpenXML.Parser as a separate module but the older version did not. I've submitted a pull request to fix this, but Jenkins (or whatever is now building things) isn't currently working for me.

sno036:
Working perfectly now thanks @zur003!

PR #10380, opened 2025-08-27
Update version of ClosedXML to be consistent with APSIM.Workflow

zur003:
Working on #10336 

APSIM.Workflow and ApsimNG were using different versions of ClosedXML. Not certain whether this will resolve the problem, as it seems a bit hard to reproduce.


sno036:
@zur003 - I am happy to test it. I would need to pull from your branch to do that.

zur003:
@sno036 - My "fix" has now gone into the main branch. Can you test to confirm that the problem has indeed been resolved?

Issue #10396, opened 2025-08-29
Exceptions are sometimes raised when Sqlite connections are closed in unit tests


zur003:
### What happened?

During Jenkins runs, we are often seeing NullReferenceExceptions raised when the Sqlite data store is closed. These errors do not occur consistently, but are frequent enough that getting pull requests through the build system can be difficult. 

### What Operating System are you on?

Windows

zur003:
Duplicate of #10346

PR #10397, opened 2025-08-29
Don't clear pools until after the connection is closed.

zur003:
Working on #10396 

The call to "ClearAllPools" is there to help ensure that a "closed" connection doesn't still hold the database file open and prevent us from deleting or renaming it, but it was probably in the wrong place.

### DIFF
--- a/APSIM.Shared/Utilities/SQLite.cs
+++ b/APSIM.Shared/Utilities/SQLite.cs
@@ -95,11 +95,11 @@ namespace APSIM.Shared.Utilities
         {
             if (_open)
             {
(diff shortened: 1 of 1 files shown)


zur003:
Why is the APSIM.POStats2 task not completing?

par456:
> Why is the APSIM.POStats2 task not completing?

Not sure, we're looking into that today. For the moment everyone else can ignore the results of POStats2, we are in the process of making sure it matches the existing system, so there are still some issues to work out.

zur003:
Closing this for now, as it is better dealt with in the context of the changes @par456 is making in conjunction with #10346.

PR #10416, opened 2025-09-04
10414 Add Binary metfile reader

### LINKED ISSUE #10414: Allow APSIM to read binary met files
### Describe the new feature

I have been working with metfiles that were compressed into a binary format to save space, but we'll need to use those in APSIM soon. It would be useful be able to read that format directly into APSIM.

We aren't looking to use this for most APSIM simulations, so we don't want a file writer for these yet, the project these are coming from just had a lot of data and a normal txt met file would cause them to run out of space.

Format details:
```
Binary Metfile
-------------------
In order to reduce file size, the following modifications are made to the data when writing:
- A start_date is stored in the constants and each row of the data is assumed therefore
  to be the next day in sequeunce. When reading out, the date must be added back to the 
  data rows
- The file has been optimized for smallest size after zipping, therefore while some of the 
  data could be made smaller, it causes zipping to be less efficient and therefore worse
  overall
- Numbers are stored as a list of symbol lookups instead of a smaller number format like 
  float16. This is to avoid any rounding or mathematical errors when doing this writing, 
  storing as symbols ensures the exact same number goes in and out.
- SYMBOLS = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", ".", "-", "/", "nan"]
- If a better version of this is developed in the future, a version string is included as
  the first bytes in the file so that versions can be handled in the future.
- Hexidecimals are used for handling memory here because the smallest unit we need is
  4-bits. This allows the script to be simplier than with binary and run faster.

The binary file is built to the follow schema:
Data Types:
- NUMBER: 4-bit int (length) + array of 4-bit int for symbol lookup
- WORD: 8-bit int (length) + array of utf-8 char

Structure:
WORD (Version)
8-bit int (Number of constants)
(WORD + WORD) x constants (Per constant: "name" + "value")

8-bit int (Number of columns)
WORD x columns (Per column: "name")
WORD x columns (Per column: "unit")

32-bit int (Number of data rows)
NUM x columns x rows
```

par456:
Resolves #10414

I've hooked this binary reader in as a pre-process on the existing string file reading. It converts the binary file to a working string version then loads it in as a memory stream which the string version did as well.

Does not seem to impact speed much (reading the file and converting it to the string version is pretty fast) even with a 50 year file that I tested it.

Need to add unit tests still.

### DIFF
--- a/APSIM.Shared/Utilities/MetBinaryReader.cs
+++ b/APSIM.Shared/Utilities/MetBinaryReader.cs
@@ -1,0 +1,235 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace APSIM.Shared.Utilities
+{
+    /// <summary>
+    /// A file that reads in a met file from binary format (extra compressed)
+    /// </summary>
+    [Serializable]
+    public static class MetBinaryReader
+    {
+        /// <summary>
+        /// Symbol dictionary for converting from a 4-bit hex to data number character
+        /// </summary>
+        private static string[] SYMBOLS = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", ".", "-", "/", "nan"];
+
+        /// <summary>
+        /// Opens a binary file and converts it to a valid string representation in a Stream object
+        /// </summary>
+        public static Stream Load(string filepath)
+        {
+            byte[] fileBytes = File.ReadAllBytes(filepath);
+            string output = Read(fileBytes);
+            return new MemoryStream(Encoding.UTF8.GetBytes(output));
+        }
+
+        /// <summary>
+        /// Converts a bytes object to a metfile string
+        ///</summary>
+        /// <param name="bytes">The bytes object from the file</param>
+        /// <returns>
+        /// Metfile as a valid met string
+        /// </returns>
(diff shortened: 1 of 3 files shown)


APSIM.Shared/Utilities/MetBinaryReader.cs
```diff
@@ -0,0 +1,235 @@
...
+                position = result.Item2;
+
+                output += SYMBOLS[index];
+            }
+
+            return (output, position);
+        }
+
+        /// <summary>
+        ///
```
ric394: Missing summary info

APSIM.Shared/Utilities/MetBinaryReader.cs
```diff
@@ -0,0 +1,235 @@
...
+                {
+                    resultString = DecodeNumber(hex, position);
+                    row.Add(resultString.Item1);
+                    position = resultString.Item2;
+                }
+                data.Add(row);
+            }
+
+            return getMetfileString(constants, columns, data);
+        }
```
ric394: See if you can simplify this by avoiding the use of Tuples.

ric394 (approved):

Issue #10432, opened 2025-09-08
Oats validation has an oats crop in replacements


hut104:
### What happened?

The oats validation set is not running the released oats model.
We should check if there is work in the released resource vs what is in the test set and work out how to capture all work.

### What Operating System are you on?

Windows

PR #10433, opened 2025-09-09
Checking for stats changes after removing replacements oats

par456:
Working on #10432

Oats had a oats crop in it's replacements, not sure if that is different from the oats in resources. Removing that from replacements and seeing if it runs differently.


par456:
Closing this so a better PR can be built.

PR #10438, opened 2025-09-10
10432 Remove Oats from Replacements in Validation

### LINKED ISSUE #10432: Oats validation has an oats crop in replacements

par456:
Resolves #10432

As stated in the issue, the Oats validation had a copy of the Oats model in its replacements, which is a pretty bad mistake. Thankfully it looks like what was in replacements was the latest version of the model, so I have been able to remove it from the validation file without it effecting stats.

However some other sections also needed to be moved.

There were a set of new cultivars defined under experiment SAE_Oats_2020, and throughout the file was a copy of "NewCultivars" containing PFR_100_05 and Coronet cultivars.

That lead to a cultivar name conflict, as I had two different copies of a new Coronet cultivar, one from SAE_Oats_2020 and one from this NewCultivars folder. For the moment to avoid causing changes in stats, I have moved both across to the Oats resource, and renamed the SAE_Oats_2020 version to Coronet_SAE_Oats_2020 and adjusted the experiment to use that one.

Lastly, I needed to edit the Locator as the Oats Initial Leaf nodes have [ ] brackets in the name. The locator thought that meant they were an array lookup, so I adjusted the code so that if the array lookup fails, it then falls back and tries to use it as normal. Shouldn't effect anything else, as this is a fallback which previously would have thrown an exception.


par456:
@hut104 

If you're happy with these changes, I'll get this merged in.

hol353:
Can you post an example path that isn't working for the locator?

Can you separate out the change to locator and then add a unit test for it?

par456:
In Oats, the new cultivars wanted to change the area value of: 
```
[Leaf].InitialLeaves[1].Area = 28.4
```

But the locator would throw an error doing this, because it thought that InitialLeaves[1] was an array lookup, not a model.

To get around this, in the Oats validation, the Oats model had instead been changed so that this leaves model was instead called "[Leaf].InitialLeaves1", but alongside it was the existing "[Leaf].InitialLeaves[2]" which did not have its name changed.

As this was the only difference between what was in the Oats replacement and what was in the resource, I chose to rename it back to having square brackets and fix the locator so it would find it. Otherwise we'll need to include a converter to fix any simulations that are using the previous oats model, and probably rename [Leaf].InitialLeaves[2] to [Leaf].InitialLeaves2 so it's consistent.

I will split the locator change to it's own PR.

Issue #9617, opened 2025-01-28
Splitting the Validation Files


par456:
### Describe the new feature

As part of the ongoing update to the build system, we are redesigning how APSIM validation files are being handled so that we can avoid the following problems:

- Merge conflicts when multiple people are adding new data to a crop model which currently requires editing the same apsimx file
- Merge conflicts when multiple people are adding new data to an excel file for validation
- The ability to run experiments within a crop validation concurrently on a cloud service

In order to do this, we need to break up the current validation files for each model, which are currently just a single big apsimx file for each model, along side weather files and excel files for the observed data. A text file input will be used to help break up these big files, so that experiments can still be grouped together when required.

After splitting these up, each experiment or group of experiment, will be placed into a sub directory within the validation/model folder, with an excel sheet that only contains records for the simulations that are connected to that new apsimx file. Weather files will not be split or put into these subfolders as they are often shared between experiments.

Some issues we need to consider:

- When working on changing a model, what file does a user work in now? Do they need to edit the base resource instead? Impacts of this.
- How does a user get a 'big picture' view of the model and graphing variables like they do in the current large files.
- When we do the split on a validation file, we have to be sure nobody is currently editing that file on a branch, otherwise they won't be able to merge their work back in.

@HamishBrownPFR @hol353 @hut104 

To do list:
- [ ] Create new project for this splitting tool outside the main apsim repo
- [ ] Investigate what needs to be split between files, what top level models and replacements are currently used in validation files
- [ ] Matching simulation names to excel rows to split up data into new sheets
- [ ] Splitting the apsimx file
- [ ] Text input file format for group experiments together
- [ ] Command line interface for program
- [ ] Unit testing to make sure that no simulations or experiments are lost during the transfer
- [ ] Unit tests to make sure no data is lost from excel sheets during transfer

HamishBrownPFR:
@par456.  Power to your arm!  This needs doing to tidy up the test set and allow us to extract maximum value from the testing system.  It is logical to have .apsimx model config files paired with observed files at an experimental level so experiments can be used to validate multiple models and custodians of experiments can easily find their experiment set ups in the testing system.  
A few immediate thoughts to consider:
[...]

PR #10463, opened 2025-09-15
9617 Add splitter for wheat validation

par456:
Working on #9617

Refactoring the splitter code so that it works in the new workflow and can be used to split up the wheat validation.

Further work in cleaning it up and creating some tests is required once I know it's working.


par456:
retest this please jenkins

PR #10465, opened 2025-09-15
Canola waterlogging data.

### LINKED ISSUE #10464: Canola Waterlogged Dataset
### Describe the new feature

We will be adding data from
Ploschuk, R. A., Miralles, D. J., Colmer, T. D., Ploschuk, E. L., & Striker, G. G. (2018). Waterlogging of Winter Crops at Early and Late Stages: Impacts on Leaf Physiology, Growth and Yield. Frontiers in Plant Science, 9. https://doi.org/10.3389/fpls.2018.01863 

This publication also covers wheat and barley, but there is pending work (not us!) on these models so we will be holding off for the time being.

ilhuber:
Resolves #10464 

Like with the other two, since the new model isn't actually present we should expect the POStats to be worse.


ilhuber:
Don't know what that error is about. Those tables exist (at least when the file is run locally), and names should be fine since they are the same table names as the soybean oxdef simulations.

BrianCollinss:
It seems the met files cannot be found when run on Azure. Have you changed their paths?

ilhuber:
Good catch! Updated through the UI (which displayed the proper relative path) but an absolute path seems to be what was written in the .apsimx file.

par456:
Is this ready for review @ilhuber? This is only adding observed data, so our team can just review and merge if ready.

ilhuber:
Yep! It is ready. Same with the maize and soybean one.

Models/Resources/Canola.json
```diff
@@ -7360,6 +7360,48 @@
...
+                "// [Grain].MaximumPotentialGrainSize.FOXgs.XYPairs.Y = 1, 1, 1.175",
+                "",
+                "// [Leaf].SenescenceRate.Vegetative.Rate.FOXSenescence.XYPairs.X = 0, 5, 10",
+                "// [Leaf].SenescenceRate.Vegetative.Rate.FOXSenescence.XYPairs.Y = 0, 0.03, 0.3",
+                "// [Leaf].SenescenceRate.Reproductive.Rate.MaximumFunction.FOXSenescence.XYPairs.X = 0, 5, 10",
+                "// [Leaf].SenescenceRate.Reproductive.Rate.MaximumFunction.FOXSenescence.XYPairs.Y = 0, 0.03, 0.3",
+                "",
+                "//    optional changes (not used) ",
+                "//   [Leaf].Phyllochron.FOXphyllo.XYPairs.Y = 1, 1.1, 1.15 ",
+                "//   [Grain].PotentialHarvestIndex.FOXhi.XYPairs.Y = 1, 0.85, 0.75"
```
par456: Can we remove the commented lines from this new cultivar?

PR #10466, opened 2025-09-16
Marianas expt

HamishBrownPFR:
Working on #7623 

Added Nitrogen data for Lincoln 2024.
Tidied up soils for Lincoln 2023 and 2024


HamishBrownPFR:
@par456 @ric394 I don't know what is causing this to throw an error.  Could someone take a look and let me know what needs to be fixed.
Thanks

par456:
Looks like the problem is being caused by the Lincoln experiments having a Soil.Physical that isn't named Physical in the tree.

That is a bug that I think we need to address (as it should be able to find by type), but if you wanted a quick fix, you could just rename the Physical nodes back to Physical and they would run.

It's a problem with Lincoln1994 and Lincoln2010

par456:
That should work now if you pull in master to update this branch.

HamishBrownPFR:
@par456 @ric394 @hol353 @hut104 please review and merge.  

Stats changed because:
1. Nitrogen data added to Lincoln2024 and some additional variables derived
2. Merge conduced to align DM and N% data which reduced N for some variables where there was duplicate (WtProportion) or replicate data (StemsPerPlant)
3. zero values removed from observed file for a few variables (WtProportion) from Lincoln2023 and Lincoln2024.

PR #10472, opened 2025-09-17
Pasture

hol353:
Working on #7381


Models/Pasture/pasture.cs
```diff
@@ -2715,7 +2717,7 @@ private void BiomassToFOM()
                 // surface residues into first layer
                 FOMData.Layer[0].FOM.amount += removed.FOM[1][1].Weight + removed.FOM[1][2].Weight;
                 FOMData.Layer[0].FOM.N += removed.FOM[1][1].N + removed.FOM[1][2].N;
-                FOMData.Layer[0].FOM.C = FOMData.Layer[0].FOM.amount * carbonFractionInDM;
+                FOMData.Layer[0].FOM.C += removed.FOM[1][1].Weight + removed.FOM[1][2].Weight * carbonFractionInDM;
```
par456, line 2720: Can I just get you to give a quick summary of why these two changes were needed.

hol353: The two lines were changed so that the match the other lines around them. They are functionaly he same as the original lines.

At first we thought we'd found a bug when looking at the code because they were different to the other lines. We don't want that to happen again.

PR #10482, opened 2025-09-18
Guard against "physical" being null

### LINKED ISSUE #10481: NullReferenceException thrown in ProfilePresenter
### What happened?

In the user interface, open the Wheat example and select "WheatSoil" (under Simulation\Field\Soil\Physical). This will cause a NullReferenceException to be thrown.

Windows

zur003:
Resolves #10481


par456:
While that's not a bad check to have, I think the actual problem is two lines higher on where it fails to find the soil parent. I'll push that fix to this branch.

PR #10484, opened 2025-09-18
Update Kubernetes and replace obsolete "IsNullOrEmpty" test

### LINKED ISSUE #10483: Use updated KubernetesClient NuGet package
### Describe the new feature

Dependabot tried to do this update with #10476, but failed because the update removes the IsNullOrEmpty method for IEnumerable (which we were using, but probably shouldn't have been).

zur003:
Resolves #10483

dependabot didn't quite get it right


par456:
I'm not overly familiar with what APSIM.Server does. When an update to a package like this comes through, how would I check that it hasn't broken that part of apsim?

PR #10498, opened 2025-09-22
Set phenology dates

### LINKED ISSUE #10497: Tidy up approach for setting germination and emergence dates and expand functionality to other phases
### Describe the new feature

For testing biomass predictions against observations we need to ensure the phenology is correct, so we are not confounding interpretations of the performance of biomass models with that of the phenology model.  
Use case 1.   Detailed biomass and N data for wheat collected at Lincoln but cultivars that have few phenology observations.  Instead of trying to cook up phenology parameters that work for this narrow dataset, it would be better just to tell the model when each stage occurred (this was observed) and keep focus on the models performance in biomass and N responses
Use case 2.  Kiwifruit prototype model, we have time coarse data for berry growth but insufficient data about the experiments to set management in a way that allows correct prediction of flowering time.  Would be better to set the flowering time for each simulation and then focus on the models predictions of berry growth.

HamishBrownPFR:
resolves #10497 

This is a straw man implementation to generate some discussion:

Do we need this functionality?  I for one would find it really useful.  If so is this the best way to go about getting it.


HamishBrownPFR:
There are a few minor diffs which I can fix.  However, I would prefer to get a higher level review of the direction of this PR before I spend any time fine tuning.

BrianCollinss:
@HamishBrownPFR I'm keen to know how it works for any stage other than emergence. Doest it scale Target thermal times, or simply skips stages to reach the target stage on a specified date?

HamishBrownPFR:
@BrianCollinss each class that implements IPhaseWithSettableCompletionDate
has a property called CompletionDate which is null by default and not set by any released code.  Each of these phases also contain if statements in their calculations of daily progression and fraction complete.  If the CompletionDate property remains null the phases proceed as normal.  If a completion date is set (this has to happen from a manager) the phase will ignore its usual calculation of daily progression and just wait until the completion date is achieved.  It will also ignore its usual calculation of fraction complete and calculate this as a fraction of days in phase relative to days to reach completion from the start and completion dates.
The intent is that this functionality can only be invoked from managers so it invisible to the general user.  There are plenty of ways unexpected things can happen.  If you set the completion date of a phase prior to the date the model reaches that phase it will sit in that phase until next year.   So if you are setting Anthesis date for example it would probably be necessary to set dates for all phases preceding this to ensure heading date is achieved prior to the specified anthesis date and the relative size of phases are sensible.  For instance, you could tell the model to complete anthesis on the 5th of October.  If HeadEmergence occurred on the 4th of October you would get an unrealistic length in the heading phase.

BrianCollinss:
Thanks @HamishBrownPFR. I was thinking exactly about

> So if you are setting Anthesis date for example it would probably be necessary to set dates for all phases preceding this to ensure heading date is achieved prior to the specified anthesis date and the relative size of phases are sensible.

I think it'd be sensible to have it throw an exception if completion date is already passed.

HamishBrownPFR:
Thans @BrianCollinss that is a good suggestion.  I will add that in

BrianCollinss:
@HamishBrownPFR given that the idea is to take phenology (at least to some extent) out of the picture, would it make more sense to make all phases ignore their normal progression up to the phase for which a completion date is set? For example, set their completion date internally based on their target TT divided by sum(target TTs) up to that phase. Of course this would mean photoperiod will be completely ignored.

HamishBrownPFR:
@BrianCollinss, that would work but it would create edge cases where you want to set multiple completion dates and would add alot of code for functionality that is not intended to be widely used.  My preference is to keep the code simple and leave the responsibility with the few users who apply the code to do so sensibly.  What do others think?

HamishBrownPFR:
I have discussed this in person with @hut104 over the last few days and done some more checking and testing as a result.
Specifically, I have merged this branch onto another where we are testing a version of the wheat model and set stage dates for an experiment where we have great biomass data but not the right sort of data for setting phenology parameters for the cultivar.  Instead, we set the completion dates for each phase to the observed values.  This worked as intended and checking a wide range of variable outputs we were not able to identify any unexpected consequences.
This remains a piece of code that will only be used in special cases by expert users and will not influence general simulations.  However, it will be very useful in model development activities where we need to remove confounding effects of phenology predictions so we think it is suitable to merge into master
@par456 @hol353 @ric394 please review

Models/PMF/Phenology/Phases/IPhaseWithSetableCompletionDate.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.IO;
+namespace Models.PMF.Phen
+{
+
+    /// <summary>Interface for a function</summary>
+    public interface IPhaseWithSetableCompletionDate : IPhase
```
par456: Don't need to extend IPhase

Models/PMF/Phenology/Phases/VernalisationPhase.cs
```diff
@@ -12,13 +14,21 @@ namespace Models.PMF.Phen
...
+        [Link]
+        private IClock clock = null;
+
         private bool firstDay = true;
         private double relativeVernalisationAtEmergence { get; set; }
+        /// <summary>First date in this phase</summary>
+        private DateTime firstDate { get; set; }
+
+        /// <summary>Flag for the first day of this phase</summary>
+        private bool first { get; set; }
```
par456: We don't need this flag, instead just check if firstDate is null and if it is, set it.

Models/PMF/Phenology/Phases/VernalisationPhase.cs
```diff
@@ -12,13 +14,21 @@ namespace Models.PMF.Phen
...
         [Link]
         private IVrnExpression CAMP = null;
 
+        [Link]
+        private IClock clock = null;
+
         private bool firstDay = true;
         private double relativeVernalisationAtEmergence { get; set; }
+        /// <summary>First date in this phase</summary>
+        private DateTime firstDate { get; set; }
```
par456: If you keep this as a string and use the dateutilities comparison method, you can allow just a day-month instead of forcing a year as well

Models/PMF/Phenology/Phases/VernalisationPhase.cs
```diff
@@ -34,12 +44,31 @@ public class VernalisationPhase : Model, IPhase, IPhaseWithTarget
 
         /// <summary>Fraction of phase that is complete (0-1).</summary>
         [JsonIgnore]
-        public double FractionComplete { get; set; }
+        public double FractionComplete 
+        {
+            get
```
par456, line 39: Since this is copied between all the phases, instead it could be written as a static function in Phenology that these all call instead.

HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
@ric394 @par456 @hol353 I am not sure why this is failing or even how to trigger a retest.  Any advise appreciated

ric394:
Just hold off for a moment @HamishBrownPFR. There's a bug in the workflow I'm fixing at this moment.

HamishBrownPFR:
OK, thanks @ric394

Models/PMF/Phenology/Phases/IPhaseWithSetableCompletionDate.cs
HamishBrownPFR: Change made

Models/PMF/Phenology/Phases/VernalisationPhase.cs
HamishBrownPFR: Change made

Models/PMF/Phenology/Phases/VernalisationPhase.cs
HamishBrownPFR: done

Models/PMF/Phenology/Phases/VernalisationPhase.cs
HamishBrownPFR: Established methods for determining Fraction complete and checking if the phase is complete to reduce code duplication in each class

HamishBrownPFR:
@par456 @ric394 @hol353 

I think this is OK to merge now but please review and let me know if there is anything further.  The stats are slightly better for a few tests.
There was one fault in the old code that I have maintained in the new set method to ensure I have got everything else correct.  Once this is merged I will fix that on a separate PR.

HamishBrownPFR:
@par456 @ric394 @hol353 can someone have a look over this.  Sunk well down the PR list so easily overlooked now

hol353:
@HamishBrownPFR Why did the stats change (for the better)?

HamishBrownPFR:
@hol353 there is a minor difference in the prediction of Phenology.Stage between germination and emergence in simulations where emergence date is set.  This seeds some minor changes through the rest of the run.  Why they are all better.  That is a fluke I suspect.  The important thing is that they are not different by much and it has only affected a handfull of the stats that involve simulations where emergence date is set.

Models/PMF/Phenology/Phases/EmergingPhase.cs
```diff
@@ -16,25 +16,24 @@ namespace Models.PMF.Phen
...
-        Plant plant = null;
-
         [Link(Type = LinkType.Child, ByName = true)]
         private IFunction target = null;
 
-        // 2. Public properties
-        //-----------------------------------------------------------------------------------------------------------------
+        /// <summary>Fraction of phase that is complete (0-1).on yesterdays timestep</summary>
+        [JsonIgnore]
+        private double fractionCompleteYesterday { get; set; }
```
hol353: Can you just make privates fields i.e. no { get; set; }. Then you don't need the [JSonIgnore]

Models/PMF/Phenology/Phases/EmergingPhase.cs
```diff
@@ -16,25 +16,24 @@ namespace Models.PMF.Phen
...
+
+        /// <summary>First date in this phase</summary>
+        [JsonIgnore]
+        private DateTime startDate;
```
hol353, line 32: [JsonIgnore] no needed.

Models/PMF/Phenology/Phases/VernalisationPhase.cs
```diff
@@ -1,5 +1,8 @@
 ﻿using System;
+using APSIM.Shared.Utilities;
+using DocumentFormat.OpenXml.Bibliography;
```
hol353: I don't think ```DocumentFormat.OpenXml.Bibliography``` is needed.

Models/PMF/Phenology/Phenology.cs
```diff
@@ -517,24 +519,79 @@ public override void OnCreated()
         }
 
         /// <summary>
-        /// Force emergence on the date called if emergence has not occurred already
+        /// Method to set the
```
hol353: Add a comment.

Models/PMF/Phenology/Phases/EmergingPhase.cs
HamishBrownPFR: done

Models/PMF/Phenology/Phases/EmergingPhase.cs
HamishBrownPFR: done.  This was in all the phases so I have fixed them all (as was the correction above)

Models/PMF/Phenology/Phases/VernalisationPhase.cs
HamishBrownPFR: Removed redundant references from all classes modified.  Visual studio has a bad habit of adding them in.

Models/PMF/Phenology/Phenology.cs
HamishBrownPFR: done

Models/PMF/Phenology/Phases/EmergingPhase.cs
```diff
@@ -29,10 +27,9 @@ public class EmergingPhase : Model, IPhase, IPhaseWithTarget, IPhaseWithSetableC
 
         /// <summary>Fraction of phase that is complete (0-1).on yesterdays timestep</summary>
         [JsonIgnore]
```
hol353: [JsonIgnore] not needed on private fields - remove.

Models/PMF/Phenology/Phases/GerminatingPhase.cs
```diff
@@ -45,15 +44,16 @@ public class GerminatingPhase : Model, IPhase
...
 
-        // 2. Private and protected fields
-        //-----------------------------------------------------------------------------------------------------------------
-
         /// <summary>The soil layer in which the seed is sown.</summary>
         private int SowLayer = 0;
 
+        /// <summary>Fraction of phase that is complete (0-1).on yesterdays timestep</summary>
+        [JsonIgnore]
+        private double fractionCompleteYesterday;
```
hol353, line 51: Another [JsonIgnore] to remove.

Models/PMF/Phenology/Phases/LeafAppearancePhase.cs
```diff
@@ -35,16 +29,23 @@ public class LeafAppearancePhase : Model, IPhase
...
+        [JsonIgnore]
+        private double fractionCompleteYesterday = 0;
 
-        private double LeafNoAtStart;
-        private bool First = true;
-        private double FractionCompleteYesterday = 0;
-        private double TargetLeafForCompletion = 0;
+        /// <summary>The target for progresson to the next phase</summary>
+        [JsonIgnore]
+        private double target = 0;
```
hol353, line 42: And a few more.

Models/PMF/Phenology/Phases/SimpleLeafAppearancePhase.cs
```diff
@@ -43,49 +38,73 @@ public class SimpleLeafAppearancePhase : Model, IPhase
...
+        [JsonIgnore] 
+        private double leafNoAtStart { get; set; }
+
+        /// <summary>relative progress through the phase yesterday</summary>
+        [JsonIgnore] 
+        private double fractionCompleteYesterday = 0;
+
+        /// <summary>The target for progresson to the next phase</summary>
+        [JsonIgnore] 
+        private double target = 0;
```
hol353, line 48: And a few more.

Models/PMF/Phenology/Phases/EmergingPhase.cs
HamishBrownPFR: Found some more and removed

Models/PMF/Phenology/Phases/GerminatingPhase.cs
HamishBrownPFR: done

Models/PMF/Phenology/Phases/SimpleLeafAppearancePhase.cs
HamishBrownPFR: done

hol353:
@HamishBrownPFR We can't seem to merge this. It is saying that Andrew request for changes needs to be resolved. Can you see anything at your end to say that you've resolved it?

hol353:
Don't worry about my last comment. We figured it out.

PR #10501, opened 2025-09-22
Fix tolerance in organ mass balance tests

### LINKED ISSUE #10500: Organ.cs mass balance check tolerance to tight
### What happened?

Organ.cs has some mass balance checks that ensure any errors in partitioning of biomass are within floating point tollerance (3e-11).  However, when organs accumulate a large mass (e.g for STRUM trunk in long simulations) errors can exceed this on occasion.  
Need to scale the tolerance relative to the size of the organ so a larger mass will tolerate a larger error

Windows

HamishBrownPFR:
resolves #10500


Tests/Validation/STRUM/RowAlley.pptx
par456, line 1: Not sure you meant to include this powerpoint?

Models/PMF/BiomassArbitration/Organ.cs
```diff
@@ -557,11 +557,11 @@ private void CheckMassBalance(double startLive, double startDead, string element
 
             double liveBal = Math.Abs(live - (startLive + allocated - senesced - reAllocated
                                                         - reTranslocated - liveRemoved - respired));
-            if (MathUtilities.IsGreaterThan(liveBal, 3e-11,0.0))
+            if (MathUtilities.IsGreaterThan(liveBal, Math.Max(live * 1e-13,1e-11),0.0))
                 throw new Exception(element + " mass balance violation in live biomass of " + this.Name + "on " + clock.Today.ToString());
 
             double deadBal = Math.Abs(dead - (startDead + senesced - deadRemoved - detached));
-            if (MathUtilities.IsGreaterThan(deadBal, 3e-11, 0.0))
+            if (MathUtilities.IsGreaterThan(deadBal, Math.Max(dead * 1e-13,1e-11), 0.0))
```
par456: The 0 at the end of this line is actually the tolerance of the comparison, so if we want to have a scaling tolerance, we should put it in here. It would also be much cleaner do your tolerance calculation on the line above and pass it in.

Then it would be like:
```
double tolerance = Math.Max(dead * 1e-13,1e-11);
if (MathUtilities.IsGreaterThan(deadBal, 0, tolerance))
```

Could we also get a comment on the line saying why we need to change the tolerance.

Also, is the Max required? Would this work if you just used = dead * 1e-13 ?

HamishBrownPFR: No problem to change the order of the comparison value and the tolerance.  It should give the same result but will be tidier.
Also fine to put tolerance calculation on separate line.
The max is required because if dead is zero the tolerance becomes zero and floating point errors will cause trigger the exception erroneously.

par456: Ah, good catch on that edge case. Just so we don't forget in the future, mention that in the comment as well so someone doesn't make that mistake.

HamishBrownPFR: comment added explaining why variable tolerance needed

HamishBrownPFR: It all gets a bit ugly with the comments and repeating the same test pattern twice.
I have put the floating point tolerence test into a method.  See below.  I have wresteled with floating point issues in tests in the pasts and have taken a variety of approaches.  Shoud this method go some where more general so it can be used in other error checking where we need to allow for floating point tolerance? 
```CS
        /// <summary>
        /// Tests if two values are diffrent beyone a tolerance to allow for floating point errors
        /// Calculates tolerence relative to the size of the values to allow for multiplication of floating point errors.
        /// </summary>
        /// <param name="V1"></param>
        /// <param name="V2"></param>
        /// <returns>true or false</returns>
        private bool AreDifferent(double V1, double V2)
        {
            //Express the difference between the two values as a positive
            double difference = Math.Abs(V1 - V2);
            //Determine the larger of the two values
            double largest = Math.Max(Math.Abs(V1), Math.Abs(V2));
            double floatingPointTolerance = 1e-11;
            //The size of errors associated with floating point variances multiplies so need to increase tolerence relative to the largest value being compared
            double largestValueTolerance = largest * floatingPointTolerance;
            //If largest = zero the tolerance will become zero so need to ensure it does not fall below the floating point tolerence 
            double tolerance = Math.Max(largestValueTolerance, floatingPointTolerance);
            return MathUtilities.IsGreaterThan(difference, 0, tolerance);
        }
```

HamishBrownPFR:
@par456 @ric394 I am not sure why this PR is failing.  Is this something in the depths of the system or a problem that has arisen with my Git identity?

par456:
We aren't sure either, but that's only on the new system, so we can ignore that for this PR.

Looks good, will merge.

PR #10503, opened 2025-09-22
Waterlogging models

### LINKED ISSUE #10502: Waterlogging models
### Describe the new feature

Now that the data has been merged, the waterlogging (oxygen deficient stress) models can be added in. Will need a converter for cultivars.

ilhuber:
Resolves #10502 

This will hopefully show model improvement. If we see differences in non waterlogging simulations we know there is some error. There may be some hidden cultivars in some of the private datasets.


ilhuber:
Looks like a canola change slipped through somehow. I think it might be phase definitions in Leaf senescence rate.

hut104:
The Canola model seems to have the Water logging functions parametersed out (ie to have no effect).  The water logging factors are only turned on for individual cultivars.
I suggest that the factors need to be turned on in the plant base model.

It is not clear as to why the converter is adding defaults into cultivars.  That is the purpose of the base model.

hut104:
Is it possible to explain what the purpose of the converter is?  This seems to be updating parameters in some cultivars.

ilhuber:
The purpose of the converter is to update a cultivar path that had targeted a location in the model tree which was moved deeper due to the waterlogging updates. It "re-targets" these paths. Looking at the diffs, apparently the only model that had default cultivars get re-targeted was Maize. Still, the converter would be necessary to preserve user's custom cultivars. 
 
Not sure where you are seeing the converter altering defaults? Those later commits are not related to the converter. Defaults in the Canola model had been altered by whoever did the work on the Canola replacement model and it was having an effect on the non-waterlogging simulations.

hut104:
Hi,
I see that there are now some missing outputs.  Was this intended?

ilhuber:
Yep! In the PRs to add data I submitted whatever I was given by @sarchontoulis. There were a few variables that I was not meant to submit, so they are being removed in this PR. 

We also added in the dataset mentioned in the initial reference panel meeting (MaizeLizaso), just putting it in this PR as discussed in the developer meeting.

ilhuber:
The soybean simulation with the off ET values is in one of those password protected files. I don't remember it, so to get a better look I might need to wait until @sarchontoulis is back.

ilhuber:
Re-posting the email @sarchontoulis sent wrt the poor Soybean ET stats:

> In the new submission of the code with the WL functions “on”, the stats for ET decreased; everything else was fine. Looking at this, we found that the cause was the Ames soybean simulation. The only dataset that we had ET data. At the time of model development (year 2018), we did not have SWIM, and we used, if you remember, an “ad hoc” representation of the water table (manager script). This is the cause of the issue; nothing to worry. We need to update the old soybean sim with SWIM. We can do that later, together with the phenology updates (spring 2026).    
> 
> Sotirios

ilhuber:
@par456 @ric394 @hut104 

What is the new way of retest-this-please-jenkinsing for errors like the above?

par456:
At the moment, if its one of the github actions. Click on the one that has failed, then in the top right of the screen, click re-run all jobs

image

ilhuber:
Looks like the PO stats system may be having some issues again?

ric394:
@ilhuber you're right. I'll give it a rerun.

par456:

**ATTENTION**

This Pull Request will be discussed at the upcoming reference panel meeting on the 7th of May. Based on the feedback from November and February, additional resources have been created for this work for your review.

The documentation for the waterlogging model has been added to the APSIM documentation and [hosted here](https://docs.apsim.info/model/Waterlogging) for you to view. This documentation will stay here after release, and will also be incorporated into the documentation of the crop models that are using this after it is released.

If you have any questions or requested changes, please respond to this email (or on [github in the pull request](https://github.com/APSIMInitiative/ApsimX/pull/10503)) **prior* to the reference panel meeting so that the conversation can be resolved ahead of a decision to release being made.

KirstenVer:
Thanks for the detailed documentation.
With the changes to maximum potential grain size in response to cumulative excess water stress>5, were these changes just based on (i.e. parameterised to) the Maize Lzaso experiment? If so, and given there were no measurements of grain size or number, how confident are you to have that now "on" for all simulations. I note that it resulted in changes to Greenthorpe 2014 Canola.Grain.Number, a small but consistent reduction of its performance. Is the confidence level suitable for making the change to all canola crops and environments (possibly affecting unsuspecting modellers in what are normally considered dryland environments (Greenthorpe), or is it something that needs to be available (just like the  crop phenology changes are off by default)? What is the status of the leaf senescence effect?

ilhuber:
Thanks for the question, Kirsten! 

Most of the maize parameterization is based around the Harbin, China and Novelty, Missouri (USA) experiments (both can be found in the overall Maize validation dataset). The Lizaso dataset was added per request after the first RP meeting on these models. There was an earlier version of the Maize model that had an effect on grain size, but this was abandoned in favor of just an effect on grains per cob.

With respect to the difference in the base canola validation dataset: Initially, our plan was to only have waterlogging associated parameters toggled on in specific cultivars. After the first RP meeting it was recommended that we should come up with a default calibration for each model we were adding waterlogging effects to. The specific changes in Greenthorpe2014 appear to be due to the AFPS effect on root growth being toggled on by default.

par456:

Based on the feedback from today's meeting, the waterlogging pull request has been approved for release and will now be merged. Please raise any further problems as an issue in the GitHub system or contact the software team.

Notification of release to come within the GUI and on the website.

PR #10515, opened 2025-09-24
Tidy up lucerne release

### LINKED ISSUE #10514: Lucerne is listed in both Prototypes and Validation
### What happened?

We've been having an intermittent problem with Lucerne on the new build system, and I've tracked the problem down to the fact that LucerneValidation.apsimx is listed in both the Prototypes/Lucerne folder and the in the Tests/Validation/Lucerne folder, causing a conflict on the build system end.

Upon having a look at these files, they both seem to contain the same number of experiments and folders, however the Tests/Validation/Lucerne folder has an additional met file "APS6.met" and excel file "ObservedExample.xlsx"

When digging further into this, the both files only read in the "ObsAllData.xlsx" file for PredictedObserved input data, despite the directory containing 14 excel files.

Not sure what's happened with this, but it looks like when it was released it wasn't then removed from Prototypes, and if the data was all merged into one excel file, the previous files were not cleaned up afterwards.

Something weird going on with this validation, looking for someone who might have some ideas as to what happened with this.

Windows

HamishBrownPFR:
resolves #10514 
-Removed lucerne protype folder and contents.
-Remove redundant excel files


par456:
Just double checking we aren't getting rid of anything we might want in the future here.

I noticed in quite a few of the excel files that they include things like Shoot Population, or Shoot Wt, and the values in those columns are not in the ObsAllData.xlsx file. Similar with the Nekia file which seems to have SW values but were never put into the combined file. I haven't looked through all of them, but it's enough just to give me pause.

I also note for future that there is some Soil.Water values that are just on Sheet1 and Sheet2 of ObsAllData.xlsx which I assume aren't being used, might need to capture that for validation at some point in the future. No need to act on that here, just leaving a note in case someone reads this later.

Considering some of that data might be a useful reference in the future, is it worth just archiving them in a folder rather than deleting them outright? We can always get them back from git if deleted, but it would be pretty hard to tell later on that they existed to recover them if we just outright delete them.

HamishBrownPFR:
There were lots of loose naming conventions used in previous files.  I think you will find that colums named things like Shoot Wt are merged into the main workbook under a name that matches the APSIM variable like Lucerne.Aboveground.Wt.  I think SW are all captured in a similar way.  I am also pretty sure Xiumei has these files archived else where but she is not here to check at the moment.  Might be better to put them into a sub folder until we can confirm nothing there is needed

PR #10518, opened 2025-09-24
CLEM ensure milking activity happens during the milking event

### LINKED ISSUE #10517: Milking happens before suckling in CLEM
### What happened?

It appears that the timing of milking is wrong in CLEM as it seems to happen before the milk is suckled during the growth event and activity.

Windows

lie112:
Resolves #10517

Set Milking activity to manual and fire during the CLEMMilking event to ensure timing with suckling is correct

### DIFF
--- a/Models/CLEM/Activities/RuminantActivityMilking.cs
+++ b/Models/CLEM/Activities/RuminantActivityMilking.cs
@@ -70,6 +70,14 @@ namespace Models.CLEM.Activities
             }
         }
 
+        /// <summary>
+        /// Constructor for milking activity
+        /// </summary>
+        public RuminantActivityMilking()
+        {
+            AllocationStyle = ResourceAllocationStyle.Manual;
+        }
+
         /// <summary>An event handler to allow us to initialise ourselves.</summary>
         /// <param name="sender">The sender.</param>
         /// <param name="e">The <see cref="EventArgs"/> instance containing the event data.</param>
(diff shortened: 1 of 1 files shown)


ric394 (approved):

hut104:
I don't think there were any tests added to this PR.

ric394:
@hut104 good call. @lie112 can you provide unit tests for these please?

lie112:
I don't have any unit tests for CLEM at present!

This is an identified weakness and a big task that I need to find some time to do. It should be in the generic approach used by activities for this case. I haven't got my head around unit testing and I would also value some assistance in writing unit tests and this would have been a great advanced session at the workshop. I was hoping AI could just do this for me :) I will start to do this when I develop new code rather than fixing the current code.

I am the sole developer of CLEM and need to be meeting urgent active user bug fixes, developing the new model for release with many users now waiting a year for this, writing all the documentation and online reference manual and user guide...... with very little allocation.

I also want to have some discussions on best practice to implement fluent validation pre-simulation in APSIM (actually across multiple events as sometimes you need some setup performed before you can validate). At present I am using the Microsoft IValidatableObject with property attributes and custom validate methods, but I would love to move the validation code out of the model in a similar way to IStructure. This discussion revolves around what 3rd party packages we are able and comfortable to use in APSIM or whether I stick with the .Net basics that are providing full validation of user parameter entry and simulation tree building in CLEM at present.

I also want to move my descriptive summary code out of the model and provide a framework for generating the plain English description of the simulation setup with any problems identified that CLEM users currently value. This is a bit of spaghetti code that is now hard to follow that I need to refactor.

These last two areas relate to the region tags in clem code. And the fact that we should not use region tags highlights that these sections of code need to be refactored out of the model code and provided using generics.

Issue #10516, opened 2025-09-24
Improve parameterisation of Wheat GxExM test sets


hut104:
### Describe the new feature

The GxExM test set needs some improving of soil properties and general recalibration to better capture inter-site, inter-season and Cv responses.

This will mostly be achieved via bringing in data collected in recent project work from historical records.

PR #10519, opened 2025-09-24
Wheat GxExM

hut104:
working on #10516


hut104:
Note that this fix is related to #9983.  There were Ks=0 in Temora dataset.

The soil water properties did not match for Junee.  These have been recalculated from the data itself.

I've fixed these parameterisation now and checked that it improves the stats.

hut104:
Please merge these changes.
I have not changed the wheat model at all.  I have just placed more appropriate soil water properties within the datasets.  This has improved soil water stats, and many plant stats.
Other stats will be fixed by later model fitting in other pull requests within the queue.

ric394 (approved):

PR #10522, opened 2025-09-25
Use Zone's albedo in AtmosphericPotentialEvaporationRate

### LINKED ISSUE #10254: Potential evaporation is based on Green Cover

BrianCollinss:
Resolves #10254

### DIFF
--- a/Models/MicroClimate/AtmosphericPotentialEvaporation.cs
+++ b/Models/MicroClimate/AtmosphericPotentialEvaporation.cs
@@ -28,24 +28,13 @@ public class AtmosphericPotentialEvaporation : Model, ICalculateEo
     public double Calculate(MicroClimateZone zone)
     {
 
-        double coverGreen = 0;
-        for (int j = 0; j <= zone.Canopies.Count - 1; j++)
-            if (zone.Canopies[j].Canopy != null)
-                coverGreen += (1 - coverGreen) * zone.Canopies[j].Canopy.CoverGreen;
-
         if (zone.SoilWater == null)
             throw new Exception("Cannot calculate atmospheric potential evaporation rate. Missing water balance or surface organic matter models.");
-
-        double residueCover = 0;
-        if (zone.SurfaceOM != null)
-            residueCover = zone.SurfaceOM.Cover;
 
         return AtmosphericPotentialEvaporationRate(weather.Radn,
                                                    weather.MaxT,
                                                    weather.MinT,
-                                                   zone.SoilWater.Salb,
-                                                   residueCover,
-                                                   coverGreen);
+                                                   zone.Albedo);
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


BrianCollinss:
@ric394 will you please help with this issue?

image

par456:
DockerHub (service where we keep our virtual system images) is currently down, so the validation system cannot work currently. It's a pretty major outage, so we are hoping it's back up before long.

par456:
Looking at this PR, I have a few questions. 

1. You are switching from calculating the Albedo in AtmosphericPotentialEvaporationRate (which uses the soil albedo and self-calculated covergreen), to using the Soil&Plant Albedo calculated in MicroClimateZone (which also uses soil albedo and covergreen), is that correct? The original issue was a bit vague on what the problem was, but it seems to me we have two different bits of code that are supposed to calculate the same thing, but are getting slightly different results?

2. I don't think you actually made any functional changes in EvaporationModel.cs, it just looks like it got auto-formatted when you've touched the file. If that's the case, it would be good to remove that file from this PR.

3. Before changing the values in PatchyMcPatchFace, I'd like to figure out why that's changing.

BrianCollinss:
Hey @par456. `MicroClimateZone.Albedo` is calculated from the soil albedo, (green?) canopy albedo, and the light extinction coefficient, accounting for the pattern of light interception across the canopy. In contrast, the original code calculates the albedo of the soil–plant system as the average of soil and canopy albedo, with canopy albedo weighted by `coverGreen`.

The change in the `PatchyMcPatch` is due to small changes in evaporation which affects leaching.

hol353:
@hut104 and @sno036 This PR makes a change in MicroClimate. Instead of using soil albedo from the Soil water model when calculating AtmosphericPotentialEvaporationRate, it uses Albedo as calculated by MicroClimate. As the 'owners' of this model, is this ok?

The PR also does a bit of refactoring which you can ignore.

sno036:
Thanks for tagging me in. I'd like a discussion with @hut104 before commenting as I can see a couple of potential immediate and longer term issues ('potential' which is why I would like the discussion). I am also concerned that there are changes showing when, according to the text (I have not looked at the code yet) the only intentional change was to MicroClimateZone. I will follow up with Neil.

sno036:
Thanks @BrianCollinss for identifying this inconsistency. @hut104 and I discussed this PR. We agree that you have identified an inconsistency. 

One of the principles that we are currently attempting to document is that before there are any changes to a model (particularly a central model like this one) there must be adequate testing data in the test suite so that the impacts of changes can be assessed at an appropriate ‘level’. By appropriate level I mean, using this particular case, on albedo and net radiation as well as crop production. Unfortunately for this PR, there is a lot of crop yield data and very little or no net radiation data in the test suite. Neil and I discussed possible mechanisms for getting the needed data into the test suite but that will take some time. 

Given that, our recommendation is that this PR is closed but the issue remain open. Once we have appropriate testing data in the system then the PR can be reopened.

PR #10529, opened 2025-09-25
Use FileName instead of FullFileName in Weather.cs

### LINKED ISSUE #10417: Adding a weather file to apsim from the same folder does not get relative path
### What happened?

When you have an apsim file saved in a folder within the Apsim repo, then adding a met file to weather from that same folder, instead of getting a relative path, we are getting the full path, which is breaking things later.

Need to investigate why.

Windows

ric394:
resolves #10417

This resolves an issue where apsim files would have the absolute path saved instead of the relative path, which can cause issues when the file is run outside of the environment where it was last saved (such as on Jenkins when model validation is run).

The issue was that FullFileName was the property being used in many places. This property always fetched the absolute path and updated FileName at the same time if it was every changed.

### DIFF
--- a/ApsimNG/Presenters/MetDataPresenter.cs
+++ b/ApsimNG/Presenters/MetDataPresenter.cs
@@ -229,10 +229,10 @@ namespace UserInterface.Presenters
                     try
                     {
                         this.weatherData.ExcelWorkSheetName = sheetName;
-                        string newFileName = PathUtilities.GetAbsolutePath(filename, this.explorerPresenter.ApsimXFile.FileName);
+                        string newFileName = PathUtilities.GetRelativePath(filename, this.explorerPresenter.ApsimXFile.FileName);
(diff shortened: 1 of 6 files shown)


par456:
If FileName should always contain the relative filename (as the code seems to suggest), should the conversion from full filename to relative filename not happen within the FileName setter?

ric394:
Very good point. The model should handle its state and not the presenter. 👍

par456:
And could I get a short unit test that proves when I give FileName an absolute path that it gives back a relative one, (and FullFileName gives back what I gave it, and RelativeFileName works), and the opposite, that giving a relative path works for all three too.

par456:
Need another unit test to make sure that %root% survives this change.

ric394:
@APSIMInitiative/administrators this one is ready for review.

PR #10536, opened 2025-09-26
WinterCereal

hut104:
working on #10543

Remove double accounting for stress on tillering rate


par456:
Does not have any validation yet as is related to work for #10531 

Winter Cereal is currently not used by any released or prototyped model.

PR #10542, opened 2025-09-29
Ensure that the CERESMineralisationTemperatureFactor is being set whe…

### LINKED ISSUE #10513: In CERESMineralisationTemperatureFactor.cs, if soil temperature is less than the base it is not being set (or leaving it the same as yesterday)
### What happened?

An omission in the if statement. About L57

Windows

sno036:
…n soil temperature is below the base temperature

working on #10513 

The routine had an if statement for "tf" calculation for when the soil temperature was above the base temperature but unless it was the first day of the simulation the factor was not being set when at or below the base temperature. This would mean that it retained the value from a previous day.

Resolved by adding an "else" statement setting the factor to zero at or below the base temperature.

Unsure how this will affect the test set as it is likely a long-standing issue. If the previous-day temperature was near the base then it might have only a minor effect, particularly as few of the simulations have a direct test for mineralisation.


sno036:
@par456, @ric394  - there error here is "Error: -29T00:52:50+00:00 [Error] Validation workflow error: One or more errors occurred. (Network is unreachable (digitalag.csiro.au:443))". Is there anything that I should be doing? Thanks!

sno036:
This is failing because “System_14” (see under WaterTableMoistureInSeason) is failing to converge. It is an Iowa simulation with a tile drain at just over 1000 mm deep. That is in a 300 mm thick layer so I thought I would try using LayerStructure to add in a few thinner layers there. When I do that I get “In MapConcentration, the number of values (16) doesn't match the number of thicknesses (18).” (see SoilUtilities.cs line 393) which is sort of bizarre. I cannot debug this so wondering if someone from the team might be able to take it on.

It is in my branch "Mineralisation error".

Note also that there is an error in the weather file with an amplitude of 29 C – probably double what it should be but one thing at a time.

hol353:
The problem is that EXCO and FIP are missing values in NO3, NH4 and Urea

image

sno036:
Thanks Dean - wonder how that managed to run initially? will pick this up again when I can.

sno036:
This fix in the code has mostly resulted in no change or an improvement in the statistics. The exceptions are:
- NZRIT simulations - and I am OK to accept these changes (they are small and the old code was clearly incorrect)
- SW_15cm in one of the Iowa SWIM (Kelly) validation simulations where other outputs were improved. @sarchontoulis you comments please.

Ready for review

hol353:
@sno036 This PR is doing 2 things, fixing a bug and changing the  SWIM.apsimx file to make it more stable. Can you put the change to SWIM.apsimx into a separate PR and we'll merge that. It will be interesting to see if the stats change. Then we can rerun this bug fix one.

sno036:
First stage PR for the SWIM simulation stability only. Note that the TF error has only been commented out so that it can be brought back in more easily, Will raise a further PR once this is merged.

resolves #10664 
working on #10513

sno036:
There were minor changes to that one simulation in the SWIM validation as below and suggest these should be accepted (primarily improvements anyway). Will now do the error fix on the TF.

image

sno036:
This PR is an error fix and, fortunately, it has in large part had positive effects on the test suite. The statistics showing red are from the NZRIT and that is a benchmark simulation and the effects are minor. As the 'owner' of that test I am happy to have these accepted.

resolves #10513

hol353:
> Note that the TF error has only been commented out 

Hmm. Doesn't look commented out to me.

ric394:
@APSIMInitiative/reference-panel ,

This is a simple defect fix to soil nitrogen mineralisation.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10513.

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10542/files).

The changes in stats can be found [here](https://postats2.apsim.info/10542).

(Note: this notification is a demonstration of the process proposed by the software team at the last RP meetings as a way to keep the RP aware of science changes. https://apsimnextgeneration.netlify.app/contribute/

PR #10548, opened 2025-09-29
FAR Harvesting Rules

### LINKED ISSUE #10546: Add harvest scripts into Wheat's FAR datasets to ensure harvest output

The FAR dataset in the wheat model test set

### What needs to be changed?

The harvest rules in these files allow simulated harvest prior to the harvest dates in the Observed data.  As a result, changes in phenological predictions can provide harvest before the observed harvest date.  In these cases, predictions become zero.
There are scripts around that will harvest on the latest of either the predicted date and the observd date.

To stop the stats being adversely affected by zeros coming in for final yield or biomass measures.

hut104:
resolves #10546


par456:
Fixing FAR Harvesting rules so that the simulations don't harvest before final measurements were taken, will add points to the datasets and change stats with the additional data.

Issue #10402, opened 2025-09-01
Developing test sets for pasture and livestock model


JPaul-Vasa:
### What happened?

Developing test sets for pasture and livestock model for AgPasture to validate the model. 

### What Operating System are you on?

Windows

sno036:
@par456  - I see that you merged PR 10562 but I don't think this is quite ready for merging. Here I'm assuming that I am looking at the version you accepted (Git thinks so).

Issues include:
- the file includes memos pasted from AgPasture.apsimx that have no relationship to this simulation
- there is no explanation of the experiments that are being simulated
- the Mutdapilly simulation is really weird - no harvesting at all for 3+ years (possibly of course but unlikely and could be covered in the memo
- sowing of cultivars in Gatton but no apparent attempt to have any different parameters to describe the cultivars
- at Gatton a graph folder labelled "Annual Pasture Yield" that is empty
- same issues with cultivars at Kyabram as Gatton, also a graph of crude protein but no simulation
- Kairi simulations are using ISRIC soils
- There is a second set of folders "MutdapillyNitrogenExperiment" - were they really sowing ryegrass on 1 Jan? Quite unlikely. First the climate is not great for ryegrass (particularly at that time of year) but also not many people working on 1-Jan. Can these be checked?
- same deal for Casino regarding the sowing etc.
- There is a folder called CombinedResults that has one plot with some very low numbers of measured AG biomass but no explanation or simulated comparison.
- NewZealand/Lincoln/Rainshelter has a permutation for a single factor experiment with several options disabled)
- BTW - excellent comparison against measured profile water!
- What is going on in the "AboveGroundWt" plot from this simulation? Most of the measured above-ground wts are <200 (and the units are kg DM /ha for that output) and that is scorched earth. Also are there missing harvests in December?
-  the Annual pasture yield plot is empty
- would be worth checking the soil data for the LandPBase simulation. The source is not given and the surface swcon is quite low given SAT-DUL. My observation here might help explain why the water content in the "ProfileWater"plot black lines are so high.
- The LandPBase  annual yield is also empty

There are very many plots - which is fine - but it is really hard to locate the PO stats plots among all the other stuff. Should they be put in some more obvious folder? 

[...]

PR #10562, opened 2025-10-01
Pasture Datasets

JPaul-Vasa:
Working on #10402 
- removed zero values from data files (these should be missing data, not zero)
- Changed rainout shelter dataset to use AgPasture instead of SPRUM

### DIFF
--- a/Tests/Validation/Ryegrass/Ryegrass.apsimx
+++ b/Tests/Validation/Ryegrass/Ryegrass.apsimx
@@ -332,7 +332,7 @@
                   "Children": [
                     {
                       "$type": "Models.Clock, Models",
-                      "Start": "1992-04-01T00:00:00",
+                      "Start": "1992-03-29T00:00:00",
                       "End": "1995-06-30T00:00:00",
                       "Name": "clock",
                       "ResourceName": null,
@@ -397,6 +397,13 @@
                         {
                           "$type": "Models.Operations, Models",
                           "OperationsList": [
+                            {
+                              "$type": "Models.Operation, Models",
+                              "Enabled": true,
+                              "Date": "1992-03-30",
+                              "Action": "[AgPRyegrass].EndCrop()",
+                              "Line": "1992-03-30 [AgPRyegrass].EndCrop()"
+                            },
                             {
                               "$type": "Models.Operation, Models",
                               "Enabled": true,
@@ -675,7 +682,7 @@
                           "TurnoverDefoliationEffectMin": 0.025,
                           "MinimumGreenLeafProp": 0.8,
                           "MinimumGreenRootProp": 0.5,
-                          "Name": "AGPRyegrass",
+                          "Name": "AgPRyegrass",
                           "ResourceName": "AGPRyegrass",
                           "Children": [],
                           "Enabled": true,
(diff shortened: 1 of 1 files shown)


JPaul-Vasa:
Retest this please Jenkins

JPaul-Vasa:
@ric394 Julian, Can you merge this please?

ric394:
Are you happy with the stats?
image

hut104:
Description of changes here:

- NConc is different because three zeros have been removed from the OBS files.  NOTE: The test system has not highlighted the difference in N.
- Dataset used to have simulations for a second species - these simulations have been removed.
- LAI data has moved from sprum output to AgPasture output - N is reduced from 312 to 126 because one species only.
- Same for cover total - half data removed with second species.
- Same for soil water data - half data removed with second species.

I just have a question about the AboveGround Wt.  Half of 210 is 105.  Add this to 31 and you should get 136, but we have 137 datapoints?????

@JPaul-Vasa  and @ric394 Can we understand this first please?

JPaul-Vasa:
@hut104 Neil, the data for AboveGround.Wt is correct. We have 106 data points for Ryegrass and 104 for Tallfescue. However, sum(Soil.Output.SWmm) does not match. The n is 528 but the datapoints are 534 for ryegrass.  Why is that?

**AboveGround.Wt**
Tall Fescue  104
Rye grass    106

**sum(Soil.Output.SWmm)**
Tall Fescue 534
Ryegrass    534

**LAI** 
Tall Fescue 156
RyeGrass    156

**Total Cover** 
Tall Fescue 156
Ryegrass    156

par456:
This PR both changes the model and changes the test data. Can I get the model change split into it's own PR as per our review guidelines. It'll make it clearer what changes the data caused compared to the model change.

par456:

Adding more data to the Pasture dataset.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10402

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10562/files).

The changes in stats can be found [here](https://postats2.apsim.info/10562).

sno036:
Thanks for the opportunity to review this Jessica! And great to see more testing on pasture species as well.

I have some comments - not very organised sorry!

Location of the file. Would it not be better to add it under Test/Validation/AgPasture as it is using that model? The simulation file includes some of the documentation from AgPasture as well.

In NZ/Lincoln (and in fact for all locations but this is where I looked first) my advice is that it is very useful to include data and simulation of amount of total biomass removed by year. Because it is a preennial looking only at standing biomasses (with lots of unders and overs) the overall impression of the utility of the model can be hard to see and  you may well have a better simulation of the data than you might think.

It is great to see some comparison against measured soil water!

There is a stray folder called "CombinedResults" with a graph with little information. Was it WiP that should be deleted? Or is something missing?

There are also many plots that are empty under the Australian folder that ought to be fixed or deleted.

Back to NZ/Lincoln/RainShelter:
- In the summary file "2006-03-15 paddock.AGPRyegrass:  Cannot sow the pasture species "AGPRyegrass", as it is already growing" which means that there are some issues with your simulation set up. AgPasture is initiated as an established crop. If we want to start with a bare soil we kill it on day 1 and then sow it whenever. There is a Manager in the toolbox for this. The error message means that the ryegrass is not being sown when you expect.
- we strongly encourage people to use the model "SimpleGrazing" for harvesting AgPasture species. It doesn't matter so much here because you are running single-species simulations but perhaps a good habit to get into.
- there is no SoilArbitrator in the simulation

What is going on at Mutdapilly? You are plotting "AboveGroundHarvestable" (but labelling it "AboveGroundWt" which is not the same thing) which appears to be in gDM/m2 but it maxes out at 700 g/m2 which is 7 t DM /ha. This is massively high! Also it does not appear to be harvested which doesn't make a lot of sense.

I think there should be some explanation as memos in the different simulations about where the data came from, the general experimental set up etc.

JPaul-Vasa:
Thank you, Val, for your detailed feedback, I appreciate it. We will take a look.

JPaul-Vasa:
Please find the following changes to the model:
1. Included annual pasture yield
2. Re-labelled graphs appropriately in Mutdapilly experiment 
3. Added a Endcrop() to kill crop to accommodate sowing in AgPasture for all experiment.
4. A general cleanup to remove stray graphs.

For the other issues it will be addressed in our other PR as this is just simulation data with model changes to come.

PR #10577, opened 2025-10-03
Fix sla units in Gatton wheat observations

### LINKED ISSUE #10576: SLA values for Gatton2014 and 2015 are in the wrong units - Wheat tests
### What happened?

Need to divide SLA numbers in Validatoin/Wheat/GxExM/LAI.xlsx

Windows

HamishBrownPFR:
resolves #10576


par456:
retest this please jenkins

par456:
Just looking at the POStats, I don't think that did what you wanted there. I think the units were correct before.

HamishBrownPFR:
Thats going to require a bit of head scratching.  If you look at the same variable in the wheat tests set, the units are consistent with what I changed them to.  But in the GxExM the predicted value for the same variable seem to be diferent by a factor of a million
image

HamishBrownPFR:
This will explain it.
@hut104, would it cause any problems if we removed the unit conversions and aliases from the report model so these simulations are consistent with the rest of the test set.

image

HamishBrownPFR:
Stats have changed here for the following reasons:
- The units of SLA for some experiments in GxExM tests for wheat have been divided by 1000000 to make them consistent with other observations.  The report variable that these were being compared with had a conversion of 1000000 which has been removed
- AboveGround.StorageWt was present in the same simulations under an incorrect alias (AboveGround.Storage.Wt).  The report variable has been changed and the name in the observed file changed also so the stats for the old variable have disappeared and the new variable have appeared
- Some rogue AboveGround.Nitrogen data in the observed Lincoln2024 file has been deleted which has reduced n and improved stats.  These values were present on dates that no N observations were present so must have resulted from a transcription error.

PR #10582, opened 2025-10-03
Add checks for PSD to Soil.cs

### LINKED ISSUE #10290: Soil particle size distribution total exceeding 100%
### What happened?

Soil particle size distribution can still contain incorrect values (e.g., totals exceeding 100%).

Example: `Kingaroy1819` experiment in Peanut validation.

Image

Windows

BrianCollinss:
Resolves #10290

### DIFF
--- a/Models/Soils/Soil.cs
+++ b/Models/Soils/Soil.cs
@@ -216,6 +216,11 @@ namespace Models.Soils
                     double max_sw = MathUtilities.Round(1.0 - physical.BD[layer] / specific_bd, 3);
                     int layerNumber = layer + 1;
 
+                    if (MathUtilities.GreaterThan(physical.ParticleSizeClay[layer] + physical.ParticleSizeSilt[layer] + physical.ParticleSizeSand[layer] - 100, 0, 0))
+                        message.AppendLine($"Clay, silt and sand percentages in layer {layerNumber} add up to more than 100%");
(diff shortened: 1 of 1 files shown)


BrianCollinss:
I'm not quite familiar with the GUI code. Is there a way to print warnings in the status panel?

zur003:
@BrianCollinss - Yes, via the ShowMessage method in the MainView. There are some instances of this in GraphView.cs (e.g., line 1721). There is also a ShowMessage method in the MainPresenter; this is used in SeriesPresenter.cs. (e.g., line 457).

Since you're asking in the context of checking PSD values, note that the ProfilePresenter has a small area for a message indicating that values in red (which typically are the PSD values) are estimates only. Perhaps you could make use of that region to display a different message if PSD values are actually provided?

BrianCollinss:
Thanks @zur003. Does this mean the checking of PSD needs to be done in ProfilePresenter.cs?

zur003:
> Thanks @zur003. Does this mean the checking of PSD needs to be done in ProfilePresenter.cs?

That depends on your intent, but if you're looking to display a message in the GUI rather than, say, writing a message to the simulation log, I think the ProfilePresenter would be a reasonable place to put it.

hol353:
@ric394 This PR looks OK to me. Can you review/merge?

PR #10588, opened 2025-10-05
Added more checks for NaN in WaterBalance and its submodels

### LINKED ISSUE #10540: newTemps(1) is outside range of -50.0 to 100.0
### What happened?

I still get this warning from time to time, @sno036. Is it worth investigating, or is the model still under development?

Windows

BrianCollinss:
Resolves #10540

Initial soil moisture is validated when entered through the GUI; however, users can still input invalid arrays via manager scripts.

### DIFF
--- a/Models/Soils/WaterModel/SaturatedFlow.cs
+++ b/Models/Soils/WaterModel/SaturatedFlow.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using APSIM.Numerics;
 using APSIM.Shared.Utilities;
 using Models.Core;
@@ -132,6 +133,9 @@ namespace Models.WaterModel
                     // drainage out of this layer goes into next layer down
                     w_in = w_out;
                 }
+
+                if (flux.Any(w => double.IsNaN(w)))
+                    throw new Exception($"SaturatedFlow model produced NaN: {string.Join(", ", flux)}");
 
                 return flux;
             }
--- a/Models/Soils/WaterModel/UnsaturatedFlow.cs
+++ b/Models/Soils/WaterModel/UnsaturatedFlow.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using APSIM.Numerics;
 using APSIM.Shared.Utilities;
 using Models.Core;
@@ -140,6 +141,9 @@ namespace Models.WaterModel
                     // when calculating theta1 and sw1.
                     w_out = flow[i];
                 }
+
+                if (flow.Any(w => double.IsNaN(w)))
+                    throw new Exception($"UnsaturatedFlow model produced NaN: {string.Join(",", flow)}");
 
                 return flow;
             }
--- a/Models/Soils/WaterModel/WaterBalance.cs
+++ b/Models/Soils/WaterModel/WaterBalance.cs
@@ -229,6 +229,9 @@ namespace Models.WaterModel
             get { return waterMM; }
             set
             {
+                if (value.Any(w => double.IsNaN(w)))
+                    throw new Exception($"Input soil moisture array (Water, mm) contains NaN: {string.Join(", ", value)}");
+
                 waterMM = value;
                 if (value == null)
                     waterVolumetric = null;
@@ -244,6 +247,9 @@ namespace Models.WaterModel
             get { return waterVolumetric; }
             set
             {
+                if (value.Any(w => double.IsNaN(w)))
+                    throw new Exception($"Input soil moisture array (SW, mm/mm) contains NaN: {string.Join(", ", value)}");
+
                 waterVolumetric = value;
                 waterMM = MathUtilities.Multiply(value, soilPhysical.Thickness);
             }
@@ -581,6 +587,9 @@ namespace Models.WaterModel
 
             // Update the variable in the water model.
             water.Volumetric = waterVolumetric;
+
+            if (waterVolumetric.Any(w => double.IsNaN(w)))
+                throw new Exception($"WaterBalance model produced NaN: {string.Join(", ", waterVolumetric)}");
         }
 
         /// <summary>


ric394:
@hol353 is this likely to cause a performance issue?

BrianCollinss:
If the change is approved, I'll add two more checks to the IF statement to check (0,1) range for SW.

hol353:
> @hol353 is this likely to cause a performance issue?

Yes possibly. We found recently that when a bunch of NaN checks were put into the pasture model it had a significant impact on run time speed. We could check lots of variables in APSIM but APSIM run time would tank. Not sure what to do. I wish .NET had a way of throwing an exception when they happen in say 'Debug' mode.

BrianCollinss:
What alternative is there? Will be happy to make the necessary changes.

par456:
@BrianCollinss Do you have an example of a manager script where people are adjusting values like this?

If we do an initial check, and we want to prevent accidental fiddling, is this instead a problem of the properties being too open and allowing people to modify something they really shouldn't be allowed to (aka a public get, private set). I haven't had time to look into these properties, so that choice would depend on how intertwined with other models they are.

Not sure of the direction needed here, but its really showing that we need more performance tests in our testing system to help track if adding something like this has an impact or not.

BrianCollinss:
I'll share it with you soon @par456. Soil moisture is one of those properties that need updating in some experiments, and at least once in most experiments.
In my opinion, this sort of checks are needed when there's a chance of bad inputs reaching deep in the model without leaving enough trace behind. Moreover, there are instances where we help these bad inputs "hide" (e.g. #10589). So, I think it's best to have the checks for a small penalty in the performance.

BrianCollinss:
retest this please jenkins

BrianCollinss:
Can one of admins review this error? The new check I added caught incorrect SW values in Chickpea as high as 242 mm/mm!

```
System.Exception
  HResult=0x80131500
  Message=Error: Input soil moisture array (SW, mm/mm) contains NaN or out-of-range values [0–1]: 0.219, 0.197, 0.261, 0.285, 0.249, 0.209, 0.212, 0.238, 0.271, 242, 242
  Source=Models
  StackTrace:
   at Models.WaterModel.WaterBalance.set_SW(Double[] value) in Models\Soils\WaterModel\WaterBalance.cs:line 251
   at System.Reflection.MethodBaseInvoker.InvokePropertySetter(Object obj, BindingFlags invokeAttr, Binder binder, Object parameter, CultureInfo culture)
```

hol353:
@BrianCollinss Good on you for looking for NaN's. I suspect quite a few of our models produce them under certain circumstances. NaN checking is a rather blunt way for finding problems. In this case, adding checks into the soil water model has found a few in the chickpea simulations. The actual problem probably won't be in soil water though. I would rather find the root cause of the problem, probably somewhere in the chickpea model, although it could be anywhere. 

I have noticed when profiling APSIM runtime speed, anything that is done on a daily timestep that involves layers or cohorts of leaves etc willl show up as being slow. I don't really want to pay the penalty if we don't have to.

I recently added some NaN checking into the pasture model because I noticed some strange outputs, but instead of committing the NaN checks I used them to track down the real problem, a divide by zero (which incidently took a LONG time to find). We don't have the time to chase NaNs down even though it is important. Do you have the time to look into it?

hol353:
One other approach we could use is to only have the NaN checks in debug builds (that our build system doesn't use). I suspect that might be possible but haven't looked into it.

ric394:
Using something like pre-processor directives might do the trick: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

You can do things like:

```c#
#if DEBUG
    Console.WriteLine("Debug version");
#endif

```

hol353:
> #if DEBUG
    Console.WriteLine("Debug version");
#endif

Hmm that'll get ugly real quick.

ric394:
> > #if DEBUG
> > Console.WriteLine("Debug version");
> > #endif
> 
> Hmm that'll get ugly real quick.

I agree but it is one way to do it. You could abstract this ugliness into classes specifically designed for doing the NaN checking 🤔

BrianCollinss:
@hol353 @ric394  The issue was in some user inputs (i.e., 242 mm/mm instead of 0.242 mm/mm). I also removed a 'reset' manager script from `GT_HatTrick_WU_2019` experiment since all the resetting of soil moisture is done via Operations.

hol353:
Let's try using Debug.Assert() for these checks - https://blog.devgenius.io/runtime-assertions-in-net-af240a4ea40f

No runtime penalty when in release mode (i.e. APSIM releases and the build system). They would trigger though when running in a debugger which might be useful.

BrianCollinss:
@hol353 @ric394 `Debug.Assert` works fine when I build APSIM in VS in the Debug mode. But, if I run an instance of APSIM that has been compiled in the Debug mode, it freezes for a few seconds and then crashes. 

image

hol353:
Hmmm you are correct. We'll have to write our own Assert function. It would be nice if we could turn the asserts on and off from the GUI / CLI i.e. defaults to off (quicker runtime) and user switches it on when their simulation produces odd outputs.

BrianCollinss:
What can I do now?

hol353:
Probably not a lot can be done on this. I think we create a GitHub Issue about creating our own Assert function. Once that is done, then we can proceed with adding NaN and other checks into the models.

hol353:
Hi @BrianCollinss

We've done some research and had a chat about this. We don't want to pay the penalty for a slower run-time, even though checking for NaNs would bring some advantages. As I said above, putting in lots of checks to look for NaNs and out of bound numbers can significantly slow down performance, particularly when you're checked layered variables like soil water, no3 etc. We looked into using Debug.Assert and we can get it to work where you only pay the performance penalty when compiled in DEBUG mode. Our build system uses RELEASE mode though to get the shortest build times possible. 

The other problem is that many users build/adapt their plant models while running in VSCode or VS, attached to the debugger so they can debug things when they run into problems. They know this means slower run-times and are ok with that but they won't be ok with that if it takes 5 times longer to run an already slow test set (particular for wheat that has 1000s of simulations).

Thanks for raising the PR - it made us think which is a good thing. For now I'll close this PR. If we, or someone else, can think of a solution we'll look into it again.

PR #10599, opened 2025-10-07
10417 Weather Path Fixes

### LINKED ISSUE #10417: Adding a weather file to apsim from the same folder does not get relative path

par456:
Resolves #10417

@hut104 ran into the GUI changing his weather paths again today, so I've gone in and spent a few hours trying to resolve this problem.

I've rewritten the Relative path function in PathUtilities so that it goes through each case correctly. It will take the given string and see if it can be a relative path, a %root% path or an absolute path, with preference in that order. If a relative path is provided that goes upwards (aka ../../file.met) it will be converted to an absolute path, unless that directory is within the apsim folder, in which case it becomes a %root% path instead.

I've then taken one of the unit tests from the last attempt at fixing this, and rewrote it with as many different use cases that I could think of. All the rules above are tested and in as many combinations as possible.

The filename Get and Set in weather have been rewritten, so any fixes are now only applied in the getter with a default setter. That will hopefully simplify this a bit as well. I also noted that in the OnSimulationCommencing that the filenames were converted to be relative, so that has also been removed as it should not be needed if that's what filename is supposed to provide.

In the GUI side, I've removed places where it was doing conversions between full and relative, and added a new comparision function into PathUtilities to help with comparing paths in the future.

Lastly I changed the ConvertSlashes function in PathUtilities to always convert to linux style / instead of using \\ when on windows. Windows will work fine with the linux style, and this will prevent all the filepaths changing if someone opens an apsimx file on linux or mac in the future. It also simplifies things when working with the paths if you always know which slash style you are dealing with.

### DIFF
--- a/APSIM.Shared/Utilities/PathUtilities.cs
+++ b/APSIM.Shared/Utilities/PathUtilities.cs
@@ -102,10 +102,7 @@
             if (relativeDirectory != null && !Path.IsPathRooted(path))
                     path = Path.Combine(relativeDirectory, path);
 
-            // Convert slashes.
-            path = ConvertSlashes(path);
-
-            return Path.GetFullPath(path);
+            return ConvertSlashes(Path.GetFullPath(path));
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


APSIM.Core/Nodes/Node.cs
```diff
@@ -443,6 +443,8 @@ internal static Node ConstructNodeTree(INodeModel model, Action<Exception> error
         head.locator = new();
         head.Compiler = compiler;
         head.FileName = fileName;
+        if (head.FileName == null)
+            head.FileName = Path.GetTempFileName().Replace(".tmp", ".apsimx");
```
hol353, line 447: Under what situation is head.FileName == null? And does it matter? It obviously doesn't on the tip revision.

ric394 (approved):

hol353:
I've pulled the branch and the code works well. I couldn't find any issues with paths - good work.

In weather.cs, I wonder whether we can have a simple  FileName property

```c#
public string FileName { get; set; }
```
rather than a complex getter (I like simple properties). FileName would store whatever it is given. When code elsewhere in APSIM wants a sanitised path, they call PathUtilities.GetAbsolutePath(weather.FileName) to get a fully qualified path. This will greatly simplify weather, ExcelInput, Input etc but it will mean you need to change code that uses the FileName property expecting a full file name to be returned.

The other weird thing about weather is that it has a field called 'fileName', a Property called 'FileName', a method called 'FullFileName' and another property called 'RelativeFileName'. That's overkill. The other models that have filenames probably have something similiar. Weather also has a property called ConstantsFile which has similar code.

I think this PR should also clean up the other models that have a FileName property.

APSIM.Core/Nodes/Node.cs
par456: It came up in several unit tests where the model was never 'saved' but was either run, or something done with the weather, and the base file didn't have a filepath for weather to compare it's filepath against.

When it went to use the weather model, it would try and get the filepath stored, but the new Getter would throw because it had nothing to compare against. I was unsure how to proceed and did that change, but the other option is that the weather get doesn't throw if it can't get the path to compare against and just returns whatever was stored.

Both have their issues.

par456:
> When code elsewhere in APSIM wants a sanitised path, they call PathUtilities.GetAbsolutePath(weather.FileName) to get a fully qualified path. This will greatly simplify weather, ExcelInput, Input etc but it will mean you need to change code that uses the FileName property expecting a full file name to be returned.

This is very much how Weather originally worked, but then there was no consistency from the GUI and other overrides or commands about what was getting stored in the filepath and depending what last touched it lead to the mishmash of relative, %root% and absolute paths in files which kept changing. I don't really want to go back to that because that's how looking at the weather in the GUI would change the file started in the first place.

Also if someone enters an absolute path into the file, I would prefer not to be storing their full system path to the file as text in our repo if we can avoid it.

What we really need is a FilePath object that is stored here instead that handles all this complexity, but that's a bit beyond the scope of this PR.

> The other weird thing about weather is that it has a field called 'fileName', a Property called 'FileName', a method called 'FullFileName' and another property called 'RelativeFileName'. That's overkill. The other models that have filenames probably have something similiar. Weather also has a property called ConstantsFile which has similar code.

Yep, it's a mess of patches that have been applied over the years. Previously these were all stored separately (and could get out of sync). So in the last fix we at least made it so that filename was the one place a value was stored, and the other properties just referenced that and did whatever conversions they expected. Only reason I avoided removing them entirely is in case of scripts/factors/whatever that might have still been referencing them.

Definitely could deal with them here in this pull request.

hol353:
@ric394 This one is ready for final review and mergining if ok.

Issue #10459, opened 2025-09-14
Command file must be able to handle property = ""


BrianCollinss:
### What happened?

For example, when someone needs to set a cultivar definition equal to "" (to use default values of model parameters):

```
[Cultivars].Australia.Janz.Command = ""
```

### What Operating System are you on?

Windows

hut104:
Are you wanting to remove the cultivar definition or simply remove the commands within it so that it inherits all parameters from the parent plant?  This is unclear in your issue description.

Note that to delete the cultivar, there is a delete command usable within the command file.

"Delete [Cultivars].Australia.Janz"

BrianCollinss:
@hut104 The goal is to be able to set any property "", including a cultivar `Command`. Currently, it is not doable. As for the cultivar `Command`, my aim is to set if "", i.e., to make cultivar adopt the default parameter values.

PR #10610, opened 2025-10-07
Fix bug in locator to allow setting a variable value that is null.

### LINKED ISSUE #10609: Cannot set the value of a variable on --apply line if that variable is null.
### What happened?

While following the --apply tutorial here: https://apsimnextgeneration.netlify.app/usage/commandline/command-line-how-to/ I got an exception "Cannot set value of [Weather].FileName". When the set command is run the existing value of [Weather].FileName is null and the locator then doesn't allow the set.

Windows

hol353:
Resolves #10609


BrianCollinss:
Is #10459 related?

hol353:
I don't think so.

Issue #10643, opened 2025-10-16
Reset maximum rooting depth in AgPasture


JPaul-Vasa:
### What happened?

The current maximum rooting depth in AgPasture for Ryegrass is capped to 1m however based on the data in the dataset it is observed that roots can reach to a depth of 2 m therefore the value should be reset to 2m to allow the model to predict accurately.

### What Operating System are you on?

Windows

lie112:
I would suggest that the 1m rooting depth is probably adequate. While the roots may have been reported down to 2m, it may not be realistic to provide the grass the full 2m profile of water. The root distribution would have to reflect this that the majority are in the upper 0-30cm where they receive most of the water for growth. Even trees have the vast majority of roots within 1m of the surface and while recognised that deep roots exist, I believe these play a more important role in survival when the top layers are dry than defining the water budget of the tree for transpiration. I would be interested to hear how trees access water in APSIM NG.

sno036:
While Adam is correct about the vast majority of the functional roots being much shallower (I'd say mostly 0.3 m), Jessica's observation is consistent with my knowledge and the deeper water can be important for survival so I'm supporting of this change.

Adam - all crops in APSIM access water/nutrients available in the soil in whichever layers they have roots.

hut104:
Thanks for the thoughts.  Our suggestion here is based on water extraction data from a rainout shelter in Lincoln.  In APSIM, the maximum rooting depth is a plant parameter just to provide a sensible upper limit if soil profile specifications are very deep (e.g. places where we have 3m deep profiles because we have lucerne in the mix).  In general situations, it is the XF parameters within the plant models that give a specific limit to rooting depth and the rate at which this is reached.

We'll increase the maximum rooting depth parameter and then see how this impacts the tests across the entire APSIM test set (not just the data we are currently calibrating) and decide if we need to revisit XF parameters within individual tests.  Interested parties will be able to check the impact of these changes before we merge the changes in.

sno036:
TBH - I'd be surprised if it had any effect because the rooting depth is specified in each simulation. Unless the value in the UI is being silently overwritten from wherever you are proposing to make this change (it is not obvious to me from the text above) I guess.

PR #10652, opened 2025-10-20
Updated MaximumPotentialRootingDepth to 2m for Ryegrass 

JPaul-Vasa:
Working on #10643

Updated `MaximumPotentialRootingDepth` from 1m to 2m in AGPRyegrass.json for Ryegrass in AgPasture


sno036:
This is interesting! I did not expect that it would make much difference as I expected that the value inserted in the UI would override any species-specific parameters. I'm not very keen on this being merged as is without a discussion so perhaps a short meeting to discuss? @rcichota may be interested as well.

sno036:
HI @JPaul-Vasa, @hut104 - so I learned something today! Just had a conversation with @rcichota and somewhere along the way, probably when the species were separated in the move from Classic to X, there were changes in the model so that there could be root growth from the initial values in the UI. So my understanding did not keep up with the model!

I've looked at the PO stats and am happy to accept the new values. OK to merge from my perspective.

hol353:
Hi @JPaul-Vasa, Neil and I have spoken to Val and we've decided we want to undo Neil's changes to your PR i.e. the 2 .apsimx files. Can you remove them? The easiest way is to get the files from another branch and copy them into this branch and then commit/push. If done correctly, they should disappear from the "Files changed" tab on this PR.

ric394:
@APSIMInitiative/reference-panel ,

This is a simple change to increase the maximum rooting depth to 2m for AgPasture Ryegrass. This is necessary because we have data showing roots below 1m. 

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10643.

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10652/files).

The changes in stats can be found [here](https://postats2.apsim.info/10652).

(Note: this notification is a demonstration of the process proposed by the software team at the last RP meetings as a way to keep the RP aware of science changes. https://apsimnextgeneration.netlify.app/contribute/

jbrider:
I clicked on the stats and got an error: Cannot find pull request #10643 in stats database

Given that @hut104 established how few of the reference panel actually see these notifications, should this really be the only method we use until we are sure the reference panel is seeing them?

ric394:
> I clicked on the stats and got an error: Cannot find pull request #10643 in stats database
> 
> Given that @hut104 established how few of the reference panel actually see these notifications, should this really be the only method we use until we are sure the reference panel is seeing them?

I've now fixed the link, it was just a small typo.

Fair point @jbrider, we may need an alternative until we get RP members correctly configured for notifications/emails.

PR #10635, opened 2025-10-14
Move more phenology only tests out of wheat.apsimx into phenology.apsimx

HamishBrownPFR:
resolves #10634

PR #10679, opened 2025-10-27
Change wheat tests to use ObservedInput file reader

### LINKED ISSUE #10678: Use ObservedInput file reader for Wheat tests sets
### Describe the new feature

Additional value can be gained from the test set using this tool that derives additional variables where the observed data exists to do so.  For example if AboveGround.N and AboveGround.Wt are present on the same date, ObservedInput will derive a value for AboveGround.Nconc and put it in the observations

HamishBrownPFR:
resolves #10678 

@par456 I think that the ObservedInput reader is good to go because it is already being used in some of the wheat test sets.  Please comment if this is not the case

Dependency #10635 must be merged before this PR.  This PR is branched off the branch that #10635 is on.  Sorry not good practice but need to keep things moving.

### DIFF
--- a/.github/workflows/create-releases.yml
+++ b/.github/workflows/create-releases.yml
@@ -14,11 +14,13 @@ permissions:
 #       - master
 #     types:
 #       - closed
+
 on:
-  workflow_run:
-    workflows: ["Build and test", "Model validations"] # Name of the workflow that needs to be successful
-    types:
-      - completed # Trigger when the referenced workflow is completed
+   workflow_dispatch:
+#  workflow_run:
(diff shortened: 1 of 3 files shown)


par456:
Might just hold off adding it in just yet, I want to get the updated version that is ready for release before we put it into too many more files.

Before the workshop I was writing unit tests for it, plan to continue that this week.

HamishBrownPFR:
Thanks @par456.  Sorry, I was getting eager to wring some more numbers out of the observed set for some of the analyses I doing.  Something unexpected has happed with the stats on this PR and I can't work it out.

HamishBrownPFR:
Will close this and leave it to @par456

PR #10683, opened 2025-10-29
Fixed AgPasture to stop NaN when DUL=LL

### LINKED ISSUE #10682: AgPasture produces NaNs for NO3/NH4 update with DUL=LL
### What happened?

When the bucket is closed by setting DUL=LL, N uptake becomes a NaN which in turn leads to the exception:
```
AGPRyegrass - Growth and tissue turnover resulted in loss of mass balance
```

Windows

hol353:
Resolves #10682


hol353:
This PR fixes the exception in PR #10652

Issue #10680, opened 2025-10-28
Further GitHub Validation Improvements


ric394:
### Where is the code located?

.github and related folders and files

### What needs to be changed and why?

- [x] To further improve the speed Eucalytpus needs to be split
- [x] The Eucalyptus validation file needs to be added to the excluded list (this will then only run the Eucalyptus split files)
- [x] Add a split file for Eucalytpus
- [x] Put all the split yml steps commands into one bash script to help with future maintenance

PR #10684, opened 2025-10-29
10680EucalyptusAndWheatPhenologyWorkflowOptimizations

ric394:
working on #10680

When #10635 was merged, the splitter given to the Wheat phenology was not set up correctly, meaning that the stats were being counted twice, once under Phenology, and again under wheat, this PR is to fix that mistake and remove them from Wheat completely.

Therefore, Wheat stats should lose a bunch of points, Phenology should disappear and a new WheatPhenology heading should appear in the stats list, with an equal number of points shifted around.

Originally this branch started as just speeding up the longer runs, and therefore Eucalyptus splitting will come in a future PR once this problem has been resolved.


BrianCollinss:
I hope someone can check these rogue points.

image

image

par456:
@BrianCollinss 

Yep, there is an existing Wheat project that has been tracking down those outliers and working out what has caused them, it'll continue to do that over time. Hamish has done a lot of work on that over the last year.

This PR is mainly just to correct a mistake from the other day.

par456:

PR #10635 incorrectly caused some of the stats to be counted twice, which was missed in the review step due to other changes in that pull request. This PR will resolve that mistake and correctly remove the stats from wheat as originally intended. 

The issue has been documented in this PR: https://github.com/APSIMInitiative/ApsimX/pull/10684

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10684/files).

The changes in stats can be found [here](https://postats2.apsim.info/10684).

Issue #10564, opened 2025-10-01
Refactor replacement system and add ability for commands (e.g. --apply cli) to be added to GUI.


hol353:
### Describe the new feature

# Current command language.

```
add [Zone] Report
add [Zone] Report MyReport                     <- add a Report node to Zone and rename it to MyReport
add [Zone] anotherfile.apsimx;[Report]

delete [Zone].Report
duplicate [Simulation] SimulationCopy

[Simulation].Name=NewName

run
save modifiedSim.apsimx
```

# Objectives

* Provide more flexibility in replacements.
* Currently replacements are done under Replacements node in treeview, loading models from resources when .apsimx is loaded, Experiments, Cultivars, ModelOverrides, Morris, Sobol. Each have their own implementation which creates a maintenance burden on the software team.
* Have one way of doing replacements in APSIM. Clean up the code and make as much as possible private in APSIM.Core
* Update the command language to something more intuitive and easy to parse. Something like this:

```
add Report to [Zone]
add Report to [Zone] name MyReport
add Report to all [Zone]
add [Report] from anotherfile.apsimx to [Zone]
delete [Zone].Report
duplicate [Simulation] rename SimulationCopy
[Simulation].Name=NewName
run APSIM
save modifiedSim.apsimx
```
Would need a way to convert the old language to the new language.

# Usecase: Add a new empty model to all zones
```
add new Report to all [Zone]
```

# Usecase: Add a Report model to all zones  (Report would be a child of the command node in the GUI)
```
add child Report to all [Zone]
```

# Usecase: Add a Report model to all zones where Report is from another .apsimx file.
```
add Report from anotherfile.apsimx to all [Zone]
```

# Usecase: Replace all wheat models in all simulations with 'NewWheat' (NewWheat would be a child of the command node)
```
replace all [Wheat] with child NewWheat
```

# Usecase: Replace all wheat models in all simulations with 'NewWheat' loaded from another .apsimx file
```
replace all [Wheat] with NewWheat from anotherfile.apsimx
```

# Usecase: Replace all wheat models in BaseSimulation1 and BaseSimulation2 with child model 'NewWheat'
```
use BaseSimulation1, BaseSimulation2
replace all [Wheat] with child NewWheat
```

# Usecase: Define a sowing by nrate factorial (based on BaseSimulation)
```
use [BaseSimulation]
define factor sow
    [Sowing].Date = 2020-06-01, 2020-07-01, 2020-08-01
define factor nrate
[...]

PR #10713, opened 2025-11-04
Change Cultivar, Experiments, ModelOverrides to use APSIM.Core replacements

hol353:
Working on #10564

### DIFF
--- a/Tests/UnitTests/CommandLineArgsTests.cs
+++ b/Tests/UnitTests/CommandLineArgsTests.cs
@@ -1,7 +1,6 @@
 ﻿using APSIM.Core;
 using APSIM.Shared.Documentation.Extensions;
 using APSIM.Shared.Utilities;
-using DocumentFormat.OpenXml.Office2010.CustomUI;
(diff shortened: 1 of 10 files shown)


Tests/UnitTests/CommandLineArgsTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using APSIM.Core;
 using APSIM.Shared.Documentation.Extensions;
 using APSIM.Shared.Utilities;
+using DocumentFormat.OpenXml.Office2010.CustomUI;
```
ric394: unused using statement

par456:
I was worried copilot would go a bit ballistic with this, but anyway.

We'll see if we can merge this in on Monday morning. It does worry me doing a PR like this, this close to MODSIM, so let's get it in as soon as we can in case an issue comes up.

hol353:
> I was worried copilot would go a bit ballistic with this, but anyway.

I only write quality code though so I'm not surprised.

PR #10715, opened 2025-11-04
Set phenology in key biomass testing experiments for wheat

### LINKED ISSUE #10714: Set phenology in some wheat validation sets to allow more precise analysis of biomass data
### Describe the new feature

This issue relates to 4 experiments in the wheat validation set:
- Lincoln2014
- Lincoln2021
- Lincoln2023
- Lincoln2024
All of these experiments have detailed canopy development, biomass accumulation and biomass partitioning data that is valuable for testing of these aspects of the model.  However, all of these experiments have issues with predicting phenology which confounds the analysis of biomass predictions.  For Lincoln2014 the cultivar 'Wakanui' was used which is well fitted but behaved differently under the rainshelter so phenology is not predicted well for this experiment.  for the others, Cultivars are used that do not have enough development data over different sowing dates to fit their phenology.  In all cases observations of phenological development were taken so the completion dates for each stage can be set to the observed to remove confounding from phenology predictions and allow full value to be extracted from the canopy and biomass data.

HamishBrownPFR:
resolves #10714 

I have used the SetModelParamsBySimulation class in each simulation to set the DateToProgress property for Emergence, TerminalSpikelet, Flagleaf, Heading and Anthesis phases to be consistent with observed values.  Setting this property forces the phenology to complete on the specified date.
This is justified because there are good observations of phenology but the model is not performing well for reasons specified in #10714.  We can get better value out of the biomass data in these tests by setting the phenology to remove this confounding.
In addition I have added a memo to top of each of these experiments highlighting that the phenology has been set (and why) and also appended '-NotModeled' to the end of Wheat.Phenology.Stage and Wheat.Phenology.Zadoc.Stage in the observed files to ensure these do not show up in the stats.  That would not provide a good measure of the models performance on these parameters as they are being set now.


HamishBrownPFR:
@par456 @hol353 @ric394.  This is failing because none of the PO tables from wheat.apsimx are being generated.  I have made changes that would affect the stats in some of those tables but I can't work out why the stats are missing altogether.  
Any suggestions appreciated.

par456:
I think it needs a fix from #10700, I'll go and get that ready to merge now so that it can fix this problem as well.

HamishBrownPFR:
Thanks Andrew

Hamish Brown
Science Team Leader - Modelling
Cropping Systems & Environment

T: +64 3 325 9394<tel:+6433259394>
M: +64 27 226 1166<tel:+64272261166>
E: ***@***.******@***.***>
W: www.bioeconomyscience.co.nz<https://www.bioeconomyscience.co.nz>
New Zealand Institute for Bioeconomy Science Limited

Postal Address:
Private Bag 4704, Christchurch Mail Centre, Christchurch 8140, New Zealand
Physical Address:
Lincoln Science Centre, 74 Gerald St, Lincoln 7608, New Zealand

par456:
The way that SetModelParamsBySimulation was handling paths was causing an issue, so I've pushed a fix to your branch here that duplicates the work we did for Weather so that paths are storing correctly. This should now run correctly.

par456:
And after I fixed that, it was parsing the dates wrong from your input file because the test system is in US language. Fixed the date reading that SetModelParamsBySimulation does.

par456:
Looks like this branch is working now, stats changes make more sense and no exceptions being thrown.

HamishBrownPFR:
Thanks for getting the working @par456 
I can see one statistic that has changed that should not have.  I will fix that

HamishBrownPFR:
@par456 I think the problem I am seeing in the stats is with the system.  Specifically, the HarvestObsPred and NZObsPred tables both show Wheat.Spike.HeadNumber is missing.  However, when I look at the graph it appears both the new and old values are there and sitting right on top of each other

HamishBrownPFR:
@par456 I have resolved the problem now.  I have accidentally deleted the Stage names that the HarvestObsPred table was matching on.  
However, I got confused as the graph does not seem to be correct.   Wheat.Spike.Head number showed up as missing in the table (which was correct, it was missing).  However, when I went to the graph to see which simulations were involved it showed the new values.  See graph below.  It appears it duplicated the old values on as new values.
image

HamishBrownPFR:
@hut104 @par456 @ric394 @hol353 this is ready for review now.

Changes in stats are as expected:  
- n has increased for spikeWt because some observed data in Lincoln2021 had the incorrect StageName against these data points.  That is correct now.  
- n has decreased for Phenology.Stage and Phenology.Zadoc.Stage because I have removed these observations from the tests as the phenology is being set (these are no longer a fair test of the phenology model).
Other changes are small ups and downs expected because phenology has been changed changed to use observed dates.

par456:
> @par456 I have resolved the problem now. I have accidentally deleted the Stage names that the HarvestObsPred table was matching on. However, I got confused as the graph does not seem to be correct. Wheat.Spike.Head number showed up as missing in the table (which was correct, it was missing). However, when I went to the graph to see which simulations were involved it showed the new values. See graph below. It appears it duplicated the old values on as new values.>

Yeah I saw that on Friday, very strange. I'll log an issue for POStats so we can fix that.

par456:


In order to further work in modelling canopy development, 4 Lincoln experiments need to have the phenology locked to recorded dates as the predicted phenology dates make it hard to look at just the canopy development.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10714

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10715/files).

The changes in stats can be found [here](https://postats2.apsim.info/10715).

We intend to merge this into the release on Wednesday the 12th unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

sno036:
Will it be really clear to someone coming back to this in the future that the phenology has been locked? Or are you planning to undo the lock once the canopy development has been improved?

HamishBrownPFR:
@sno036 its a good question.  I have made it as clear as possible by adding memos at the top of affected simulations explaining that phenology has been set and providing justification.  The parameter setting includes a model specifically for doing the sets in the simulations with a distinct icon so that also makes it clear there is something unconventional in the simulation configurations.  
It is possible the sets could be undone in the future but well intended plans don't always pan out so we need to assume it will be set indefinitely.  I have removed the phenology observations from the statistics so it is not giving misleading tests of the models performance.
I had thought about having the model that does the sets write a message to the summary file as well but that is not really a good way of alerting people about what is happening.
Open to other suggestions

sno036:
Thanks @HamishBrownPFR  - this is fine with me (and you are right - things don't always pan out and memories fade!). Happy to proceed as you already have planned. The memo should help.

PR #10758, opened 2025-11-11
Output a new variable in ZadokPMFWheat to get Zadoks as DAS

### LINKED ISSUE #10757: Export DAS for a specific Zadoks Scale in wheat model
### Describe the new feature

Current `ZadokPMFWheat` model reports `Stage` as the Zadoks Score from 0 to 90. 

It is not easy to get 

* actual date or DAS for a specific Zadoks Score, e.g. Z51. 
* calculate compare RMSE with other key phenology stages (e.g. Heading or Anthesis) when Zadok is directly used in the validation, as Zadoks Score is not a continuous scale without an unit.

Would be useful to directly output DAS for a specific Zadoks Score.

byzheng:
resolves #10757

## Implementation

A new output variable `Z` is defined in the model `ZadokPFMWheat` to output DAS for a specific Zadoks when it is reached at first occurring. For example, Z51 as the first day Zadoks Stage is more than 51. 

Changes as follows.

* `Phenology.cs`: add a new variable DaysAfterSowing to use in the `ZadokPFMWheat`. If it is not suitable, we can modify the link of `plant` in `ZadokPFMWheat` from `IPlant` to `Plant` which also have an variable `DaysAfterSowing`
* `ZadokPFMWheat.cs`: Add new output variable `Z`. See below about how to use.
* `ZadoksStage.apsimx`: A new simulation to test the output variables.

## How to use

In the Report model, `Z(Index)` can be used to report the specific Zadoks Score, e.g. `Z(65)` returns the day after sowing when stage 65 occurred in first day.

### DIFF
--- a/Tests/UnitTests/Plant/ZadokPMFWheatTests.cs
+++ b/Tests/UnitTests/Plant/ZadokPMFWheatTests.cs
@@ -141,7 +141,7 @@ namespace UnitTests.PMF.Phenology.Scales
             sim.Prepare();
             // The exception should be thrown when trying to run the simulation with an invalid stage index
             Assert.Throws<SimulationException>(() => sim.Run(), 
-                "StageDAS(91) should throw an exception as it's out of valid range (1-90)");
+                "StageDAS(0) should throw an exception as it's out of valid range (1-90)");
         }
 
         /// <summary>


hut104:
This is an interesting addition that may make things simpler for some users.

I wonder if the reporting function should be called something other than "Z".

e.g. [Wheat].Phenology.Zadok.DaysAfterSowing[30]

hut104:
@ric394 and @par456 , Where should we put this sort of test?  Could this be a unit test?  Should we also try outputting a value outside of the range of the index (e.g. 100 or 0).  The code has checks for these.  We probably should test that the exceptions also work (ie they throw).

byzheng:
> DaysAfterSowing



> This is an interesting addition that may make things simpler for some users.
> 
> I wonder if the reporting function should be called something other than "Z".
> 
> e.g. [Wheat].Phenology.Zadok.DaysAfterSowing[30]

DaysAfterSowing might be confusing here as it is the unit not the variable name. How about StageDAS? So the full name is [Wheat].Phenology.Zadok.StageDAS(65)

Models/PMF/Phenology/Phenology.cs
```diff
@@ -212,6 +212,11 @@ public IPhase CurrentPhase
         /// <summary>Gets the current zadok stage number. Used in manager scripts.</summary>
         public double Zadok { get { return zadok?.Stage ?? 0; } }
 
+        /// <summary>Days after sowing</summary>
+        [JsonIgnore]
+        public int DaysAfterSowing{ get { return plant.DaysAfterSowing; } }
```
par456: This isn't really needed. ZadokPMFWheat already has a reference to plant, so it can use "plant.DaysAfterSowing" instead of asking phenology for it.

ric394:
> > DaysAfterSowing
> 
> > This is an interesting addition that may make things simpler for some users.
> > I wonder if the reporting function should be called something other than "Z".
> > e.g. [Wheat].Phenology.Zadok.DaysAfterSowing[30]
> 
> DaysAfterSowing might be confusing here as it is the unit not the variable name. How about StageDAS? So the full name is [Wheat].Phenology.Zadok.StageDAS(65)

I agreed with your new suggestion @byzheng, StageDAS is appropriate here. This keeps it consistent with other outputs such as `FloweringDAS`.

Models/PMF/Phenology/Scales/ZadokPMFWheat.cs
```diff
@@ -42,6 +42,8 @@ public class ZadokPMFWheat : Model
         /// </summary>
         public static readonly double[] GROWTH_STAGE_NUMBERS = [5.0, 5.99, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0];
 
+        private double[] zadokDays = new double[100];
```
par456: If Zadok is from 1-99, then you only need an array length of 99.

However, if you wanted to avoid confusion between a 0-based array and a 1-based array, you could keep it as a length 100 array, and then ignore the 0th position value in your code. Then you wouldn't need to do the index-1 all through this code.

Neil also says that the 0 position can be "sowing" or pre germination.

Models/PMF/Phenology/Scales/ZadokPMFWheat.cs
```diff
@@ -77,5 +79,50 @@ public double Stage
...
+        {
+            double currentStage = Stage;
+            // Progressively check from the last recorded stage onward
+            for (int i = lastRecordedStage + 1; i < zadokDays.Length; i++)
+            {
+                if (zadokDays[i - 1] > 0) continue; // Already recorded
+                if ((int)currentStage < i) continue; // Not yet reached
+                if (i > currentStage + 1) break; // No further stages can have been reached
+                zadokDays[i - 1] = Phenology.DaysAfterSowing;
+                lastRecordedStage = i;
```
par456: This is really hard to try and understand with the continues and break. You've defined the 3 situations where lines 99 and 100 will not run, but I can't figure out the case where it does run. We generally don't allow continue and break for this reason, they make the code much harder to follow.

Even if it's not quite as performant, can I get you rewrite this in a simpler way? I don't think the extra complexity here actually saves all that much time, as these are pretty operations.

Tests/Simulation/Zadoks/ZadoksStage.apsimx
par456, line 1: Thanks for the test file!

Models/PMF/Phenology/Phenology.cs
byzheng: ZadokPMFWheat  link to IPlant, instead of Plant. IPlant does not have a variable DaysAfterSowing. Might be simple to link to Plant in ZadokPMFWheat

Models/PMF/Phenology/Scales/ZadokPMFWheat.cs
byzheng: I implement this way to reduce the run time which don't require the loop from 0 to 100 at each day. Try to add some comments in the codes

par456:
Hi Bangyou, just pushed up some adjustments to your PR, thanks for those comments, that helped a lot.

Can you just confirm I didn't accidently make a logical mistake with that, and if not, I'll get this merged in.

byzheng:
Thanks @par456 for your changes. It works as I expected. Should we have any unitest?

byzheng:
@par456 I have an unitest for ZadokPMFWheat. Hope it is no any errors.

Tests/UnitTests/Plant/ZadokPMFWheatTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            Models.Report report = sim.Node.FindChild<Models.Report>(recurse: true);
+            report.VariableNames = [
+                "[Clock].Today",
+                "[Wheat].Phenology.Zadok.StageDAS(0)"
+            ];
+
+            sim.Prepare();
+            // The exception should be thrown when trying to run the simulation with an invalid stage index
+            Assert.Throws<SimulationException>(() => sim.Run(), 
+                "StageDAS(91) should throw an exception as it's out of valid range (1-90)");
```
ric394: Exception message should be `StageDAS(0) should throw an exception...`

byzheng: Thanks for fixing. Forget to update message after pasting.

PR #10762, opened 2025-11-12
Wheat Stem Number Per Plant

### LINKED ISSUE #10761: Wheat StemsPerPlant output is incorrect
### What happened?

The wheat model has an output for the number of stems per plant but it only accounts for tillers.  It ignores the main stem and so is always off by one.  It should be [Structure].BranchNumber + 1.

Windows

hut104:
resolves #10761


hut104:
This simple bug fix improves the stats for the parameter being repaired and changes nothing else.

@HamishBrownPFR are you happy with this change to the currently released model.  I was having issues comparing results between the released wheat model and the model under development.

HamishBrownPFR:
This is a good fix thanks @hut104.  I have checked all the observed data under Leaf.StemsPerPlant and it is tiller number plus one so the output should reflect this

par456:

A bug in the wheat model meant that the main stem was not counted when number of stems was reported (only tillers counted), leading to the value always being one less than it should. This pull request fixes that mistake in the wheat model and improves the predicted observed fit.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10761

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10762/files).

The changes in stats can be found [here](https://postats2.apsim.info/10762).

We intend to merge this into the release on Thursday the 11th unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

PR #10768, opened 2025-11-13
10767 Remove duplicate volumetric array

### LINKED ISSUE #10767: Changing the initial water and resetting does not update water levels
### What happened?

If I use a manager script to increase the amount of initial water at the start of the simulation, it is reset back to the original level on the 2nd day of the simulation. 

I suspect another model is overwriting the change with a copy of the original initial water.

Windows

par456:
Resolves #10767 

Upon investigating, I found that both Water and WaterBalance both keep a copy of Volumetric water to work from, and that WaterBalance will overwrite the values in water each day after it's done its calculations.

Looking at how this works, it's clear that Water should not have a copy of the volumetric data and should instead be just referencing the SW stored in the sibling ISoilWater class (WaterBalance or SWIM), instead of this weird code that just passes the values back and forth to try and keep the two arrays in sync.

While I was in the WaterBalance code, I also went through some of the sub-models (SaturatedFlowModel, UnsaturatedFlowModel  and WaterTable) it uses for its daily calculations and converted them to static methods that are given the soil water values they need, instead of linking directly back to them. This will make unit tests on these calculation functions easier in the future and just generally reduces the amount of cross linking between models.

This should not change stats or any functionality if I've done this correctly, but should now allow me to adjust and reset the initial water in my simulation without needing to reset both Water and WaterBalance to do so. It also generally just cleans up some chunks of duplicated code between the two classes.


par456:
Took a bit of debugging, but I think I've worked out the stats change. ISoilWater has a function called RemoveWater which removes water from the arrays stored in ISoilWater. This is used by a variety of models to extract water through the simulation each day, normally as root organs or plant models doing water uptake.

In the case of WaterBalance however, when this function was used, it did not pass those changes onto the Water model, meaning that until WaterBalance ran the next day, any water extracted by plants was not reflected in the report at the end of the day when reporting using Soil.Water.Volumetric. 

While there aren't too many models that directly asked for the water level from Water, all of those would have had the same problem, getting the value at the end of the water DoSoilWaterMovement stage, not what had changed during later steps in the day.

Models/Soils/Water.cs
```diff
@@ -25,14 +25,23 @@ public class Water : Model, IStructureDependency
         [field: NonSerialized]
         public IStructure Structure { private get; set; }
 
+        /// <summary>Finds the 'Soil' node. Try parent first, then a soil in scope.</summary>
+        private Soil Soil => Node?.FindParent<Soil>() ??
+                            Node?.WalkScoped()
+                                ?.FirstOrDefault(n => n.Model is Soil)
+                                ?.Model as Soil;
```
hol353: A parent model is in scope. So can't you just look in scope. Also, if you use a find method rather than walkScoped the above becomes something like.
```c#
private Soil Soil => Structure.Find<Soil>();
```
In Models, always use an IStructure. Models are guarenteed that it won't ever be null. Never use Node - too low level. UI / testing code will need to use Node but not anything derived from Model.

Tests/Simulation/Volatilisation/Volatilisation.apsimx
```diff
@@ -7257,7 +7257,7 @@
                 "",
                 "        ///<summary>Gets or sets volumetric soil water content (mm/mm)(</summary>",
                 "        //[Description(\"Soil water content (mm/mm), comma separated values\")]",
-                "        public double[] SW { get {return water.Volumetric;} set {water.Volumetric = value;} }",
+                "        public double[] SW { get {return water.InitialValues;} set { throw new NotImplementedException(); } }",
```
hol353: Hmm. Not sure this is right. water.InitialValues isn't the same as water.Volumetric. Is this change necessary? Why?

Models/Soils/WaterModel/WaterBalance.cs
```diff
@@ -557,14 +560,14 @@ private void OnDoSoilWaterMovement(object sender, EventArgs e)
             Water[0] = Water[0] - es;
 
             // Calculate unsaturated flow of water and apply.
-            Flow = unsaturatedFlow.Values;
+            Flow = UnsaturatedFlowModel.CalculateFlow(Thickness, ll15, Water, dul, DiffusConst, DiffusSlope);
```
hol353: UnsaturatedFlowModel.CalculateFlow is a static method call, which is different to line 534 above where saturatedFlow is an instance method call. Looks a bit weird.

hol353 (commented):
The original concept was that the child models of WaterBalance could be swapped in/out with other implementations in the same way we do for PMF. For example, there has been talk about developing a new approach to soil evaporation and just swap it in/out without having to touch WaterBalance. Of course, this future may never happen. I'm not in favour of having complexity in place for a possible future that may never happen.

What I don't like is treating the child models differently e.g. satflow and unsatflow. Part of me says just make every thing static as they are easier to test. I assume the only reason Sat Flow is different is because it has a user settable parameter 'backedUpSurface' (which isn't settable via the GUI, only a manager script). We could just put this into WaterBalance and pass it down as an argument though the static method call. Thoughts?

par456:
> What I don't like is treating the child models differently e.g. satflow and unsatflow. Part of me says just make every thing static as they are easier to test. I assume the only reason Sat Flow is different is because it has a user settable parameter 'backedUpSurface' (which isn't settable via the GUI, only a manager script). We could just put this into WaterBalance and pass it down as an argument though the static method call. Thoughts?

Yep, that's correct, except it's a user readable property that it set during the calculation (public get, private set). I was thinking the same, WaterBalance could just provide that parameter instead to properly static the function.

Models/Soils/Water.cs
par456: Ah, that was existing code that I just moved its location, I'll simplify that down.

Tests/Simulation/Volatilisation/Volatilisation.apsimx
par456: This is within a manager script in a sensibility test, the manager script is implementing a dummy instance of ISoilWater where most of the properties are just "throw not implemented". Since it's up to ISoilWater to store the running Volumetric, this manager script can't just ask Water for it anymore.

I think the 'correct' implementation would be to have it hold a copy of the initialwater values (like how other ISoilWater make a copy when initialised), but then it never actually modifies the SW value, so I ended up shortcutting and just pointing it to initialwater and privating the set command. I can implement this with a copy initialisation if that's better.

hol353: Probably best to get this right. You can't return InitialValues if the caller is expecting SW (assuming that SW is actually being called in that simulation)

par456:
I had a think about how to deal with WaterBalance's sub models, and I've gone with a better solution I think. I've created a new interface IWaterCalculation that just defines the one 'Calculate' function that they all should use, then did a bit of re-jigging so that all the current sub models have that same way for running them and pulling the required values back out. Now it's a lot more clearer that these are all kind of related to each other which I missed previously.

There might be some future work in cleaning up some of the other models, but I'm going to avoid that here since this is already changing stats a little bit. They also need some specific unit tests for the static functions, and the other models need their calculations converted to static functions. This we keep the modularity for PMF, but improve the consistency between these models and bring them closer to our modern formatting standards.

I've also edited the DummyWaterBalance script so that it now initialises its SW correctly in the same way that WaterBalance does.

hol353:
Why are there stats changes?

par456:
I've got two stats here that aren't just minor changes, in Soybean I have Soil.SoilWater.WaterTable that looks completely different and in WEIRDO, MeasuredProfile looks way off.

@hol353 Any ideas why that might be on just those two files?

par456:
> Why are there stats changes?

Because we nearly always report Soil.Water.Volumetric which was only updated in the DoSoilWaterMovement event, but plants would continue to draw water out of WaterBalance throughout the rest of the day that then wouldn't be syncronised with Water until the next day.

I could just patch that bug in another PR without the other changes to show the change, that might help me rule out if any of the refactoring is causing stats changes.

hol353:
Yes, create another PR with the bug fix. I just looked at WaterBalance.RemoveWater which is what the plant models call to do there uptake. Like you say, it was setting WaterBalance.Water but not Water.Volumetric. The two variables would have been out of sync. What a mess. I'm glad you're fixing it.

par456:
This looks good now. There are some very minor stats differences caused by double precision errors, but that's fine.

ric394:

This pull request fixes an issue where two volumetric arrays where being maintained for soil water which caused outputs in some instances to get out of sync.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10767

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10768/changes).

The changes in stats can be found [here](https://postats2.apsim.info/10768).

As this is a bug fix this will be merged immediately. Please comment on this PR if you think this should be reverted.

PR #10778, opened 2025-11-14
Account for N reallocation from leaf in senescence code

### LINKED ISSUE #10777: SimpleLeaf does not account for NReAllocated from senescing leaf when adding live to dead pools.
### What happened?

The code in SimpleLeaf for moving senescence from live to dead pools does not account for the N that was reallocated from the senescing leaf to other organs as part of the N arbitrator's calculations.

Error is in the following code.  This code needs to be moved to the N retrans code (as in GenericOrgan) so that the sums can account for reallocation.

https://github.com/APSIMInitiative/ApsimX/blob/118a3d4bd1e39e810b61511411c548140c389852/Models/PMF/Organs/SimpleLeaf.cs#L981-L988

Windows

hut104:
resolves #10777


hut104:

Issue https://github.com/APSIMInitiative/ApsimX/issues/10777 identified an error in the code of Simple Leaf. The fix to this bug (ie this Pull Requst) improved the stats for almost all models using this code. The exception is the soybean model.

The soybean model needs to be checked for recalibration especially for leaf.live.nconc.  An issue (https://github.com/APSIMInitiative/ApsimX/issues/10864) has been raised to ensure that this is followed up by one of the soybean modellers. 

The improvements to the model should be provided to the users of the other crop models using SimpleLeaf.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10777
The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10778/files).
The changes in stats can be found [here](https://postats2.apsim.info/10778).

We intend to merge this into the release on Tuesday 16th unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

PR #10781, opened 2025-11-14
10767 Update Water Volumetric when removing Water

par456:
Working on #10767

Test to see what the stats changes are.

### DIFF
--- a/Models/Soils/WaterModel/WaterBalance.cs
+++ b/Models/Soils/WaterModel/WaterBalance.cs
@@ -820,6 +820,8 @@ namespace Models.WaterModel
         public void RemoveWater(double[] amountToRemove)
         {
             Water = MathUtilities.Subtract(Water, amountToRemove);
+            // Update the variable in the water model.
+            water.Volumetric = waterVolumetric;
         }
 
         /// <summary>Sets the water table.</summary>


par456:
Comparing this against my other pull request, the stats here make sense. Can put this up for review now. Updating Water from the RemoveWater function in WaterBalance seems to fix this bug without the changes caused by the other refactors. I'll update the Issue with the bug information.

hol353:
Are the stats in this PR the same as the other one, or are there other bugs to fix?

par456:
There are some additional changed stats in the code refactor, most of which seem to be some sort of precision error (can't see any difference on the graphs or in the stats numbers, but flagged as 'changed')

Except for:
- Soybean Soil.SoilWater.WaterTable 
- WEIRDO MeasuredProfile

Which are radically different and doesn't show up in this PR. That seems to be an introduced bug in the code refactor I'll need to fix first.

hol353:
Then I'm happy for this to be merged. @ric394 can you merge this one?

par456:

Water and ISoilWater are both keeping a copy of the current water values stored in the soil, however a function in WaterBalance was not updating the Water model when water was removed from the soil. This caused the models to get out of sync causing a slight error when reporting that in our stats.

This PR fixes the error in the code to bring them into sync again, while another PR [here](https://github.com/APSIMInitiative/ApsimX/pull/10768) will refactor the larger design mistake in having duplicate values.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10767

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10781/files).

The changes in stats can be found [here](https://postats2.apsim.info/10781).

As this is correcting a coding mistake, it will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #10783, opened 2025-11-14
Fix create-release workflow bugs

ric394:
Removed unnecessary condition checks for 'resolves' in release creation steps.

This should fail correctly when there are no resolves comments now.


ric394:
This can be merged without status checks - purely workflow related

PR #10785, opened 2025-11-14
fix units of nconc obs data

### LINKED ISSUE #10784: Canola Test Set N errors
### What happened?

Observed data in the Canola Test set has observed N concentrations expressed as % when the model uses g/g.  As a result, test stats are fairly insensitive to change.  Obs vs Pre graphs are meaningless.

Windows

hut104:
resolves #10784


par456:


A mistake in the units for NConc were found in two of the Canola datasets. Instead of a percentage between 0 and 1, it was between 0 and 100. This has been corrected in this PR.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10784

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10785/files).

The changes in stats can be found [here](https://postats2.apsim.info/10785).

As this is only correcting a data error, it will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

Issue #10799, opened 2025-11-18
APSIM Advanced Training 2025 Changes and Additions


ric394:
### Describe the new feature

This is a placeholder issue to create pull requests against for APSIM Advanced Training Event.

PR #10800, opened 2025-11-18
Add manager script for sowing and harvesting event labels

ric394:
working on #10799

Introduce a manager script that captures sowing and harvesting events, allowing them to be displayed on graphs using the EventNamesOnGraph model. This enhances the visualization of crop events in the management toolbox.


ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
```diff
@@ -996,6 +996,64 @@
...
+            "    [Serializable]",
+            "    public class Script : Model",
+            "    {",
+            "    ",
+            "        [Separator(\"\\nHOW TO USE THIS MANAGEMENT SCRIPT: \\n\\nThis management script can be used to capture sowing and harvesting events.\\n\"+",
+            "        \"These can be displayed on a graph with the help of the EventNamesOnGraph model which can be\" +",
+            "        \"added to the relevant Graph's Series model.\\n\")]",
+            "        ",
+            "        [Description(\"Variable name to use in your Report model - DO NOT MODIFY - LEAVE BLANK\")]",
+            "        public string CropEventName { get; set;}            ",
```
hol353, line 1021: If you don't want the user to modify the parameter then don't make it a parameter i.e. remove the [Description] attribute.

hol353 (commented):
Why is this manager model needed? I thought the EventsOnGraph feature worked without needing a helper manager script? Have you found a simulation where EventsOnGraph is used e.g. in wheat validataion and seen how it works?

ApsimNG/Resources/Toolboxes/ManagementToolbox.apsimx
ric394: I was looking for a way to add a summary or description to explain it's usage and this is the way I was able to do it. But I can tidy it up and remove the separator and description if required.
Note, the separator only works if it has properties with description attributes. This is why I have a user mutable property.

ric394:
> Why is this manager model needed? I thought the EventsOnGraph feature worked without needing a helper manager script? Have you found a simulation where EventsOnGraph is used e.g. in wheat validataion and seen how it works?

If you have say a 100 year simulation you get an illegible graph using the same variable used in the wheat validation graphs. See the screen shot below:

image


Another inconvenient side effect is if you have a very long running simulation the graph can take 5 mins+ just to load 😞. Because this one only shows sowing and harvest, it has much fewer labels and is much easier to use and read.

If you think this is not needed I'm happy to close the pull request.

hol353:
I've only ever seen events on graphs for 1 year simulatons, never any longer e.g. this from barley validation:

image

For this graph, Hamish just output [Barley].Phenology.CurrentStageName and then selected this variable in EventNamesOnGraph:
image

Don't need a manager script - keep it simple.

PR #10805, opened 2025-11-20
Sensitivity documentation for Morris/Sobol and related sims

### LINKED ISSUE #10804: Include installation instructions for Sobol/Morris simulations
### Describe the new feature

Including installation instructions for R in these simulations and also in the tutorial documentation will help users interested in running sensitivity analyses.
This is particularly important for the APSIM Advanced Training Event which has a session related to these simulations.

ric394:
resolves #10804
working on #10799

This adds a document with instructions on what to install when a users wants to run sensitivity analysis simulations.

This also removes documents and a section related to cloud services that are no longer supported.

### DIFF
--- a/Examples/Sensitivity/Morris.apsimx
+++ b/Examples/Sensitivity/Morris.apsimx
@@ -1,17 +1,24 @@
 {
   "$type": "Models.Core.Simulations, Models",
-  "ExplorerWidth": 300,
-  "Version": 100,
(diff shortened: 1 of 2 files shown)


par456:
Can I get you to add that detail to a memo in the tutorials themselves?

ric394:
Can do. Though I thought about this, but you already get warned in the message box to install R if it's missing. So it may be redundant.

PR #10811, opened 2025-11-23
Correct units for soybean data.

### LINKED ISSUE #10788: Shell N Conc values in Soybean tests are incorrect units
### What happened?

Some of the Shell N conc observed data within the soybean test set are in units of % but should be in g/g.
The incorrect data appears to be in the Iowa datasets.

Windows

ilhuber:
Resolves #10788 

The values that had originally been in the table were the pod nconc *
100. Sotirios and I were able to find the original data source and correct the values.


par456:
Is the 10 less Shell.NConc values because when you looked at the original data, it had less points for that variable than the pod Nonc's that were there?

ilhuber:
Yep. These are in season measurements, and I am told it is not possible to separate the pod and shell that early in the plant's lifecycle. These columns are adjacent in the excel table, and it looks like when the data got added the values from the original dataset for pod (in g/100g) were pasted into the shell column. After pod was computed with a formula the data that was pasted was not removed.

We found some additional later data points (the values at harvesting, plus another in season pod observation) that Sotirios wanted added, hence the extra 8 for pod.

PR #10818, opened 2025-11-24
Removed Encrop() in the simulation and made changes to the code 

### LINKED ISSUE #10402: Developing test sets for pasture and livestock model

JPaul-Vasa:
Resolves #10402
Removed Encrop() in the simulation and made changes to the code to reflect fallow before sowing.


par456:
This is looking good and is replicating the results. 

Can I just get you to go back into the Ryegrass file and remove the EndCrop operation lines entirely instead of commenting them out. Looks like we don't need them there anymore and it's better to remove completely rather than just commenting it out.

PR #10820, opened 2025-11-25
Small changes to fix Gtk Warnings

### LINKED ISSUE #10809: Gtk warnings should be dealt with
### What happened?

When ApsimNG is run from a LInux command line (or from an MSYS2 command line on Windows), a number of Gtk Warnings (and even a few Criticals) are reported on the console. Many of them are of the form ` Gtk-WARNING **: 21:34:35.211: GtkPaned does not have a child property called expand`, with similar warnings about the `fill` and `position` properties. These should be fixed, probably by updating various .glade files.

The `CRITICAL` errors don't crash the application, and generally look like ` Gtk-CRITICAL **: 21:34:03.352: gtk_text_buffer_apply_tag: assertion 'GTK_IS_TEXT_TAG (tag)' failed`. My impression (which may be misleading) is that these occur when rapidly moving through Models in the simulation tree, perhaps when closing a view with a GtkTextView which hasn't had time to fully initialise? These errors don't seem to occur consistently.

Linux

zur003:
Resolves #10809

### DIFF
--- a/Examples/SCRUM.apsimx
+++ b/Examples/SCRUM.apsimx
@@ -32,7 +32,7 @@
         {
           "$type": "Models.Climate.Weather, Models",
           "ConstantsFile": null,
-          "FileName": "%root%/Tests/WeatherFiles/lincoln.met",
+          "FileName": "%root%/Examples/WeatherFiles/NZ_Lincoln.met",
           "ExcelWorkSheetName": null,
           "Name": "Weather",
           "ResourceName": null,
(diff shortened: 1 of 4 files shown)


zur003:
Not clear to me why the tests failed, as my changes should only have impacted the GUI, not any models.

par456:
Is this ready to merge @zur003?

zur003:
@par456 - Yes, I think so, though a programmer's work is never done.

PR #10846, opened 2025-12-08
Cotton Model

hut104:
working on #4631


hut104:
@par456 and @ric394 I think this can be merged.  This model is not yet released.  This change is purely to capture changes made in the last few months.  increases in N are due to simulations being added.  decreases are due to experiments being disabled because of significant plant growth regulator use.  We may include these again at a later date.

PR #10847, opened 2025-12-08
Cotton Validation File changes

hut104:
working on #4631

1) Addition of a sensibility test showing detailed analysis of Biomass and N accumulation/recycling within the plant.
2) Removal of a few memos to clean up the documentation.


hut104:
@par456 I think this can be merged.  These are cosmetic changes only. If we merge these now, we can then start changes to the model itself and reviewers can then see just those changes in isolation.

PR #10856, opened 2025-12-11
10855 Return no stress if Live.MetabolicNConc is 0

### LINKED ISSUE #10855: Leaf.Fn reports 0 (maximum stress) when crop has no N demand, while Leaf.Fw does the oppersite.
### What happened?

Doing graphing of N and water stresses with Fn and Fw, I've noticed that while Fw reports a 1 when there is no water demand in the crop (aka, pre/post harvest), Fn reports the opposite with 0 meaning complete stress.

This makes it hard to use reporting functions like:
```
min of [Wheat].Leaf.Fw from [Wheat].Sowing to [Wheat].Harvesting
min of [Wheat].Leaf.Fn from [Wheat].Sowing to [Wheat].Harvesting
```

As Fn will always end up as 0 since it has no demand prior to emergence, while the Fw version does work.

Fn should be fixed so that both Fn and Fw report stress the same way for situations where the crop has no demand.

Windows

par456:
Resolves #10855

Testing to see if this impact any stats


par456:
This seems to work and does not affect the results on Acceptance Tests, but as it's changing the results of the leaf model, I will need someone familiar with it to give the ok before it's merged.

PR #10872, opened 2025-12-16
10871 Add double precision check into PerennialLeaf.CoverDead

### LINKED ISSUE #10871: PerennialLeaf.CoverDead has a double precision error
### What happened?

Found when investigating the stats problems with #10768 

Due to a double precision error in PerennialLeaf.CoverDead, it can return a value slightly above 0 instead of actual 0 when the pasture in TropicalPasture has died. Leading to it continuing to draw water from the soil even when the model suggests it is dead.

Windows

par456:
Resolves #10871 

This should change some stats of crops that use PerennialLeaf, we'll have to check what the damage is from this previously working slightly wrong.

### DIFF
--- a/Models/PMF/Organs/PerennialLeaf.cs
+++ b/Models/PMF/Organs/PerennialLeaf.cs
@@ -303,7 +303,15 @@ namespace Models.PMF.Organs
         public double LAIDead => cohort.LaiDead;
 
         /// <summary>Gets the cover dead.</summary>
-        public double CoverDead { get { return 1.0 - Math.Exp(-ExtinctionCoefficientDead.Value() * LAIDead); } }
+        public double CoverDead { 
+            get {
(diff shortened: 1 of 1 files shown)


par456:
These results are what was expected. Some minor differences in the models using PerennialLeaf, with a pretty big difference in the Dably Tropical Pasture experiment where this bug was occuring.

ric394:

This pull request fixes an issue where TropicalPasture will continue to take up water when PerennialLeaf.CoverDead returns a near zero value, when it shouldn't because it is effectively dead.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10871

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10872/changes).

The changes in stats can be found [here](https://postats2.apsim.info/10872).

As this is a bug fix this will be merged immediately. Please comment on this PR if you think this should be reverted.

PR #10876, opened 2025-12-17
Add default option of the dropbox for crop types in the frostHeatDamageFunctions

### LINKED ISSUE #10875: Default option of the dropbox for crop types in the frostHeatDamageFunctions
### What happened?

Please refer to [comment](https://github.com/APSIMInitiative/ApsimX/pull/9061#issuecomment-2224228948) from @BrianCollinss 

Windows

PengchengHU:
Resolves #10875

### DIFF
--- a/Models/Functions/FrostHeatDamageFunctions.cs
+++ b/Models/Functions/FrostHeatDamageFunctions.cs
@@ -461,14 +461,23 @@ namespace Models.Functions
             }
 
             // Check if the selected crop type matches the plant type in the simulation
-            string actualPlantType = Plant.Name;
+            string actualPlantType = Plant.PlantType;
             string selectedCropType = CropType.ToString();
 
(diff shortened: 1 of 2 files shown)


par456:
Looks like it broke a unit test for FrostHeatDamageFunctions, you'll need to check what happened there and why it tried to set it to Wheat in a Barley simulation.

PengchengHU:
Hi @par456 , the test has passed.

PengchengHU:
Hi @par456 , could you please look at this one? Thanks. The stats make no sense to me, as I just modified the scripts for damage functions and the unit tests, which should not affect cotton and so on.

par456:
Stats should be fixed after this run, no need to worry about that.

I've also amended your check to use PlantType instead of name which should make it more robust (in case someone renames their Wheat/Canola node, it should now still work).

PR #10887, opened 2025-12-19
Add more PO figures as in the PO Stats

PengchengHU:
Working on #10170


par456:
Note for clarification: this just adds more graphs to the apsim file, not more comparisons into the POStats system so doesn't change the stats.

Issue #9996, opened 2025-05-23
Nuget packages could be updated


zur003:
### Where is the code located?

All ApsimX projects

### What needs to be changed?

Many of the Nuget packages used by ApsimX are not current. The NuGet manager for the solution reports 45 updates are available.

### Why does the code need changing?

Some of the currently used packages have been deprecated or have known security vulnerabilities. Keeping this up-to-date is a maintenance issue that need to be done periodically.

PR #10893, opened 2025-12-22
Update Nuget packages

zur003:
Working on #9996

This PR brings nearly all references to NuGet packages up-to-date, which should be helpful in preparing for the transition to .Net 10 (which ought to be done before the end of 2026, when active support for .Net 8 will end). Most changes here are reasonably straightforward, but there are a few points to note:

- Mapsui has been updated to version 5. This required (and enabled) a few minor changes to the mapping code. Specifically: (1) the "plan B" base map to be used if OpenStreetMaps cannot be accessed is now in the form of an embedded GeoJson file rather than an external shapefile; (2) the markers placed on the map are now Mapsui in-built markers rather than being generated from a bitmap embedded resource; (3) I've deleted MapRenderer.cs from APSIM.Interop, as it was identical to the code in APSIM.Shared (which APSIM.Interop references).
- WriteTableToSqlServer now uses Microsoft.Data.SqlClient rather than the now obsolete System.Data.SqlClient. This change may break some usage, as connection security is more stringent. I have been unable to test the changes, as there is no unit test for it, nor any example simulations that make use of it.  Perhaps @hol353 knows someone who may actually be using this and can run some tests.
- I was not able to update GtkSharp and CairoSharp because of some quirky issue with OxyPlot.GtkSharp3. Apparently the latter was built against an earlier version of GtkSharp and can't make use of the new one, though I don't fully understand the issue. I wonder whether we should just make our own fork of OxyPlot.GtkSharp3, which hasn't received any updates since our own @hol430 made some changes more than 4 years ago.

### DIFF
--- a/APSIM.Shared/APSIM.Shared.csproj
+++ b/APSIM.Shared/APSIM.Shared.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>net8.0</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


zur003:
@hol430 - I see you've raised the GtkSharp version problem as an issue on the oxyplot-gtksharp github repository. It will be interesting to see whether you get a response, given that the code base has been almost static for the past four years. I suspect it may be best if we just fork the code for OxyPlot.GtkSharp3 and maintain it ourselves. We shall see.

hol430:
Hey Eric - yeah that repo doesn't see much activity so I'm not sure how much interest there will be in improving the situation. That said, this will be a problem for anyone wanting to use oxyplot with the latest GtkSharp, and the fix is pretty simple and doesn't appear to break anything.

par456:
That might force us to accelerate the timeline of dealing with the graphing code if that's going to be a major hurdle switching to dotnet 10, good catch on that, would have been a pain to learn about that in 6 months time when trying to switch.

hol430:
Is a more recent GtkSharp actually needed for the .net 10 migration? It would be nice to be able to take advantage of the latest upstream bugfixes, but as far as I know (having not actually tested this), the current version of GtkSharp and OxyPlot should work fine under .net 10.

zur003:
I just had a (very) quick attempt at building for .Net 10 and while I don't think there's a problem with GtkSharp, there is a problem with SixLabors.ImageSharp, which has a known security vulnerability. We don't reference that directly, but it's pulled in by PdfSharpCore.  Glancing at the PdfSharpCore repository, a lot of users have raised this as an issue. Apparently they haven't moved to an updated version of ImageSharp because of changes in the ImageSharp licensing, and no-one has yet taken the time to port things over to SkiaSharp.

zur003:
Further update on PdfSharp: it appears to have been forked and moved, so with luck we should just be able to switch to an updated package from a different source. I shall do some testing.

ric394:
While trying to run your branch locally I get these errors.

image

zur003:
@ric394 - The Humanizer error is due to a bug in older versions of NuGet. See [https://github.com/NuGet/NuGet.Client/pull/6124](https://github.com/NuGet/NuGet.Client/pull/6124) for more information. Updating Visual Studio should, I think, fix the problem.

ric394:
We're using VS Code over here. For now I'll revert Humanizer back to version 2.14.1 so it still runs through that IDE. I'll update Humanizer before we move to .NET 10.

Issue #10896, opened 2026-01-02
Bad data in BFG experiment


hut104:
### What happened?

The data in the BFG experiment for the pinus model seems incorrect (or at least different to that in a paper for modelling of the site).
We recommend redigitising the data and adding this into the test set in its place.  This may make the results different, but at least the data should supposedly be more trustworthy.
A later pull request will be required to fit the model to the new data.  
BTW, the current predictions of the model for the site are not too good in that the parameterisation appears to have incorrect soil properties.
So,
1) we merge in the correct OBS data
2) a subsequent PR will merge in better soil properties to match the new soil data.

### What Operating System are you on?

Windows

PR #10897, opened 2026-01-02
Changed soil water data for BFG experiment

hut104:
working on #10896

NOTE: @par456 and @ric394 this PR just updates the data.  A subsequent PR will be required to refit the model to this corrected data.
We don't allow PR's to change both data and model and so we'd probably need to merge this one first.


par456:


The data in the BFG experiment for the pinus model seems incorrect (or at least different to that in a paper for modelling of the site). This PR adjusts the data to reflect the published data with further work needed to revise the model based on this.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10896

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10897/files).

The changes in stats can be found [here](https://postats2.apsim.info/10897).

As this is only correcting observed data and not modifying an existing model, it will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #10899, opened 2026-01-05
Ensure some experiments harvested after specific dates

PengchengHU:
Working on #10170 

It adds the "HarvestAfterDate" to ensure that some experiments are harvested after specific dates, thereby avoiding missing data points in PO Stats after tuning the phenological parameters.


PengchengHU:
Hi @par456 , I am trying to fix the missing maturity dates in [this PO Stats](https://postats2.apsim.info/10302) of #10302  by adding "HarvestAfterDate" in the barley.apsimx file. It introduced slightly worse performance for maturity date estimations in the current PO Stats. Is it possible to merge this PR? Ta.

PR #10901, opened 2026-01-06
Put all unit tests inside the UnitTests namespace and improve Sqlite …

### LINKED ISSUE #10900: Unit tests "temporary" files are not all being cleaned up properly
### What happened?

Running the full set of unit tests on Windows leaves over 600 "temporary" files behind. These are meant to be deleted by the `OneTimeTearDown` logic in `UnitTests.Utilities`, but this is failing for two reasons:

1.  A path for holding temporary files is being set in `UnitTests.Utilities.OneTimeSetup` but this does not happen until code in the `UnitTests` namespace is executed. As a result, all the tests in the `APSIM.Core` and `APSIM.NG` namespaces - which are executed first - are not using the intended temporary folder.
2. Microsoft.Data.Sqlite is notorious for keeping files open even after they are nominally closed. The actual database file closure may not occur until disposal and/or garbage collection has occurred. When this happens, the folder holding the database files cannot be removed in the `OneTimeTearDown` code (which may execute before garbage collection has completed). This is not normally obvious because the resulting exception is not raised until after all the unit tests have been run successfully.

This should be fixed, since it can cause a fairly rapid buildup of unwanted files on whatever system is running the tests.

Windows

zur003:
…cleanup

Resolves #10900

### DIFF
--- a/Models/Graph/SeriesDefinition.cs
+++ b/Models/Graph/SeriesDefinition.cs
@@ -323,12 +323,7 @@ namespace Models
                     if (columnNames.Contains("SimulationID"))
                         filter = AddToFilter(filter, $"SimulationID in ({simulationIdsCSV})");
                 }
-
-                //Filter out any string columns with empty values that might cause doubles to be read as strings after being filtered in the dataview
-                foreach (DataColumn field in data.Columns)
-                    if (field.DataType == typeof(string) && field.ColumnName != "CheckpointName"&& field.ColumnName != "SimulationName")
-                        filter = AddToFilter(filter, $"{field.ColumnName} <> \"\"");
-
+                
                 //cleanup filter
                 filter = filter?.Replace('\"', '\'');
                 filter = RemoveMiddleWildcards(filter);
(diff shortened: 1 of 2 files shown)


ric394:
@zur003 what is the rationale for the addition of global to the APSIM.Core using statements. 
From my quick read up on it's usage it doesn't seem like this will make a difference. Does it help with the cleanup issue?

zur003:
@ric394  I found that the `global:` qualifiers were required to get things to compile, but I must confess that I don't understand namespace resolution well enough to explain why that was the case. I think part of the problem was that (for example) by changing the namespace `APSIM.Core.Tests` to `UnitTests.APSIM.Core.Tests`, other references to `APSIM` in namespaces (such as `APSIM.Shared`) were being sought as something within `UnitTests.APSIM` rather than globally.

There may well be better ways to handle the overall issue, but nesting everything withing a single `UnitTests` namespace does simplify handling the "cleanup" procedures..

Issue #9918, opened 2025-04-28
Add simulation metadata table to datastore


par456:
### Describe the new feature

As part of the new POStats system, we need to know certain metadata about the simulations being run so that the data can be filtered and displayed correctly. To do this, we'll add a new hidden _metadata table to the datastore that will record a bunch of information about each simulation as it is running.

To start with, we'll record:
Crops Sown
Cultivars used
Years Run
Latitude and Longitude

If there are any obvious metadata that I've not thought of, record it here and I can add it to the list.

This will be more than just interrogating an apsim file, this would happen during simulations so you don't need to account for replacements or factors changing the base simulation, or scripts running and setting what cultivar is being planted. This should record what actually happened, not what the file outlined should happen.

hol353:
Instead of Crops Sown, maybe list all models because POStats has soil water / nutrient / MicroClimate / Stock validations as well.

HamishBrownPFR:
It is a bit hard to know what will be useful to have here as it will depend on what is easily placed elsewhere.

Currently we can put a meta data variable into a report and if the test file is set up accordingly with replacements it is easy to get that variable output for all simulations in a test set.  As the test sets are broken up this becomes harder to do so I guess it then becomes sensible to tend toward putting more into the meta data in this database so it can use used for post simulation analysis.

Here is a list of the things I use in the wheat data for analysis 

'IWeather.Latitude',
'IWeather.Longitude',
'Wheat.SowingData.Cultivar',
'Wheat.SowingData.Population'
'Wheat.SowingDate'
'Experiment',
'FolderName'

Perhaps it would be a good idea to compile a list of SowingData objects so there is a record of what crops are sown, and other associated information such as cultivar and population.   If SowingDate was added to SowingData it could hold the when component too.

BrianCollinss:
- Irrigated / Rainfed
- Field / Controlled Environment

PR #10913, opened 2026-01-19
9918 Metadata

par456:
Working on #9918

Testing and planning work around adding metadata to our validation data. This work requires changes on both the apsim and postats sides to make it work, so this isn't a merging PR yet, just something to do testing with.


par456:
Closing this for now, will reopen new PR with additions.

PR #10918, opened 2026-01-21
Report stress factors

PengchengHU:
Working on #10170


PengchengHU:
Hi @par456  and @ric394 , could you please merge this one? Just added the stress factors and their graphs to the barley.apsimx. Ta.

It is weird that it caused changes in wheat performance as shown in the POStat.

par456:
I think what happened here was that there was two runs of the validation happening at the same time on this PR. Those shouldn't interfere with each other, but I have seen it happen before. I reran your PR and the stats are fine now.

PR #10938, opened 2026-01-27
Wheat test tidyups

### LINKED ISSUE #10937: Some configuration errors in wheat test set
### Describe the new feature

- Pask soils have summer and winter dates for changing u and cona inverted
- VP values in 325_52775.met (used in Pask simulations) are extremely low.  Possibly wrong units
- Grain fill duration is too short for NZ genotypes Claire and Wakanui.

HamishBrownPFR:
resolves #10937

### DIFF
--- a/Tests/Validation/Wheat/Pask/PaskExperiments.apsimx
+++ b/Tests/Validation/Wheat/Pask/PaskExperiments.apsimx
@@ -1288,7 +1288,6 @@
             "[Wheat].Root.RootFrontVelocity.PotentialRootFrontVelocity",
             "[Wheat].Phenology.Photoperiod",
             "[Wheat].DaysAfterSowing",
-            "[Wheat].Leaf.TillerProsperity",
             "[Wheat].Leaf.WaterDemand",
             "[Wheat].Leaf.Transpiration"
           ],


HamishBrownPFR:
retest this please jenkins

HamishBrownPFR:
@par456 @ric394 I am not sure why this PR is failing.  Any pointers appreciated

par456:
This is a weird one, I'll have a look right now.

par456:
There was a 2nd missing comma in the wheat.json file with the edits you made, but it only effected apsim on linux, the windows serialiser was able to handle the issue.

HamishBrownPFR:
Thanks @par456

HamishBrownPFR:
Trivial changes to stats as a result of changing the grain full duration of Claire and Wakanui.  @par456 @ric394 ready for review.

par456:
What changes happened in the excel file? 

This PR has both data and model changes, so could I get you to put the Pask LC07 TT0607.xlsx‎ into its own PR so we can see those changes separate to the changes caused by the model.

HamishBrownPFR:
The changes to the excel file are changing simulations names.  The Pask tests have 3 simulations with nitrogen treatments.  2 had the factor called N and the other had it called nutrition so I changed to make consistent.  Need to submit the excel file and the pask.apsix file together to keep the simulation names in sync for matching observed to predicted.
What is potentially more interesting is why the stats have not changed for the pask tests.  I was expecting they would

par456:
Ah right, cheers for that, then yes, this PR will be fine to keep together.

Looks like the Pask experiments threw an exception which is why they aren't showing up (under the logs on POStats), it's complaining about not finding Leaf.TillerProsperity in a report.

HamishBrownPFR:
Thanks @par456 I will fix that.  How come it did not cause run-model-validations to fail?

par456:
A limitation of the new system, the github workflow doesn't currently see the result, it just 'completes' if everything is started correctly. The Acceptance Tests is the only status that checks final results now, exceptions and stats together. I'll add this to the list of things to address with future work on this.

HamishBrownPFR:
Now I understand what is happening, the new approach is fine.  It would be useful if validation failures were listed clearly somewhere to make them easier to identify

HamishBrownPFR:
Stats look as expected.  Some up and some down but on balance the number and magnitude of the ups is greater.

PR #10943, opened 2026-01-28
Wheat test tidyups files only

par456:
Working on #10937

Doing a PR to merge all the other changes from #10938 without the model change.

### DIFF
--- a/Models/Resources/Wheat.json
+++ b/Models/Resources/Wheat.json
@@ -7358,8 +7358,7 @@
                     "[Phenology].CAMP.FLNparams.VxPLN = 4",
                     "[Phenology].HeadEmergenceLongDayBase.FixedValue = 150",
                     "[Phenology].CAMP.EnvData.VrnTreatTemp = 6",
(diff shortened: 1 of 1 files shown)


HamishBrownPFR:
Changes to stats are what I would anticipate.  Substantial improvement in biomass predictions because water stress is not over predicted now.  Still issues with low grain yields because grain growth phase is too short for the genotypes used

par456:

A number of errors were found in the setup of the Pask wheat experiment, this pull request will resolve them. This includes summer and winter u and cona being around the wrong way and vp values in met files in the wrong units.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10937

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10943/files).

The changes in stats can be found [here](https://postats2.apsim.info/10943).

This is only repairing mistakes in a simulation and will therefore be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #10946, opened 2026-01-28
Wheat test tidyups

par456:
Resolves #10938

Continuation of #10938 because github closed both PRs before. This has just the cultivar changes in it.


par456:


This pull request increases the grain fill duration for Claire and Wakanui cultivars in the wheat validation. They were previously too short. 

Stats show generally better results on important variables.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/https://github.com/APSIMInitiative/ApsimX/pull/10938

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10946/files).

The changes in stats can be found [here](https://postats2.apsim.info/10946).

We intend to merge this into the release on Monday 2nd unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

PR #10974, opened 2026-02-06
MarkdownView now fetches remote images

### LINKED ISSUE #10973: MarkdownView should be able to show remote images
### Describe the new feature

The MarkdownView, which is used for showing documentation does not currently show images outside APSIM. 
Changes should be made to make this available.

ric394:
resolves #10973

If a markdown image such as `example image` is present, APSIM will attempt to fetch and save the image in a temporary location for use in the MarkdownView


par456:
Can you wrap this web code into a function in WebUtiltiies (and check if there is anything in there that might already do this for you).

Rather than having web fetching code just hanging out in a view class.

PR #10986, opened 2026-02-11
4631 Enable all Cotton Sims

par456:
Working on #4631 

Only some of the simulations are enabled in the cotton prototype validation, going to enable them all to try and sort out the number changes in postats on the other PRs.


par456:
This POStats looks more sensible. Increased N values with all the sims enabled. No changes to the model in this one, so changes are purely from having more data.

Ready for merging

PR #10987, opened 2026-02-11
4631 Cotton PR

par456:
Working on #4631 

This brings in all the graphs, reporting and simulation changes from the new cotton, but using the existing cotton model.


par456:
Does not include simulation changes, those will be next

PR #10733, opened 2025-11-06
Ndvi model

HamishBrownPFR:
resolves #9901

Source code for this has already been merged.  

This adds a converter that changs NDVImodel in manager to replace it with the compiled model.  
Changes the name of Variable NDVIModel.Script.NDVI to Spectral.NDVI in all excel files containing this variable

Fixes a bug in the JSONUtilities so the SearchReplaceGraphVariableNames works

PR #10992, opened 2026-02-12
Fix yield and water uptake calculations in STRUM

### LINKED ISSUE #10991: STRUM calculation improvements
### What happened?

Two errors have been identified in STRUM:
1. The calculation of Fresh Yield Per Ha is based on the area of the canopy, not the area of ground which is confusing.  Needs to be changed.
2. In some cases, water uptake by the understory exceeds supply and demand.  This has been traced to an error in the way the report variable is calculated where the area of zones sometimes switches around so the wrong area is applied.

Windows

HamishBrownPFR:
resolves #10991 

The issue resolved here will fix problems in models using the classes in BiomassArbitration (currently only STRUM and SPRUM in release).
These classes use PlantWaterOrNDelta so agregate resource uptake, supply and demand variables over multiple zones.  However, each plant model was retrieving zones in a different order which mean sometimes the areaPerZone property of the PlantWaterOrNDelta class were set in the wrong order.


HamishBrownPFR:
@par456 @ric394 This PR fixes a bug which is why the stats change.  The bug fix has made the fits substantially worse for the Prosser experiments but I think this should be merged and then any further changes needed to refit the model to that data set made on a seperate PR.  Please let me know if you disagree.

ric394:
I've had a look at this one and it looks fine. Any further comments @APSIMInitiative/administrators ?

HamishBrownPFR:
On reflection, I think it will be better to address the reduction in accuracy as part of this pull request.

HamishBrownPFR:
@ric394 @par456.

The current commit on this PR fixes a bug that had been introduced that was causing a substantial change in the stats for the Prosser experiments (which are now returning the same results).

This should be good to merge.

HamishBrownPFR:
The yield stats for Avocado experiments have changed.  This is because there was a bug in the calculation of yield per ha.  It was being calculated as kg/ha of canopy rather that kg/ha of land.  Fixing this has changed the stats for this test in Avocado.  Stats for other yield observations have not changed because they are kiwifruit orchards where the canopy cover is full (grown on pergolas) so canopy area and land area are the same thing.

HamishBrownPFR:
@ric394 @par456 Please let me know if anything further is required on this.

HamishBrownPFR:
@par456 @ric394.  I have some follow on work that requires this PR to be finalized.  Please advise if there is anything further I need to do to complete it.

par456:


This pull request solves two errors have been identified in STRUM. First the calculation of Fresh Yield Per Ha is based on the area of the canopy, not the area of ground which is confusing. Second, in some cases, water uptake by the understory exceeds supply and demand. 

This has some small effects on the stats for STRUM.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10733

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10992/files).

The changes in stats can be found [here](https://postats2.apsim.info/10992).

As this is fixing a bug in the model, this will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #10996, opened 2026-02-12
Sow operation pasture model

### LINKED ISSUE #10993: Add sow operation in Pasture Model

\ApsimX\Models\Pasture\Pasture.cs

In order to implement IPlant interface Sow operation is required in Pasture.cs class.

JPaul-Vasa:
Resolves  #10993 

Added Sow method to trigger when there is a Sow event.

### DIFF
--- a/Models/Pasture/pasture.cs
+++ b/Models/Pasture/pasture.cs
@@ -563,13 +563,15 @@ namespace Models.GrazPlan
         {
             get { return null; }
         }
+       
         /// <summary>
-        /// Not implemented
+        ///IsAlive flags plant status alive or dead
         /// </summary>
-        public bool IsAlive
-        {
-            get{return false;}
-        }
+
+        public bool IsAlive { get; private set; }
+
+       
+
         /// <summary>
         /// IsReadyForHarvesting: Not implemented
         /// </summary>
@@ -597,7 +599,7 @@ namespace Models.GrazPlan
             throw new NotImplementedException();
         }
 
-         private bool isAlive = false;
+         
 
         /// <summary>
         /// Sow: Not implemented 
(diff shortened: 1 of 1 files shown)


Models/Pasture/pasture.cs
```diff
@@ -595,6 +595,9 @@ public void Harvest(bool removeBiomassFromOrgans = true)
         {
             throw new NotImplementedException();
         }
+
+         private bool isAlive = false;
```
par456: IPlant has a property definition for IsAlive, I would suggest using that instead of creating a new private variable that does the same thing.

Models/Pasture/pasture.cs
```diff
@@ -2418,6 +2535,38 @@ private void OnDoPastureWater(object sender, EventArgs e)
...
+        //                 {
+        //                     if (value > 0)
+        //                         return true;
+        //                 }
+        //             }
+        //         }
+
+        //         return false;
+        //     }
+        // }
```
par456: Do you still need this? 

If not, would be better to delete commented out code rather than committing it to the repo.

Models/Pasture/pasture.cs
```diff
@@ -609,10 +612,124 @@ public void Harvest(bool removeBiomassFromOrgans = true)
...
+                        green[idx].stress_index = greenInit.StressIndex;
+                        green[idx].frosts = greenInit.Frosts;
+
+                        // ABOVE-GROUND
+                        green[idx].herbage = new Herbage[2];
+
+                        // LEAF
+                        green[idx].herbage[0] = new Herbage(1, new TPlantElement[] { TPlantElement.N });
+                        green[idx].herbage[0].dmd = greenInit.LeafDMD;
+                        green[idx].herbage[0].weight =
```
par456: For styling purposes, we generally try to avoid using the = ? : syntax for single line if statements, it just reads easier as a proper if statement.

Could you change this one and the one below to get rid of the shorthand.

PR #11037, opened 2026-02-23
Standing fraction set

### LINKED ISSUE #11036: Set FractionStanding when crop biomass is removed
### Describe the new feature

SurfaceOrganicMatter has two pools of biomass, standing and lying.  The later is not connected to the surface so does not degrade and represents stubble that has not been mown or incorporated.

Currently in next gen, StandingFraction is assigned a default value of zero and is not possible to set an alternative value when biomass removal takes biomass from the plant model and gives it to the surface organic matter model.

Need to make StandingFraction an arguement in BiomassRemovals.RemoveBiomass() so it can be set

HamishBrownPFR:
resolves #11036 

Adds code needed to set standing fraction on a biomass removal event.
Add code to SCRUM so standing fraction can be set as part of a standard SCRUM simulation
Add sensibility test to make sure is setting standing and lying surfaceOM fractions as expected

### DIFF
--- a/Docs/content/Usage/server.md
+++ b/Docs/content/Usage/server.md
@@ -20,7 +20,7 @@ The recommended way to send these commands to the server is to use a client API.
 
 ## Zmqserver Behaviour
 
-Likewise the server will initially prepare a simulation to run, and then wait for a client connection. In the oneshot mode, the commands (run, get) are similar to the V1 server. In interactive mode, the server will pass the simulation the address of a controlling port that a manager component within the simulation will connect to. The protocol (get/set/do/resume) for this connection is implemented within the manager component. Samples of clients in R and Python are in the [Test Directory](../../../Tests/Simulation/ZMQ-Sync/).
+Likewise the server will initially prepare a simulation to run, and then wait for a client connection. In the oneshot mode, the commands (run, get) are similar to the V1 server. In interactive mode, the server will pass the simulation the address of a controlling port that a manager component within the simulation will connect to. The protocol (get/set/do/resume) for this connection is implemented within the manager component. Samples of clients in R and Python are in the [Test Directory](https://github.com/APSIMInitiative/ApsimX/tree/master/Tests/Simulation/ZMQ-Sync).
 
 ## Notes on running/invoking the server
 
--- a/Tests/UnderReview/Cotton/Cotton.apsimx
+++ b/Tests/UnderReview/Cotton/Cotton.apsimx
@@ -7566,84 +7566,35 @@
                       "ReadOnly": false
                     },
                     {
-                      "$type": "Models.Functions.LinearInterpolationFunction, Models",
-                      "Name": "WaterFactor",
-                      "ResourceName": null,
-                      "Children": [
-                        {
-                          "$type": "Models.Functions.SoilWaterScale, Models",
-                          "LLModel": 1,
-                          "Name": "SoilWaterScale",
-                          "ResourceName": null,
-                          "Children": [],
-                          "Enabled": false,
-                          "ReadOnly": false
-                        },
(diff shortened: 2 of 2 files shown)


HamishBrownPFR:
@par456 @ric394 can someone have a look at this please.  Stats are failing because some simulations are missing.  Not sure if they are not being run or just not matching with observations.  The simulations that are missing in the test all run, appear on obs/pred graphs and give the same results when run locally.

HamishBrownPFR:
@par456 @ric394 can this one be merged now?  Let me know if there are any further requirements

par456:
I had a look at this the other day, but needed a bit of time to think about it and get some second opinions.

Is it just organ that is actually taking advantage of the addition of FractionStanding to the interface? or are there other classes as well that I've missed. It seems like a lot of changes to add this to the remove biomass interface if most of the classes are just going to ignore the input.

Also I would prefer not to use _fractionStanding as a property in Organ, that underscore notation means temporary variable and it's not something we normally use in apsim classes.

HamishBrownPFR:
Thanks @par456 

More detailed explanation of what I have done and why.  
Fraction standing is primarily an argument in SurfaceOrganicMatter Add method which is unchanged by this pull request.
```csharp
public void Add(double mass, double N, double P, string type, string name, double fractionStanding = 0, double no3 = -1, double nh4 = -1)
{
```
Any model implementing the IHasDamageableBiomass interface uses the above Add method through their  RemoveBiomass methods.  This is most plant organs and a few others.  

To enable the setting of fractionStanding I have added it as an argument to the RemoveBiomass method in IHasDamageableBiomass.

This is the step that makes it possible for any plant model to specify a non zero standing fraction when biomass is removed.  Thus far I have set standingfraction to zero in all (but one exception) organs so the current behaviour is maintained.  
The case that required standingFraction to be set were some SCRUM simulations so I have implemented code inSCRUMCropInstance and SCRUMManagerInstance to allow users to set non-zero standing fractions and tested to ensure it is working as intended.
Functionality should be added to other crop models to set standingFraction but this would require more thinking and testing that I have time to cover at the moment.  So most things are ignoring it at the moment.  However, the change to the interface means the plumbing is there for them to deal with it when it is required.  

regarding _fractionStanding in organ.  It is used in the scope of a method that has fractionStanding as an argument so I can't use that because it clashes.  I could make it into FractionStanding but was of the understanding we should not start private fields with a capital.  I could also make it another name like fracStanding but I find it confusing when we get a multiplicity of subtly different variable names that have the same value.   Please advise what the preferred approach is for private fields that have a direct relationship to public property.

par456:
Hi Hamish, thanks for that description that does help out.

I've gone in and made some small changes to get this merged. I see what you mean about the interface, I don't think there is a good solution around so we'll put that change in.

However rather than have an optional from RemoveBiomass getting passed into an optional in SOM, I've made fractionStanding non-optional now on the Add function, and added an override function for Add that uses 0 if fractionStanding is not passed in. That'll avoid errors in the future where the optional on RemoveBiomass isnt the same as the one on Add and cause issues.

On the _fractionStanding property. I have had a look through the way the code is written, and worked out that other than when AddSOMtoZones is called in the RemoveBiomass function, fractionStanding is always 0, since it starts as 0, and is reset back to 0 after the call in RemoveBiomass and never set anywhere else. Since BiomassRemoval is where we have access to fractionStanding, I've added that as a optional on AddSOMtoZones and removed it as a property so we don't have a name conflict in that class anymore.

Julian will merge this if you're happy with that change once it's finished running through the tests.

HamishBrownPFR:
@par456 changes look good to me.  Thanks for taking the time to tidy this up further.  Please go ahead and merge.

PR #11044, opened 2026-02-26
11043 Reduce check to 0.001 and see what changes

### LINKED ISSUE #11043: Pinus dies when dropping below 1 plant per 100m2
### What happened?

Due to safety protection on the plant population setter, if a plant drops below 0.01 per m2 than it will be killed with EndCrop.

Some new Pinus validation does have plant density this low however, and shouldn't kill the crop at that point.

Windows

par456:
Resolves #11043

Dropping the check from 0.01 to 0.001 to see what changes happen in the validation.


par456:
This had no effect on the validation, so should be a safe solution to that problem.

Issue #11054, opened 2026-03-02
Ability to read detailed composite factors for Experiment from an input file


lie112:
### Describe the new feature

There are situations where experiment factors contain many CompositeFactors with each level being a large list of properties to change, such that each simulation is based on a series of properties set to those of the case simulation. For example, a feeding trial may have many individuals with different characteristics (initial age, weight, SRW, body fat etc), different feed types (grain, hay, supplement) and regimes (daily rates, feed till satisfied), different lengths of feeding (simulation start and end date) etc. So, this doesn't follow the standard breed x feed type x feeding regime x trial length approach we can currently implement in APSIM within the Experiment. While the grouping may still exists it isn't set individually for each level of the factor rather than for groups of simulations. CLEM is able to run all these simulations if we can build all the CompositeFactors for hundreds of individuals in the feedlot trial being used in model validation.

I have written Python scripts to read a formatted excel spreadsheet as a template, modify a base APSIMx json simulation file and run an CMD line APSIM simulation. At the end of all simulations, I then join all the individual db output files into a single database file for analysis. 

This approach works but is a bit clumsy for other researchers to use and is outside the Experiment logic. 

I know there was talk of maybe looking at the Experiment code in future to simplify and add more functionality. 
I always thought nesting Factors would be a nice way of including permutation in the experiment. In this case, there is no permutation component, but the GetSimulations of a Factor responds to subfactors to build permutations.

So, I propose this initial functionality I require is provided within APSIM UI with a new component for the Experiment called FactorsFromFile which has a property that points to a excel spreadsheet that provides all the information to create the factors.  This reads the file and creates new Factors with child CompositeFactors that get added to the simulation tree before the simulation is run. Once the factors are created, the experiment works as normal and sees the newly added factors either below the Factors or below Permutation depending on the location of the FactorFromFiles.

The spreadsheet has at least two sheets. The first called "properties" defines the property paths to use in any compositeFactors.

Properties Sheet
Column "label": a unique label to identify the property to use.
Column "path": the APSIM node path to the property as would be provided in the Specifications list of the CompositeFactor

Then a sheet is provided for each factor to include where the sheetname is used for the factor name. This spreadsheet provides a row for each level or CompositeFactor of the Factor where the columns identify the property and the column name matches a property label, and the column values are the settings for the property for each factor level.
[...]

PR #11055, opened 2026-03-02
Factors from file component

lie112:
Working on #11054 

Here is the implementation of new feature #11054 for consideration.

You can add the new FactorsFromFile component under any Experiment in the Factors or Permutation children
The example spreadsheet provided simply changes the start and end dates of two levels.

[FactorsFromFileSpreadsheet.zip](https://github.com/user-attachments/files/25673388/FactorsFromFileSpreadsheet.zip)


lie112:
This is a preliminary pull, for consideration to include this approach and there is some discussion to be had about ways to report the outcome of this component to the user when Models does not talk to APSIMNG UI. This may result in the StringWriter code being deleted, and the discussion comments at the end of CreateFactors method removed.

par456:
I've had a brief look at your code last week and I think what you are trying to do is sensible, but the software team needs to get together and have a session to talk about it to think it through some more.

I suspect there may be some tools already in apsim that we can take advantage of to reduce the complexity of this class a bit.

What I think we do need is an example of the file that this would read in, that'll help in figuring out some more of your requirements for this.

lie112:
The issue is when you need to create a different compositeFactor with lots of properties to set for each level of a factor and there are potentially hundreds of levels to consider, and these might change or be updated regularly between simulations. I've done this externally in Python by building the .apsimx json file, running the simulation with Console, and then combining all db files into a single file at the end of analysis. This is very clunky when an error is thrown by one of the APSIM simulations that is hard to trap and present to the user. The experiment approach in APSIM UI currently works for changing a single factor at a time, which also works through permutation to give all combinations. But there are times where each simulation is a detailed set of parameter and setting changes, across multiple APSIM and CLEM components that can't be simply build from the experiment structure. It would be nice if I could also assign factor levels to these, such as Sex, Breed, FeedStyle, .... that would be included as columns in the report, which should be possible, but I never fully worked out this aspect of experiment reporting. In effect, I am creating the combined list of simulations all at once representing a large list of compositeFactor components below a factor. The approach is able to handle any number of factors to include, and thus when added below a permutation it becomes extremely flexible and able to read simulation data setup from a provided spreadsheet and perform many custom simulations and run a further set of factorial runs on the simulations, all with one component and an excel file.  

This simple class adds much needed functionality I have done manually in the past with Python, but I need to be able to include hundreds of simulations in the same file for validation purposes and didn't want to add hundreds of individual simulations somehow tied to validation data into anywhere in APSIM validation. By, creating the Factor and CompositeFactor components that are saved with the apsimx file this approach allows Experiment and Permutation to work without change, but is simply a way to read in a list of factors with associated compositeFactors that are otherwise imposible to get into APSIM simulations without a lot of manual work and hours of searching for issues when one property was copied and pasted without changing the value correctly. Plus the output can be viewed in DataStore and APSIM graphs.

I know there is talk of refactoring the whole experiment handling in future, but I can't see this happening for a long time and with not wanting to cause non-breaking changes this will be difficult. I always thought nested factors in the tree structre would be a great way of undertaking permutation rather than adding a special component with factors below. So, I have been coming up with my own solutions to running detailed multi-CLEM simulations and thought I'd throw a simple component together so the researchers can undertake this analysis as they are really struggling to use Python to do the task. I am happy for this to be included in Prototypes and tried to write it in a very generic fashion such that any APSIM user might find it useful. I acknowledge there might be ways to clean up the reading of data sheets and analysis to generate the factors from the spreadsheet, but this doesn't need to be anything special and only runs once at the start of "Run".  I used the latest Structure and Model.Node.AddChild() to perfrom the tasks and spent two days fixing up all new CLEM code to correctly use Model.Node where possible and correct the errors in applying the new structure across APSIM code where some "relativeto" or "recurse" arguments had been omitted and caused CLEM simulation to fail.

Here are some example files. Unfortunately, all my files are in my dev version which won't open for you, but I could just give you the cut down experiment if you wanted the aspimx files. 
- SheepSimulations.xlsx is the input file for my python script that the new code also handles. Here there are a heap of individual sheep with different starting ages etc that need to be run through simulations to compare with validation data.
- HeiferValidation. This is a screen shot of the experiment setup done manually for heifer growth validation. The Treatment/HeiferMonthly shows what needs to be set for each individual in the treatment. this is across a range of components so not easy to add copies of components into the experiment to use. There are other validations where we need lots of individuals. It is so much easier and more transparent to manage this data from the field data in spreadsheets than miss a seeing in the experiment factor of UI tree.
- GrowPF_MLA_Validation is another example.

[Examples.zip](https://github.com/user-attachments/files/25891613/Examples.zip)

lie112:
PS. I left the logging detail code for now just in case there was an easy way to display this to the user, but all the StringWriter details will likely be deleted before accepting.

lie112:
I guess this could be an "import factors" right click pop-up menu option off a Factor, or Permutation node in the tree and just create the factors below. This wasn't quite suitable as they need to be deleted and the process repeated to update the settings, while this component overwrites the factors when the simulation is run.

par456:
Spent most of last week working on this, so apologies for completely rewriting the class, but it is much more straightforward now, and more customisable for some other use cases we want it for. 

The big changes from the original code:
- FactorsFromFile now builds a command script to do the file editing using the command language. This means it doesn't need to touch the structure itself and simplifies the code.
- The csv entries can have commands like getting a model from another file (see example below) allowing you to build your experiment from template files defined elsewhere
- There are now three properties to define, the name of the factor that will be created, the name of the column that acts as the factor labels, and the filepath to the csv file.
- The GUI now updates the tree structure when the csv file is updated, but currently only if FactorsFromFile is the active presenter.
- Switched the presenter to use the QuadPresenter so it can show the contents of the csv file, and the properties.

The biggest change here is that this is the first model implementing a way to generate models that are added to the tree structure. Managers do this, but they have previously been an exception, so I had to do some design sessions with @ric394 to figure out how we intend to do this going forward, since this is something we have wanted in other situations previously.

This was my test case, and how it looks in the GUI now: 

image


@lie112 Can I get you to pull down this branch and do some testing to make sure it still fits your use cases, and if not, what I need to fix so that it will. In the meantime I've got some changes to clean up and tests to write for this.

lie112:
@par456 Can I get a copy of the test simulation file and the csv file you are using as when I tried to add the component I just got errors about no file provided and then stuck in UI breakpoints and errors.

I must say I have no idea how you went about implementing this from looking at the code.
It would be nice to understand how you are doing all this updating of apsim files from UI.

From the image, is the grey section the nodes that have been added dynamically from the file?
How do you implement a factor (say AnimalID) that has a range of properties to be set for each level within the composite component such as Age, Sex, Wt, ....... through to parameters of the breed?
It looks like you have to specify a factor name and the name of the column in the csv file, which automatically assigns the values in the file to that factor. 
The attached excel spreadsheet (Examples) from message above shows that a factor (SimLength, taken from the sheet name),  has three levels (1yr, 5yr, 10yr) and creates three composite factors with start ([Clock].StartDate) and end ([Clock].EndDate) properties with the row values used for each level, where the property path is provided in the Properties tab. This was a very simplified test case to show the concept.
This allows users to be working with results and link the spreadsheet up to the factors of the simulation. It also means there can be additional calculation columns not considered in the composite components (i.e. StartAge leading to StartAgeD in example spreadsheet). 
Can we build on top to allow multiple factors to be specified in the file and read in automatically. This was designed to allow the spreadsheet to control the experiment which was a lot easier to work with than the experiment presenter when you might want to ignore one factor level etc as I started to add the Include column to implement this. 

Here's the start of a more detailed version of all the sheep feedlot trial validation runs that are needed.
image

par456:
> @par456 Can I get a copy of the test simulation file and the csv file you are using as when I tried to add the component I just got errors about no file provided and then stuck in UI breakpoints and errors.

It must be trying to do things when the properties are still null, I'll have to add some more checks in to stop that.

>I must say I have no idea how you went about implementing this from looking at the code. It would be nice to understand how you are doing all this updating of apsim files from UI.

To change the GUI involved going deep and refactoring/fixing some of the deep GTK code. At the model level, it just has a timer on the property presenter that tells the gui to refresh the tree if the file has been saved more recently than it was last read.

To actually change the models, the GetCommands() function is building a list of text commands on how to modify the apsim file. These commands are described here: https://apsimnextgeneration.netlify.app/usage/commandline/commandlanguage/

Previously these could only be passed in as a command line argument to Models.exe, but this model is now using them within the GUI to update the models. The plan is to move more of our 'replacements' systems over to using this over time, so we don't have models touching the structure. I considered adding the commands output to the presenter, that might be good way to demonstrate what this model is doing behind the scenes.

> From the image, is the grey section the nodes that have been added dynamically from the file?

Correct. Any nodes that are generated like this are being marked as read only so they are not saved to the file, kind of like being temporary models.

> It looks like you have to specify a factor name and the name of the column in the csv file, which automatically assigns the values in the file to that factor.

Correct, your column headings define what factor should be overridden with the provided values in that column. So your start and end columns would need to be changed to Clock.StartDate and Clock.EndDate .

> How do you implement a factor (say AnimalID) that has a range of properties to be set for each level within the composite component such as Age, Sex, Wt, ....... through to parameters of the breed? It looks like you have to specify a factor name and the name of the column in the csv file, which automatically assigns the values in the file to that factor.

This might be where we are starting to diverge on functionality a bit. I was approaching this as one row of the sheet connected with a single permutation within APSIM. If we want to allow for further permutations based on a list of values from a cell, we'll need some way to differentiate that from a List of values for an array input (which is what it currently expects).

> Can we build on top to allow multiple factors to be specified in the file and read in automatically. 

That is indeed how this works. Based on the columns and rows, each row is built into a CompositeFactor with all the values in that row assigned to the correct properties. It should end up with exactly as many simulations as rows in the file.

> as I started to add the Include column to implement this.

That's a good point, the column should be called "Enabled" instead, and simply disable the simulation if set to false. That way it better mirrors how experiments currently work. Will add that in.

Let me do some fixes so you can get it working and play around with it some more. I'll also get my test file so you can see what I did.

par456:
Worked out a bunch of the gui issues and tidied some more bits up, should be testable now Adam.

Here is the files I've been working with to build this, might give you an idea of how I'm setting things up:
[factorfile.zip](https://github.com/user-attachments/files/26556266/factorfile.zip)

lie112:
Here are a few fixes to get what I was after

- Allow FactorsFromFile to be added below a Permutation component as well as a Factors component. This requires the parent to be used in the script rather than "[Factors]" (I assume this is the Factors node regardless of whether the user has renamed it). Set the valid parent attribute to be set to "Factors" and "Permutation" for this component which I had already done. 
- Rename to FactorFromFile as this no longer supports adding multiple factors from a single file as far as I can tell (That would be great from the spreadsheet read). This would need a symbol to identify header rows in the CSV file, or work with spreadsheets and all or a number of named sheets.
- I don't like having to have the full property path as the column header in the csv as these can get very long and make the file hard to read when working with the data. There is a difference between a specific file such as the csv you work with and a data table the user is working with such as in Excel. The ability to work with a working table and select the columns of interest is what I had developed. Could we have an additional file filename.props or something where the column names to use and their associated property path are provided in comma delimited manner. This would also need the sheet name for excel, and sheet name could be used as the factor name. This is equivalent to the Properties sheet in my spreadsheet. The system could look for properties sheet when a spreadsheet is provided and not include this sheet in the other factor tables. It would revert to your approach is the properties sheet weas not provided. The properties sheet also provides a means of specifying a sub-set of columns to be included. 
- There is no way of ignoring any columns or rows (see previous point). an Include column could specify the rows to include. 
- I tried to do this from a spreadsheet.  I was getting there, but I got a "Column 'level' does not belong to table 'Duration'.  This was expected but the property entry panel collapsed and the table view was above the properties until I realised I had to resize all the panels to see things again and it was hard to find the resize handles unless you realised what you needed to do.  
- The error message window needs to be cleared occasionally. After you get an error, it is hard to know when you have addressed it and it has gone away. There used to be a "Clear" menu popup option. It would be nice if errors had an owner and the owner could clear them when happy. Then if the presenter worked it could remove the error. 
- Altering the excel file does not result in presenter update
- Data table display only updates when you leave and come back to the component
- The readonly Factor and children don't update. Maybe this was after I changed the Factor Name. I deleted the readonly Factor, but I can't get it re populated with the new data. Oh I think I now continuously get a "Object not set to instance of object" but I don't know why. This happens if I open the file again too. Is there a way to rerun the Commands?
- This will need to be pretty straightforward abd bug proof for most users to understand otherwise it is quite black box. 
- The commands script is great. Where are the commands designed to be used from? Can we use these for other purposes in modifying the sim file, or are these only for use in the UI in presenters?

par456:
> Allow FactorsFromFile to be added below a Permutation component as well as a Factors component. This requires the parent to be used in the script rather than "[Factors]" (I assume this is the Factors node regardless of whether the user has renamed it). Set the valid parent attribute to be set to "Factors" and "Permutation" for this component which I had already done.

Done

> Rename to FactorFromFile as this no longer supports adding multiple factors from a single file as far as I can tell (That would be great from the spreadsheet read). This would need a symbol to identify header rows in the CSV file, or work with spreadsheets and all or a number of named sheets.

Done, but may require you to hand edit your test file to open it again.

> There is no way of ignoring any columns or rows (see previous point). an Include column could specify the rows to include.

I'll make it so that columns that don't match a variable are not used. That'll solve the next point too then. I started doing this at one point, so it simple to change back.

> I don't like having to have the full property path as the column header in the csv as these can get very long and make the file hard to read when working with the data.

This sounds like you need an alias list, to match table headers to apsim variables. Let me have a think about how the best way to do that is.

> The error message window needs to be cleared occasionally. After you get an error, it is hard to know when you have addressed it and it has gone away. There used to be a "Clear" menu popup option. It would be nice if errors had an owner and the owner could clear them when happy. Then if the presenter worked it could remove the error.

Done

> Altering the excel file does not result in presenter update

The excel file will suffer from the same issue as all excel files in apsim, in that apsim cant read them if excel has them open. Excel puts a block on files it has open that our reading library is stopped by. When I tested though, after editing and closing the excel file, and with FactorFromFiles still selected, the tree did update.

Errors on reading may have also caused an issue, so hopefully that's resolved now.

> Data table display only updates when you leave and come back to the component

I shall have to look into this further, may review a page refresh.

> The readonly Factor and children don't update. Maybe this was after I changed the Factor Name. I deleted the readonly Factor, but I can't get it re populated with the new data.

I'll make it so that when FactorsFromFile is selected, it forces a tree update so it can refresh this. In the future it might be worth making the refresh timer for components like this global so they update at any time, but that seems a bit dangerous still right now.

> This will need to be pretty straightforward abd bug proof for most users to understand otherwise it is quite black box.

Yeah. I think it helps that you can see the structure it is making, but if you don't understand experiment nodes that well it does seem a bit black boxish. The GUI has been a real issue for this PR the entire time, I do worry all the changes are going to have knock on effects that I don't know about yet.

> The commands script is great. Where are the commands designed to be used from? Can we use these for other purposes in modifying the sim file, or are these only for use in the UI in presenters?

They were originally designed to be 'acted upon' an apsim file like a runtime modifier. The GUI kind of sits on top of the models and structures, so all I'm doing here is using the commands to change the underlying structure, then telling the gui to refresh itself to see the changes.

This presenter is passing them into the CommandProcessor.Run function to change the loaded model, but normally we just have a list of commands in a text file, and run them using the --apply flag on Models.exe,

par456:
Starting to get this much more stable now, though the scope has obviously crept a little.

Added a manual refresh button to the property view, this is a new type of property that is just a boolean behind the scenes that can be used to trigger model changed events. I've then removed the auto-loading feature, as it just straight up isn't working well with excel files (since excel is blocking the reading).

Lots of small fixes around inputs that I found when implementing this with some project work, so shouldn't break as easily now. GUI is now also handling the adding/deleting of the read only nodes much better and throwing less gtk errors now.

Since excel files can have multiple sheets, I've added that the factor name must match the sheet name it is read from in excel, let me know if that was dumb and I can separate that back into two fields.

Error messages are now displayed in the commands view if it fails, a good way to do user feedback when it fails, without throwing an exception all the way out.

Permutations should now work correctly with this, was a number of issues to work through on that.

Add a new Move command to the command language as I needed to be able to make sure the generated nodes were beside FactorFromFile, as order can be important when doing experiment factors.

On your topic about having a way for the columns to have non-apsim variable names for the column names. I think the best approach is going to be a child node that has a list of the mappings. I am hesitant to put the remapping as an extra sheet in the excel file, as that would prevent that functionality on csv files, and it also hides the remapping outside of the apsim file. So, I'll do up a small model for that tomorrow.

lie112:
Ah, such a simple add-on! You can see all the challenges I faced getting this working in the approach I implemented. 
Thanks for your work on this. I think it will be super useful, especially for large validation runs that can now run quickly in a single simulation.
A few comments from your response before I pull and try again

- One option is to add the new factors below the FactorsFromFile component which you know in the code from the presenter. This makes clearing to rebuild the tree very easy. When you go to find all the factors in the experiment or permutation you can use a recursive children search and this should maintain the order they are provided in and maintain what you need and avoid needing the move command.
- Can I ask that multiple sheets in the excel file are all considered and allow multiple factors to be added? 
- I use the EnableDisplay and VisibleDisplay method arguments to show properties based on a user value entry. I would make the filename the first property. If the user selects an excel spreadsheet (file with xls or xlsx extension) then it displays the SheetNames property to specify the sheets you want to include as separate factors. A blank entry defaults to all sheets or the only sheet present. Then you don't need the factor name property, and it can be hidden. It would be nice if you could provide a common factor level column name that would apply across all sheets and even have a default value "Name" (ignorecase). Then I could leave the SheetNames blank and enter Name (or ignore if default) as the column label and assume each sheet had Name to define the CompositeFactor name
- My code had got permutations working with very minimal changes elsewhere. 
- I agree with your last point. My assumption was if the specified sheet name for property mappings was not provided it reverted to the original approach, and could use the original approach for csv files. the same functionality could be provided for csv files by having a matching additional csv file with expected extension, or a way of adding the mapping at the top of the csv as identified header rows
- Multiple factors could be supplied across multiple csv files by supporting a wildcard expression in the file name
- Oh I just realised multiple factors breaks your data display. I guess I can add multiple FactorsFromFile components but its not as dynamics and simple.
- Oh, can you manipulate the data display? Is that read in once and saved with the APSIM simulation, or does it get read again from the input file when the simulation is open. Is this why things were not updating? I'd say it must come from the file  or changes would get lost.

lie112:
P.S. I do dislike the limited functionality of csv files!

lie112:
I was looking at some validation setup they are using for stock model validation (see attached) 
This is an interesting case that I was trying to implement in my first attempt. This actually gets to the heart of what I was trying to implement, but I never got to understanding the factor reporting in reports. 

Here we have a nested design of trial (Trial) and treatment (TRT), but the treatments are different for each level of the Trial factor. We want to be able to setup all the individual runs, but also feed values back into the Trial and Treatment factors so they can be correctly reported in reports as actual factor columns. This allows for anlaysis of the data such as average for each trial without having to deconstruct a combined trial/treamtment label in the analysis software before doing the calculations. This would also allow labelling of points on graphs representing the individual treatments. Alternatively, another file might actually have the same treatment level labels for each trial, but even this isn't possible at present. Effectively we need a pseudo permutation style where all the individual simulations are defined, but it knows the labels to use for factor defined columns (Trial and TRT).

image

par456:
Had to spend a bit of time debugging various presenters due to event hooks misfiring, but this seems to be stable again now.

What I've done is we now have two new models, FactorFromFile which does what we talked about before, and FactorsFromFile which is closer to your original use case. FactorsFromFile will read an excel file and create a FactorFromFile for each tab in the excel sheet, and you can either provide a list of tabs to include, or leave it empty to try and run it on all tabs. Haven't done aliasing for variable names yet, but I think that could be put into the FactorsFromFile as a table, will look at that next.

> Here we have a nested design of trial (Trial) and treatment (TRT), but the treatments are different for each level of the Trial factor.

Yeah this gets tricky when you want to send information through apsim, but not something that goes into a model.

In the past when I've done this, I've added a Blank memo to the simulation with the name of the thing I wanted to store (so in this case TRT or DaysOnFeed) then in my experiment factor I put "[TRT].Text=HGWP", then in my output report, I put "[TRT].Text as TRT". 

It's a bit clunky, but it does work. I've played around with the idea of a metadata model to hold this sort of info, and make it with setable variables, but that's a bit of a ways off yet.

lie112:
Thanks, I'll have a look when it builds successfully. 
I was writing some ideas until I get stuck realising something needs to store these values to report them and they are above the level that we create. Could each composite factor also have a list of overarching factor names and values that get propagated when it is created by the FactorFromFile (another user specified entry of level columns), otherwise they would be empty by default and never considered.

These could be injected onto the factor tree values reported in reports before the level that the factor is reporting for itself.
The manager would need to go through all components and add these factors as items in the reporting list, then the component returns multiple values instead of just the one describing the levels of its simulation. 

Where does the report get the factors levels? I assume each factor knows its name for the column name and the current level it is simulating. I did go looking at one point but stopped heading down the rabbit hole.

Enough for now. This space is a never-ending consumer of new functionality requests as the experiment is potentially so powerful in using APSIM linked to real datasets.

jbrider:
Is this going to be looked at again soon? - have a use case for it so was keen to try it out.

par456:
Yep, I'm working on this week and testing it on a big site based simulation I have. Hoping to get it merged this week.

jbrider:
still keen to have a look at this when it's working.

lie112:
Where are we at with getting this implemented so I can do some test runs before we get it included in APSIM?

APSIM.Shared/Utilities/FileUtilities.cs
```diff
@@ -0,0 +1,78 @@
...
+                return propsTable;
+            }
+            else
+            {
+                throw new Exception($"Unable to read '{filepath}', extension '{extension}' not recognised");
+            }
+        }
+
+        /// <summary>
+        /// Get the names of all the sheets in an excen file
```
ric394: excen should be excel

ApsimNG/Presenters/EditorPresenter.cs
```diff
@@ -9,15 +9,13 @@
 
 namespace UserInterface.Presenters
 {
-
-
     /// <summary>
     /// A presenter class for showing a cultivar.
```
ric394: Comment needs updating

ApsimNG/Presenters/EditorPresenter.cs
```diff
@@ -9,15 +9,13 @@
...
     /// </summary>
     public class EditorPresenter : IPresenter, ISubPresenter
     {
         /// <summary>The cultivar model</summary>
```
ric394: Comment needs updating

ApsimNG/Presenters/EditorPresenter.cs
```diff
@@ -9,15 +9,13 @@
...
-        private ILineEditor model;
+        private ICodeEditor model;
 
         /// <summary>The cultivar view</summary>
```
ric394: Comment needs updating

ApsimNG/Presenters/QuadPresenter.cs
```diff
@@ -110,12 +127,65 @@ private void DisconnectEvents()
...
             }
             explorerPresenter.CommandHistory.ModelChanged -= OnModelChanged;
         }
 
         /// <summary>
-        /// The mode has changed (probably via undo/redo).
+        /// Destroys any existing presenters and rebuilds everything depending 
+        /// on the model type.
+        /// </summary>
+        private void createPresenters()
```
ric394: Make method name uppercase

ApsimNG/Presenters/QuadPresenter.cs
```diff
@@ -110,12 +127,65 @@ private void DisconnectEvents()
...
+            catch (Exception exception)
+            {
+                explorerPresenter.MainPresenter.ShowError(exception, overwrite:true);
+            }
+        }
+
+        /// <summary>
+        /// Destroys all the created presenters by detatching them
+        /// </summary>
+        private void destroyPresenters()
```
ric394: Make method name uppercase

ApsimNG/Views/ExperimentView.cs
```diff
@@ -35,31 +57,31 @@ public ExperimentView(ViewBase owner) : base(owner)
         }
 
         /// <summary>Grid for holding data.</summary>
-        public IListView List { get; }
+        public IListView List { get; private set; }
         
         /// <summary>Gets or sets the value displayed in the number of simulations label./// </summary>
         public ILabelView NumberSimulationsLabel { get; set; }
 
         /// <summary>Filename textbox.</summary>
```
ric394: Comment needs updating

Models/Factorial/FactorFromFile.cs
```diff
@@ -0,0 +1,450 @@
...
+    /// 
+    /// | SimName | ColumnLabel | [ModelToReplace]                             | [Model].Property | [Manager].Script.Property |
+    /// |---------|-------------|----------------------------------------------|------------------|---------------------------|
+    /// | Sim1    | First       | [Folder].ModelToCopy                         | 10               | 20                        |
+    /// | Sim2    | Second      | [Folder].ModelToCopy from anotherfile.apsimx | 20               | [SomeModelReference]      |
+    /// 
+    /// This table is converted to a list of APSIM commands that describe everything that must be done to the file in 
+    /// order to create the simulation modifications that you've provided. These commands are run whenever the file is 
+    /// loaded, is run, or the FactorFromFile is refreshed. These generated nodes are not saved into your file, and are 
+    /// recreated whenever the file is openned again.
```
ric394: `openned` is mispelled

Models/Factorial/FactorsFromFile.cs
```diff
@@ -0,0 +1,239 @@
...
+                _filename = value;
+            }
+        }
+
+        /// <summary>
+        /// The column in the data table that is used to label the composite 
+        /// factor that is built from that row.
+        /// </summary>
+        [Description("Include Sheets (Comma seperated)")]
+        public string[] Sheets { get; set; }
```
ric394: comment needs updating

Models/Factorial/FactorsFromFile.cs
```diff
@@ -0,0 +1,239 @@
...
+            List<string> createdNodes = new List<string>();
+
+            List<string> matchingSheetNames = new List<string>();
+            if (Sheets == null || Sheets.Count() == 0)
+                matchingSheetNames = sheets.ToList();
+            else
+                foreach(string sheet in Sheets)
+                    if (sheets.Contains(sheet))
+                        matchingSheetNames.Add(sheet);
+            matchingSheetNames.Reverse(); //do the children in oppersite order
```
ric394: `oppersite` is mispelled here

ric394 (changes requested):
Just comment updates and spelling corrections. Otherwise, it looks good.

lie112:
I thought I received a post that this was ready for testing. 
It throws an "object reference not set to instance of an object" error on creating the FactorsFromFile from adding the component and clicking on it. 

   at UserInterface.Presenters.ListPresenter.Refresh() in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Presenters\ListPresenter.cs:line 90
   at UserInterface.Presenters.ListPresenter.Attach(Object model, Object view, ExplorerPresenter parentPresenter) in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Presenters\ListPresenter.cs:line 53
   at UserInterface.Presenters.QuadPresenter.AddList(WidgetPosition position) in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Presenters\QuadPresenter.cs:line 378
   at UserInterface.Presenters.QuadPresenter.CreateLayoutFactorFromFile() in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Presenters\QuadPresenter.cs:line 451
   at UserInterface.Presenters.QuadPresenter.createPresenters() in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Presenters\QuadPresenter.cs:line 157

I thought it might be that I didn't have spreadsheet open so I opened one and immediately got an "object reference not set to instance of an object" error in PropertyView.cs at line 858 in the IEnumerable<Widget> AllChildren() method.
"Parent was null" in foreach child in parent.AllChildren

par456:
I've just pulled out FactorsFromFile off this branch for the moment to get the FactorFromFile part running and in first, planning to put up a 2nd PR with FactorsFromFile as soon as this goes in, so I suspect that's what you've run into there Adam.

lie112:
I still get the same error. I assume FactorFromFile handles both single factors and multiple factors as I only see the one FactorFromFile in the add model list. I was hoping to be able to use this in our meeting this afternoon as we have to get our validation completed for a central Qld cattle project.

par456:
> I still get the same error. I assume FactorFromFile handles both single factors and multiple factors as I only see the one FactorFromFile in the add model list. I was hoping to be able to use this in our meeting this afternoon as we have to get our validation completed for a central Qld cattle project.

FactorFromFile can only do one sheet at a time, but for the moment you can add multiple for each sheet (until FactorsFromFile goes in)

I've not been able to reproduce that error of yours Adam, do you have a file you can send me to test with?

ric394 (approved):

lie112:
Here is how I can reproduce my FactorFromFile and PropertyPresenter errors.
In the attached cut down file.

1) Click on the FactorFromFile. I get an object not found error in ListPresenter.Refresh where data is null
2) Delete the FactorFromFile and add a new one with Add Model - same error on add.

3) Add a clock to the Simulation with Add Model. Error thrown in property presenter GetAllProperties as the object passed in the argument is null. Exception on obj.GetType().GetProperties(flags)

[experiment.zip](https://github.com/user-attachments/files/29446928/experiment.zip)

par456:
Those should be resolved for you now Adam.

Now to implementing the next part of this work, to do multiple sheets and/or aliasing the column names. 

Is reading multiple sheets in, each to a different factor important, or is it just as easy to define multiple FactorFromFile in your experiment that each point to the same file, but with different sheets? I guess what I'm asking is that how often would a sheet be added to an experiment file, and is that often enough to need an entire automated system to do that, when they could just add another factorfromfile to the apsim file.

I would also recommend having the alias mapping as part of the apsim file/interface rather than a sheet in the excel, unless you have a good reason to keep it in the excel sheet. If its stored in the file, it much more likely to break if a pathing or locator change happens in the future, whereas if its in the file, we can update it with converters.

Was there other functionality that the current version is missing that you originally wanted as well? I wanted to get the basic version in yesterday, so that we can properly extend it now to cover all your use cases.

lie112:
I've got to get my head back around where we're at. I haven't got to play with a version that works yet. 

As long as we can get a factor that creates multiple levels each containing a CompositeFactor of numerous properties based on columns with the values provided in the columns. 

Multiple sheets would be nice in future, but this just automates multiple instances of a the FactorFromFile based on sheets specified (or all if not specified). My thought was that the excel file could drive the entire process by providing the columns to use (properties), sheets to use (factors) and just having one component in the tree. But this was doing everything dynamically in the background. The alternative that just happened in the way I needed to build the tree in my code meant the input files built the experiment in the tree that is able to be saved with the json file. This was updated on run so kept the full experiment used to derive the output and was therefore a permanent record that didn't require the input file to run again if not present by being passed to another user. 

The example I am working with is James' working validation document in excel. There are calculation columns to get things in the right format that should be ignored. This is a working document that I had the information provided to describe how it should be built into an experiment. Its easier to specify columns to include rather than ensure the input file is in the exact clean state.

I often find it difficult to edit what parts of a factorial are going to be included in the experiment other than disabling factors or levels. In a very old version of APSIM all iterations were presented in a list and you could enable or disable particular ones to cut down big runs.  It is really useful to be able to ignore specific combinations of factors and levels in a factorial analysis. The FactorFromFile should handle both separate factors and factorials. This was always a bit clumsy. I had coded it so that nested factors operate as factorials while siblings are separate factors, as this seemed logical to me and matches the approach I use throughout CLEM.

I liked the idea of having an include column that if present needs a true value in the cell for a row to be included.  I haven't quite thought how this could be done for rows unless like the APSIM output it checks if the second row is true/false in a way similar to the units row in output. This means you can edit the spreadsheet to dictate the simulation. I realise this is moving away from transparency in that this information is not included in apsim json, but it's also quite hard to implement in the UI without a lot of time. I like the idea of the simplest approach using csv or very simple excel, that you can add onto with other built-in functionality. I like that the excel spreadsheet is effectively the data and config. I don't see that this would ever need to be changed with converters for new versions, other than a change in a property name. You would need to modify the spreadsheet to implement this.

par456:
> The example I am working with is James' working validation document in excel. There are calculation columns to get things in the right format that should be ignored. This is a working document that I had the information provided to describe how it should be built into an experiment. Its easier to specify columns to include rather than ensure the input file is in the exact clean state.

That'll good to test. It does some filtering of the columns, but I suspect will bring in columns you don't care about. When it does, let me know how it breaks things and I might look into ways to exclude those columns in a sensible way.

> In a very old version of APSIM all iterations were presented in a list and you could enable or disable particular ones to cut down big runs. 

You can still do this if you go to the Experiment node, it will list all the permutations that are to be run, and you can enable or disable them there. Not sure if that is saved to file though.

>I liked the idea of having an include column that if present needs a true value in the cell for a row to be included.

With the correct column name, it should actually be possible to have FactorFromFile disable the composite factor that it build. I'll have a play with that to see if I can get it working and give you an example.

Issue #11038, opened 2026-02-24
Refactoring of Pinus Model


hut104:
### Where is the code located?

pinus.json

### What needs to be changed and why?

1) The model as written is not easily used with parameter sensitivity tools with many parameters included in equations as text.
Expression function models will be refactored to have the parameter values externalised to child constants through which statistical methods can easily change their values. Model outputs should not be affected by these changes.

2) Furthermore, many linear interpolation sets are either superflous, or have more pairs than is required, making it difficult to use for parameter techniques.
Superflous XY pairs will be removed to simplify the functions.  Model outputs should not be affected by these changes.

3) Finally, in many cases, the interpolations could be replaced with other simpler functional forms (e.g. power function) to reduce the number of actual parameters.
Simpler functional forms will be evaluated.  Changes will need to be evaluated against test datasets and endorsed by the reference panel.

PR #11056, opened 2026-03-02
Refactor SpecificLeafArea function in pinus model

hut104:
working on #11038


par456:
Removes expressions from cultivars and adds child constants SLAPotential to allow optimising.

PR #11058, opened 2026-03-02
Add Lizaso Maize dataset.

### LINKED ISSUE #11057: Add in new Maize waterlogging dataset.
### Describe the new feature

Relevant to #10503. This experiment was requested in the September reference panel meeting, and added in #10503, but we would like that PR to reflect the improvement so the data are being added "ahead" of time here.

ilhuber:
Resolves #11057 

This was also added in #10503 but we would like that PR to show improvement.


par456:

New simulations that were created for the waterlogging model, but should be merged in before the water logging model is added in order to show changes.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/10503

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11058/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11058).

This is only adding new simulations and not modifying the model this be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

Issue #11059, opened 2026-03-02
More datasets are available in Australia for soybean phenology


hut104:
### Describe the new feature

These data include data for modern cultivars of industry importance.  Their inclusion will allow work to improve our simulation of the development of these modern cultivars.

PR #11060, opened 2026-03-02
Soybean phenology

hut104:
working on #11059


par456:

New simulations added to Soybean dataset from various trials around Queensland and NSW. Their inclusion will allow work to improve our simulation of the development of these modern cultivars.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11059

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11060/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11060).

This is only adding new simulations and not modifying the model this be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

Issue #11062, opened 2026-03-03
Updated and new Pinus model test data


hut104:
### Describe the new feature

There are additions/corrections to existing obs data in the pinus model test set and a new test data set for SilverCreek.
The obs data should be updated to check that changes to Obs data are desireable.
Then new simulation added to give extra predicted-observed points.
There should be no changes to predictions for existing test sets.

Review of the BFG experiment publications indicate some changes required for soil and management specification.

PR #11063, opened 2026-03-03
Updated Pinus obs data file

hut104:
working on #11062


hut104:
Closing PR because data set has undertaken further work.

PR #11065, opened 2026-03-03
Display a mean value with boxplots.

### LINKED ISSUE #11064: Boxplot mean is NaN.
### What happened?

Boxplot means are NaN.

Image

This is because the BoxPlotItem constructor sets Mean to NaN and we never set this field.

Windows

ilhuber:
Resolves #11064 

Do we want a test for this? The code change is just that we are setting a field that wasn't being set before.


par456:
Not for a view change like this, it's pretty hard to do unit tests for them.,

PR #11073, opened 2026-03-09
4631 Update root front velocity change (version 2)

par456:
Working on #4631

newer version of the same change as the other PR, making ot compare differences.


par456:
@ric394 This is ready for merge. The drops in stats are expected, and should return back up once the next PR is added which addresses those changes.

PR #11076, opened 2026-03-10
Wheat GxExM work

hut104:
working on #10516

This PR just fixes an error in the anthesis dates for the Gatton 2015 trial.  This is just a bug fix and I suggest merging @par456


par456:

A mistake in the Wheat GxExM gatton trials was found, this PR fixes the anthesis dates.

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11076/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11076).

As this is only correcting a mistake in a simulation configuration, this will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #11081, opened 2026-03-11
Store a DateTime with each Operation to speed up date comparison.

### LINKED ISSUE #11080: Operations date comparison can be quite slow.

Models/Management/Operations.cs

I've got an 8 year simulation with ~40 operations as the only management, and checking whether or not each operation triggers on a given date dominates the running time with repeated calls to DateUtilities.GetDate(). This is a SWIM and SoilTemp2 simulation with 18 soil layers, typically one would expect those to take the most time.

As a quick fix, storing a DateTime (with some signal indicating the absence of a year) along with the operation would probably lead to faster comparisons.

ilhuber:
Resolves #11080 

I am using year 1 to indicate a missing year value in the stored DateTime (valid range is 1-9999), but maybe a different approach is better?

### DIFF
--- a/Models/Management/Operations.cs
+++ b/Models/Management/Operations.cs
@@ -18,11 +18,9 @@ namespace Models
     [Serializable]
     public class Operation
     {
-        private const int NO_YEAR = 1;
+        private DateTime? ActionDate { get; set; }
 
-        private DateTime _dt;
-
-        private string _date;
+        private bool HasYear => ActionDate?.Year != DateTime.MinValue.Year;
 
         /// <summary>
         /// Default constructor.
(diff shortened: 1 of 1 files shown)


par456:
I think we can make this a little tidier, but the idea is very good.

We keep the Date property as a string, but underneath we have the DateTime as a private property (using get; set;) and we don't store the string version (it's already in the Line property anyway). Then if you want the date, you get it converted back from a DateTime to the correct string representation (based if it has a year or not).

Instead of defining a 'NO_YEAR' constant, you can instead use DateTime.MinValue.Year instead. That value is still 1, but then we don't need to track it seperately.

I'd also break the single line TriggersOnDate into an if else checking the year date, we try to avoid using shorthand like that for maintainability purposes. It makes it harder to debug a problem when multiple conditions are on the same line.

ilhuber:
> We keep the Date property as a string, but underneath we have the DateTime as a private property (using get; set;) and we don't store the string version (it's already in the Line property anyway). Then if you want the date, you get it converted back from a DateTime to the correct string representation (based if it has a year or not).

To support the previous behaviors of this property we will need to either make the private DateTime property nullable or choose an appropriate value for if the property is unset (this can't be DateTime.MinValue because the string 01-Jan would also get a date of DateTime.MinValue). I'll go with nullable.

PR #11084, opened 2026-03-12
Wheat GxExM

hut104:
working on #10516

Note that this change moves us to specifying traits as experimental factors rather than using individual cultivars.  Therefore, all treatments plant Hartog, but set phenology and phyllochron


hut104:
@par456 I think this can be merged in.
The changes fall into two categories.  
1) We are no longer including the simulations for Minnipa because of complexities at the site.  For this reason, N is reduced for a lot of attributes, and this changes the stats for measurements as well.
2) Fitting on soil water parameters has been undertaken for the southern sites.  This improves the stats for the soil water data at two of the sites.
Again, changes are only made to soil inputs or model apsimx files.  No changes to any model (eg. wheat model) are included with this PR.

par456:

This PR includes updates to the Observed data for the Wheat GxExM validation. 

Changes include:
- We are no longer including the simulations for Minnipa because of complexities at the site. For this reason, N is reduced for a lot of attributes, and this changes the stats for measurements as well.
- Fitting on soil water parameters has been undertaken for the southern sites. This improves the stats for the soil water data at two of the sites.

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11084/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11084).

As this PR is only correcting observed data it will be merged today. Please comment if there is an issue.

PR #11086, opened 2026-03-12
Dookie wheat2024

### LINKED ISSUE #11085: Add UoM experiments to wheat test set
### Describe the new feature

Project lead by James Hunt looking to understand the effect of breeding on biomass partitioning of modern wheat genotypes.  We will need to make changes to the model but the first step is the get the test set into release.

HamishBrownPFR:
resolves #11085


HamishBrownPFR:
@par456 @ric394 please review.  I think this is OK to merge.  Stats have changed in Dookie wheat test because we have changed the way we are configuring these simulations.  Added some new simulations as well.  These tests are still a work in progress but it will make the process easier if we can get current progress into the build.

par456:
Can I get you to add some [gitignore rules for R](https://github.com/github/gitignore/blob/main/R.gitignore) into the .gitignore file. I'm noticing a bunch of temporary R files getting included in this PR.

HamishBrownPFR:
Need to tidy up some things here.  Will re-open under a more accurate name soon

PR #11087, opened 2026-03-12
Properly get number of jobs in information panel

### LINKED ISSUE #11052: The simulations counter includes disabled simulations
### What happened?

The simulation counter displaying (Simulations running 1 of 2]) in the information panel includes some simulations have been disabled in the tree structure. I think this only happens for disabled Simulation components, and seems to respect disabled CompositeFactor components.

Windows

ric394:
resolves #11052

Shows only the actual total number of jobs to be run in the information panel.


par456:
I think this can be done in a much simpler way. Rather than extend IRunnable, how about instead in SimulationGroup, check if the job is of type Simulation, if it is, cast it to a Simumlation and check the enabled status. If it is not, just add it to the list. That'll make this much cleaner.

ric394:
I'll create another PR with a much cleaner implementation just like you suggested @par456.

PR #11088, opened 2026-03-12
Ensure all resource variable references are valid 

### LINKED ISSUE #11046: Wheat component misspelled
### What happened?

Under the wheat model there is a mistake, [Wheat].Grain.WtProportion.LiveWt = [Grian].Wt instead of [Grain].Wt

Windows

ric394:
resolves #11046

Corrects a typo in the Wheat model.

#### Further Updates to PR

This has been expanded into a tidy up of model variable references and the addition of a unit test to ensure the variable references in model resources are valid.


par456:
Huh, I'm guessing this is just used as an output and isn't being used by part of the model or validation, hence no stats changes.

Before we merge this, can you find out why that wasn't throwing a 'could not find' error, and what it was doing in that case? I feel like this REALLY should have thrown an exception.

ric394:
I was thinking it was odd it didn't throw as well 🤔.
I'll investigate.

ric394:
Yep, you're correct. It wasn't used. Either way this should be corrected.

ric394:
I'm writing a test as discussed to test all the resources with the validation simulations to locate the broken variable references. I'll mark this PR for review when this test is submitted.

Issue #11095, opened 2026-03-17
Renaming of Maize organs to better make industry standards


hut104:
### Where is the code located?

Maize.Json and all apsimx files using the maize model

### What needs to be changed and why?

The maize model has multiple organs to simulate the reproductive structure on the plane (ie grain, rachis, husk).
The changes should be as follows:

1) The rachis should be renamed to cobb
2) A new composite organ 'GrainPlusCobb' be added to allow comparison against field data

A converter will be required to rename reporting variables. 
The observed data should be studied to find datasets currently using different ways of accounting for GrainPlusCobb

PR #11098, opened 2026-03-17
Maize organ name changes

hut104:
working on #11095


par456:
I notice you've got 6 Maize.Rachis.Wt measurements that should become Cob.Wt in the future, I assume because it's still Maize.Rachis.Wt in the observed file that will need to be updated.

We also picked up a few Maize.Ear.Wt from the microclimate which must have been previously mislabelled in the observed files there, so that's good.

I think we can merge this and do the data updates separately if you are happy with this now.

Issue #11094, opened 2026-03-17
Extra Datasets for Maize Model Test Set


hut104:
### Describe the new feature

New datasets are available for testing of the maize model.  These datasets should be added to the APSIM test processes to provide extended testing across a wider range of conditions and with modern cultivars.

The datasets to be added include the following
1) 5 cultivars x 5 N rates x 5 Populations at Ames in Iowa, USA
2) 5 cultivars with  differing times to maturity at Ames in Iowa, USA
3) Source-Sink manipulation experiment conducted in Lincoln (NZ), Ames (USA) and  Gois (Brazil)

These datasets have been chosen to target key growth processes requring model improvement.

PR #11099, opened 2026-03-17
11094 maize data

ilhuber:
Working on #11094


ilhuber:
The data has been added, but we maybe should not merge this until the changes from #11098 and #11103 have been accounted for in this.

par456:
The problem it was throwing on before was that there was a factor just called _ which apparently is illegal for a column to be called in sql.

I renamed the factor to _Factor_ on the NZ experiments, and updated the observed data to match. I also switched the simulation from using ExcelInput to Observations to help me debug that problem which I think has derived some additional stat points for you.

ilhuber:
Interesting, something to keep in mind for the future. 

I removed a few of the redundant variables (e.g. Clock.Today.Year), and accounted for the changes in the other two PRs. There are a few composite variables left over that I am inclined to leave for now since there aren't easy [Maize].Whatever values for them.

par456:

New simulations created for development of updated Maize model.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/pull/11094

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11099/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11099).

This is only adding new simulations and not modifying the model this will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #11102, opened 2026-03-17
Put leaf size by position outputs into leaf

### LINKED ISSUE #11101: Add leaf size by position outputs to leaf model
### Describe the new feature

This is common data that is very useful for model testing.  At the moment we construct these variable outputs in a script which is a bit fickle.
The variables should be calculated in the leaf class

HamishBrownPFR:
resolves #11101


HamishBrownPFR:
Mistake made

PR #11103, opened 2026-03-17
Add leaf size by position outputs

### LINKED ISSUE #11101: Add leaf size by position outputs to leaf model

HamishBrownPFR:
resolves #11101


par456:
Added "MaximumMainStemLeafNumber" = 150 to fodderbeet so that the array in Leaf will work, however FodderBeet does not seem to use that property and has no effect on stats.

PR #11111, opened 2026-03-18
Simulation updates

### LINKED ISSUE #11110: Recently merged simulation maize updates

Tests/Validation/Maize/RM and EarManipulation

People would rather the data be in one table.

In the case of EarManipulation there was a definition issue with Stem weight that may cause some small changes in POstats.

ilhuber:
Resolves #11110


ilhuber:
Looks like my initial hope was wrong and we will have to redo the predicted observed tables in these simulations.

ilhuber:
Looks like the POStats system did not run for this most recent push?

par456:
I'll just get you to have a look over the stats here @ilhuber, and if they are correct, I'll merge this.

ilhuber:
> I'll just get you to have a look over the stats here @ilhuber, and if they are correct, I'll merge this.

Looks like this is what it should be, go ahead and merge.

par456:

Updating the new Maize simulations that were added the other day. Observations data has been merged together into a single file, and updates were made to property names.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11110

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11111/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11111).

As this is only moving data around and not modifying the model this will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #11113, opened 2026-03-18
Maize Outputs

### LINKED ISSUE #11112: Extra Maize model outputs
### Describe the new feature

Many simulations calculate common agronomic metrics, and these can include mistakes or inconsistancies - especially in organs are added or removed.  We should add outputs for stover (ie a stover composite organ), harvest index, nitrogen harvest index and grains per ear.
The wheat model has also included simpler output names for leaf tips and ligules.  These could be added for consistency.

hut104:
resolves #11112

### DIFF
--- a/Models/Resources/Maize.json
+++ b/Models/Resources/Maize.json
@@ -1979,6 +1979,87 @@
               ],
               "Enabled": true,
               "ReadOnly": false
+            },
+            {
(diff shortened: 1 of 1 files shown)


hut104:
@par456 @ric394 This PR is ready for review

Issue #11100, opened 2026-03-17
Soybean Height Model


hut104:
### Describe the new feature

The soybean model predicts canopy height with a simple model based on growth stage.  In reality, final crop height is related to crop duration (ie final main stem Node number).  The rate of height increase also changes with growth stage with slower early growth and faster height increase later in the crop.

Data is now available in the test set to evaluate this across a range of cultivars and site-years.

The proposed model will see height increase modelled as growth stage-dependant rate (thermal time basis) modified for water stress.  Early height growth (up to node 6)  proceeds at 0.66 mm/oC.d with a rate of 1.5mm/oC.d thereafter.

PR #11114, opened 2026-03-18
Soybean height

hut104:
working on #11100


hut104:
I think this model change is suitable for release.

Predictions for canopy height are significantly improved when compared to the previous model.  The improved predictions of height also change a large number of other model outputs.  Only a few other outputs are adversely impacted.  In all cases, the changes to other outputs are minor. 

The changes to outputs other than height itself are due to the influence of height on aerodynamic conductance terms within the penman-monteith model in the Microclimate model.  This will cause changes in daily transpiration and therefore water balance.  In most cases, it appears that the changes are beneficial.

par456:

This pull request improves the modelling of the height property on soybean. The proposed model will see height increase modelled as growth stage-dependant rate (thermal time basis) modified for water stress.

The improved predictions of height changes a large number of other model outputs. Only a few other outputs are adversely impacted. In all cases, the changes to other outputs are minor.

The changes to outputs other than height itself are due to the influence of height on aerodynamic conductance terms within the penman-monteith model in the Microclimate model. This will cause changes in daily transpiration and therefore water balance. In most cases, it appears that the changes are beneficial.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/pull/11100

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11114/files).

The changes in stats can be found [here](https://postats2.apsim.info/11114).

We intend to merge this into the release on Monday 23rd unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

PR #11132, opened 2026-03-19
11131 new wl data

### LINKED ISSUE #11131: Waterlogging datasets for Barley and Wheat
### Describe the new feature

We would like to add waterlogged datasets for Wheat and Barley in anticipation of future work. The data source is the same as in #10465.

ilhuber:
Resolves #11131

### DIFF
--- a/ApsimNG/Presenters/QuadGraphPresenter.cs
+++ b/ApsimNG/Presenters/QuadGraphPresenter.cs
@@ -77,7 +77,6 @@ namespace UserInterface.Presenters
             {
                 Water water = null;
                 Physical physical = null;
-                SoilCrop crop = null;
                 if (model is Water)
                 {
                     water = model as Water;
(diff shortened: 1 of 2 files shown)


ilhuber:
Looks like merging #11121 may have thrown off the the PO stats system? Its change is being reflected here as a change from this PR. Similar in #10503

par456:
Yep, something has gone out of order. I'll put in a bugfix for the gui and reset postats on that.

ric394:

This PR adds new Wheat and Barley Waterlogging validation datasets.

The issue has been documented [here](https://github.com/APSIMInitiative/ApsimX/issues/11131)

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11132/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11132).

This is only adding new simulations and not modifying the model this will be merged today. If you think there has been a mistake, please respond so that this change can be reverted.

PR #11145, opened 2026-03-23
Correct the default value of k_AFPV

### LINKED ISSUE #11143: The default value of k_AFPV in NH3Volatilisation is not as intended
### What happened?

k_AFPV is currently defaulted to 300 but it should be 75.

Windows

sno036:
resolves #11143


par456:
I'm surprised changing a default like this doesn't cause a stats change, is this because it only effects a sensibility test currently and not used in other validation?

sno036:
Correct @par456! But merging the change will give users a better picture of what values to use.

Issue #11147, opened 2026-03-24
ColdVrnResponse is not zero when air temperature below zero in the CAMP Model


byzheng:
### What happened?

As Figure 1 in [Wang et al 2025](https://nph.onlinelibrary.wiley.com/doi/10.1111/nph.70651), the cold response is equal to Zero when temperature is below zero.

This figure shows the cold response in the Figure 1 of Wang et al 2025 without dividing 24 h.

<img width="748" height="509" alt="Image" src="https://github.com/user-attachments/assets/82e7c5be-7f37-4ed4-9db8-3f18f75c5b73" />

An UnitTest is setup in the APSIMX master branch to test the cold vernalisation response for wheat with following fixed temperature profile for daily minimum and maximum temperature in three days:

| Scenario | Daily Minimum Temperature (°C) | Daily Maximum Temperature (°C) |
|----|----|----|
| -5/0 | -5 | 0 |
| 0/5 | 0 | 5 |
| 5/10 | 5 | 10 |
| 10/15 | 10 | 15 |

The UnitTest run and generate the hourly temperature, cold response, and the summarised daily cold response.

This figure shows the hourly temperature inteploted by HourlySinPpAdjusted method.

<img width="748" height="509" alt="Image" src="https://github.com/user-attachments/assets/a0b50e82-a46e-422d-baf5-efc66631037f" />

This figure shows the cold vernalisation response for each scenario. The cold response is not zero when temperature below zero (i.e. scenario -5/0).

<img width="748" height="509" alt="Image" src="https://github.com/user-attachments/assets/d983e290-469f-4155-951e-961825c3d633" />

This figure shows the summarised daily cold response for each scenario. The cold response is not zero when temperature below zero (i.e. scenario -5/0).

<img width="748" height="509" alt="Image" src="https://github.com/user-attachments/assets/49c3b141-2ae4-4837-924b-ad19ccfe4cb6" />

### What Operating System are you on?

Windows

PR #11148, opened 2026-03-24
Cold Response returns zero when air temperature below zero

byzheng:
resolves on #11147

Changes in this pull request

* in ColdVrnResponse, return zero if dX is below zero
* Add an unitttest to check the cold response

### DIFF
--- a/Models/Resources/Soybean.json
+++ b/Models/Resources/Soybean.json
@@ -1477,51 +1477,79 @@
               "ReadOnly": false
             },
             {
-              "$type": "Models.Functions.LinearInterpolationFunction, Models",
+              "$type": "Models.Functions.AccumulateFunction, Models",
+              "StartStageName": "Emergence",
+              "EndStageName": "StartGrainFilling",
+              "ResetStageName": null,
+              "FractionRemovedOnCut": 0.0,
+              "FractionRemovedOnHarvest": 0.0,
+              "FractionRemovedOnGraze": 0.0,
+              "FractionRemovedOnPrune": 0.0,
               "Name": "HeightFunction",
               "ResourceName": null,
               "Children": [
                 {
-                  "$type": "Models.Memo, Models",
-                  "Text": "",
-                  "Name": "Memo",
+                  "$type": "Models.Functions.MultiplyFunction, Models",
+                  "Name": "DailyIncrease",
                   "ResourceName": null,
-                  "Children": [],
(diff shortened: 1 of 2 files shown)


HamishBrownPFR:
@byzheng @par456 @hut104.  I am not sure this test is working properly.  The stats are showing changes to the soybean stats.  In particular a large change to height predictions.  The obs pred graph for this looks like the changes Neil showed us last week.

HamishBrownPFR:
@byzheng good catch.  @par456 @ric394 @hut104 this error needs to be corrected.  Excepting the confusion over soybean stats (see comment above) this change only has minor impacts to stats.  Most are blue ticks.  The red crosses are all small and mostly on variables of secondary importance.  Importantly, the HeadingDate, FloweringDate and FLN tests in the Phenology set are not impacted.

byzheng:
Hi @HamishBrownPFR The soybean height is confusing to me as I cannot find soybean model to use ColdVrnResponse.

par456:
Ah, I think that PR for soybean height went in while this PR was running. Will update to master and rerun and that should fix the stats.

par456:
Yep, soybean stats are the same now. 

@hut104 you might want to have a think about this change as well before it's merged.

Models/PMF/Phenology/CAMP/ColdVrnResponse.cs
HamishBrownPFR, line 1: This change makes sense

HamishBrownPFR:
@par456, thanks for confirming the issue with the soy tests.  The stats make more sense now.  
The most red crosses are in the Pask test which makes sense as that is the coldest environment we have in the test set.  All reductions in stats are trivial.
Reductions in stats in other sets are also trivial.

My recommendation is to merge this PR as it fixes a fault in the code and only has trivial affects on the test stats.

par456:


A bug was found in the ColdVrnResponse allowing it to go negative when the temperature dropped below 0, it is now capped at 0 to prevent it going negative.

Has some small stats changes in simulations in cold environments.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/pull/https://github.com/APSIMInitiative/ApsimX/issues/11147

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11148/files).

The changes in stats can be found [here](https://postats2.apsim.info/11148).

PR #11153, opened 2026-03-25
Add a switch for windspeed to drive volatilisation

### LINKED ISSUE #11152: Volatilisation should be driven by windspeed not Eo
### What happened?

The history behind the code to use Eo is 'lost to memory' and also makes little sense. As an interim measure add in windspeed as well with parameters to allow testing.

Windows

sno036:
resolves #11152 

This is to allow testing to proceed with a switch for the testers to switch between windspeed and Eo as a diver. One of these methods will eventually be removed.

### DIFF
--- a/Models/Soils/Nutrients/NH3Volatilisation.cs
+++ b/Models/Soils/Nutrients/NH3Volatilisation.cs
@@ -98,6 +98,10 @@ namespace Models.Soils.Nutrients
         [Description("Factor for soil/atmosphere gas exchange (AFPV/mm):")]
         public double k_AFPV { get; set; } = 75.0;
 
+        /// <summary>For testing - apply k_AFPV to windspeed? If false then apply to Eo</summary>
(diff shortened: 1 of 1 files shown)


par456:
I have a couple of questions on this one.

- NH3Volatilisation is attached to Nutrient in the file I had a look at, yet its class is listed as having Soil as its allowed parent. Not sure what happened there, but you might want to check if that was intended or not.
- Despite this being for a prototype model, I am still a bit uncomfortable adding a flag like this into the model, these have a bad habit of sticking around. We should really provide the function a child IFunction that is calculated from an expression instead and keep the built in calculation as the default if there is no child.

sno036:
Thanks @par456 - some tidying will be needed once testing progresses. On your first question this is likely a legacy issue related to if volatilisation should be under Nutrient or directly under Soil - this has implications for the patching mechanism. Let's resolve that as it gets closer to being releasable. And for the second, Jody is pulling from my branch now so we don't need this merged. I'll close it after this comment.

sno036:
Closing as not needed.

PR #11157, opened 2026-03-26
11142 Ag Pasture mass balance fix

### LINKED ISSUE #11142: APSIM failing mass balance tolerance tests under Linux but passing under Windows
### What happened?

AgPasture (like I assume many models) has some internal mass balance tests. Under we are finding (under r8012 but this may not be relevant) that there are some simulations that are failing the mass balance tolerance test when run on our HPC (Linux) but when we run the same simulation under Windows it is passing.

Has anyone else experienced this?

My solution will be to relax the tolerance a bit for the Linux simulations and then compare the outputs to those under Windows with the default tolerance - but first I wanted to see if this is more widespread than AgPasture.

Linux

par456:
Resolves #11142

The problem ended up being in the AgPasture root model, where a looping increment of two doubles could cause a value that was supposed to go to 0, to end up not quite at 0. Linux seems to be more sensitive, and then over time it caused the value to get more and more out of sync until it broke the mass balance.

We'll see what this does to the validation, or if further work is needed.

### DIFF
--- a/Models/AgPasture/RootTissue.cs
+++ b/Models/AgPasture/RootTissue.cs
@@ -116,10 +116,10 @@ namespace Models.AgPasture
                 for (int layer = 0; layer < dmByLayer.Length; layer++)
                 {
                     dmByLayer[layer] -= DMTransferredOut * prevRootFraction[layer];
-                    if (MathUtilities.FloatsAreEqual(dmByLayer[layer], 0))
+                    if (MathUtilities.FloatsAreEqual(dmByLayer[layer], 0, 0.000000000001))
                         dmByLayer[layer] = 0;
                     nByLayer[layer] -= NTransferredOut * prevRootFraction[layer];
-                    if (MathUtilities.FloatsAreEqual(nByLayer[layer], 0))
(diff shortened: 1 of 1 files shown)


par456:
@sno036 

This seems to have solved the issue. 

There are some minor stats changes as I found it took simulation time (roughly 6 months from the first time it happened) for the error to build up enough to throw off the mass balance, meaning some of the simulations may have had this occurring but not enough to be detectable. 

If you're happy with this I'll get Julian to review and merge it in.

sno036:
Very happy with that @par456! Thanks for picking up the cause of the error. Stats are fine (slightly better than they were).

ric394:

A bug in AgPasture was found that was caused by floating point precision that effected stats. This caused a minor improvement in stats.

The issue has been documented [here](https://github.com/APSIMInitiative/ApsimX/issues/11142)

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11157/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11157).

PR #11165, opened 2026-03-30
11164 Remove partial 1995 data from Ames met file

### LINKED ISSUE #11164: AmesSoilTemperature met file is non-contiguous
### What happened?

In the work for auditing metfiles, I have found a mistake in the AmesSoilTemperature file where values from 1995 were append after 1990. The simulations did not use the 1995 values as they finished at the end of 1990, however because the file did not provide a Tav and Amp value as constants, they were calculated using the partial 1995 data causing some weirdness.

I need to address this invalid metfile prior to finishing the work on the metfile PR, so am doing this change separately since it will effect the stats.

Windows

par456:
Resolves #11164 

Removes the 8 months or so that was appended to the end of the file and causing Tav and Amp miscalculations. No tests needed as this sort of error will be checked with the changes in #11140

Will change stats as this was a bug in the validation.


ric394:

This PR removes non-contiguous weather data that was causing incorrect TAV and AMP value calculations.

The issue has been documented [here](https://github.com/APSIMInitiative/ApsimX/issues/11164)

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11165/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11165).

PR #11170, opened 2026-03-31
Add flag leaf data for Temora

hut104:
working on #10516


hut104:
@par456 and @ric394 
The change simply adds three observations to a single experience in the GxExM test set.
The stats are different because N has increased by 3 and the new data are added to show that the final leaf number was lower than predicted (therefore the goodness of fit is reduced).
A following PR will show changes to the parameterisation to address this.
I suggest merging this change if you agree.

ric394:

This PR adds 3 additional observations to a single GxExM datasets.

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11170/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11170).

As this PR is only adding data it will be merged today. Please comment if there is an issue.

Issue #11193, opened 2026-04-10
Restructure Livestock Tests


hut104:
### Where is the code located?

Stock.apsimx

### What needs to be changed and why?

We now have test datasets for beef, dairy and sheep.  

1) With many people working on this same file, risks of merge conflicts increase
2) statistical analysis is difficult when all three are combined. It would be good to know how each type are simulated
3) Statistical acceptance tests will be undermined if retained together.  e.g. a decrease in predictive capacity for lamb growth may be not picked up when stats are dominated by data for beef cattle which are an order of magnitude larger.

PR #11194, opened 2026-04-10
Livestock test restructure

hut104:
working on #11193


hut104:
Will try a different approach using OBS Pred tables within a structured file

PR #11199, opened 2026-04-13
Disabled Children Stay Disabled When Replaced

### LINKED ISSUE #9343: Disabled children of Model factors get enabled when running an experiment.
### What happened?

When running an experiment in which a model is being replaced by other models via a factor, any disabled children of the replacement models will be enabled before the simulation is run. For instance, a Soil with disabled CERESSoilTemperature and Enabled SoilTemperature children will have the CERESSoilTemperature re-enabled (and possibly take precedence if higher in the simulation tree). 

I've attached an example simulation demonstrating the behavior. Culprit code can be found at 
https://github.com/APSIMInitiative/ApsimX/blob/8ab38c5f076cd9e035fba9bccf02e00b28306bcd/Models/Core/Model.cs#L84-L91
and 
https://github.com/APSIMInitiative/ApsimX/blob/8ab38c5f076cd9e035fba9bccf02e00b28306bcd/Models/Core/ApsimFile/Structure.cs#L215

[FactorExample.zip](https://github.com/user-attachments/files/17120679/FactorExample.zip)

Windows

### LINKED ISSUE #11191: One can sow a disabled cultivar
### What happened?

If you have a cultivars folder in replacements or a resource with a cultivar folder, and one of the cultivars is disabled, you can still sow that cultivar and the model will run.

Seems counter intuitive.

Windows

ric394:
resolves #11191
resolves #9343

This fixes an issue where replaced model's children would have their Enabled property set to true. 

Instead of changing this behaviour in `Model.cs` it makes sense to do it when the replacement occurs to mimimise unintended edge cases.


APSIM.Core/CommandProcessor/Commands/ReplaceCommand.cs
```diff
@@ -98,13 +98,35 @@ INodeModel IModelCommand.Run(INodeModel relativeTo, IRunner runner)
...
+
+        var originalChildren = originalModel.GetChildren()?.ToArray();
+        var replacementChildren = replacementModel.GetChildren()?.ToArray();
+
+        if (originalChildren is null || replacementChildren is null || originalChildren.Length == 0 || replacementChildren.Length == 0)
+            return;
+
+        int childCount = Math.Min(originalChildren.Length, replacementChildren.Length);
+        for (int i = 0; i < childCount; i++)
+            CopyEnabledStateRecursively(originalChildren[i], replacementChildren[i]);
```
par456: This is only matching on index number, not if the children models actually match. Needs to match up all the children and skip children that don't have a match.

par456:
what was the change in canola?

ric394:
> what was the change in canola?

@par456 I thought there was a meaningful change, but I was mistaken. I've removed it now.

PR #11210, opened 2026-04-16
11054 Fix Composite Factors in Validation

par456:
Working on #11054 

When working on #11208 I found several composite factors that were set up in a dangerous/illegal way in our validation. Under the new model these would throw, but since fixing them requires editing of the simulation, I wanted to do this first in a separate pull request.

Only the change to Ryegrass should effect stats, the others should be functionally the same.

Changes made:
- SPRUM: Added a specification to a composite factor that had no specification. The specification added is just to replicate what the base simulation does (which is an empty date field in a manager script)
- Chickpea: CompositeFactor was overwriting a Sowing manager with an Operation. Removed the Manager and replaced it with an empty Sowing Operation that can be replaced.
- WheatDamageFunction: CompositeFactor was overwriting a Sowing manager with an Operation. Removed the Manager and replaced it with an empty Sowing Operation that can be replaced.
- Ryegrass: Composite factor was overwriting an Irrigation manager with an Operation of fertilisation operations. Changed the factor to instead overwrite the Fertilisation Operation (like similar experiments alongside it). May cause a stats change.
- Lucerne: Added a specification to a composite factor that had no specification. The specification added is just to replicate what the base simulation does (which is an empty Operation). Also cleaned up the base simulation a bit so it better reflects the first options in the experiment setup.


par456:
Copilot doesn't really understand the apsim json format, these notes aren't really things I would act upon.

ric394 (approved):

ric394:

This PR primarily corrects the behaviour of Experiment Composite Factors. 
This PR also effects Ryegrass stats after fixing an error in its prototype simulation.

The issue has been documented [here](https://github.com/APSIMInitiative/ApsimX/issues/11054)

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11210/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11210).

As this PR is fixing simulation errors and model bugs, it will be merged today. Please comment if there is an issue.

Issue #11200, opened 2026-04-13
Supplement Presenter needs to be Generalised


ric394:
## Phase 1 Changes/Fixes

Supplement uses a custom presenter and view which looks to have been ported from older software.
To decrease the maintenance, this class should be changed to use existing views and presenters.
The view and presenter also have other issues and simplifying this should help resolve these issues.

- [x] Make StoreTypes children of Supplement
- [x] Make Supplement use GenericPresenter
- [x] Type up the summary remarks in Supplement.cs (for view)
- [x] Hide Ash property
- [x] Hide Passage Rate Property

## Phase 2 Changes/Fixes

- [ ] Make Supplement StoreType property `IsRoughage` show dropdown control properties `roughage` and `concentrate`.
- [ ] Replace sulphur property and calculate using GrazPlan functions to simplify supplement usage.
- [ ] Replace phosphorus property and calculate using GrazPlan functions to simplify supplement usage.
- [ ] Replace ADIP property and calculate using method James Dougherty used to calculate
- [ ] Allow StoreTypes to be addable classes rather than classes created from supplement.txt within SupplementLibrary.cs to mimic old behaviour of SupplementView.cs

PR #11213, opened 2026-04-16
Simplified Supplement View

ric394:
working on #11200

Makes changes to the model structure to allow simpler PMF-style structure to Supplement rather than use a bespoke view to make it look like a GrazPlan product UI. 

This also has the advantage of removing some UI bugs/quirks when making supplement changes.

Some Supplement class refactoring has occurred to tidy up some files and to make it more compliant with C# standards.

This may need some real-world usage to make sure it behaviours as intended.

| New Explorer View Structure and Supplement property view | Old Explorer View Structure and Property View |
| :--------------------------------------------------------------------: | :-----------------------------------------------------: |
| image | image |
| image | image |


par456:
I suspect you'll need to open up stock on both an older version of apsim, and the one on this branch and look at an experiment that has moved in the stats. Just check that the values being reported in the supplements has converted across correctly.

Otherwise you'll have to dig further into the work that was refactored and work out where the difference is coming from.

ric394:
> I suspect you'll need to open up stock on both an older version of apsim, and the one on this branch and look at an experiment that has moved in the stats. Just check that the values being reported in the supplements has converted across correctly.
> 
> Otherwise you'll have to dig further into the work that was refactored and work out where the difference is coming from.

I spent the better part of yesterday trying to track down the location of the changes in this manner. I haven't substantially changed the inner workings of the Supplement or related models, so I'm scratching my head as to the cause. 

I'm seeing that supplement is still being applied to paddocks at the correct rates however the amounts they're uptaking are reduced which is impacting that stat, as well as the weight related stats.

I'll keep working on this.

PR #11223, opened 2026-04-27
11222 Throw on Duplicate Cultivars

### LINKED ISSUE #11222: Duplicate Cultivars in Wheat resource
### What happened?

Doing some debugging on Wheat and discovered that we have the same cultivars listed in multiple folders, being Cesario, Stockade and Whistler.

Windows

par456:
Resolves #11222 

Add in check to throw when a cultivar is defined multiple times. Which it was previously possible to 'override' a cultivar by defining it again with the same name in the simulation, this will not throw and require users to either use the Replacements folder to do the override, or give their cultivar a different name so it isn't confused with the Release cultivar.

This is also important to prevent name conflicts in the resource system and the wrong version of a cultivar getting run or referenced accidently.

Already found a duplicate placeholder Rockstar defined in the FAR validation that has been removed, however other parts of the validation might need to be fixed to solve this correctly.


par456:
@HamishBrownPFR 
In the Barley model, we have two definitions for Dash, one in the New Zealand folder and one in the Australia folder. The Australia Dash is supposed to be used in the NPIField experiments, but currently its using the New Zealand one.

Can you have a quick look and advise on what should happen to resolve this conflict?

par456:
A note later for @hut104 as well, in Lucerne we have "southern_special" defined as an alias for both FD5 and FD7. FD5 is a cultivar with no changes, while FD7 has cultivar changes.

In the LucerneValidation, southern_special was used only in the Griffith experiment, so its just a matter of if that was supposed to be using FD5 (Lucerne with no cultivar changes) or FD7.

ric394:
keep open GitHub

Issue #5197, opened 2020-05-25
Rice model developement


andomariot:
The development of a rice model using the plant modelling framework of APSIMX is a project led by USQ in collaboration with different partners that have been working in the development of APSIM-ORYZA. The project is open to partnership and is partially supported by the broadacre cropping program of the DAF and USQ partnership. 

This project is looking to provide a common platform for rice modeling science community  to exchange and  strengthen collaboration across research institutions. It is open to contribution and inputs in addition of  data, consultation, test and use.

 The proposed plan of development is composed of three steps:

1- Compiling existing datasets for the model development and validation and identifying gaps where additional data may be needed. Among these gaps are rice crop establishment under direct seeded rice,  phenology and  biomass partitioning responses to stresses, genotypic variability of cold tolerance, rice quality. To start,  a  dataset of reference for upland rice from the Northern Queensland Rice project will be used  and  the set of lowland rice data used by Gaydon et al, 2017 in APSIM-ORYZA evaluation . 

2- Describing the rice plant phenology and growth in aerobic condition using the Barley crop template for phenology and biomass production.  The approach would be to calibrate a crop reference plant to mimic a rice plant then define any processes where the two crops differ and implement the needed change of representation . At this stage new representations that need to be designed specifically for rice  crop are  : phenology, respiration and photosynthesis process, stress responses.  

3 - Describing water balance and nutrients dynamics in ponded system and Revisiting validated formalism under aerobic rice from  2 to account for factors driving variability of  rice crop growth and development  under ponded conditions among these are the rice pond process modeling, the transplanting shock period with biomass dilution and phenology delay, water and nutrients balance.

This project on development of rice model in APSIMX  will open new opportunities to account for model improvement from the existing gaps limiting previous use of the APSIM-ORYZA model. The APSIM-Next Gen platform will facilitate the use of the model in large scale simulations and to support decision making tools development using  mobile and web based applications that are for use for researchers, extensions agents , farmers and as well policy makers. 

For more information, update and progress, please reach me out ando.radanielson@usq.edu.au

LenLon:
What's the current state of this? We might have resources for development, as we desperately need a rice model soon.

HamishBrownPFR:
@hut104 @hol430 @par456 @ric394.  Do you guys know if anyone is actively working on the rice model or if there are plans to get it moving again?

hol353:
[...]

PR #11225, opened 2026-04-28
Rice Model prototype

hut104:
working on #5197


par456:
This work will be replaced by #11233, but we need to capture it first.

PR #11232, opened 2026-04-29
Allow linux distro users to see apsimx files

### LINKED ISSUE #11231: Linux users cannot see apsimx files from file dialog
### What happened?

The file dialog on linux systems doesn't work with the current file type filters.
This needs to change to handle it for different operating systems

Windows

ric394:
resolves #11231

Linux users should now be able to see apsimx files again in File Dialog windows.

This has been tested on Ubuntu only. I would appreciate if anyone had a moment to test on another Linux distro.


par456:
Does the dialog on linux still show both apsimx and json?

ric394:
> Does the dialog on linux still show both apsimx and json?

Yes, it would I'll fix that quickly.

Issue #11246, opened 2026-05-05
Incorrect Value for Energy Content of Rumen Undegradable Protein in Fermentable Energy Calculations


DocJamesD:
### What happened?

While writing the ‘plain language explainer’ for how the Freer model (as defined in the 2012 report) calculates feed intake, I noticed that the coefficients for calculation of fermentable energy were rather odd but also that it said to subtract off the energy in rumen undegradable protein but did not provide a formula for that either. Looking in the APSIM code to see how it approached it, I identified an error in how APSIM (and AusFarm) calculates fermentable energy – it uses 14 mj/kg as the energy content for undegradable protein but that is about half of the actual value. I then reached out to Hutton Oddy in case he had any insight as to how hat coefficient was derived and he agreed that it was an incorrect value. He thought it maybe might come from an assumption that the rumen undegradable protein was ~60% digestible but even so if that was the case:

a.	the digestibility of RUP is calculated elsewhere in the model
b.	even if it was only 60% digestible on average, it’s digested in the lower gut and therefore wouldn’t be potentially fermentable in the rumen regardless so it’s still not the correct value
 
My recommendation is that this value should be updated to the 23.6 MJ/kg value used elsewhere for protein. The fat value is fine in the code but incorrect in the Freer paper, though the 36 MJ/kg value should probably also be changed to 39.3 MJ/kg, for consistency.  

To do this, I suggest changing GrazType.ProteinE2DM from 14 to 23.6 and GrazType.FatE2DM from 26 to 39.3. See image below for where these terms are called in the code. 

<img width="1111" height="521" alt="Image" src="https://github.com/user-attachments/assets/57f685e1-b8d9-4a98-9b08-150b6d8af154" />

### What Operating System are you on?

Windows

hut104:
@DocJamesD 
The relevant code is found here
https://github.com/APSIMInitiative/ApsimX/blob/64a4d2e6fe3b51255604fee9f840679157ad7cef/Models/Stock/graztype.cs#L739

Note: To add a bit of complexity here, the value in APSIM for FatE2DM is currently 36, not 26.  There may be a difference here between the models here already.

DocJamesD:
The 36 is what I meant to type for fat, sorry; 26 is a typo. However, the Freer report has the value for fat at 23.4, but the code has 36, which is at least closer to the 39.3 that it should be. So there is already a difference in the report vs the code, and the report lacks values for protein for this adjustment as well.

PR #11247, opened 2026-05-05
Constants change for converstion factors for fat and protein to energy

hut104:
working on #11246

### DIFF
--- a/Models/Stock/graztype.cs
+++ b/Models/Stock/graztype.cs
@@ -736,12 +736,12 @@ namespace Models.GrazPlan
         /// <summary>
         /// Conversion factor
         /// </summary>
-        public const double FatE2DM = 36.0;
+        public const double FatE2DM = 39.3;
 
         /// <summary>
         /// Conversion factor
         /// </summary>
-        public const double ProteinE2DM = 14.0;
+        public const double ProteinE2DM = 23.6;
 
         // Various constants with biological meanings
 


sno036:
Hi @hut104 - these are pretty big changes. I assume (while you are there it would be good to add some units/description!) that these are DM that is fat/protein and the conversion of DM to energy in MJ? That value seems rather high to me for fat and that is nearly double the value for protein (and I don't have sensible values in my head for protein). Are you incorporating some sort of efficiency in these that was not in the original?

DocJamesD:
@sno036 The units are MJ/kg DM. The values for fat and protein been changed to that used by SCA (2007) for the energetic content of fat and protein gain/loss in animals.  These are also derived from the average enthalpies of fat and protein, converted from Mcal to MJ. We decided that are also suitable for estimating the energy density (MJ/kg DM) of protein and fat in feeds - this is the method Hutton Oddy and I were using in the with our modelling efforts (albeit with slightly higher numbers from Jim Oltjen's maths, 23.8 for protein and 39.6 for fat) but I decided to use 23.6 and 39.3 for consistency with the rest of the model.  

The SCA (1990) book doesn't use fermentable energy, it calculates things from OM apparently digested in the rumen (OMADR) or digestible OM intake, or from MD (MJ ME/kg DM) but without adjustment for fat or RUP, so this was a change in the logic/formulation between the 1990 and 2007 version. Unfortunately the 2007 SCA doesn't define FME via an equation, it just says MEI minus the energy contained in fat, UDP, and silage acids, but doesn't provide coefficients for estimating those values. The 2007 SCA book does provide a regression-based approach for estimating MD from DMD and  fat (0.134 DMD + 0.235 EE + 1.23) which is similar to Eqn 32 in the Freer 2012 report but I can't find a value for RUP there either.  U

I did reach out to Hutton to ask if he knew where the 14 value came from for protein and his theory was that it was possibly the 23.6 value with an estimate of 60% digestibility in the lower gut, but agreed that that wouldn't really affect fermentable energy in the rumen per se and said to use the higher values.  He said that unfortunately it's hard to tell where Freer got the values from but in his opinion the protein value in particular was an underestimation. However, if there's more information as to why Freer took that approach that would be great - I have been asking around and trying to do some sleuthing as to where the numbers came from but it's not clear. 

I am also working on a broader document unpicking the Freer intake model and exploring the coefficients/structure and trying to figure out where the different coefficients came from. It's a work in progress but I'm making good headway; I've had to set it aside for a few weeks due to an MLA deadline at the end of the month but have been working through the equations and doing some sensitivity analyses that I think folks will find useful.

PR #11272, opened 2026-05-12
Root depth should not be greater than the actual maximum potential depth

### LINKED ISSUE #11271: Error in the logic of setting the maximum rooting depth in AgPasture plants
### What happened?

There is a gap in the logic used in setting the maximum rooting depth where, in some cases, a deeper depth could result.

@rcichota 

Windows

sno036:
Fixing a bug in the logic of setting the maximum rooting depth.

@rcichota 

resolves #11271

### DIFF
--- a/Models/AgPasture/PastureBelowGroundOrgan.cs
+++ b/Models/AgPasture/PastureBelowGroundOrgan.cs
@@ -297,7 +297,7 @@ namespace Models.AgPasture
                 { // root depth limited by some soil issue
                     if (z > 0)
                     {
-                        MaximumAllowedRootingDepth = soilPhysical.ThicknessCumulative[z - 1];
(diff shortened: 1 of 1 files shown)


sno036:
The differences are minor and this is fixing an error. Can this be merged please?

par456:

This PR resolves a bug in AgPasture where the Maximum rooting depth property was always overridden instead of using its given value. Minor stats changes were caused as the rooting depth is now being set correctly.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11272

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11271/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11271).

As this is only a small bugfix, it will be merged today. Please comment if there is an issue.

PR #11275, opened 2026-05-13
11041 report name collisions

### LINKED ISSUE #11041: Name collisions in different reports can unexpectedly throw off expressions.

VariableExpression method FillVariableNames(), Locator method GetInternal()

When filling variables in a VariableExpression that belongs to a report column, variables from different reports can be added if they share a name. As an example, I might have a variable defined in a daily report and a variable with the same name in an annual report (say b), and subsequent column in that annual report that is an expression involving this variable (say b+1 as a). The locator will find whichever b comes earlier in a depth first search of scoped IVariableSuppliers rather than the 'nearest'.

This can produce unexpected results in simulations with several reports containing many different column definitions. It may be better to disallow distinct reports from being able to get each others variables.

ilhuber:
Resolves #11041 

The MockStorage class may be worth looking into a little, it can't seem to handle multiple tables right now.

### DIFF
--- a/APSIM.Core/Locate/Locator.cs
+++ b/APSIM.Core/Locate/Locator.cs
@@ -185,25 +185,11 @@ internal class Locator
         //if it doesn't start with those characters, it might be a report variable, so we need to search for that in the report columns
         else if (includeReportVars)
         {
-            // Try to avoid reading other report's values if we are a report ourselves.
             if (relativeTo.Model is IVariableSupplier supplier && supplier.Get(namePath, out object supplierValue))
             {
                 var comp = new VariableComposite(namePath);
                 comp.AddInstance(supplierValue);
                 return comp;
-            }
-            // Try a variable supplier.
-            foreach (IVariableSupplier variableSupplier in relativeTo.WalkScoped()
-                                                                     .Where(n => n.Model is IVariableSupplier)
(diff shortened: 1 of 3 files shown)


par456:
Does this completely block a report from reading the variables from another report? (if so, that's good, I don't want to allow that)

Could I also get a unit test showing it throwing an error for that test case

ilhuber:
It didn't, but I have removed that functionality altogether. Looks like there was just one simulation in the validation set that was impacted (in a non PO reported variable).  

The original test I wrote was a bit superfluous with the behavior removed, so I updated to test if an error is thrown when the variable is in another report.

par456:
With the locator change, I notice it has an outer if for "IncludeReportVariables"

Would it be cleaner to just set that to false whenever the asking object is a report? Never sure on that locator code, it's such a tricky thing to edit.

ilhuber:
That comes from a LocatorFlag (LocatorFlags.IncludeReportVars). This flag might be vestigial, it seems like the only areas of APSIM that set it are downstream of being within a Report.

par456:
Interesting, I suspect this is a feature someone has added at some point but only used once or twice. Good reason to remove it though considering it adds a lot of complexity for little gain.

Right, I'll merge this for now then and might come back to look at it again next time I have to deep dive the locator.

PR #11276, opened 2026-05-14
Move models up and down in GUI via Nodes rather than using children d…

### LINKED ISSUE #11263: Cannot reorder stack panel series
### What happened?

I am using stack panel in the Graphs. The order is important for displaying the graph correctly. The stack graph does not update when you reorder the series in the graph in the UI tree. This is shown in the legend which also does not represent the order items are provided after moving them around. This can be fixed by deleting and adding the series in the correct order but has to be repeated to add a new series.

Windows

zur003:
…irectly.

Resolves #11263


par456:
I don't think there are any other displays that change based on their order, at least not off the top of my head.

PR #11289, opened 2026-05-20
11284FixAddCommandResourceBug

### LINKED ISSUE #11290: Add a plant model to simulation using a command
### Discussed in https://github.com/APSIMInitiative/ApsimX/discussions/11284

<div type='discussions-op-text'>

<sup>Originally posted by **BrianCollinss** May 19, 2026</sup>
How can I add a Plant model to a simulation using a command like this (which fails)?

`add new Wheat to [Simulations].Base.Field name Wheat`

error: `Unknown model type Wheat`

And how to assign a Plant model to a Manager Script's Crop parameter of type IPlant?

`[Sowing].Script.Crop=[Wheat]` and `[Sowing].Script.Crop=Wheat` don't work!

**UPDATE**

Answer to the second question: `[Manager].Script.Crop.Name=Wheat`</div>

ric394:
resolves #11284
resolves #11290

This fixes a bug where users were unable to add a resource using the command line add command.

A unit test has been included to exercise the bug fix.

### DIFF
--- a/APSIM.Core/ModelRegistry/ModelRegistry.cs
+++ b/APSIM.Core/ModelRegistry/ModelRegistry.cs
@@ -40,15 +40,16 @@ internal class ModelRegistry
         if (typeToCreate == null)
         {
             // Find and return a resource.
-            var resourceName = modelsAssembly.GetManifestResourceNames().
-                FirstOrDefault(r => r.Equals($"Models.Resources.{modelNameToCreate}.json", StringComparison.InvariantCultureIgnoreCase));
+            string resourceName = modelsAssembly.GetManifestResourceNames().
+                FirstOrDefault(r => r.Equals($"Models.Resources.{modelNameToCreate}.json", 
+                StringComparison.InvariantCultureIgnoreCase));
             if (resourceName != null)
                 return CreateResourceModelFromName(modelNameToCreate, resourceName);
         }
-        if (typeToCreate == null)
-            throw new Exception($"Cannot find a model with name: {modelNameToCreate}");              
-        var model = (INodeModel)Activator.CreateInstance(typeToCreate, true)
-                    ?? throw new Exception($"Cannot create a model of type {typeToCreate.Name}");
+
+        INodeModel model = (INodeModel)Activator.CreateInstance(typeToCreate, true);
+        if (model == null)
+            throw new Exception($"A {typeToCreate.Name} model could not be found or could not be created.");
         return model;
     }
 
@@ -60,8 +61,9 @@ internal class ModelRegistry
     /// <returns>The newly created instance.</returns>
     private static INodeModel CreateResourceModelFromName(string modelNameToCreate, string resourceName)
     {
-        using var stream = modelsAssembly.GetManifestResourceStream(resourceName)
-            ?? throw new Exception($"Could not find model or resource with the name: {resourceName}");
+        using var stream = modelsAssembly.GetManifestResourceStream(resourceName);
+        if (stream == null)
+            throw new Exception($"Could not find model or resource with the name: {resourceName}");
         using var reader = new StreamReader(stream);
         string json = reader.ReadToEnd();
         
@@ -74,7 +76,7 @@ internal class ModelRegistry
         INodeModel resource = FileFormat.ReadFromString<INodeModel>(jObject.ToString()).Model.GetChildren().FirstOrDefault();
         
         List<INodeModel> children = resource.GetChildren().ToList();
-        foreach (var child in children)
+        foreach (INodeModel child in children)
             resource.RemoveChild(child);
         return resource;
     }
--- a/APSIM.Core/ModelRegistry/ModelRegistry.cs
+++ b/APSIM.Core/ModelRegistry/ModelRegistry.cs
@@ -21,7 +21,7 @@ internal class ModelRegistry
     internal static Type ModelNameToType(string modelNameToCreate)
     {
         DiscoverModels();
-        var modelTypes = ReflectionUtilities.GetTypeWithoutNameSpace(modelNameToCreate, modelsAssembly);
+        Type[] modelTypes = ReflectionUtilities.GetTypeWithoutNameSpace(modelNameToCreate, modelsAssembly);
         if (modelTypes.Length != 1)
             return null;
         Type typeToCreate = modelTypes.First();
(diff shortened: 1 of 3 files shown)


APSIM.Core/ModelRegistry/ModelRegistry.cs
```diff
@@ -36,14 +35,45 @@ internal static Type ModelNameToType(string modelNameToCreate)
...
-        Type typeToCreate = ModelNameToType(modelNameToCreate)
-                            ?? throw new Exception($"Unknown model type {modelNameToCreate}");
+        Type typeToCreate = ModelNameToType(modelNameToCreate);
+        if (typeToCreate == null)
+        {
+            // Find and return a resource.
+            var resourceName = modelsAssembly.GetManifestResourceNames().
+                FirstOrDefault(r => r.Contains(modelNameToCreate, StringComparison.InvariantCultureIgnoreCase));
+            if (resourceName != null)
+                return CreateResourceModelFromName(modelNameToCreate, resourceName);
```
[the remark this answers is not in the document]
ric394: Good point. I'll implement that.

Tests/UnitTests/APSIM.Core/CommandTests.cs
```diff
@@ -738,4 +738,21 @@ public void EnsureReplaceOnNameOnlyMatchesFirst()
...
+    {
+        Simulation simulation = new();
+        Node.Create(simulation);
+
+        IModelCommand cmd = new AddCommand(modelReference: new NewModelReference(modelName),
+                                           toPath: "[Simulation]",
+                                           multiple: false);
+        cmd.Run(simulation, runner: null);
+
+        Assert.That(simulation.Children[0], Is.InstanceOf(typeof(Plant)));
```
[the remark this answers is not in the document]
ric394: This won't do this in this instance.

BrianCollinss:
Shouldn't this be linked to https://github.com/APSIMInitiative/ApsimX/issues/11290?

APSIM.Core/ModelRegistry/ModelRegistry.cs
```diff
@@ -36,14 +36,49 @@ internal static Type ModelNameToType(string modelNameToCreate)
...
+        Type typeToCreate = ModelNameToType(modelNameToCreate);
+        if (typeToCreate == null)
+        {
+            // Find and return a resource.
+            var resourceName = modelsAssembly.GetManifestResourceNames().
+                FirstOrDefault(r => r.Equals($"Models.Resources.{modelNameToCreate}.json", StringComparison.InvariantCultureIgnoreCase));
+            if (resourceName != null)
+                return CreateResourceModelFromName(modelNameToCreate, resourceName);
+        }
+        if (typeToCreate == null)
```
par456: This will never be true. If the typeToCreate is null, it will either return a resource, or throw on line 46.

Might be worth wrapping lines 50 - 52 into an else statement to show that it's an explicit this or that and that the code will never leave the if block.

APSIM.Core/ModelRegistry/ModelRegistry.cs
```diff
@@ -36,14 +36,49 @@ internal static Type ModelNameToType(string modelNameToCreate)
...
 
+    /// <summary>
+    /// Create a model from an embedded resource.
+    /// </summary>
+    /// <param name="modelNameToCreate">The name of the model to create.</param>
+    /// <param name="resourceName">The name of the embedded resource.</param>
+    /// <returns>The newly created instance.</returns>
+    private static INodeModel CreateResourceModelFromName(string modelNameToCreate, string resourceName)
+    {
+        using var stream = modelsAssembly.GetManifestResourceStream(resourceName)
```
par456: Rather than doing this one-liner across two lines, why not just do the if null check below it.

APSIM.Core/ModelRegistry/ModelRegistry.cs
```diff
@@ -36,14 +36,49 @@ internal static Type ModelNameToType(string modelNameToCreate)
...
+            ?? throw new Exception($"Could not find model or resource with the name: {resourceName}");
+        using var reader = new StreamReader(stream);
```
par456: Just double check that these using statements should be put into a using block { }

APSIM.Core/ModelRegistry/ModelRegistry.cs
```diff
@@ -36,14 +36,49 @@ internal static Type ModelNameToType(string modelNameToCreate)
...
+        // Inject the ResourceName into the JSON before deserializing
+        var jObject = JObject.Parse(json);
+        // Get the first child of the root object and set its ResourceName property.
+        // The actual resource always has a simulations Parent, so the first child of the root object is always the resource.
+        jObject["Children"][0]["ResourceName"] = modelNameToCreate;
+        
+        INodeModel resource = FileFormat.ReadFromString<INodeModel>(jObject.ToString()).Model.GetChildren().FirstOrDefault();
+        
+        List<INodeModel> children = resource.GetChildren().ToList();
+        foreach (var child in children)
```
par456: just generally, try and use defined types instead of var for declaring variables, just makes it easier to know what something is supposed to be

APSIM.Core/ModelRegistry/ModelRegistry.cs
ric394: No longer required since .NET 8. I think it improves readability too.

APSIM.Core/ModelRegistry/ModelRegistry.cs
ric394: Agreed. I'll add the type.

BrianCollinss:
@ric394 I still get this error. I can see APSIM Releases stage has failed. Does it mean this pull wasn't merged into release?

for `[SowingHarvest].Script.Crop=[Wheat]`
```
Invalid property type: Models.Core.IPlant
```

for `[SowingHarvest].Script.Crop.Name=Wheat`
```
Invalid property type: Models.Core.IPlant
```

image

ric394:
Even if the release fails, once another 'release-able' PR is merged, it will be included.
I can see you're on 8046 at the moment, so it's concerning that you're still seeing the issue.

par456:
> @ric394 I still get this error.

This is a bit of a tricky one, since I don't think the language is set up to have a model reference on the left side yet. My suggestion would be this:

1. Have a string property in your script for the name of the Crop
```
public string CropName { get; set; }
```

2. In your commands, use that to set which crop it should link to:
```
[SowingHarvest].Script.CropName = Wheat
```

3. In your script where the crop reference needs to be used (or in a StartOfSimulation method)
```
Crop = Node.Find<IPlant>(CropName);
```

PR #11312, opened 2026-05-31
11285 total biomass across 12 DMD classes

### LINKED ISSUE #11285: Total biomass across 12 Dry Matter Digestibility (DMD) classes are not current in Pasture Model

C:\Jessica\GIT\ApsimX\ApsimX\Models\Pasture\grazpastpopn.cs

The HerbageMassGM2() method retrieves biomass based on Comp(Live , Dead), Part(Leaf, Stem) and DMD(12 classes). However the sum of biomass across all classes not current thereby not updating biomass pools when herbage is removed resulting in mass balance error. In order to keep it consistent and current it is necessary to refactor  HerbageMassGM2().

JPaul-Vasa:
Resolves #11285 

The PR keeps the biomass pools current by adding biomass across DMD classes thereby keeping it consistent


par456:
Does the GetHerbageNutr function have the same issue as the DM one? Don't have to fix that here, but it might be a very similar formula to do in the future.

JPaul-Vasa:
Totals are updated but there is a delay. This delay affects the mass balance checks when biomass is removed. N is not balanced so doesn't error out. Probably Nutrients (N,P ect) in the DMPools  needs refactoring at some stage.

PR #11321, opened 2026-06-05
WinterCereal CHO reserves

hut104:
working on #10516


par456:
This changes a constant in the model, was that delibrate?

EDIT: Ah yes, I see that it was, the other changes were from file updates.

PR #11349, opened 2026-06-16
Fix calculation for QMax

### LINKED ISSUE #11348: QMax is calculated incorrectly
### What happened?

When doing some weather analysis, we found that when graphing the QMax value with an Alpha of 0, the line would dip between 8th december to 29th of december each year instead of being a smooth curve like expected.

Copilot suggested the error is in the Q0i calculation

Windows

ric394:
resolves #11348

This fixes an issue in the ordering of variable calculations in the Q0i method which is used in QMax calculations.

### DIFF
--- a/APSIM.Shared/Utilities/MetUtilities.cs
+++ b/APSIM.Shared/Utilities/MetUtilities.cs
@@ -250,16 +250,14 @@
             double LATr = 0;
             double HSmax = 0;
 
+            DEC = 23.45 * System.Math.Sin(2.0 * System.Math.PI / 365.25 * (day - 79.25));
+            DECr = DEC * 2.0 * System.Math.PI / 360.0;
+            LATr = lat * 2.0 * System.Math.PI / 360.0;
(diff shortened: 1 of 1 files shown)


par456:

A mistake was found in the QMax calculation for daily radiation in the APSIM weather code, which would cause a few days in the year to calculate slightly lower than it should. This has been corrected so that the correct values are being used, rather than always using values of 0. 

This error has existed in the code prior to a pull request in classic in 2012 and was probably an accident when the code was originally written.

This has an effect on the Diffuse light calculation within the Rice model, which causes some stats to change very slightly.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/pull/11348

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11349/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11349).

As this is a bugfix with minimal effects on the models, it will be merged today. If you think there has been an error, please respond and we can address any issues.

PR #9955, opened 2025-05-12
Implement new Ruminant Growth Model in CLEM

lie112:
Working on #9954 

DO NOT MERGE!

THIS PULL REQUEST IS TO ALLOW TIME FOR REVIEW OF THE CONSIDERABLE CHANGES IN THIS PULL REQUEST

- Additional commits for Grazing and Breeding are still required to complete this update.
- A full check and validation are required before this merge can be accepted and available to users in APSIM.

PR #11359, opened 2026-06-23
CLEM Major upgrade (i.e. V2.0)

### LINKED ISSUE #9954: Implement sub-monthly time-step and new Ruminant Production with fat and protein tracking in CLEM
### Describe the new feature

This issue represents the upcoming MAJOR CLEM upgrade to address limitations of the initial model release and ensure CLEM is suitable for future development and integration with other APSIM models as the CLEM userbase increases. This is a current work in progress entering the final stages of validation and checking before release for testing by CLEM users.

For more details see [CLEM Online User Guide and Reference Manual](https://www.apsim.info/clem/Content/Details/MajorUpdate2025.htm)

This outline provides details of the MAJOR upgrade to CLEM for feedback by users

### Reason for major upgrade.
It was noted through the first couple years of users experience that current and near future user requirements are not possible in CLEM model that are available in APSIM. These restrictions would need to be addressed at some stage and require considerable refactoring of the CLEM codebase that was restricted by the simplifications of monthly time-step used in the code when ported from IaT and NABSA. Now was the time to implement large scale changes while the userbase is relatively small and several expert users are currently not using the model in projects. Also, time was available with animal physiology experts to update the ruminant production model to track fat and protein stores which will improve the animal growth approach and provide valuable information on individual and herd health in management simulations. This has occurred while animal production was ported from GrassGro to APSIMStock model.

This was also an opportunity to refactor code for optimal performance with future increases in computation expected and ensure the minimal amount of setting of parameters was undertaken for large herds. It does not appear that the changes have reduced the fast running of CLEM simulations.

A position was taken that CLEM would maintain its core approach of “simplicity until needing more detail” and would not introduce any additional overhead and burden of parameterisation to the user if not needed. All CLEM simulations using the monthly time-step must run in the new version.
### Challenges reported 
* The monthly time-step used by IAT, NABSA and initial release of CLEM allowed for many simplifications in approach but this was agreed to be was too coarse for some management and processes, so the idea is to allow the time-step to be defined by the user for the simulation case and even allow for dynamic timing of activities in future by linking CLEM to end of day rather than end of month events.
* Some though was needed to ensure a seamless integration of CLEM with other APSIM models such as crops and pasture. This related to the firing of CLEM events on the EndOfMonth event which may be too coarse for full integration.
* Can’t go completely to daily as per APSIM as this is not suitable for the WholeFarm approach of IAT and NABSA and would be an enormous CPU overhead for 20,000 head individual herds. But also recognised the monthly time-step with all actions tied to the first day of the month was too coarse for herd dynamics such as conception, births etc. 
* Conception in breeders is a tricky aspect to simulation using only monthly, or daily probabilities and a new approach was needed to allow full breeding cycles in an individual based model with consideration of animal condition imparts on breeding success.
* The individual animal model was ready for genetic variation but a rethink of ruminant parameters used across a range of different activities was needed and separated to reduce overhead of storing individual values. This approach allows for future individualisation required by the user with the understanding that fully implemented will require a large memory allocation but that is the cost of undertaking such analysis.

Upgrade process
* All CLEM components (activities, resources, timers, groups etc) will require a MinimumTimeStepPermitted attribute with a default value of “Monthly”. This provides the permitted time-step level to be provided in the code and handled in CLEM validation and error checking on running a simulation. An error thrown if a component is used that does not supported the specified time step. This means the model can be brought into the new time-step slowly. Trying hard to allow all previous simulations to keep running.
* Several Upgrade methods are applied to update user apsimx files to implement new property names and styles and avoid needed to recreate simulations for the new approach.
General changes implemented
* Animal production and dynamics
  * Improve the animal production as test of latest understanding and assessment of the AusFarm equations based on research of James Dougherty and Thomas Keogh
  * Allow CLEM to run on sub-monthly time-step, down to daily, for a sub-set of activities. For example if just using feeding and grow activities there is no reason why a daily time-step cannot be used. This allows for the simulation of animal production in feedlots for validation of the growth model against field data.
  * Add tracking of feed quality with additional parameters
  * Implement mixing of feed
  * Move properties to Information classes (e.g. Weight, Energy, Output, etc) which allows users to easily locate properties for reporting.
  * New code to handle allocation of fat and protein for new individuals and births in the model.
  * Link code to AusFarm reported equations.
  * Opportunity to fully understand where the IAT/NABSA animal production equations were from.
  * Allow a range of enteric methane emission equations to be included in the enteric methane activity.
  * Include a force feeding to override code and fully represent feedlot feeding rates for testing and validation.
  * Refactor to separate ruminant creation by cohort vs births.
  * Shift mortality from Grow activity to separate activities and warn if death not included in simulation. Death can therefore be removed for feedlot trials by not including activities rather than having to set parameters in ruminant space to 0.
  * Implement CLEM grow, Oddy approach, SCA07, and new GrowPF (Protein and Fat)
  * Refactor repeated code across growth models into static classes to be shared.
  * Update wool production and shear activity. Remove cashmere.
  * Allow feeding 0 kg with warning displayed as needed in experiments to turn off feeding..
  * Clean up all naming of weight properties and ensure they are consistent and as expected by users. E.g. live vs base weight with and without gut fill etc.
  * Add toxaemia and dystocia mortality
  * Implement and check sheep
  * Improve sterlisation handling. Castration, Freemartin, Webbed, Spayed.
* Animal parameters
  * Opportunity to clean up naming for better understanding of units required (e.g. Proportion vs Percent in property name)
  * Improve categorisation with the category and subcategory attribute
  * Include UI display of CrossBreed category to identify properties that should change between breeds
  * Implement user access levels and restrict visibility of parameters for users based on experience
  * Added male specific min age and weight for maturity
  * Fix filter by property for new parameter approach
  * Shifted properties to sub groups and unified naming
* Sub monthly time-step
  * New CLEMEvents component
    * Remove all CLEM events from APSIM Clock as requested by APSIM Users to simplify Intellisense when users are not using CLEM components.
    * CLEM will use CLEMEvents for timing which is a component required below the APSIM Clock for CLEM simulations.
    * Implement Interval, TimeStepStart and End Date, and Interval index to track time-steps.
    * Fire CLEM Events at EndOfDay on appropriate days in Clock
    * Actual number of days in a month is now considered in the monthly time-step as well as leap years.
  * Timing
    * Update all hard coded time-step (month length of 30.5 days) to use the CLEMEvents.Interval
    * Refactor actions that had assumed monthly step for simplicity
    * Provide new CalendarTimer that provides the functionality of previous DateRange, MonthRage and MonthInterval timers.
    * Implement dayOfYear entry for users.
  * Change animal age to be in days
    * Provide a new class AgeSpecifier which allows the user to define ages/time in days using the familiar convention of years, months, days for entry and reading which is converted into days
    * This required new functionality to allow a property of a class (YMDValues of AgeSpecifier) to be the entry style to define a complex class property in the PropertyPresenter using attribute another associated property. For example the Y, M, D entry is used and converted to the class. See [Core.Display(SubstituteSubPropertyName = "Parts")]
  * Change timing stores used in CLEM to be DateTime rather than day, or month count integer values. 
  * Allow sub timestep actions
    * This includes births in middle of time step affecting days pregnant and days lactating as we cannot assume the entire time step is pregnant or lactating.
    * Improve weaning functionality to allow sucking and weaned actions withing time step
    * These require a rethink of tracking animal state and the timing of updating details with respect to production, intake and action activities.
* Maintain or improve model performance
  * All changes considered the implications of going to a finer time-step on an individual based model where large herds (20,000 breeders) are simulated.
* Crops
  * Rework cropping to allow input based on any time-step
  * Fix the crop harvest timers to allow 
  * Implement new IFeed interface for feed quality inputs from crop data
  * Fixed mixed/nested crops for example
* Labour
  * Implement age in days for people
  * Update timing
  * Fixed labour activity task
  * Implement AgeSpecifier in other animals and update breed, feed etc to handle changes.
  * Fixed labour tasks and labour allocation presenters to work with time step
* Grazing 
  * Allow fine pasture growth input
  * Link pasture quality to new feed quality implemented for animal production (IFeed interface)
  * Remove old equations to handle pasture biomass and quality impacts on intake as now handled by physiological and intake quality processes.
  * Refactor to reduce duplicate code between GrazeAll, GrazePaddock and GrazePaddockBreed and behind scenes creation of all grazing activities required for GrazeAll.
* Improve validation and reporting to the user
  * Add additional property validation checks
* General fixes
  * Fixed URI links to help system
  * Fixed reporting to Summary on error and therefore not stopping on validation errors due to timing of database write in APSIM
  * Update base ruminant activity to work with multiple brreds.
  * Redefine which activities allow multiple ruminant breeds
  * Allow RuminantMove activity to define start or end of time-step.
  * Started implementing unit tests
    * Trap issues in Ruminant constructor in unit tests
  * Refactor manure production to one place
  * Improved working with individuals in herd avoiding excessive filtering and sorting.
  * Update descriptive summaries as changes made
  * Removed old unneeded properties.
  * Clean up code with recommendations of APSIM and latest .Net coding style guides
  * Ignore duplicate warning and error messages in Messages display
  * Changed some Iemumerable to use Yield
  * Update ruminant filtering to allow nested parameters.
  * Add model associations
    * Help identify issues for the user
    * Implements ModelAssociations using ValidParent to avoid duplication
  * Add/adjust all new tree icons for UI
  * Update help system with all new changes

lie112:
This is the current Branch for CLEM V2 major upgrade.
It is provided for software review
Changes and documentation can be found at 
[CLEM Major Upgrade 2026](https://www.apsim.info/clem/Content/Details/MajorUpdate2025.htm)
The CLEM User guide and reference manual is now updated to reflect these changes (i.e. Ruminant parameters and growth etc)

Resolves #9954

I never know whether to address all the code style suggestions as it's not clear if they are Visual Studio or APSIM and often relate to new styling conventions and code simplification of the latest C# version.


par456:
Things to address:

- CLEM Examples aren't running in validation and neither is Validation/CLEM/CLEM_Sensibility_GrowCrop.apsimx
- Can you remove global.json (and it's copy)
- Remove Models.csproj‎ as we do want warnings as errors (and fix any warnings that throw)
- Not sure why Apsimx.sln has changed
- AnimalGroup can be reverted, functions made public, but then not used
- Can you turn the comment in ModelWithdigestibleBiomass into an issue instead
- ModelAssociationStyle looks like it wants to be a generic class, but then has specific Ruminant references, can you give me a rundown what this is needed for?

Need to look into the propertypresenter changes some more, but these are the easiest changes.

lie112:
Examples working. The Upgrade 117 just had to be enabled (method changed and current increment increased). Removed validation files as not validation

global.json removed. Not sure where this came from. I think it was a while ago. Sometimes you have to play around to get the code to compile when moving between major versions. Usually a Clean and Rebuild of solution is enough but not always.

I don't understand how to remove models.csproj. Isn't this the master project file for models that is needed? Why do I have it and you don't? I have never added this. Do I just delete the file from the folder while Visual Studio is closed? Was this previously there and not been carried through pull requests from ASPIM/master after being removed by someone else?

What is the difference in APSIM.sln? I haven't changed it but curious as to what has.

Do I just change the code back manually in AnimalGroup or is the a revert in Git I need to use. I have removed the comment and taken the "static" from the methods. I think they were both always public. Done. changed back and committed changes

Removed the additional comment. I have lots of pages of notes created as I tried to understand the logic of SimpleGrazing and Stock, plus we have pulled apart many of the equations behind Stock and checked them as we added them to CLEM. This is a complicated space for the person who picks up Stock and we have lots of info that hasn't gone in as issues. This isn't currently my work.

ModelAssociationStyle is used with the ModelAssociationAttribute. This is a more advanced customisable version of ValidParent attribute that I use throughout CLEM. This allows me to automate the checking to ensure required components are present in the simulation in the correct scope. This allows me to define other required components at the top of each component in an easy to see location similar to ValidParent. The check of ModelAssociations is done during my model setup and validation so a need error message is written to Summary for the user to correct the issue. This can be a generic class. I was merely trying not to create too many extra CLEM specific files in other areas but accept nested classes in a file is bad practice. The link to Ruminants is that this is a special condition also required by a heap of components in CLEM. All these have removed an enormous overhead of having to try and fix users simulations and it is really easy to forget why things might break and I have to spend ages debugging only to find they are missing a component needed. This also neatly traps the changes needed for the new upgrade such that if the user hasn't added the new RuminantParameters component and all children a clean error is generated rather than a nasty null object error that none of my users can interpret.

I didn't realise I made so many changes to PropertyPresenter. They seem much bigger when looking at track changes in Git. I could convert this to a CLEMPropertyPresenter and leave the other PropertyPresenter alone. These add functionalty to PropertyPresenter which is Wrapped by a CetegoryPropertyPresenter to show category tree structure. This adds a list of properties categories and subcategories to the presenter that are obtained from Attributes associated with each properties. This somewhat mimics the grouping provided in PropertyPresenter but is more powerful. Properties can appear in more than one class and depending one where in the tree the user clicks will determine the properties displayed. I also have a user role that is set in the CLEM component. This will hide/disable properties that that don't need to know about. This saves me having to explain that just because a properties has been provided and not hard coded doesn't mean you should change it for your herd. This means simple model users can reduce the vast parameter space presented to just the parameters they'd be expected to change. This helps with searching for parameters. I am going to add a smart search box one day to further filter. This will be useful for the Stock animal parameters of which there are hundreds. Plus, it is nice to be able to filter by description on not have to know the dumb parameter naming convention like dhrmsCI12. This also allows all the properties of nested components to be seen in the one location and filtered together. See image below. All of this functionality is available now for any APSIM model by adding the Category Attribute to Properties

`        [Category("Task:Herd size", "Breeding females:Breeding females")]
`
and setting the presenter and view to wrap the property presenter
`    [ViewName("UserInterface.Views.PropertyCategorisedView")]
    [PresenterName("UserInterface.Presenters.PropertyCategorisedPresenter")]`


image

All tasks and details are now completed.

I have found some other issues while doing this task but I am not going to be able to get them fully fixed at present.

Should I open an issue relating to how I can modify or work with SimpleGrazing? I need to either 
(a) Have multiple simpleGrazing components fully parameterised in my simulation (the current preferred method of the users). One for each Zone, with their functionality turned off using the ByOtherModels enum for the request for pasture. I also need to disable the animal rotation logic. I will then need to add a public method that allows me to add urine that will go through the ZoneWithForage class managed by SimpleGrazing and do the fertilisation. I can then directly interact with all the sub-zones created in the background from my models. I think the method might need some changing as it looks for SimpleCow (which it can ignore if not present) which I don't need as I am providing the daily amount of urine from CLEM.
(b) Build a new Adapter that manages the creation of all sub-zone urine patches in soil for any Zone (paddock) in the simulation tree. This would have methods to access the paddock of interest like Forages and ForageProviders and then ant model can take feed (biomass) from the plants and deposit dung and urine via the patch urination method. Care would be needed to ensure that SimpleGrazing is not affected or the code duplicated for now. I had originally used the methods of Stock to place urine, but this doesn't allow for the patch urination of SimpleGrazing to be implemented in any livestock model.

lie112:
Fixed merge conflict from new upgrade added while I was merging CLEM

lie112:
I fixed the code to pass all of my timer unit tests
I have just disabled (commented) the ruminant constructor unit tests for now. The code is ok, but I need to better understand how to run the model to see object structure or work out the initialisation methods needed to generate the herd for unit tests.

lie112:
I'm not sure the latest issue is due to CLEM. I now have all my unit tests passing and don't use browntop anywhere I know of.

lie112:
Understand what you meant by the models.csproj. I have removed the line that allowed me to keep working while the issue was reported and fixed. This is the only way to keep working on dev version when these issues occur. You still get the warnings and red icons in error list but it doesn't stop the compile. Just have to remember to remove this line if added.

I don't know what to do about the APSIM.sln file. How can I revert to your version. Should I get the code from git and replace what is in mine? I don't know why this has been changed.

par456:
Hi adam, I'll pull it down now and have a look. I'll also do some cleanup if needed.

lie112:
I hope you have an automated way of running all the resources through the latest upgrade in converter! This needs to be part of the build process if the last version number changes.

ric394:
> I hope you have an automated way of running all the resources through the latest upgrade in converter! This needs to be part of the build process if the last version number changes.

If you want to do this, you can do it from the main menu when run from your IDE.

image

lie112:
Oh, that's what that menu button does. Do resources copy with the pull from APSIM/master request? Why would another resource I am not using be out of date when I had pulled the latest master? This stopped the build process.

Are the 2 commits on my origin/RuminantV2Working repo the two just made by Andrew. So, if I pull these, I will have Andrew's changes to check. Just confirming I have that right and I'll pull those to commits to my local and be able to see them. I am always very cautious about anything in Git not part of my normal workflow as I have seen disastrous outcomes. 

I'll check the new substitute property approach as I can't see how it is deciding to use a substitute and which property it uses.

ric394:
> Oh, that's what that menu button does. Do resources copy with the pull from APSIM/master request? Why would another resource I am not using be out of date when I had pulled the latest master? This stopped the build process.
> 
> Are the 2 commits on my origin/RuminantV2Working repo the two just made by Andrew. So, if I pull these, I will have Andrew's changes to check. Just confirming I have that right and I'll pull those to commits to my local and be able to see them. I am always very cautious about anything in Git not part of my normal workflow as I have seen disastrous outcomes.
> 
> I'll check the new substitute property approach as I can't see how it is deciding to use a substitute and which property it uses.

1. If you pull from APSIMInitiative/master you'll get the most recent file versions in your resources.
2. Yes, these are from Andrew. You can see it here in your PR (as shown in the below screenshot). For the second part of your second question, if you pull from your local branch, it will bring in Andrew's changes.:
image

lie112:
Ok. So, it was actually my pull request that broke that resource as I incremented the latest version to 218. It's odd that the error wasn't thrown for all resources, maybe it just crashed on the first encounter. So, from now on we have to do a resource update whenever we increment the version number in converter?

I'll pull the changes and check it's all working.

lie112:
Using the Parts property of AgeSpecifier to receive the user input for the class is working in the UI. What is actually needed to implement this elsewhere in future? It seems the first property with a description attribute of the Class considered as a property in the UI is used as the sub-property. Does it undertake this whenever the Parent Property is not of the types expected and thus has no user entry style? So, you just made this the default functionality rather than the specified approach I used. I would not have added a default functionality as I can't know how that might affect other models. My approach to date has been to use attributes to identify special CLEM functionality.

lie112:
Is that all that is needed from me for now with respect to this pull request? I am nervous as this has been a more rushed pull than I anticipated. 

I am now after feedback on #11362 and what I am permitted to change/implement moving forward as this will be the next branch I work on after CLEM v2 is accepted.

par456:
> Does it undertake this whenever the Parent Property is not of the types expected and thus has no user entry style? So, you just made this the default functionality rather than the specified approach I used.

That is correct, however previously it would have just thrown errors when trying to do that, since the Property presenter didn't know what to do with a class object, so this is safer anyway. If we get to a case where this needs to be more complex we'll revisit it then, but we only have one use case for this right now, so we'll limit the changes to prevent complexity.

> Is that all that is needed from me for now with respect to this pull request? I am nervous as this has been a more rushed pull than I anticipated.

Should be, I'm just doing some additional cleanup on some files where I'm happy for your changes to be the default, should be merged within the next hour

par456:
@ric394 When this is finished running, it should be ready to merge.

ric394:
> Ok. So, it was actually my pull request that broke that resource as I incremented the latest version to 218. It's odd that the error wasn't thrown for all resources, maybe it just crashed on the first encounter. So, from now on we have to do a resource update whenever we increment the version number in converter?
> 
> I'll pull the changes and check it's all working.

Yes. If you increment the number in the converter, you must also update the resources.

ric394 (approved):

lie112:
Resolves #9955

PR #11400, opened 2026-07-02
The runoff parameters should be settable but not on the UI!

### LINKED ISSUE #11383: Make some runoff parameters settable
### Describe the new feature

Needed for the testing and sensitivity analysis of the runoff process

sno036:
resolves #11383 

Don't put the parameters on the UI

### DIFF
--- a/Models/Soils/SWIM/Swim.cs
+++ b/Models/Soils/SWIM/Swim.cs
@@ -125,12 +125,12 @@ namespace Models.Soils
         private const double slcerr = 0.000001;
 
         /// <summary>Depth in the soil to assess water content for runoff effect (mm).</summary>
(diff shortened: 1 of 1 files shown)


par456:
Hi Val, the ai is pretty much on the mark for this one. I would recommend deleting the description tag rather than hiding it with a callback, as that's getting very complicated when the solution is to just not tag it with attributes. The bound attribute also only effects the GUI (so is useless if its not to be shown in the GUI), so that should go as well so it's not confusing.

I've pushed up that change to your branch.

If there is something from the description or bounds that you want to keep on the property, just add it into the summary field instead.

ric394 (approved):

PR #11420, opened 2026-07-09
9918 Metadata

par456:
Working on #9918 

New PR branch for this.

Getting the infrastructure behind metadata into the current release to test with POStats system. For the moment, the metadata table only includes the Category (None, Example, Prototype or Validation), Directory of the apsimx file and the filename of the apsimx file.

Also included in this branch is a new metadata table for PredictObserved tables, so that POStats can get that information from the database instead of trying to read the json going forward.

As this PR only adds two new tables to the datastore, it won't effect the function of POStats yet, once this is merged, I'll put up a PR to have POStats use this new information.

Need to add some more tests and comments to the PR before this is merged.


par456:
This should be good for a review. Unit test added for the new PredictObserved metadata table.

ric394 (approved):

PR #11430, opened 2026-07-14
Fixes timing issue of CLEM pasture taken and fixes negative urine nitrogen request.

### LINKED ISSUE #11429: Problem with offtake and urine interacting with AgPasture
### What happened?

A few issues have been identified with the interaction between CLEM and AgPasture in initial simulations

- A negative amount of urine is reported from CLEM
- Incorrect amount taken due to timing between models issue.
- Need to change CLEM timing to start of CLEM interval rather than previous end of interval.
- Pending resource tracking was not returning unneeded feed and pasture to resources before grazing.
- Error in creating temporary ruminant for descriptive summaries.

Windows

lie112:
Resolves #11429


lie112:
I don't this the error preventing this request is due to anything CLEM

lie112:
I have checked and the test\validation\clem\AgPasture_CLEM.apsimx runs successfully in the latest version so this doesn't appear to be the issue. But the last commit has the X beside it.

par456:
We were all looking at the wrong file, it wasn't the new test in validation that was having the problem.

The CLEM example CLEM_Example_RuminantFiltering seems to be getting stuck in an infinite loop eventually causing a stack overflow and killing the process. I was able to see the same thing when running that locally on your branch.

lie112:
Thanks for identifying this. I'll look into it. I did ned to make some changes to the event firing and ordering in CLEM to work with APSIM and this is the likely cause.

lie112:
Interesting. This is a really cut down simulation. During validation a report steps through the initial 15 individuals to report details of each individual prior to the simulation start and somehow generates 3.4 Gb of memory usage but just writing properties. I'll sort this out before I can understand where an infinite loop exists which may be a bug in starting and ending the simulation in a single month for a monthly timestep in CLEM. It seems to just be jumping between the progress bar update in GUI, but I can see how CLEM would stop the clock progressing to finish the simulation.

lie112:
Attn @par456 
Narrowing it down after 2 hours of searching and still some odd behaviour. I was asking what the normal process is when a class object is provided as a Report variable? I had accidently left off the property name I wanted to report. I used 
`[ReportHerd].ReportDetails.Individual.Weight`
instead of 
`[ReportHerd].ReportDetails.Individual.Weight.Live`
And for my 15 individual cows that got passed to the report twice each during the simulation this resulted in 22GB of memory being used and the simulation hanging. 

The GUI eventually shows "Finding simulations to run..." while it was running the only simulation present which normally finished in less than a second. If I hit stop I get an error
 I get two Sequence contains no matching element errors if that could be involved. 
`System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.ThrowHelper.ThrowNoMatchException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at Models.Core.Run.SimulationDescription.get_Storage() in C:\Data\Source\Repos\lie112\ApsimX\Models\Core\Run\SimulationDescription.cs:line 78
   at Models.Core.Run.SimulationDescription.Cleanup(CancellationTokenSource cancelToken) in C:\Data\Source\Repos\lie112\ApsimX\Models\Core\Run\SimulationDescription.cs:line 123
   at APSIM.Shared.JobRunning.JobRunner.Stop() in C:\Data\Source\Repos\lie112\ApsimX\APSIM.Shared\JobRunning\JobRunner.cs:line 113
   at Models.Core.Run.Runner.Stop() in C:\Data\Source\Repos\lie112\ApsimX\Models\Core\Run\Runner.cs:line 333
   at UserInterface.Commands.RunCommand.Stop() in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Commands\RunCommand.cs:line 154
   at UserInterface.Commands.RunCommand.OnStopSimulation(Object sender, EventArgs e) in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Commands\RunCommand.cs:line 135
   at UserInterface.Views.MainView.OnStopClicked(Object o, EventArgs e) in C:\Data\Source\Repos\lie112\ApsimX\ApsimNG\Views\MainView.cs:line 1151`

This is the problem when the datastore is empty. 
If however, there is already data in the datastore then I do get columns created based on 66 properties of the weight class (well actuallt including the properties of the properties that are also classes (TrackingItems), but they never receive any values on reporting so never get to reporting. Adding the .Live gives the correct value. I note that the 67th property should be from the Cashmere RuminantTrackingItem that should work like the WoolClean trackingItem. So, I am not sure why it stopped adding properties unless there is a limit coded.  It should then go on to add the weight properties. I trie

I also got a System.AccessViolationException: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.' on one run.

If we leave it running, so trying to propagate the columns in the report from Weight it just keeps consuming memmory and Ive seen it up to 28 GB. I tried to let it keep running for longer and stop it to see what column it got up to, but I have not been able to repeat the process that give me the column names. 

I'd like to get to the bottom of this so I can fix it in CLEM\Resources\RuminantInfoWeight or provide an error in APSIM to show there's a problem. This could be a common mistake by users having to manually enter report variables in reports.
It doesn't really explain what was using up all the memory and hanging the application and I wasn't able to identify the issue in CPU or memory diagnostics in Visual Studio.

I do have some additional fixes to commit to this pull request now as I have changed some timing again so reporting still happens at the end of a CLEM time step and I automatically alter the clock.end date if not suitable for a monthly time step during the setup to avoild reporting issues in CLEM. I also put an error count in two while statements to make sure they never hang.

You should be able to run the offending CLEM example if you go into the Reports\Report and add .Live to the end of the Weight report variable entry. This change can identify what the issue is. Is it in report, or is it in RuminantInfoWeight class? There are some RuminantTrackingItem properties in RuminantInfoWeight and these are set to new() at initialisation. There did seem to be too many of these created in the memory comparison, so maybe the Report keeps trying to test and call the object and it keeps creating these and they are not handled correctly. 

My feeling is it's in the generation of report headings (maybe locator) from nested classes when a class is added as a report variable (by mistake or on purpose). 
I though someone might understand this better than me as same me going the next step to look into the report column generation.

PS Report view doesn't update between simulation runs where report variables have been changed (added). You have to go elsewhere or up to the datastore to get it to refresh.

So, the issue arose from me not setting a reporting property correctly, but the issue is real and needs to be addressed. I just don't know if it is my class that I happened to display all properties of, or the issue is in the backend generation of the columns to report. Given it keeps adding more and more memory allocation, I'd suggest it might be in the locator for the report columns.

par456:
That is some good investigation there Adam, and I think you're right, I suspect when given an object, its trying to dig through and work out all the columns, and depending on how things are connected, either looping back on itself, or just has too many properties to display.

I'll have a look into this myself and see if I can trace what its doing and put some sort of safety throw in here to stop it from happening again.

par456:
I've pushed in a fix on the reporting side so that it should throw when encountering a class it shouldn't try to report on. We'll see if that has knock on effects, but that might be a solution to the problem. I would like you to pull the branch down and test if it runs green to make sure I didn't bust anything else in Clem by adding this restriction.

PR #11441, opened 2026-07-17
Add hourly DCaPST reporting

### LINKED ISSUE #9833: DCaPST hourly variable output
### Describe the new feature

This will allow users to view DCaPST hourly outputs such as canopy light interception, photosynthesis, water use, conductance, and transpiration, leaf temperature.

yufenghe22:
## Resolves https://github.com/APSIMInitiative/ApsimX/issues/9833
- Add a DCaPST `IntervalStep` event for native hourly reporting.
- Expose hourly sunlit and shaded assimilation, water use, temperature, VPD, and LAI values through the Report model.
- Add LAI-weighted canopy temperature and VPD outputs.
- Retain the calculated sunlit and shaded LAI for each DCaPST interval.

## Rationale

DCaPST calculates hourly values internally from 06:00 through 18:00, but these values could not previously be exported by an APSIM Report. The new event is emitted after the actual water-limited biomass calculation so reported intervals reflect the values used for daily biomass production.

Reports can subscribe to `[DCaPST].IntervalStep` and select the required hourly properties.

This PR could have an `Improvement` label if someone could add that for me please.

### DIFF
--- a/Models/DCaPST/DCaPSTModelNG.cs
+++ b/Models/DCaPST/DCaPSTModelNG.cs
@@ -150,6 +150,116 @@ namespace Models.DCAPST
         public DCAPSTModel DcapstModel { get; private set; } = new();
 
         /// <summary>
+        /// Invoked once for each DCaPST calculation interval after the day's
+        /// water-limited biomass calculation has completed.
+        /// </summary>
+        public event EventHandler IntervalStep;
+
+        /// <summary>Date and time of the current DCaPST interval.</summary>
+        [JsonIgnore]
+        public DateTime IntervalDateTime { get; private set; }
+
+        /// <summary>Hour of the current DCaPST interval.</summary>
+        [JsonIgnore]
+        [Units("hours")]
+        public double Hour { get; private set; }
+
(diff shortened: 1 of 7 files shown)


par456:
Hi, just doing a review on this work.

I'm pretty happy with the functionality, just require one change to how it is structured. In this PR the DCaPSTModelNG has a lot of properties added to it that are filled in for each interval, then the whole object is passed to an event for reporting. This leads to the last interval's value remaining on the class until the next day.

Could I instead get you to make a new class that holds these properties, and pass that in the event instead? That way the values will only exist at reporting time and not stored.

yufenghe22:
Thanks for the suggestions.

- Added `DCaPSTIntervalOutput.cs` containing all interval outputs.

- Changed `IntervalStep `to pass `DCaPSTIntervalOutput `as its event arguments. Replaced the numerous interval properties on `DCaPSTModelNG.cs` with one transient `CurrentInterval`.

- `CurrentInterval `is cleared in a `finally `block, including when a subscriber throws.

- Updated tests to verify event data, nested Report lookup, and cleanup.

Issue #11421, opened 2026-07-09
EmergingPhase ignores any child ThermalTime function, instead using the ThermalTime property of the phenology model directly.


ilhuber:
### Where is the code located?

Models/PMF/Phenology/Phases/EmergingPhase.cs

### What needs to be changed and why?

Many published models currently contain a superfluous ThermalTime variable reference child in their EmergingPhase. This child is actually never evaluated during a normal APSIM run, which is confusing. For consistency's sake EmergingPhase should maybe follow a Progress child typically linked to the phenology's thermal time?

PR #11444, opened 2026-07-20
Make emergingphase follow its child.

ilhuber:
Working on #11421 

Making this PR to see if validation goes through with this change. Hoping there are no simulation differences.


par456:
Looks like the only problem is that the prototype lentil model and TropicalPasture doesn't have a ThermalTime child. 

The least impactful option I can think is for each of these models to put a variable reference node as a child (with the name ThermalTime) pointing back to where it currently looks for ThermalTime.

ilhuber:
Looks like the tropical pasture one might have just been a typo.

ilhuber:
With these changes it looks like there weren't any science problems, which is good. Are we okay with this child being named ThermalTime or do we want it to be called Progression as it is in GenericPhase?

par456:
There does not seem to be a standard for this, but looking at the other Phase classes, it does seem to use just a link to ThermalTime rather than the name progression most of the time.

Let's stick with ThermalTime for now.

ilhuber:
In that case, this should be ready to go.

PR #11446, opened 2026-07-20
Use full path to read excel sheet names.

### LINKED ISSUE #11445: GUI can fail to display excel met file sheets with an error.
### What happened?

It looks like when excel weather input is using a relative path, the GUI will search for the file wherever it was launched from rather than the simulation location. The simulation will still run properly, but this UI component can become unusable. 

Windows

ilhuber:
Resolves #11445 

Changes where full path is declared, and use that to read excel sheet names.


par456:
One issue we always run into when doing this, is where by opening the file, suddenly the file names start getting saved as absolute paths again, the filepath code in things like weather are really sensitive to breaking.

Can you open up some apsim files, both in the repo and outside it, click/edit the weather paths and make sure that they are still saved as relative into the file.

ilhuber:
Looks like same directory and sub directories will still be saved as a relative path, but a parent directory becomes an absolute path? I think that is how it behaved before but I am not sure.

PR #11458, opened 2026-07-27
Volatilisation - unit conversion error fixed

sno036:
working on #9763

Still working but please merge this so it is available to testers!

There was an error in the conversion of mg/m2 to kg/ha found by @hrpasl. This code fixes that.

@KirstenVer 
@JodyBig


ric394:
@sno036 @hrpasl for the same reasons provided in the previous PR #11153, I'm uneasy to merge this one.

I've discussed a path forward with @par456 and this involves using an `IFunction` to accomplish the same thing, but in a safer manner.

sno036:
Thanks @ric394  - can you give me an example of how you would want this done?

par456:
Hi Val

From what I see here, you have two changes you want to test on this model, first is calculating PotGasExchangeNH with either weather.Wind or waterBalance.Eo, and then at the bottom, changing 10E to 1E

I'd be happy if you just want the 10E to 1E change merged, since that's an obvious units bugfix, so if you went back master and made a new branch with just that change, you could put that up as a PR and we can put that in easy.

If you need the test flag merged in as well, I'll outline how we can do that in a more sensible way without using a flag.

PR #11459, opened 2026-07-27
9617 Virtual Model

par456:
Working on #9617 

Continuing the work on improving our test system, this new model allows us to read in the contents of one apsimx file into another at runtime and in the GUI. Some of this work was started back in FactorFromFile, and has now been expanded into this Virtual model.

Virtual models contain the structure referenced in its settings as read only children, that can then be run, overridden or viewed (like with graphs) in the current APSIM file. The nodes are created whenever the file is loaded, run or the update button on the model screen is clicked in the GUI.

This feature is very much still in the testing stage, but will be used with the changes to POStats last week to start tracking prototyped changes to Released models, so that we can more frequently merge in development work for model changes, and track the stats for that work, without needing to change the released model until it is ready.

Feel free to play with this model once it is merged in, however there may be breaking changes on in the near future as I roll it out to more situations, so use with caution.


APSIM.Core/Path/APSIMFilePath.cs
```diff
@@ -0,0 +1,53 @@
+using System.Text.Json.Serialization;
+using APSIM.Shared.Utilities;
+
+namespace APSIM.Core;
+
+/// <summary>
+
```
ric394: Add a summary here

ApsimNG/Menus/MainMenu.cs
```diff
@@ -209,8 +210,24 @@ public void OnRun(object sender, EventArgs e)
...
         {
+            IModel currentNode = explorerPresenter.CurrentNode;
+            
+            //Prior to running, we need to generate any virtual nodes that have 
+            //been changed since opening the file
+            bool treeUpdated = false;
+            foreach (Node node in currentNode.Node.Walk())
+                if (node.Model is IGenerateNodes generator)
+                    if (generator.RequiresUpdating())
+                        treeUpdated = true;
```
ric394: Should this be the other way around? 
```
if (!generator.RequiresUpdating())
```

ApsimNG/Presenters/UpdatePresenter.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                _view.Clicked -= OnClick;
+                _eventsConnected = false;
+            }
+        }
+
+        /// <summary>Refresh the list</summary>
+        public void Refresh()
+        {
+        }
```
ric394, line 72: Is this meant to do nothing?

Models/Factorial/FactorFromFile.cs
```diff
@@ -169,14 +169,6 @@ public string FileName {
...
         public string NameColumn { get; set; }
 
-        /// <summary>
-        /// Property to update in order to trigger OnModelChanged events.
-        /// </summary>
-        [Description(" ")]
-        [Display(Type = DisplayType.Button)]
-        [JsonIgnore]
-        public bool Refresh {get;set;} = false;
-
```
ric394, line 179: Is this meant to be included in this PR?

ric394 (changes requested):
Just a few suggestions but otherwise this is good to go.

APSIM.Core/Path/APSIMFilePath.cs
par456: good catch, will fill in

ApsimNG/Menus/MainMenu.cs
par456: if statement is correct, but will rename variable to make more sense

ApsimNG/Presenters/UpdatePresenter.cs
par456: it is, refresh is needed for the interface, but a button has nothing to refresh. will add comment saying that.

Models/Factorial/FactorFromFile.cs
par456: Yes...ish

FactorFromFile had a refresh button done that way, but I've implemented a safer version with Virtual and changed it over to use the safer version. It was a small change so I didn't want to bother with an extra PR just to do that.

ric394 (approved):

PR #11334, opened 2026-06-10
Update the parameter values and the related simulations listed in the frost heat damage functions paper

PengchengHU:
Working on #11333

PR #11462, opened 2026-07-28
Update the parameter values frost heat damage functions paper

### LINKED ISSUE #11333: Update the parameter values of the damage functions as listed in the published paper
### Describe the new feature

The damage function paper is published, and people ask to update the new parameter values.

The reference:

Hu, P., He, D., Zheng, B., Whish, J., Kirkegaard, J., Bell, L., Leske, B., Chen, S., Uppal, R., Biddulph, B., Trethowan, R., Beletse, Y., Lilley, J., 2026. Event-based frost and heat damage functions improve yield predictions of APSIM canola and wheat: formulation, calibration, and evaluation. Agricultural and Forest Meteorology 386, 111239. https://doi.org/10.1016/j.agrformet.2026.111239

par456:
Resolves #11333

This is the 2nd part of the PR which includes the model changes. First part with the simulations and data can be seen #11334

### DIFF
--- a/Models/Functions/FrostHeatDamageFunctions.cs
+++ b/Models/Functions/FrostHeatDamageFunctions.cs
@@ -44,7 +44,11 @@ namespace Models.Functions
     /// </para>
     /// </remarks>
     [Serializable]
-    [Description("Damage functions of frost and heat stresses are developed for canola and wheat by the GRDC Frost and Heat Management Analytics (FAHMA) project.")]
+    [Description("The development of frost and heat damage functions for canola and wheat was supported by the Frost and Heat Management Analytics (FAHMA) project " +
+        "via funding from the Grains Research and Development Corporation (GRDC; Grant No. CSP2204-009RTX).\n" +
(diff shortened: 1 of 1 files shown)


par456:
@PengchengHU 

Are the parameters that we are updating this in PR coded into the APSIM file because they are changeable by the user? 

If so, should we be forcing people using this model to update to the newer parameter values that you've set here? Otherwise these changes will only effect newly created simulations which is a problem.

PengchengHU:
Hi Andrew @par456 , thanks for splitting the PR. I think the parameter values can be treated as default values, and the new parameter values are editable from the UI so that it keeps the flexibility for user to explore their own values. What do you think about this approach? Should we make them uneditable?

par456:
What should we do in the case of existing files that are using the old defaults? Say they have used the model previously (which has saved the old defaults to their file), but not adjusted any of the values, should we update them to the new defaults, or leave them as they are.

PengchengHU:
I think it is better to just leave them as they are.

PR #11463, opened 2026-07-28
Refactor upload-release-assets.sh to simplify process

ric394:
Removed cache update and Netlify build steps from script, as these are no longer needed due to the Netlify site being retired, and the registration API being rewritten.


par456:
Is it time to pull the entire Netlify folder from the repo with this? or is that something you think should be handled in its own PR?

ric394:
> Is it time to pull the entire Netlify folder from the repo with this? or is that something you think should be handled in its own PR?

Let's do that in a subsequent PR.

PR #11464, opened 2026-07-28
11222 Remove duplicated Wheat Stockade and Cesario

par456:
Working on #11222

There were two versions of Stockade and Cesario in the wheat model, one under Preliminary and another under New Zealand. Got confirmation that the ones under Preliminary were the most updated version, so am deleting the ones under New Zealand.

This may change stats, as I suspect the New Zealand versions were used. If so, stats will need to be updated to use the correct cultivar now.


par456:
@HamishBrownPFR @hut104 

Just some questions from Neil on this so that we don't lose something you may have fitted. 

In the duplicate cultivars we are removing, we had a bunch of non-phenology changes. Now with the proper phenology, the cultivars most fit better and don't need those extra adjustments, but if they were measured, we should keep them in there.

They are:
```
[Leaf].ExtinctionCoeff.VegetativePhase.FixedValue
[Grain].NumberFunction.GrainNumber.GrainsPerGramOfStem.FixedValue
[Grain].MaximumPotentialGrainSize.FixedValue
[Leaf].CohortParameters.MaxArea.AreaLargestLeaves.FixedValue
```

par456:

There were two copies of the Stockade and Cesario cultivars in the Wheat model, causing outdated parameters to be used when picking one of these cultivars. This PR deletes the old cultivar so that the newest version is used.

This mostly changes the stats for the better, as the newer version was fitted on the validation better.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11464

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11222/files).

The changes in stats can be found [here](https://postats2.apsim.info/11222).

We intend to merge this into the release on Monday 3rd unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

PR #11466, opened 2026-07-28
10531 Wheat Winter Cereal Prototype Validation

par456:
working on #10531

This PR uses the new Virtual model to apply the new Wheat model as a prototype dataset over the existing Wheat dataset, so both the release and prototype can be tracked separately in POStats without duplicating the wheat validation files.

This will also start showing the Wheat / WheatWinterCereal comparision table on POStats.

This will cause a stats change, as it is adding 'new' data.


par456:

This PR adds two apsimx files that use the in development wheat model (based off WinterCereal), and applies it to the existing Wheat and GxExM validation datasets in a way that they will appear on our stats pages.

The aim of this work is to allow our model developers to work on and track changes to a prototype model that could replace an existing released model, and have it run on the existing validation datasets that the released model is using. For example, in the case of the wintercereal wheat, having separate stats tracked showing how a change effects the prototype, and against the Released model at the same time.

To go along with this PR, the POStats website has also had a small update, with outputs now shown in a tabbed interface, and a new comparision page to show how prototyped models using this system compare stats wise to the original models.

This PR does not change the functionality of any model, or the data of existing validation sets, so we will merge this today, but we wanted to inform the RP on this change as it will hopefully change how our developers work going forward.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11466

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/10531/changes).

The changes in stats can be found [here](https://postats2.apsim.info/10531).

ric394 (approved):

PR #11467, opened 2026-07-29
Volatilisation

sno036:
working on #9763 

Still working but please merge this so it is available to testers!

The parameter k_AFPV was pointing to Eo rather than wind speed for reasons 'lost to history' (something from Classic days). This has now been corrected. Testing has ensured that it is robust against wind data not being in the weather file.


par456:
Looks like it didn't have an effect on the stats, is that okay to merge?

sno036:
Yes please!

PR #11484, opened 2026-08-04
11222 Remove duplciated Cesario and Stockade from WinterCereal Validation

par456:
Working on #11222

Following on from #11464

Making the same change there into the WinterCereal validation to avoid confusion.


par456:

There were two copies of the Stockade and Cesario cultivars in the Wheat model, causing outdated parameters to be used when picking one of these cultivars. We fixed the released Wheat model last week, applying the same change to the in development Wheat model in this PR.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/pull/11484

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11222/files).

The changes in stats can be found [here](https://postats2.apsim.info/11222).

This will be merged today as it is for a prototype model that is still in development.

ric394 (approved):

PR #11508, opened 2026-08-05
Tidy up solution files

### LINKED ISSUE #11482: Remove unused solution files
### What happened?

It would be nice if visual studio code could automatically select ApsimX.sln as the default solution. Because there exist other solution files in the repo, a user must select this every time to get IntelliSense, type highlighting and other useful features. This is tedious. As these other solutions look not to be used, I suggest we remove them to benefit visual studio code users.

Windows

ric394:
resolves #11482

The repo had three solution files that made using VS code IDE a little tedious. These extra solution files have now been removed.

As a part of doing this, another project was added to the main ApsimX solution that had its own solution file.

Also another .slnf file was removed. It looked like a 'convenience solution' for just building the Models and ApsimNG projects which I'm assuming is unused.


lie112:
I'll put my hand up and own the slnx file. It was some time ago and I didn't think the local Visual Studio files ended up in the repository. This was needed during a changeover in .Net version as I needed to ignore rebuilding some projects to keep working on code and the cut down solution ran a lot faster without needing to load all the other projects. Happy for it to be deleted.

ric394:
> I'll put my hand up and own the slnx file. It was some time ago and I didn't think the local Visual Studio files ended up in the repository. This was needed during a changeover in .Net version as I needed to ignore rebuilding some projects to keep working on code and the cut down solution ran a lot faster without needing to load all the other projects. Happy for it to be deleted.

Thanks for the update. I was wondering about its purpose/history 😄

PR #11509, opened 2026-08-05
Allow outflow to be outputted even if inflow is not provided

### LINKED ISSUE #11461: Lateral outflow is not calculated when lateral inflow is null
### What happened?

APSIM Next Generation version: 2026.1.7969.0

I am using the WaterBalance/LateralFlowModel with non-zero KLAT, slope, discharge width, and catchment area.

Expected behaviour:
Lateral outflow should be calculated when soil water is above DUL, even when no lateral inflow is supplied.

Observed behaviour:
No lateral outflow values are produced in the DataStore. The OutFlow array appears to remain empty when InFlow is null.

Relevant code in LateralFlowModel.Calculate():

if (InFlow != null)
{
    if (OutFlow.Length != InFlow.Length)
        OutFlow = new double[InFlow.Length];

    double[] SW = MathUtilities.Add(swmm, InFlow);

    // lateral outflow calculation
}

Because the full outflow calculation is inside `if (InFlow != null)`, no outflow is calculated when lateral inflow is not supplied.

This seems inconsistent with the documentation, which describes lateral inflow as optional and gives the lateral outflow equation as a function of KLAT, water above DUL, slope, discharge width, and catchment area.

Possible expected logic:
- initialise OutFlow using the number of soil layers;
- treat null InFlow as zero;
- calculate lateral outflow regardless of whether lateral inflow is supplied.

Also, the code comment states that lateral flow currently does not move solutes.

Windows

ric394:
resolves #11461

This simply creates an Inflow array if the inflow property is null of the same length as physical layers.

It also adds some protection for when KLAT values are not provided in WaterBalance.

### DIFF
--- a/Models/Soils/WaterModel/LateralFlow.cs
+++ b/Models/Soils/WaterModel/LateralFlow.cs
@@ -55,34 +55,38 @@ namespace Models.WaterModel
         public void Calculate(double[] swmm)
         {
             // Lateral flow does not move solutes. We should add this feature one day.
-            if (InFlow != null)
+            if (InFlow == null )
+                InFlow = new double[physical.Thickness.Length];
+
+            if (OutFlow.Length != InFlow.Length)
+                OutFlow = new double[InFlow.Length];
+            double[] SW = MathUtilities.Add(swmm, InFlow);
+            double[] DUL = MathUtilities.Multiply(physical.DUL, physical.Thickness);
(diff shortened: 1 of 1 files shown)


par456:
Probably just hold off doing too much with this one at the moment. I couldn't find any simulations in the validation set that actually used this model, so we might need to address that first to ensure making this change doesn't cause an issue.

PR #11513, opened 2026-08-07
11510 SimulationResetTests

### LINKED ISSUE #11510: SimulationResetTests
### What happened?

One we might need some info from @hol430 

I've had a unit test called TestSimulation in SimulationResetTests, throw when I changed the start date of the OilPalm example. The test looks to check if the simulation object changes before and after it is run, and was created as part of the work for APSIM.Server #6529 

All the examples except OilPalm were commented out, and most of them fail if I try to run them. @hol353 turned them off in #8217 for some reason.

@hut104 thinks this may have been from an attempt to do sensitivity analysis, where instead of loading and running the same file over and over, it would instead just slightly change a value for each run one after another, but that would require the contents of the simulation don't change when its run.

Was this work ever completed? Did I accidently stumble across a bug here that the unit test wasn't capturing?

Windows

par456:
Resolves #11510 

This was a pain to get all the unit tests running again. This PR address a whole lot of little reset issues that a bunch of models had. In summary:

- Update RedClover example to fix DefoliationManager not resetting
- Added completed event listener to PastureSpecies so it can reset at end of simulation
- Go in reverse when undoing overrides/cultivar changes
- Add completed event listener to Leaf so it can reset at end of simulation
- If you try to get the VegetativePhaseCalculationWheat on a model without a HaunStage, it will return -1 instead of null reference error
- Change the soil KL back to original values after scrum simulation finishes and have scrum clean up if it finishes without harvesting
- Have SurfaceOrganicMatter listen for Completed event and reset
- Have Plant listen for completed event and reset

This might break validation or change stats, will debug further if thats the case. Might also be worth adding more examples to the unit test too.


hol430:
Wow, well spotted - I bet you had fun chasing those down!

ric394 (approved):

PR #11536, opened 2026-08-25
Try other matching methods when multiple exist

### LINKED ISSUE #11535: Operations should attempt to try multiple methods if named the same
### What happened?

The operations model will not try to use another matching method if multiple exist such as the method `SurfaceOrganicMatter.Add()`. It instead only uses the first one it finds. This should not be the case.

Windows

ric394:
resolves #11535

The Operations model will now attempt to get parameters for each method until a non-null set is found.


sno036:
@ric394  - I grabbed your new file and tested it. Good news is that it works! One thing that I suspect to be inconsistent with elsewhere is that the case of the label seems to matter. For example "no3" works but "NO3" does not. Thanks for your help with getting this sorted!

Issue #11519, opened 2026-08-16
AgPasture overstimates N fixation


rcichota:
### What happened?

All legume species in AgPasture are fixing N at their maximum rate (e.g. 60% of plant N demand in white clover). This rate should vary between a minimum (20% for white clover) and the maximum depending on the balance between plant demand and soil supply capacity. Checking the code revealed a bug in the calculations following the end of the Runge-Kutta loop to get the soil N uptake. Fixation is recalculated on ```SetActualNitrogenUptakes```, despite it already been computed on ```GetNitrogenUptakeEstimates```, which is within the Runge-Kutta loop. This seems to be needed to harmonise the estimates of fixation (and remobilisation) with the final value of N supply and uptake. The issue is that at this point the data being passed through ```ZoneWaterAndN``` does not include soil moisture, and this results in soil n supply being recalculated as zero and thus fixation as maximum...
This recalculation of soil N supply needs to either get soil moisture or to not require those values. Fixing this should fix the corresponding harmonisation of fixation and remobilisation (the method to compute the later should also be split into calculation of potential and actual change in remobilised N). This will have implication to the performance of AgPasture models...

### What Operating System are you on?

Windows

PR #11541, opened 2026-08-27
Fix AgPasture N fixation bug

rcichota:
Working on #11519 
In AgPasture, the methods to compute N fixation and remobilisation (reallocation) are intertwined with that evaluating soil N supply. They are all called on ```GetNitrogenUptakeEstimates```. There is also a final call to those methods on ```SetActualNitrogenUptakes```, at which stage no water information is passed resulting in the evaluation of N supply as zero and consequently maximising N fixation. This 'final' call seems unnecessary and results in overestimation of N fixation (although a call to the method evaluating N fixation is needed to match those estimates with the outcome of the Runge-Kutta procedure for N uptake). I am changing the logic a bit to simply set N supply to be equal to the values of N uptake and only then compute/update the values for fixation and remobilisation. A this will change current model behaviour, significant diffs are expected. Some may be smoothed by changing the N fixation parameters


rcichota:
@par456 @ric394 
Could this please be merged. The changes are mostly exposing some variables and lots of tidying up. Results from validation seem to be a bit better even... I fear this PR is getting too big (too many small changes) to keep track. I think it would be best to merge things up to here then complete the changes needed to resolve the issue.
I've made the changes as I went through the code to find out what was going on with N fixation/uptake and the allocation of new growth to roots (I found a small bug in there). I will commit changes to actually fix the issue afterwards.

par456:
Hi @rcichota

I'm not opposed to the idea of getting this merged in so your work isn't out on a branch too long, however there are a few things we need to do first in this PR.

First off you have a mix of c# files and apsimx files in this PR, and we generally only want one or the other in a PR. Most of the changes in the apsimx files seem to be from Test stats nodes, which is not something we really need to do in the PR here. If you could revert the apsimx files that don't have actual changes, that can clear them out of this.

I do note that it looks like you're renaming a few properties in some of the classes/resources. If that's the case and you need to update that in the apsimx files, we normally do that through a converter rather than manually updating, that way everyone using those models will also get their files updated.

Let me know if you need some help with any of that and I can give some demonstrations on how to do it.

rcichota:
Thanks @par456 
I have reverted the changes to the .apsimx files (I had committed them to make easier to compare/identify new changes) and had a go at the converter. I though the later would not be necessary as the changes are on parameter not really widely used (or not tinkered with by users), but its more than fair to have to do... However, I am not sure my implementation is working properly; the two simulations that failed (Narayen.apsimx and ThinzarExperiment.apsimx) do so because those changes in parameter name: for some reason the values are now zeroed (see snapshot below). The difference I can see from these simulations is that the PastureSpecies model has been renamed within those simulations (e.g. from AGPRhodes to Panicum in Narayen.apsimx). For simulations that kept the model's name there was no problem. Do I need to code the Coverter differently or is some different issue linked to these two simulations?
image

rcichota:
Question/clarification on how to deal with updates to the test simulations. They are never really updated when there are changes in them due to new APSIM versions (they work because the infrastructure / converter handles the changes). However, when making changes like I am doing here, it becomes really frustrating as is very difficult to see what I am changing from the changes accumulated over the last year or two... that is why I was committing them with the updates (mine and previous). Should I do a PR specific for these changes (update to latest version) before start making new changes? Should I then do a separate one in the end to incorporate any changes then that I made? Or updating these simulations for this reason is discouraged?

par456:
Just having a look at this converter issue now, see if I can work out whats going wrong. Thanks for fixing the merge conflict.

rcichota:
I've 'sneaked in' a couple more commits as they are basic tidying ups and again do not change the stats. The separation of calculations of tissue turnover and remobilisable N follow good practice of having a method doing one thing only and will make easy implementing some changes need to actually solve the main issue of this PR...

@par456 , not sure why the fail triggered for WheatWinterCereal.apsimx (Prototypes). The message says it is new, but the simulation is already in the Master branch... Did the change in microClimate trigger it to run when it wasn't before, or was it not supposed to??

par456:
Yep, its been happening with all the PRs since late Thursday, going to diagnose what's going wrong this morning so we can get these merged

par456:

Some refactoring work on AgPasture. Some trivial stats changes are made, but these do not actually affect the results (caused by floating point differences).

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11551

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11541/files).

The changes in stats can be found [here](https://postats2.apsim.info/11541).

This will be merged today as it does not make a functional change.

Issue #11546, opened 2026-08-31
Interference between independent paddocks


rcichota:
### What happened?

When using multiple paddocks (that is the basic ```Zone```) in a single simulation, there seem to be some interference between them with results changing depending on order as well as which paddock is enabled...
In the attached simulation, three paddocks are used to grow ryegrass, white clover, and paspalum, each in their separate paddocks. When one of the paddocks is disabled, the result from the other changes when it shouldn't have. In the example, an experiment was setup with the main Factor being enable/disable the paddock with ryegrass, and it is clear that the results from the other two crops differ. Changing the order of paddocks does change the results as well...
We noted that intercepted radiation for each crop changes from day one, potentially giving a clue that MicroClimate is not behaving as expected, but there might be other issues related to scoping...

[TestAgPastureInPaddocks.zip](https://github.com/user-attachments/files/31666686/TestAgPastureInPaddocks.zip)

### What Operating System are you on?

Windows

par456:
There are a few things here.

There was some incorrect linking happening in MicroClimate that causes it to reference only the first zone's soilwater, I resolved that pretty quickly to make it use the soilwater for each zone instead of a link, the problems persisted.

But the 2nd issue is more complex. Your problem is here:

<img width="1225" height="977" alt="Image" src="https://github.com/user-attachments/assets/e2b14d1a-50cb-4a36-98f3-4bc932f0bef4" />

When microclimate is providing the light profile, PastureSpecies is doing some extra calculations for fractionGreenCover and swardGreenCover, but is doing some illogical things. Its dividing the provided canopyLayer.AmountOnGreen over the microClimate.RadiationInterceptionOnGreen, which doesn't make much sense here, since realistically those should be the same.

However in a multi-zone simulation microClimate.RadiationInterceptionOnGreen is only going to give the amount on green for the first zone every time.....but dividing it by that zone's amount on green (which is provided through the light profile), means that would always just equal 1.

lie112:
OFFICIAL

It might be worth noting that I believe multiple paddocks (Zones) in the simulation will be treated as manual  urine and manure patches in SimpleGrazing rather than these being created dynamically in the code. This could be part of the problem as to why the paddocks appear linked when you expect them to be separate paddocks. This functionality should be handled by nested paddocks being treated as urine and manure patches and sibling paddocks treated as separate paddocks with independent pasture and grazing.

[...]

PR #11547, opened 2026-09-01
11546 Fix links in Microclimate

par456:
Working on #11546

This resolves the incorrect linking that is happening, but does not solve the issue. Seeing if this changes stats.


par456:
This luckily didn't change stats and resolves some mistakes, but not the route cause of the issue.

This can be merged and @rcichota can act on the other issue found in a separate PR.

PR #11553, opened 2026-09-02
Update CO2 values for values from NOAA

### LINKED ISSUE #11538: CO2 record needs updating
### What happened?

The datafile provided to users for global co2 level is now out of date.  The data for 2025 should be added to the end of the record.

Windows

ric394:
resolves #11538

This PR updates the example CO2 values for ones sourced from https://gml.noaa.gov/ccgg/trends/gl_data.html. These are global values for each year from 1979 to 2025.

The old values were close but these should be more up to date.


ric394:
@hut104 any thoughts on if this is a suitable way forward? 
I've searched the git history of the original file (CO2.csv) and its original issue for a source for the original values with no success. 
The upside of moving ahead with these values is we can refer to this source for future updates.

par456:
How we'll resolve this PR:

- Make a new .csv file from the source you have, this should be global based co2 level, not a northern or southern number.
- Call that new csv a name that describes where the data came from.
- Create an .md file that outlines what CO2.csv is (a co2 file with an unknown source that is being archived here now), and the new co2 file, where it came from, and the process for updating it.
- Update the co2 example to point at the new file.

PR #11560, opened 2026-09-04
Expand the delete command

### LINKED ISSUE #11518: Comand line delete can delete models from resources that are not in the loaded .json.
### What happened?

Specifically, when I include:
delete all [Cultivar] 
in an apply file (with the intention of deleting cultivar descriptions in a replacements node)and run the file, the model throws an error:
"Cannot find cultivar of name ..."
I am guessing, but the only explanation I can think of is that it has deleted the cultivar models from the crop model in memory.  The crop model is not in the json, it is loaded as a reference but it appears the delete action is affecting it too.

I have found 
delete all [Replacements].[Cultivar]
works as intended, deleting cultivar models from the replacements model but not from the base crop model and everything runs fine.
for the record:
delete all [Cultivar] from [Replacements]
delete all [Cultivar] to [Replacements]
do not delete the cultivar models from replacements but they dont raise an error either, just fail silently.

Windows

ric394:
resolves #11518

The delete command can now handle commands like:

```text
delete all [Cultivar] from [Replacements]
```

This should help add some needed precision when deleting models.

### DIFF
--- a/APSIM.Core/CommandProcessor/Commands/DeleteCommand.cs
+++ b/APSIM.Core/CommandProcessor/Commands/DeleteCommand.cs
@@ -3,21 +3,26 @@ namespace APSIM.Core;
 /// <summary>A delete model command</summary>
 internal partial class DeleteCommand : IModelCommand
 {
-    /// <summary>The name of the model to delete.</summary>
+    /// <summary>The name or path of the model to delete.</summary>
     private readonly string _modelName;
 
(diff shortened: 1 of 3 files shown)


par456:
What does the parent flag do in this?

Also, what does "delete all to" do?

ric394:
The `delete all to` is not a valid command. I'll remove that.

Issue #11551, opened 2026-09-02
SimpleLeaf class does not handle non-structural biomass correctly.


hut104:
### What happened?

If a model developer configures a simple leaf class to include non-structural DM within a leaf, the model does not account for reallocation of dm from leaves during senescence.
The code will need to be able to account for DM retraslocated from leaves to other growing organs during canopy senescence.

### What Operating System are you on?

Windows

PR #11562, opened 2026-09-06
Account for reallocation from senescing material moving to the dead pool in simple leaf

hut104:
working on #11551


hut104:
@par456 and @ric394 The code changes deal with a deficiency in the model in that reallocation of biomass from leaves as they senesce was not removed from the mass moved to the dead pool.  To allow users to check this, some outputs have been added for future use.
The changes are all positive.

par456:

An error was found in the SimpleLeaf model causing senesced leaves not being moved to dead biomass pools. This PR fixes that mistake by moving them there correctly.

All stats are improved.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/pull/11551

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11562/files).

The changes in stats can be found [here](https://postats2.apsim.info/11562).

We intend to merge this into the release on Wednesday 9th unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

HamishBrownPFR:
We need this bug fixed and the changes to stats are trivial so recommend merging

PR #11569, opened 2026-09-10
Fix the issue that FrostHeatDamageFunctions silently ignores one crop's output in a rotation

### LINKED ISSUE #11568: FrostHeatDamageFunctions silently ignores one crop's output in a rotation
### What happened?

Thanks to Di @hedycau for finding this issue.

When a simulation rotates between two crops that each have their own FrostHeatDamageFunctions child (e.g. Wheat in year 1, Canola in year 2, both present as siblings under the same Zone), only the first crop's damage function output is ever reported. The second crop's FrostHeatDamageFunctions instance computes correctly, but any Report/Graph that references it by name (e.g. [FrostHeatDamageFunctions].FrostHeatYield, as used in the shipped example files) keeps reading the first crop's values for the entire simulation.

Linux

PengchengHU:
Resolved #11568 

Tried to dynamically auto-rename the name of damage functions by adding a prefix (i.e., crop name) to it, but seems that ApsimNG does not refresh a tree node's label on a generic property edit; the rename wouldn't show up until you saved and reopened the file.

Instead, I simply added a runtime check at simulation start, which will throw a clear error if two `FrostHeatDamageFunctions` share a name, telling the user exactly what to rename them to give each crop's instance a unique name and reference it via a fully-qualified path in Reports. Also, I added a "How To Use This Model" Memo to cover usage, outputs, the rotation-naming guidance, and funding/citation info.

### DIFF
--- a/Models/AgPasture/PastureSpecies.cs
+++ b/Models/AgPasture/PastureSpecies.cs
@@ -65,10 +65,6 @@ namespace Models.AgPasture
         [Link]
         private ISoilTemperature soilTemperature = null;
 
-        /// <summary>Link to micro climate (aboveground resource arbitrator).</summary>
-        [Link]
-        private MicroClimate microClimate = null;
-
         ////- Events >>>  - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
         /// <summary>Invoked for incorporating surface OM.</summary>
@@ -193,9 +189,11 @@ namespace Models.AgPasture
                 {
                     InterceptedRadn = 0.0;
                     myLightProfile = value;
+                    double totalRadiationInterceptionOnGreen = 0;
                     foreach (CanopyEnergyBalanceInterceptionlayerType canopyLayer in myLightProfile)
                     {
                         InterceptedRadn += canopyLayer.AmountOnGreen;
+                        totalRadiationInterceptionOnGreen = canopyLayer.AmountOnGreenTotal;
                     }
 
                     // stuff required to calculate photosynthesis using Ecomod approach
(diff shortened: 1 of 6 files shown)


PengchengHU:
Hi Andrew @par456  and Julian @ric394 , could you please help me with the stats? I'm not sure why there are so many numbers related to `WheatWinterCereal` in the Stats. Thanks.

par456:
Hi Pengcheng,

I've been getting that on PRs all day, not sure whats going on, I'll need to investigate some more tomorrow.

par456:
Some comments on this PR:
- Adding a child node when creating the node isn't something we normally do in APSIM. I think a better solution would to use a view that allows the optional displaying of that information like some other models do that need instructions.
- The ambiguity issue should already be throwing an exception, as that is something Report is checking for (if it can match multiple models to the same line).

Let me look into both those to see if we can deal with this in a better way.

PengchengHU:
Hi Andrew, thanks for the comments. I tried to dynamically auto-rename the name of damage functions by adding a prefix (i.e., crop name) to it, but it seems that ApsimNG does not refresh a tree node's label on a generic property edit; the rename wouldn't show up until you saved and reopened the file. You must have a better way.

par456:
I've gone through and done a general cleanup of code and solved your problems in some different ways.

Report wasn't throwing on ambiguous references like it should, so this might have some fallout to other files in the validation. Will see in the results and if its bad, i might need to add that check in a separate PR.

par456:
@PengchengHU 

I've gone through and done a bit of a clean and lots of changes to this. 
- The reporting issue has been separated off and merged in with #11583, so that exception will happen on all models now instead of needing something specific for FrostAndHeat.
- Your instruction text will now appear in the GUI alongside the property display, so it doesn't need to be put in as a memo in the structure, so that made all this much simpler.
- I've also generally formatted the code to the current C# standard which is where most of the changes are.

Julian is going to give this a review, and if both you and he are happy, we'll get this merged.

ric394 (approved):

PR #11570, opened 2026-09-11
Add a constraint to the range of FractionOfDungUrineOffPaddock in Sim…

### LINKED ISSUE #11567: SimpleGrazing is allowing net export of dung and urine outside the range of 0-1
### What happened?

It should only allow a fractional values

Windows

sno036:
resolves #11567

### DIFF
--- a/Models/AgPasture/SimpleGrazing.cs
+++ b/Models/AgPasture/SimpleGrazing.cs
@@ -502,6 +502,12 @@ namespace Models.AgPasture
                     throw new Exception(errors);
             }
 
+            // FractionOfDungUrineOffPaddock should only lie between 0 and 1. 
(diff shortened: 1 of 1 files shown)


sno036:
The exception is just new data for Wheat as far as I can see. Mergable?

ric394:
@sno036 The exception is related to the code changes. See below:

```text
System.AggregateException: One or more errors occurred. () (Error while performing simulation cleanup)
---&gt; ERROR in file: /wd/Examples/SWIM.apsimx
Simulation name: SWIMExample
Models.Core.SimulationException
---&gt; System.Exception: FractionOfDungUrineOffPaddock should be between 0 and 1.
```

Looks like it's getting a non-fractional value in a few places.

PR #11572, opened 2026-09-14
Fixing AgPasture N balance bug

rcichota:
Working on #11519 
Continuing the work done in #11541 

A few issues affecting N balance in PastureSpecies (AgPasture) have been found and need fixing. The issues are somewhat related and encompass the estimation of N fixation (or its handling during Runge-Kutta process for N uptake), reporting of soil available N (after Runge-Kutta), allocation/remobilisation of N in roots, and weighting of N allocation to organs under stress. 
The methods to compute N fixation and remobilisation (reallocation) are intertwined with that evaluating soil N supply. They are all called on GetNitrogenUptakeEstimates. There is also a final call to those methods on SetActualNitrogenUptakes; this 'final' call seems unnecessary and results in overestimation of N fixation (although a call to the method evaluating N fixation is needed to match those estimates with the outcome of the Runge-Kutta procedure for N uptake). I have revised and am changing the logic a bit to re-evaluate N supply (and fix output as now this is zero) ensuring it is not smaller than N uptake, and only then update the values for fixation and remobilisation. 
While fixing this I noticed an issue with the allocation of N that follows the processes mentioned above. There were instances where N allocated to some organ could lead to N concentration going below minimum or above maximum (and no test for it), this was more prominent for roots where partition also need to consider layers... I have revised the method to partition N between organs, making the calculation to go more explicitly in stages (fist fill in to minimum, then to optimum and only then luxury, with test to avoid going over), also converting all the related variables (turnover, remobilisation, etc.) to be arrays for the roots, so loss of mass balance can be easily catch...
As these will change current model behaviour, significant diffs are expected. Some may be smoothed by the interplay of the various processes (e.g. change in N fixation on any given day can be large, but the cumulative effect over time seems to be fairly small in general), some changes to the value of parameters may help (as some may reflect 'bad' behaviour of old code)...


rcichota:
@sno036 
These are the latest changes for you to check. The diffs are similar to what they were when we talk the other day, in the main validation simulation, some results are better and a few slightly worse (only clover content for the Radcliffe set is considerably worse). The prototype sims are also mixed or better. The two that have worse stats are RITTemplate2025 and TreadingDamageTest. Also, now that Andrew fixed the patch tests, we now have an exception raised for PatchyMcPatchFace...

sno036:
Thanks @rcichota. @par456 @ric394  - After discussion with Rogerio he is going to commit an updated set of observeds for the Treading model. 

This presents an improvement to the model so I am happy it be merged. But how to deal with the Patching statistics? I don't want these now statistics accepted. Happy to discuss!

par456:
If we just merge in the observed data separately, then the patching statistics won't be changed (they are only effected by model changes). Merging in observed data changes is very straightforward.

Then the patching statistics changes can be put in alongside the model changes when those are due to actually go in.

rcichota:
Thanks @par456 
Just to clarify, the observed data I have updated is for one of the 'system' simulations (TreadingDamageTest), not for patchiness. Me and Val checked the former and noted that the diffs were due to changes in biomass production, while the behaviour the simulation was testing (treading damage) remains broadly the same. Hence updating the observed should be enough (this are simulated data against simulated data). Please let me know if you want a separate PR to send this in or is ok just adding here...
Now, for the PatchyMcPatchFace, although the situation is similar for the changes due to this PR, the background is different. Because that simulation was not actually run properly for a while, its stats had drifted prior to this PR. Val does not want to lose that and so is not prepared to simply replace the observed values for it (not now, without understanding where they came from). So, will be possible to merge this PR without fixing the patchiness test?

par456:
> So, will be possible to merge this PR without fixing the patchiness test?

Not without turning the patchy test off once again, otherwise once it's in the release everyone PRs will throw that exception and it'll get very confusing for them.

PR #11575, opened 2026-09-14
Constrain SimpleGrazing's dung/urine sent off-paddock to 0-1

sno036:
resolves #11570 

Strange conflicts in the original branch so blew it away and started again.

Added a constraint to the excreta export.

Fixed the SWIM example that had a -89 appear in the fraction of dung/urine N sent off paddock. This (a value of -89) has appeared in one other simulation - both for no apparent reason and I have been unable to replicate it.


sno036:
I am unsure what is going on here. My branch was up to date before the pull request but is showing as having new data included. Any advice please?

par456:
Don't mind the stats change, the logs are saying that the SWIM validation and example both threw exceptions when there were run.

PR #11580, opened 2026-09-16
Report NO3 and NH4 components.

### LINKED ISSUE #11579: NO3 and NH4 fertiliser application amounts
### Describe the new feature

Certain projects might like nitrogen fertilisers broken down into their solute components for reporting. While this could be done as a post processing step, it would be nice if the model could report this as well.

ilhuber:
Resolves #11579 

Maybe we would like something more general?


ric394:
Are you able to make something more general that perhaps uses the Solutes name to track the amount applied?

par456:
As Julian said, I think this will need a more generalised solution. An array holding this value for each solute, and a function that takes a string for the get call.

Also, might be worth calling this a daily applied or similar, so its clear the value gets reset each day, otherwise people might assume its applied over the whole simulation.

ilhuber:
Hmm, there are some issues with the present NitrogenApplied variable. It will be set even if the fertiliser being applied does not have any Nitrogen (e.g. Dolomite). Also, its units can be thrown off with something like UreaGranular.

Maybe it would make more sense to store this variable in the solutes being applied by the fertiliser? Simulation solutes can be a bit dynamic; they can be added or removed but I am pretty sure that the default Nutrient model will not function if Urea, NH4 and NO3 are not present.

Issue #11577, opened 2026-09-16
Spectral model to support multiple crops


hut104:
### Describe the new feature

The spectral model calculates indices (e.g. NDVI) for comparison to remote or proximal sensing.  The current implementation only supports monocultures (ie wheat).  Emerging usage is for forestry systems where there is both overstory and understory components.  The model would be required soon for rotations or pasture systems, all of which will require the model to support more than monocultures.

The suggested change would be best implemented by using the information on light interception from MicroClimate.

PR #11584, opened 2026-09-17
Calculate NDVI from microclimate outputs

hut104:
working on #11577


hut104:
@par456 I think that this can be merged in.  I have spoken with @HamishBrownPFR about this.

I will raise a a second issue regarding the wheat model and ways in which it represents canopies when communicating with MicroClimate.

par456:

When adding the NDVI to a multi-crop simulation with Pinus and Pasture, it was found the Spectral model only looks at the first plant to do its calculations. To resolve this the model has been changed to ask MicroClimate to calculate the canopy of the entire zone, instead of Spectral trying to do that.

This however has caused some stats changes as MicroClimate does its calculations at start of day and Spectral did it at end of day previously. The formula is also slightly different, as Spectral assumed that all dead biomass was below green biomass, whereas MicroClimate does not.

Further work will be needed to improve this model, however this change reduces duplication of functionality and allows Spectral to start being used in more complex simulations.

This PR addresses the problem, having Spectral ask Microclimate 

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11577

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11584/files).

The changes in stats can be found [here](https://postats2.apsim.info/11584).

We intend to merge this into the release on Tuesday 22nd unless there is any objection. There is no need to reply unless you wish to suggest alternative action.

HamishBrownPFR:
This is a necessary change to expand the functionality of APSIM and the effects on existing stats are minor.

Recommend merge

sno036:
@par456 - I'm fine with the coding and validation changes but I am not seeing the documentation file updated or any comments that users could use to understand what the code is doing. I realise that this is not just the NDVI work but the science documentation has not been updated for two decades despite the changes in zones etc. Where should the line be drawn here on undocumented changes?

par456:
I agree with you there Val, I'm going to get some documentation added to the Spectral model as part of this PR since it only has a single line right now.

The question around models like Microclimate is an ongoing issue that I don't really know how we address. I think there needs to be some effort to document how it currently functions, but that doesn't solve the problem of it continuing to change over time.

PR #11585, opened 2026-09-17
11038 Make a Pinus v2 prototype for model development

par456:
Working on #11038

A new file for prototyping the Pinus model so it can be worked on without effecting the released model. Uses Virtual to share the validation set from pinus.


par456:

Adding a new prototype file for work on the Pinus model. This will allow for stats tracking and merging for this project with it causing changes to the released Pinus model.

The issue has been documented https://github.com/APSIMInitiative/ApsimX/issues/11038

The file changes are [here](https://github.com/APSIMInitiative/ApsimX/pull/11585/changes).

The changes in stats can be found [here](https://postats2.apsim.info/11585).

This does not actually change stats, but does add new stats for the starting point of the Pinus model. This will be merged today.

PR #11589, opened 2026-09-22
11587 Check children of IFunctions for correct types

### LINKED ISSUE #11587: APSIM does not throw when an IFunction under a multiply is replaced by something thats not an IFunction
### What happened?

Just ran into an issue where a manager script in Replacements had the same name as an IFunction in a plant model, which caused the model to incorrectly do the calculation as it simply skipped/ignored the IFunction turned manager script.

This is both the danger and power of a replacement model, but really the only non-IFunction below something like  the multiply function should be a memo, so we might just fix this by making the calculation stricter.

Windows

par456:
Resolves #11587 

This PR is looking to do two things, first is to check that IFunctions in models only have IFunction or memos as children, and throw if they do not.

It then also adds a rule into the replacements folder code to prevent memos in replacements from overwriting other models in the file. 

This should hopefully stop this sort of error happening again.

Some additional cleanup is needed, but want to run this against the test set before going forward further.


par456:
To solve this problem I have added a new method to a bunch of the IFunctions to ensure they only have children of type IFunction or IText (memos etc), that way if a replacements changes them, it should throw then.

I put this check in as a static utility functions so that it wasn't copied everywhere.

I've also added a section to replacements to prevent memos from replacing anything, as that prevents the same bug still happening with a memo.

Lastly, I removed Structure references from the IFunctions I edited since we are moving away from that.

Unit test were also written.

Issue #11591, opened 2026-09-22
PatchyMcPatchFace is failing locally but not in the test system


sno036:
### What happened?

The simulation ...ApsimX\Tests\Simulation\SoilNitrogenPatch\PatchyMcPatchFace.apsimx is designed to fail when the pseudo patch simulations drift away from the equivalent explicit patch simulations. It is currently failing locally but is not triggering the test system. This has probably been happening for a while now as the plots have deviated substantially (for the worse) since the publication (Figure 2 in https://doi.org/10.1016/j.mex.2026.103790).

### What Operating System are you on?

Windows

ilhuber:
I am also finding different numeric behavior depending on running environment (differing number of SWIM failures between Linux and Windows runs of the same simulation). Will look deeper into this.

par456:
Spent an hour this morning tracking the Patchy changes, we have 3 in the past year.

May 13: #11272 Root depth should not be greater than the actual maximum potential depth
Patchy didn't change until this PR, (confirmed numbers didn't change between November 1st to May 12th)

August 25: #11517 Updating AgPasture initialisation
Then changed again with this one

September 14: #11541 Fix AgPasture N fixation bug
And again in this one, but this also caused Patchy to stop throwing as the stats moved back into the tolerance set last year.

PR #11592, opened 2026-09-22
11591 Lower Tolerance on PatchyMcPatchFace to ensure it throws

par456:
Working on #11591 

Just checking that if forced it will actually throw in the test system.


par456:
@sno036 

I'm confident now that Patchy is running correctly and the test is showing its currently within tolerance. I'll get this merged in so we know if any other PRs effect this now.

ric394 (approved):

